Issue #25, opened 2022-05-04
Add notebooks to fill the basic CLV grid


ricardoV94:
The idea is to write a notebook with pure PyMC model(s) for each of these CLV scenarios. We can start with the Lifetime part (not-value yet), but ideally we will include value as well by the end. This might be a constant with time-decay penalty in the simplest cases.

Hopefully this will give us a good picture of the building blocks that are necessary for a minimum viable package, and can also serve as the base documentation. Overtime we would replace the custom PyMC code with imports from the CLV sub-package.

![image](https://user-images.githubusercontent.com/28983449/166695993-688174ec-04b8-4ac4-ade0-7caba1919816.png)

- [x] Continuous Non-contractual #16 
- [x] Continuous Contractual #36
- [ ] Discrete Non-contractual (up for grabs)
- [x] Discrete Contractual #32

larryshamalama:
> Is the goal to add more than one model to each quadrant? The CNC quadrant now has @larryshamalama 's BG/NBD distribution (#16), but there are several other models available in that quadrant. I wasn't sure if the idea was to get one per quadrant max or if the goal was broader than that. Also, the distribution added in #16 is named `ContNonContract`, which seems to imply that there aren't going to be other continuous non-contractual distributions.

Sorry for not replying to your question initially. Effectively, there are other continuous non-contractual distributions, as I realized later on (e.g. ParetoNBD). We briefly touched upon this point in #128

cluhmann:
Discrete contractual is covered by https://github.com/pymc-labs/pymc-marketing/pull/133.  Looks like a PR to address https://github.com/pymc-labs/pymc-marketing/issues/176 would cover the final quadrant.

BBDS-Colt:
We have a distribution block for Continuous Contractual, but no modeling options per https://github.com/pymc-labs/pymc-marketing/issues/279. I've gotten questions on LinkedIn about modeling purchases made through active memberships and/or phone app sessions, so there's demand for it.

ColtAllen:
We have a distribution block for Continuous Contractual, but no modeling options per https://github.com/pymc-labs/pymc-marketing/issues/279. I've gotten questions on LinkedIn about modeling purchases made through active memberships and/or phone app sessions, so there's demand for it.

ColtAllen:
Related: https://github.com/pymc-labs/pymc-marketing/issues/64

PR #32, opened 2022-05-06
sBG model notebook for discrete/contractual setting

drbenvincent:
Addressing #25, here is a notebook to demonstrate the sBG model for the discrete/contractual setting.

As well as a general review, here are a few questions or things to think about
- feedback on the maths and implementation would be appreciated. **We seem to get estimation bias.**
- ~I'm happy to develop a `pm.TruncatedGeometric` distribution as part of this pull request, but equally it could be a separate thing.~ (EDIT: I think this can be done after the notebook is at least semi finished and we have faith that it's working as intended)

On my TODO list is:
- [ ] Add a summary, particularly highlighting the (possibly wrong) interpretation of a posterior over theta means.
- [ ] Add code + plot to actually calculate lifetime value distribution.
- [x] Better function to generate synthetic data, with actual randomness
- [x] add assertion that we have >1 observation
- [ ] Resolve the estimation bias for multiple cohorts
- [x] Add model: `theta ~ cohort`
- [ ] ~Add model: `theta ~ year`~
- [ ] ~Add model: `theta ~ cohort + year`~
- [ ] explain why adding year as a predictor would totally change the model


ricardoV94:
Closing this in favor of #133

PR #44, opened 2022-06-22
Improve MMM base example notebook

juanitorduz:
Add more effect plots to explain the cumulative and share channel effects. This is a followup of https://github.com/pymc-labs/pymmmc/issues/21


juanitorduz:
I will close this one as it diverged from the initial notebook draft. Most of the plots are already in master.

Issue #46, opened 2022-07-05
Marginalization over dichotomous variable


larryshamalama:
I know that this is ongoing work raised in [issue 21](https://github.com/aesara-devs/aeppl/issues/21) of AePPL, but is there a quick way to marginalize over a discrete parameter? Akin to the example provided in the issue linked above:

$$p(Y=y | X=0) * p(X=0) + p(Y=y | X=1) * p(X=1)$$

for some continuous $Y$ and dichotomous $X$. In this case, $X$ would be the variable indicating a customer churning or not and the difference between a contractual and non-contractual likelihood would be that one is the marginalized version of the other. I have not revisited the math in several weeks, but I'm fairly certain of this and the ability to marginalize likelihoods as such may provide better model building blocks rather than define a distribution class for each quadrant. Just an idea so far...

ColtAllen:
Is there a status update on this? I see the linked issue was made before the `pytensor` fork.

ColtAllen:
Any reason for this issue to remain open?

PR #65, opened 2022-11-05
Creating a distribution class for the Gamma-Gamma distribution

larryshamalama:
In this (so far WIP) PR, I implement the Gamma-Gamma distribution class, akin to `ContContract` and `ContNonContract` (not to be confused with the `GammaGammaModel`. This builds on #61. Given the CLV API for customer retention and spending models, the creation of such an underlying log-likelihood graph for mean transaction inference is based on `pm.Potential` rather than a "proper" distribution class.

### DIFF
--- a/pymmmc/clv/models/gamma_gamma.py
+++ b/pymmmc/clv/models/gamma_gamma.py
@@ -8,6 +8,7 @@ import pymc as pm
 from aesara.tensor import TensorVariable
 from pymc import str_for_dist
 
+from pymmmc import GammaGamma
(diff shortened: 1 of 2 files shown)


ColtAllen:
I'm curious - what was the conclusion of these discussions? And was it in any way related to https://github.com/pymc-labs/pymc-marketing/issues/46? I recall the question of marginalization coming up with the [ParetoNBD model](https://github.com/pymc-labs/pymc-marketing/issues/127), which has its own distribution class.

PR #67, opened 2022-11-09
Adding BetaGeoModel to clv submodule

### LINKED ISSUE #39: Implement pre-built `BG/NBD`/`BetaGeoFitter` model
It would be good to add a `BetaGeoFitter` function that returns a `ContNonContract` with some default priors. A signature that resembles what is provided in the `lifetimes` package would be a good idea. Something along the lines of the following snipet.

```python
def BetaGeoFitter(name, a, b, r, alpha, T, T0, *, observed, **kwargs):
    p = pm.Beta(f"{name}_beta", a, b, size=size, shape=shape)
    lam = pm.Gamma(f"{name}_gamma", r, 1/alpha, size=size, shape=shape)
    return ContNonContract(name, lam, p, T, T0, size=size, shape=shape, **kwargs)
```

We should also add some useful summary stats / plots. If they are not specific to the `BG/NBD` the better!

larryshamalama:
Closes #39.

Currently, the progress is mostly a template. Akin to the `GammaGammaModel` in #65, the likelihood function stems from a distribution class around which the API is wrapped.


pymmmc/tests/clv/models/test_beta_geo.py
```diff
@@ -0,0 +1,205 @@
...
+            2.17059763,
+            3.29959668,
+            4.06284244,
+            4.72380438,
+            1.73223875,
+            3.1645582,
+            4.08454744,
+            5.20603817,
+            6.34220761,
+        ]
```
ricardoV94, line 155: Where are you getting these numbers? Can you generate them in the `setup_class` method?

pymmmc/tests/clv/models/test_beta_geo.py
```diff
@@ -0,0 +1,205 @@
...
+            8.18144426,
+            8.50640294,
+            8.81155417,
+            9.0994203,
+            9.37205729,
+            9.63116452,
+            9.8781637,
+            10.11425711,
+            10.34047107,
+        ]
```
ricardoV94, line 185: Same here

pymmmc/tests/clv/models/test_beta_geo.py
```diff
@@ -0,0 +1,205 @@
...
+            "b_log__": (),
+            "alpha": (),
+            "alpha_log__": (),
+            "r": (),
+            "r_log__": (),
+        }
+
+    def test_model_convergence(self):
+        # b parameter has the largest mismatch of the four parameters
+        # rtol=0.3 should be enough
```
ricardoV94, line 99: Can you confirm that if you increase your data size the mismatch gets smaller?

pymmmc/tests/clv/models/test_beta_geo.py
```diff
@@ -0,0 +1,205 @@
...
+        a_true = 0.8
+        b_true = 2.5
+        alpha_true = 3
+        r_true = 4
+
+        # Number of subjects
+        N = 500
+        # Subject level parameters
+        p = pm.draw(pm.Beta.dist(a_true, b_true, size=N), random_seed=rng)
+        lam = pm.draw(pm.Gamma.dist(r_true, alpha_true, size=N), random_seed=rng)
```
ColtAllen, line 25: n = 500 is a small dataset; perhaps too small for model convergence.

I think it would make more sense to test this with the CDNOW dataset, which is a benchmarking dataset frequently used in research. This is what lifetimes uses for testing, and being able to reproduce research results is the closest thing we have to a ground truth for model assessments.

It would be worth adding a datasets utility that includes CDNOW in a separate PR.

pymmmc/tests/clv/models/test_beta_geo.py
larryshamalama: I am able to decrease `rtol` with increasing `N`

ricardoV94: Good enough, no need to change it

pymmmc/clv/models/beta_geo.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def _to_xarray(self, *arrays, **kwargs):
+        if "coord_label" in kwargs:
+            coord_label = kwargs["coord_label"]
+        else:
+            coord_label = "customer_id"
+
+        num_customers = len(arrays[0])
+        dims = (coord_label,)
+        coords = {coord_label: range(num_customers)}
```
ricardoV94, line 203: We should ask users to pass the `customer_id`. If they want the expected_num_of_purchases of customers 10 to 100, those should be the coords, not 0 - 90

pymmmc/tests/clv/models/test_beta_geo.py
larryshamalama: Resolving this in light of issue #100 and PR #104 that you just opened

pymmmc/tests/clv/models/test_beta_geo.py
larryshamalama: I already added it, it can be easily removed. Hopefully it doesn't add too much computation to our tests. I'll force push the change this afternoon.

pymmmc/clv/models/beta_geo.py
larryshamalama: The sampled parameters aren't specific to an individual if I remember correctly. What effect will these coords have on the output other than renaming the DataArray coords?

larryshamalama: Do you mean that we want to have `expected_num_purchases` to have a method signature that takes input only `customer_id`? Are you saying that we want to provide users to option to call

```python
expected_num_purchases(some_t_of_interest, customer_id=range(10, 90))
```
instead of
```python
expected_num_purchases(some_t_of_interest, frequency[10:90], recency[10:90], T[10:90])
```

pymmmc/tests/clv/models/test_beta_geo.py
larryshamalama: Yes! I am now using the function provided in `BetaGeoFitter` from the `lifetimes` package

pymmmc/clv/models/beta_geo.py
larryshamalama: Or are you suggesting that we ask users to provide `customer_id` to clarify the coords of the DataArray? Sort of like what is being done in the Gamma-Gamma model currently

ricardoV94: > Or are you suggesting that we ask users to provide `customer_id` to clarify the coords of the DataArray? Sort of like what is being done in the Gamma-Gamma model currently

This

larryshamalama:
Most recent updates:

- `TestBetaGeoModel` now uses `BetaGeoFitter` methods from `lifetimes` for testing purposes
- `customer_id` added as an argument for `BetaGeoModel` and its methods
- type hints added

ricardoV94 (approved):

PR #103, opened 2022-12-08
Work on streamlining docs

OriolAbril:


OriolAbril:
As there are no rtd previews yet, this is how it looks now with both themes:

imatge

imatge

There should also no longer be any warnings or errors, and I have removed the distributions module, the two contract ones are on the top level namespace so that repeated them and I don't think RVs should be documented there.

ricardoV94:
Looks good. I prefer the white theme, it's also more in line with the Labs website. Anything else that's needed here?

ricardoV94 (approved):

OriolAbril:
https://github.com/pymc-labs/pymc-marketing/blob/c06b1e782858f4d8f667704a2d2a3e5187fd0c69/docs/source/conf.py#L144 makes it default to white independently of the browser theme,but it can be changed from the navbar button. If we want we could delete that button so the dark mode becomes unaccessible.

The only thing missing is rtd config (see discord)

OriolAbril:
Everything working already: https://pymc-marketing--103.org.readthedocs.build/en/103/

juanitorduz (approved):

OriolAbril:
Given the 2 approves I decided to go on and merge to avoid the need for further PRs like #106 and potential git conflicts

Issue #99, opened 2022-12-08
Consider using "penalized" priors by default in CLV models


ricardoV94:
In conversation with @ColtAllen I realized the reason `find_MAP` gives worse results than lifetimes is (might be?) that lifetimes is often doing penalized MLE. We should investigate if we can add better priors so that `find_MAP` is more competitive with the old lifetimes results in terms of stability/speed.

This is important as we will probably want to provide users access to MAP/MLE fits.

ricardoV94:
About API, should we rename the current `model.fit()` to `model.sample()` and leave `fit()` for the MLE/MAP approach? Or give the latter a separate name `fit_MAP()`?

twiecki:
I would go with `fit_MAP()` just because our 95% default will be sampling and `.fit()` I think is what people would expect as the main fitting method.

larryshamalama:
Would `model.sample()` be confused with `pm.sample()`? Even though we're calling the latter implicitly

ricardoV94:
> Would `model.sample()` be confused with `pm.sample()`? Even though we're calling the latter implicitly

I don't see a problem with that

ColtAllen:
> I would go with `fit_MAP()` just because our 95% default will be sampling and `.fit()` I think is what people would expect as the main fitting method.

This gets my vote as well. When in doubt, go with the `scikit-learn` conventions.

I do wonder if MAP warrants its own separate fitting method, or if it could be specified as a parameter in `.fit()` instead.

twiecki:
> I do wonder if MAP warrants its own separate fitting method, or if it could be specified as a parameter in .fit() instead.

The only thing I dislike about this approach is that the return type will be different depending on the input argument.

ricardoV94:
We can convert the map result to an Xarray so it works seamlessly with other methods that expect a posterior array. It just had 0 chains and samples

ColtAllen:
> In conversation with @ColtAllen I realized the reason `find_MAP` gives worse results than lifetimes is (might be?) that lifetimes is often doing penalized MLE. We should investigate if we can add better priors so that `find_MAP` is more competitive with the old lifetimes results in terms of stability/speed.
> 
> This is important as we will probably want to provide users access to MAP/MLE fits.

[...]

PR #104, opened 2022-12-09
Datasets Module

ColtAllen:
This PR mostly copies over the datasets module from `lifetimes`, providing loading functions and benchmarking datasets for continuous, noncontractual, and discrete, noncontractual modeling. Pytest fixtures are also added for use in testing.

These will be useful for notebook examples, and particularly testing new models because these same datasets are used in research. Being able to reproduce research results is the closest thing we have to a ground truth assessment for model performance. 

Four datasets are added:

- CDNOW_master.txt
- cdnow_summary.csv
- CDNOW_sample.txt
- donations.csv

'CDNOW_master.txt' is a larger file at 1.9MB and currently does not have a loading function. It is added only for completeness, and if testing on a larger dataset is desired.

'cdnow_summary.csv' contains customer data in RFM summary format, the required format for most BTYD/CLV models. (I'm still not sure if CLV is the most appropriate name in this library).

'CDNOW_sample.txt`' contains raw transactions of customer data. Some models only accept data in this format, but this can also be used for testing a utility function aggregating data in RFM format - an excellent addition for a future PR.

'donations.csv' is for discrete, noncontractual modeling. This data aggregates customers with identical characteristics by count, so it was necessary to add some code to convert it back into one customer per row.

Lastly, I've added some AzureML files to .gitignore because I sometimes work in AzureML, which has an annoying habit of auto-adding its own proprietary temp and ignore files which are invisible until pushed to a remote repo.


juanitorduz:
Shall we add a reference to the `lifetimes` package and / or data sources?

lucianopaz (changes requested):
Thanks so much @ColtAllen! There are two main problems with the current PR. The first one is that the loading functions are all implicitly looking for the files in the current working directory, and as such, they will likely fail when being called in the wild. I left a comment on how to address this problem using the `__file__` magic attribute.
The second problem is that the txt files need to be included in a manifest.in file for them to be shipped along with the pymmmc package. If the manifest isn’t there, then the data files will be missing in any install that didn’t clone the repo, and do a development pip installation. You can read the official documentation on this topic [here](https://packaging.python.org/en/latest/guides/using-manifest-in/). If you need help setting up a manifest, let me know.


The last thing that I wanted to say is that this PR assumes that we **do want to ship datasets with the package**. Do we want to do that? Or should they be left outside the package and downloaded upon request via some helper function? This isn’t completely trivial but it’s doable, and very typical, and we’ve done it with pymc-examples and other places, so the approach can be reproduced.

pymmmc/datasets/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    ----------
+    **kwargs
+        Additional keyword arguments passed into pandas.read_csv function.
+
+    Returns
+    -------
+    DataFrame
+    """
+
+    return pd.read_csv("cdnow_summary.csv", **kwargs)
```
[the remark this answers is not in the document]
ColtAllen: Whoops 😆 I've added some `resource_filename` calls to these functions which should achieve the same end.

pymmmc/datasets/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    **kwargs
+        Additional keyword arguments passed into pandas.read_csv function.
+
+    Returns
+    -------
+    DataFrame
+    """
+
+    return pd.read_csv(
+        "CDNOW_sample.txt",
```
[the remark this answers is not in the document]
ColtAllen: See above

pymmmc/datasets/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    ----------
+    **kwargs
+        Additional keyword arguments passed into pandas.read_csv function.
+
+    Returns
+    -------
+    DataFrame
+    """
+
+    donations = pd.read_csv("donations.csv", **kwargs)
```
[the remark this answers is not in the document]
ColtAllen: See above

pymmmc/tests/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+    """Create a test dataset from the CDNOW RFM summary for use in multiple tests."""
+
+    return cdnow_summary()
+
+
+@pytest.fixture(scope="module")
+def cdnow_trans() -> pd.DataFrame:
+    """Create a test dataset of CDNOW transactions for use in multiple tests."""
+
+    return cdnow_transactions()
```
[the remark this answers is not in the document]
ColtAllen: I thought about this, and felt that such tests would be redundant when these fixtures are used, which will probably be the case in https://github.com/pymc-labs/pymc-marketing/pull/67. However, if 100% code coverage is desired, it's an easy addition.

ColtAllen:
> If you need help setting up a manifest, let me know.

I could use some help with this. It's interesting though - my [btyd](https://github.com/ColtAllen/btyd) library shipped these same files without a manifest or dev pip install. I wonder if it's because I used poetry.

> The last thing that I wanted to say is that this PR assumes that we **do want to ship datasets with the package**. Or should they be left outside the package and downloaded upon request via some helper function?

Total memory for these datasets is about 2.2 MB, 1.9 of which is CDNOW_master.txt, which is honestly optional. I added it because some of these models would benefit from a larger dataset. They'll still converge with CDNOW_sample.txt, but testing tolerances will need to be widened.

These files are available online, but speaking from professional experience, some companies have firewall rules set up that will block outgoing HTTP calls. If the file size is acceptable, I think shipping these with the package is preferred.

ColtAllen:
> Shall we add a reference to the `lifetimes` package and / or data sources?

I think so, for the data sources. Should we add this reference in the function docstrings, or in the documentation somewhere? Because it'd be a good idea to include a link to [Bruce Hardie's website](https://www.brucehardie.com/) in the docs, as it's a treasure trove of information for these models.

pymmmc/datasets/__init__.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 import pandas as pd
+from pkg_resources import resource_filename
```
lucianopaz, line 3: Oh! I'd never seen `pkg_resources` before, I had to google it up. From its documentation it looks like its use is discouraged
image

I found a [migration guide](https://importlib-resources.readthedocs.io/en/latest/migration.html#migration-guide) that says that you can replace the `resource_filename` with this:

```python
ref = importlib_resources.files('my.package') / 'resource.dat'
with importlib_resources.as_file(ref) as path:
    # Do something with the open path
```

pymmmc/datasets/__init__.py
```diff
@@ -24,7 +25,9 @@ def cdnow_summary(**kwargs) -> pd.DataFrame:
     DataFrame
     """
 
-    return pd.read_csv("cdnow_summary.csv", **kwargs)
+    return pd.read_csv(
+        resource_filename("pymmmc", "datasets/cdnow_summary.csv"), **kwargs
+    )
```
lucianopaz, line 30: As per my previous comment about not using `pkg_resources`, the recommended way to do this is:
```suggestion
ref = importlib_resources.files("pymmmc") / "datasets" / "cdnow_summary.csv"
with importlib_resources.as_file(ref) as path:
        return pd.read_csv(path, **kwargs)
```

pymmmc/datasets/__init__.py
```diff
@@ -44,7 +47,7 @@ def cdnow_transactions(**kwargs) -> pd.DataFrame:
     """
 
     return pd.read_csv(
-        "CDNOW_sample.txt",
+        resource_filename("pymmmc", "datasets/CDNOW_sample.txt"),
```
lucianopaz, line 50: This should be adapted to use a similar approach to the above suggestion

pymmmc/datasets/__init__.py
```diff
@@ -68,7 +71,9 @@ def donations(**kwargs) -> pd.DataFrame:
     DataFrame
     """
 
-    donations = pd.read_csv("donations.csv", **kwargs)
+    donations = pd.read_csv(
+        resource_filename("pymmmc", "datasets/donations.csv"), **kwargs
```
lucianopaz, line 75: This should be adapted to use a similar approach to the above suggestion

lucianopaz (changes requested):
Thanks again @ColtAllen. I didn't know about the `pkg_resources` package, but it looks like using it is discouraged! I managed to find a migration guide to replace it and posted a suggestion.

Regarding the MANIFEST.in file, it looks like poetry took care of adding figuring out the package data for you! If you don't want to rely on poetry, you can do one of two things according to the [docs](https://packaging.python.org/en/latest/guides/using-manifest-in/)

1. Add a `package_data` or `data_files` keyword argument to the `setup` call in `setup.py`, listing all the datasets
2. Add a file called MANIFEST.in. The file should have a line that reads `graft pymmmc/datasets/`

Either approach is fine

pymmmc/datasets/__init__.py
ColtAllen: Thanks for pointing this out. The changes have been made.

ColtAllen:
I've removed the CDNOW_master.txt file, so the total size of the added files is only about 300k now.  Datasets have also been added to `setup.py`, 'importlib_resources' is being used, and library name change to `pymc_marketing` has been reflected.

Been a while since I've worked on this though, and have introduced some merge conflicts. I'll dig into these more tomorrow.

ColtAllen:
After considering https://github.com/pymc-labs/pymc-marketing/issues/99, I've decided to close this PR for now. These datasets are far smaller than what is usually encountered in practice, so reproducing the research results derived from MLE will require unrealistically informative priors and wide testing tolerances. However, I've made an effort to determine these priors for the BG/NBD, Gamma-Gamma, and Modified BG/NBD models in my [BTYD library](https://github.com/ColtAllen/btyd), which could be useful in the future for optimizing choice of priors.

Proceeding with testing new models on synthetic data via the CLV distributions module is my preferred approach, but [the distributions need to be model-specific](https://github.com/pymc-labs/pymc-marketing/issues/128). Those distribution objects will also be quite valuable for inference against observed data, so resolving https://github.com/pymc-labs/pymc-marketing/issues/128  will kill two birds with one stone.

PR #105, opened 2022-12-10
Refactor validation and preprocessing to rely on mixin classes

### LINKED ISSUE #70: Instantiate `pymc.Model` after data validation and preprocessing in MMM base class
The `MMM` base class instantiates an empty `pymc.Model` during its `__init__`. It does this before having gone through data preprocessing and validation. We should move the model instantiation later in the `__init__`, closer to the call to `_build_model`

### LINKED ISSUE #71: MMM `adstock_max_lag` and `control_data` are model specific
If we intend our `MMM` class to be a suitable generic base for many media mix models, we should have as generic arguments to `__init__` as possible. `adstock_max_lag` is completely specific to a particular way to apply a convolution operation on a vector, and `control_data` might be handled differently by different `MMM` subclasses (e.g. it could handle continuous controls differently than categorical controls). We should remove as much model specific arguments from the `__init__` signature, and leave the rest as `kwargs` that are forwarded into `_build_model`

lucianopaz:
Closes #70
Closes #71

This PR was based off of a comment from @juanitorduz, where he suggested to use mixin classes to implement the validation and preprocessing steps of the models. After playing a bit with the idea, I liked it. I think that it makes things more modular and easier to customize in other kinds of models.

### DIFF
--- a/pymmmc/mmm/preprocessing.py
+++ b/pymmmc/mmm/preprocessing.py
@@ -1,0 +1,58 @@
+from typing import Callable
+
+import pandas as pd
+from sklearn.pipeline import Pipeline
+from sklearn.preprocessing import MaxAbsScaler, MinMaxScaler, StandardScaler
+
+__all__ = [
+    "preprocessing_method",
+    "MixMaxScaleTarget",
+    "MaxAbsScaleChannels",
+    "StandardizeControls",
+]
+
+
+def preprocessing_method(method: Callable) -> Callable:
+    if not hasattr(method, "_tags"):
+        method._tags = {}
+    method._tags["preprocessing"] = True
+    return method
+
+
+class MixMaxScaleTarget:
+    @preprocessing_method
+    def min_max_scale_target_data(self, data_df: pd.DataFrame) -> pd.DataFrame:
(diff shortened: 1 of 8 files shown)


juanitorduz (approved):
This looks super nice @lucianopaz! Just left one comment regarding categorical variables which we can tackle in a next iteration :)

pymmmc/mmm/preprocessing.py
```diff
@@ -0,0 +1,58 @@
...
+        transformers = [("scaler", MaxAbsScaler())]
+        pipeline: Pipeline = Pipeline(steps=transformers)
+        self.channel_transformer: Pipeline = pipeline.fit(X=channel_data.to_numpy())
+        data_df[self.channel_columns] = self.channel_transformer.transform(
+            channel_data.to_numpy()
+        )
+        return data_df
+
+
+class StandardizeControls:
```
juanitorduz, line 48: At the moment this works for continuous controls and ordinal numeric encoded right? Maybe we want to specify this in the name as will probably want to add a default one-hot-encoding for categorical variables?

lucianopaz: I’ll create an issue to keep track of that

PR #106, opened 2022-12-11
Fix readthedocs yaml file

lucianopaz:
Should fix failing action from #105


OriolAbril:
I used the old way of setting the python version, sorry. It should be like https://github.com/arviz-devs/xarray-einstats/blob/1e032006d32c145a832d1fb25d2e5d9e48ecdafb/.readthedocs.yaml#L5

juanitorduz (approved):

PR #110, opened 2022-12-12
Ignore E501 Globaly (fix lint)

juanitorduz:
fix lint

### DIFF
--- a/pymmmc/distributions.py
+++ b/pymmmc/distributions.py
@@ -94,7 +94,7 @@ class ContNonContract(PositiveContinuous):
 
     ========  ===============================================
     Support   :math:`t_j > 0` for :math:`j = 1, \dots, x`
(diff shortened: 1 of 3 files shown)


lucianopaz:
I think that it’s better to ignore E501 throughout the entire code base and just rely on black to automatically reformat code relevant lines. Docstrings and comments shouldn’t have any bearing on the line length lint test

OriolAbril:
I think I could also update that to use multi-line math and not need to ignore the line length check, not sure what option do you prefer

juanitorduz:
I do not have any strong opinion. Maybe we want our doc-strings to be quite complete with formulas and explanations where we might not want to be limited by this lint checker. Again, no strong opinion :)

lucianopaz:
> I think I could also update that to use multi-line math and not need to ignore the line length check, not sure what option do you prefer

I think that the best is to disable that flag globally, but you could also use multi line math in the docstrings

juanitorduz:
so shall we merge as it is :) ?

lucianopaz (approved):

PR #114, opened 2022-12-14
Implement vectorized adstock transformations

### LINKED ISSUE #22: Adstock transformation without `for` loop
We would like to write a more efficient implementation of the adstock transformations so that we do not use a for loop. An attempt with `scan` was (unsuccessfully) implemented in https://github.com/pymc-labs/pymmmc/pull/15 

**Requirements:**
- We should be able to add the `l_max` parameter to truncate the size of the effect. 
- Be vectorised
- Should be faster than the current implementation.

lucianopaz:
Closes #22 

- [X] Convert the old adstock transformations to use the vectorized convolution
- [x] Write tests for the vectorized convolution with multiple dimensions and broadcasting patterns
- [x] Maybe add an assert to prevent segfaults due to passing a incorrect axis?

### DIFF
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -1,7 +1,60 @@
 import pytensor.tensor as pt
 
+from pymc_marketing.mmm.utils import params_broadcast_shapes
 
-def geometric_adstock(x, alpha: float = 0.0, l_max: int = 12, normalize: bool = False):
+
+def batched_convolution(x, w, axis: int = 0):
+    """Apply a 1D convolution in a vectorized way across multiple batch dimensions.
+
+    Parameters
+    ----------
+    x :
+        The array to convolve.
+    w :
+        The weight of the convolution. The last axis of ``w`` determines the number of steps
+        to use in the convolution.
+    axis : int
+        The axis of ``x`` along witch to apply the convolution
+
+    Returns
+    -------
+    y :
+        The result of convolving ``x`` with ``w`` along the desired axis. The shape of the
+        result will match the shape of ``x`` up to broadcasting with ``w``. The convolved
+        axis will show the results of left padding zeros to ``x`` while applying the
+        convolutions.
+    """
+    orig_ndim = x.ndim
+    axis = axis if axis >= 0 else orig_ndim + axis
+    w = pt.as_tensor(w)
+    x = pt.moveaxis(x, axis, -1)
+    try:
+        l_max = w.shape[-1].eval()
+    except Exception:
+        l_max = None
+    x_shape, w_shape = params_broadcast_shapes([x.shape, w.shape], [1, 1])
+    x = pt.broadcast_to(x, x_shape)
+    w = pt.broadcast_to(w, w_shape)
+    x_time = x.shape[-1]
+    shape = (*x.shape, w.shape[-1])
+    padded_x = pt.zeros(shape, dtype=x.dtype)
+    if l_max is not None:
+        for i in range(l_max):
+            padded_x = pt.set_subtensor(
+                padded_x[..., i:x_time, i], x[..., : x_time - i]
+            )
+    else:  # pragma: no cover
+        raise NotImplementedError(
+            "At the moment, convolving with weight arrays that don't have a concrete shape "
+            "at compile time."
+        )
+    conv = pt.sum(padded_x * w[..., None, :], axis=-1)
+    return pt.moveaxis(conv, -1, axis + conv.ndim - orig_ndim)
+
+
+def geometric_adstock(
+    x, alpha: float = 0.0, l_max: int = 12, normalize: bool = False, axis: int = 0
+):
     """Geometric adstock transformation.
 
     Adstock with geometric decay assumes advertising effect peaks at the same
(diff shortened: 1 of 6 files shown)


pymmmc/transformers.py
```diff
@@ -31,29 +82,28 @@ def geometric_adstock(x, alpha: float = 0.0, l_max: int = 12, normalize: bool =
...
-        at.concatenate(tensor_list=[at.zeros(shape=x.shape)[:i], x[: x.shape[0] - i]])
-        for i in range(l_max)
-    ]
-    x_cycle = at.stack(cycles)
-    x_cycle = at.transpose(x=x_cycle, axes=[1, 2, 0])
-    w = at.as_tensor_variable([at.power(alpha, i) for i in range(l_max)])
-    w = at.transpose(w)[None, ...]
-    w = w / at.sum(w, axis=2, keepdims=True) if normalize else w
-    return at.sum(at.mul(x_cycle, w), axis=2)
+    warnings.warn(
```
twiecki, line 92: Just delete it, we haven't released yet.

pymmmc/transformers.py
```diff
@@ -1,7 +1,58 @@
...
+        The result of convolving ``x`` with ``w`` along the desired axis. The shape of the
+        result will match the shape of ``x`` up to broadcasting with ``w``. The convolved
+        axis will show the results of left padding zeros to ``x`` while applying the
+        convolutions.
+    """
+    orig_ndim = x.ndim
+    axis = axis if axis >= 0 else orig_ndim + axis
+    w = at.as_tensor(w)
+    x = at.moveaxis(x, axis, -1)
+    x_shape, w_shape = at.random.utils.params_broadcast_shapes(
```
lucianopaz, line 31: @ricardoV94, is there another way to implement this shape broadcasting?

- If I’m not explicit about it, broadcasting of symbolic `x` and `w` fails and takes the shape of `x` as the correct one (maybe symbolic isn’t the correct term, I mean tensors that have `None` in their shape).
- When I am explicit but the arrays fail to broadcast at run time, I get segfaults (probably due to the C backend trying to read or write to memory locations that aren’t part of the array)

drbenvincent:
Could I recommend:
a) placing an assert at the start of the function to ensure sizes of the data and parameters are aligned
b) is it worth doing some timing tests, even if informal. I'm finding that inference is slowing down significantly when going from [4, T] sized inputs (ie 4 time series) to [4, 150, T] (ie 4 time series each of which has 150 geographical regions.

pymmmc/transformers.py
lucianopaz: From a closer look at [the source](https://github.com/pymc-devs/pytensor/blob/main/pytensor/tensor/random/utils.py#L55) I think that there should be a couple of assert Ops there.

drbenvincent:
Feature request: At the moment there's the assumption that if our data is shape `[4, 150, T]`, then we have parameters of shape `[4, 150]`. This makes sense in the case that we want independent parameters for each predictor*region combination, whether we place a hierarchy on that or not. But, this is likely to significantly slow down inference. 

Would it be possible to have the option (or an alternative function) which allows you to just have different priors over predictors which is applied over all regions. So in this example, that would be data of size `[4, 150, T]`, but parameters of size `[4]`

lucianopaz:
> Feature request: At the moment there's the assumption that if our data is shape `[4, 150, T]`, then we have parameters of shape `[4, 150]`. This makes sense in the case that we want independent parameters for each predictor*region combination, whether we place a hierarchy on that or not. But, this is likely to significantly slow down inference.
> 
> Would it be possible to have the option (or an alternative function) which allows you to just have different priors over predictors which is applied over all regions. So in this example, that would be data of size `[4, 150, T]`, but parameters of size `[4]`

This is already supported, just pass the parameters with shape `(4, 1)` and the same parameter value will be used for every entry in the second axis of `x`. The logic behind this is that the T dimension of `x` ids ignored, and the rest of the axis are broadcasted following the same rules that numpy does. That’s why you got a failure to broadcast when the parameter had shape `(4,)`, because 4 and 150 don’t broadcast together

pymc_marketing/mmm/transformers.py
```diff
@@ -31,29 +82,29 @@ def geometric_adstock(x, alpha: float = 0.0, l_max: int = 12, normalize: bool =
...
-        pt.concatenate(tensor_list=[pt.zeros(shape=x.shape)[:i], x[: x.shape[0] - i]])
-        for i in range(l_max)
-    ]
-    x_cycle = pt.stack(cycles)
-    x_cycle = pt.transpose(x=x_cycle, axes=[1, 2, 0])
-    w = pt.as_tensor_variable([pt.power(alpha, i) for i in range(l_max)])
-    w = pt.transpose(w)[None, ...]
-    w = w / pt.sum(w, axis=2, keepdims=True) if normalize else w
-    return pt.sum(pt.mul(x_cycle, w), axis=2)
+    warnings.warn(
```
twiecki, line 93: lets just remove this, were still pre-release

lucianopaz: Done

lucianopaz:
@ricardoV94, I'd like to ask for your input on the `params_broadcast_shapes` function that I added here. As I mentioned in some older comments, the `pytensor` version is not robust to broadcasting failures (shapes that cannot broadcast together are taken to broadcast just fine and can cause segfaults down the line). This version might be suitable for `pytensor` itself, if you think that it's OK. My main question is whether the approach that I followed for getting "concrete shapes" (shapes that are not `TensorVariable`s themselves) is fine, or if there is something that I might have missed.

lucianopaz:
I opened pymc-devs/pytensor#152, related to the problem I found with `params_broadcast_shapes`. If the fix from this PR looks ok, I can also apply it over on `pytensor`

lucianopaz:
Pinging @ricardoV94 to have a look. I updated the `params_shape_broadcast` to rely on the tensors `broadcastable` attribute. If this isn't desirable, I can undo the last commit.

ricardoV94:
You can't set the broadcastable flag manually for the time being (it's a read only property, except for the Unbroadcast Op which was kept just for some edge cases) so I am not sure it's worth copying the method here. When we reintroduce static broadcasting the Pytensor method will be adapted accordingly.

juanitorduz:
Hey! Any blockers for this one? Anything I could do to support ?

ricardoV94:
Would be good to see if we can remove the custom `params_broadcast_shapes` and use the one in PyTensor, now that we fixed `BroadcastTo`

ricardoV94:
Closed in favor of #221

PR #119, opened 2022-12-23
Make channel_contributions contain a channel axis and exploit that in plots

### LINKED ISSUE #118: MMMs should provide per channel/control contribution variables
At the moment, the MMM class [sums the contributions across channels](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/delayed_saturated_mmm.py#L98) and [across controls](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/delayed_saturated_mmm.py#L115). This makes it harder to plot contributions per channel/control since the curves have to be recomputed almost from scratch.
A better approach is to have `MMM`s guarantee that there will be a deterministic variable that tracks contributions across channels and time. This variable need not be limited to only 2 dimensions, but could also include region or brand, but the API should guarantee to have at least channels and time in it. That way, we can make plotting functions that are region/brand agnostic, and then reduce all potentially irrelevant things.

lucianopaz:
Closes #118 

This PR does two things.
1. It adds a `channel` and `control` dimension to the `channel_contributions` and `control_contributions` respectively. These deterministics are then used in the `BaseMMM` plotting methods.
2. The plotting functionality tests were moved to the `mmm.test_base.py` script and were made to run a proxy fitted toy MMM.

### DIFF
--- a/pymc_marketing/mmm/base.py
+++ b/pymc_marketing/mmm/base.py
@@ -12,12 +12,13 @@ import arviz as az
 import matplotlib.pyplot as plt
 import matplotlib.ticker as mtick
 import numpy as np
-import numpy.typing as npt
 import pandas as pd
 import pymc as pm
 import seaborn as sns
 from pymc.util import RandomState
-from xarray import DataArray, Dataset
+from sklearn.pipeline import Pipeline
+from sklearn.preprocessing import FunctionTransformer
+from xarray import DataArray
 
 from pymc_marketing.mmm.validating import (
     ValidateChannelColumns,
@@ -51,7 +52,6 @@ class BaseMMM:
             self.validate(self.data_df)
         self.preprocessed_data = self.preprocess(self.data_df.copy())
 
-        self.model: pm.Model = pm.Model()
         self.build_model(
             data_df=self.preprocessed_data,
             **kwargs,
@@ -88,6 +88,13 @@ class BaseMMM:
             if getattr(method, "_tags", {}).get("preprocessing", False)
         ]
 
+    def get_target_transformer(self) -> Pipeline:
+        try:
+            return self.target_transformer
+        except AttributeError:
+            identity_transformer = FunctionTransformer()
+            return Pipeline(steps=[("scaler", identity_transformer)])
+
     def validate(self, data_df: pd.DataFrame):
         for method in self.validation_methods:
             method(self, data_df)
@@ -102,11 +109,14 @@ class BaseMMM:
         raise NotImplementedError()
 
     def get_prior_predictive_data(self, *args, **kwargs) -> az.InferenceData:
-        with self.model:
-            prior_predictive: az.InferenceData = pm.sample_prior_predictive(
-                *args, **kwargs
-            )
-        return prior_predictive
+        try:
+            return self._prior_predictive
+        except AttributeError:
+            with self.model:
+                self.prior_predictive: az.InferenceData = pm.sample_prior_predictive(
+                    *args, **kwargs
+                )
+            return self._prior_predictive
 
     def fit(
         self,
@@ -190,10 +200,10 @@ class BaseMMM:
         )["likelihood"]
 
         if original_scale:
-            likelihood_hdi_94 = self.target_transformer.inverse_transform(
+            likelihood_hdi_94 = self.get_target_transformer().inverse_transform(
                 Xt=likelihood_hdi_94
             )
-            likelihood_hdi_50 = self.target_transformer.inverse_transform(
+            likelihood_hdi_50 = self.get_target_transformer().inverse_transform(
                 Xt=likelihood_hdi_50
             )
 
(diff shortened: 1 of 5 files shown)


lucianopaz:
This PR should also fix the flaky tests that popped up in #114

pymc_marketing/mmm/base.py
```diff
@@ -231,48 +242,86 @@ def plot_posterior_predictive(
...
-            y=az.extract(self.fit_result, var_names=["intercept"]).mean().item(),
+            y=intercept.mean().data,
             color=f"C{i + 1}",
             ax=ax,
         )
         ax.fill_between(
             x=self.data_df[self.date_column],
             y1=intercept_hdi[:, 0],
-            y2=intercept_hdi[:, 1],
+            y2=intercept_hdi[:, 0],
```
juanitorduz, line 324: seems `y1` is the same as `y2` now? maybe this is not intentional?

lucianopaz: Oops, that was a typo

pymc_marketing/mmm/base.py
```diff
@@ -88,6 +88,14 @@ def preprocessing_methods(self) -> List[Callable[[pd.DataFrame], pd.DataFrame]]:
             if getattr(method, "_tags", {}).get("preprocessing", False)
         ]
 
+    def get_target_transformer(self):
+        try:
+            return self.target_transformer
+        except AttributeError:
+            null_scaler = MaxAbsScaler()
+            null_scaler.fit([[0], [1]])
```
juanitorduz, line 96: Is this to create a "trivial" scaler?

lucianopaz: Yes. I wanted a scaler that did nothing. Is there a better way?

tests/mmm/test_base.py
```diff
@@ -30,6 +33,123 @@
...
+        def build_model(self, data_df, **kwargs):
+            pass
+
+    mmm = ToyMMM(
+        toy_df,
+        target_column="y",
+        date_column="date",
+        channel_columns=["channel_1", "channel_2"],
+    )
+    mmm.control_columns = ["control_1", "control_2"]
```
juanitorduz, line 48: Can we also test when there are no control variables ?

pymc_marketing/mmm/base.py
juanitorduz: I think this is ok! I wonder if is an overkill just adding a `IdentityTransformer` just for the sake of readability...

juanitorduz:
The changes are quite nice! specially regarding the better use of `xarray` API. I once the plotting tests also handle the "con conrtol variables" case  we are good to go :)

lucianopaz:
I discovered that the failing tests were due to a bug in `matplotlib` that was caused by a change introduced to `numpy` 1.24. The [matplotlib bug](https://github.com/matplotlib/matplotlib/issues/24106) was fixed in version 3.6.1, but seaborn still hasn't made a release that is compatible with that version. At the moment, the only thing that we can do to avoid this bug is to downgrade numpy in the requirements. I'll add an issue that keeps track of this, so that when seaborn cuts a release, we can unpin numpy and everything else.

tests/mmm/test_base.py
lucianopaz: Done

pymc_marketing/mmm/base.py
lucianopaz: I don't see any `IdentityTransformer` in the sklearn documentation. From what I saw, it looks like a [`FunctionTransformer`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html#sklearn.preprocessing.FunctionTransformer) with `func=None` would do the job.

juanitorduz: As you wish :) No strong opinions here.

juanitorduz:
> I discovered that the failing tests were due to a bug in `matplotlib` that was caused by a change introduced to `numpy` 1.24. The [matplotlib bug](https://github.com/matplotlib/matplotlib/issues/24106) was fixed in version 3.6.1, but seaborn still hasn't made a release that is compatible with that version. At the moment, the only thing that we can do to avoid this bug is to downgrade numpy in the requirements. I'll add an issue that keeps track of this, so that when seaborn cuts a release, we can unpin numpy and everything else.

Thanks for documenting this!

pymc_marketing/mmm/base.py
```diff
@@ -88,6 +88,13 @@ def preprocessing_methods(self) -> List[Callable[[pd.DataFrame], pd.DataFrame]]:
             if getattr(method, "_tags", {}).get("preprocessing", False)
         ]
 
+    def get_target_transformer(self):
```
juanitorduz, line 91: add return type hint?

juanitorduz (approved):
LGTM 👍 (just a small type hint comment)

pymc_marketing/mmm/base.py
lucianopaz: Done

PR #125, opened 2023-01-05
Add utility to compute customer lifetime value

### LINKED ISSUE #20: LTV computation
After having an implementation of https://github.com/pymc-labs/pymmmc/issues/18 and https://github.com/pymc-labs/pymmmc/issues/19 we would like to combine them to get an LTV estimation at user level following the approach in the `lifetimes` package. Se should be able to reproduce all the results of the documentation https://lifetimes.readthedocs.io/en/latest/Quickstart.html

### LINKED ISSUE #96: Distinguish between distributional and expectation methods
For some CLV models we provide methods that return an expectation (with some degree of uncertainty emanating from the posterior parameters), whereas for others we provide a [full distribution](https://github.com/pymc-labs/pymmmc/blob/fe114062dae21362f469560130ae78465f48a84f/pymmmc/clv/models/gamma_gamma.py#L62-L68):

We should reflect this distinction clearly in the method names/docstrings. It's always possible to obtain an expectation from the histogram, by taking the mean.

ricardoV94:
Closes #20 
Closes #96


pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -67,8 +68,35 @@ def expected_customer_spend(
...
-        """Predict expected transaction value of new unobserved customers"""
+    def expected_customer_spend(
+        self,
+        customer_id: Union[np.ndarray, pd.Series],
+        mean_transaction_value: Union[np.ndarray, pd.Series],
+        number_transactions: Union[np.ndarray, pd.Series],
+    ) -> xarray.DataArray:
+        """Expected transaction value per customer
+
+        Adapted from: https://github.com/CamDavidsonPilon/lifetimes/blob/aae339c5437ec31717309ba0ec394427e19753c4/lifetimes/fitters/gamma_gamma_fitter.py#L117  # noqa: E501
```
juanitorduz, line 79: I suggest we also add the paper reference as in the `lifetimes` docstrings. Something like
```
Equation (5) from:
        http://www.brucehardie.com/notes/025/
```

juanitorduz (commented):
I think this look nice (I have not checked the test failing). I left some suggestions to improve the docstrngs.

pymc_marketing/clv/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    time: int = 12,
+    discount_rate: float = 0.01,
+    freq: str = "D",
+) -> xarray.DataArray:
+    """
+    Compute the average lifetime value for a group of one or more customers.
+    This method computes the average lifetime value for a group of one or more customers.
+    It also applies Discounted Cash Flow.
+
+    Adapted from lifetimes package
```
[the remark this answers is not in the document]
ricardoV94: Added in the main util

pymc_marketing/clv/utils.py
```diff
@@ -0,0 +1,104 @@
...
+def customer_lifetime_value(
+    transaction_model,
+    customer_id: Union[pd.Series, np.ndarray],
+    frequency: Union[pd.Series, np.ndarray],
+    recency: Union[pd.Series, np.ndarray],
+    T: Union[pd.Series, np.ndarray],
+    monetary_value: Union[pd.Series, np.ndarray, xarray.DataArray],
+    time: int = 12,
+    discount_rate: float = 0.01,
+    freq: str = "D",
```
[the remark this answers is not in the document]
ricardoV94: Thanks, it was missing from lifetimes as well. Added

pymc_marketing/clv/utils.py
```diff
@@ -0,0 +1,118 @@
+from typing import Union
+
+import numpy as np
+import pandas as pd
+import xarray
+
+
+def to_xarray(customer_id, *arrays, dim: str = "customer_id"):
```
ColtAllen, line 8: I'm curious why this is no longer an internal method for the `CLVModel` base class?

pymc_marketing/clv/__init__.py
```diff
@@ -1,12 +1,13 @@
...
     GammaGammaModel,
     GammaGammaModelIndividual,
 )
+from pymc_marketing.clv.utils import customer_lifetime_value
 
 __all__ = (
     "BetaGeoModel",
     "GammaGammaModel",
     "GammaGammaModelIndividual",
+    "customer_lifetime_value",
```
ColtAllen, line 12: Is it intended for users to call this function via `GammaGammaModel`, or as a separate utility function? In [btyd](https://github.com/ColtAllen/btyd/blob/main/btyd/models/gamma_gamma_model.py) I eliminated this redundancy and made it an internal method of the model.

pymc_marketing/clv/utils.py
ricardoV94: It was an internal method of the `BetaGeoModel` only

ColtAllen: My mistake. Do you think this would be more appropriate as an internal method?

pymc_marketing/clv/__init__.py
ColtAllen: > The `GammaGamma` model provides a wrapper method as well. This is more general in that you could theoretically use a model other than the GammaGamma to estimate the monetary value.

I like the flexibility of this approach. Google has a guide on building a DL variant of the Gamma-Gamma model:

https://cloud.google.com/ai-platform/docs/clv-prediction-with-offline-training-train

ColtAllen: > We can also make it just part of the CLV models that estimate monetary_value, maybe as a Mixin

I definitely want to explore continuous-time CLV calculations in a future PR, as we've discussed before:

https://github.com/ColtAllen/btyd/issues/76

The calculation itself is specific to the model predicting expected transactions (such as the BG/NBD or Pareto/NBD), because the expected future transactions themselves are discounted to a present value. However, it'd be easy enough to callback this function via the Gamma-Gamma model.

ricardoV94:
I need an approving review :)

twiecki (approved):

PR #130, opened 2023-01-23
Add `fit_MAP` to CLV models

### LINKED ISSUE #99: Consider using "penalized" priors by default in CLV models

ricardoV94:
Closes #99


pyproject.toml
```diff
@@ -6,7 +6,6 @@ addopts = [
     "-v",
     "--strict-markers",
     "--strict-config",
-    "--cov=mmm",
-    "--cov=pymmmc",
+    "--cov=pymc_marketing",
```
ricardoV94, line 9: CC @juanitorduz

ricardoV94: Not sure if it should be `pymc-marketing` either :P

pymc_marketing/clv/models/basic.py
```diff
@@ -12,10 +14,27 @@ def __init__(self):
             pass
 
     def fit(self, *args, **kwargs):
+        """Draw samples from model posterior using MCMC sampling"""
         with self.model:
             self._fit_result = pm.sample(*args, **kwargs)
         return self._fit_result
 
+    def fit_MAP(self, **kwargs):
+        """Find model maximum a posterior using scipy optimizer"""
```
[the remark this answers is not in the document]
ricardoV94: Yes it uses `scipy` under the hood. Thanks for flagging the typo

ColtAllen:
Apart from the docstring typo, this looks good! We floated the idea of rolling both `pm.sample()` and `pm.find_MAP()` into a single `fit()` method, but these return `InferenceData` objects that are quite different, so it's probably best to keep them separated.

ricardoV94:
> Apart from the docstring typo, this looks good! We floated the idea of rolling both `pm.sample()` and `pm.find_MAP()` into a single `fit()` method, but these return `InferenceData` objects that are quite different, so it's probably best to keep them separated.

They return very similar objects, so that the summary statistic methods work the same. We could use the single `fit` method indeed, with a keyword argument.

```
model.fit(method=["mcmc"| "map"|"vi"])
```

Do people like that more? I think I like it more.

ColtAllen:
> They return very similar objects, so that the summary statistic methods work the same. We could use the single `fit` method with a keyword argument:
> 
> ```
> model.fit(method=["mcmc"|"map"|"vi"])
> ```
> 
> Do people like that more? I think I like it more.

If the return objects are functionally equivalent, then this gets my vote.

I only briefly tested ADVI when I started working on BTYD, but the posteriors it generated did not impress me.

zwelitunyiswa:
Bambi has an "inference_method" parameter in its .fit function. In mcmc, and vi, it also allows "nuts_numpyro" and "nuts_blackjax". Will the .fit function here allow sampling using jax too?

ricardoV94:
The PyMC `sample` function will soon allow to choose one of those variations of mcmc via a kwarg so that should be available automatically here:

https://github.com/pymc-devs/pymc/pull/6422/files

zwelitunyiswa:
@ricardoV94 very cool. Thanks for the clarification.

ricardoV94:
Updated to use a keyword in the user facing `fit` method

PR #132, opened 2023-01-24
Release automation

lucianopaz:
This PR adds two things.

1. It adds a release.yml that serves as a template for release notes. It establishes the same labeling convention that is used in PyMC.
2. It adds a workflow that is only triggered by publishing a release. This will first create the sdist and bdist for the package, test that installing them provides the package **with the same version as the one set in the commit tag**, then attempts to upload to test.pypi (it ignores already existing uploads which could have happened if the action needs to be re-run for whatever reason), and if successful, finally uploads to pypi.


ricardoV94 (approved):
Let's try it out ;)

lucianopaz:
We need an admin to set the pypi token secrets for the package btw

PR #138, opened 2023-01-31
MMM Notebook: fix z values which should be x

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/137

Improve mmm example notation.

PR #141, opened 2023-02-03
fix lint 🙈 

juanitorduz:
When I rebased and merged https://github.com/pymc-labs/pymc-marketing/pull/138 the lint problem persists. I hope this fix it :)

### DIFF
--- a/pymc_marketing/clv/distributions.py
+++ b/pymc_marketing/clv/distributions.py
@@ -20,7 +20,6 @@ class ContNonContractRV(RandomVariable):
     _print_name = ("ContNonContract", "\\operatorname{ContNonContract}")
 
     def make_node(self, rng, size, dtype, lam, p, T, T0):
-
         T = pt.as_tensor_variable(T)
         T0 = pt.as_tensor_variable(T0)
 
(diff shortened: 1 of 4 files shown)


ricardoV94:
Is your pylint more strict than ours/CI? Mine is not deleting single empty lines. Maybe it's a new version?

juanitorduz:
We are not using `pylint` right? In the make file:
```
check_lint:
	pip install -r lint-requirements.txt
	flake8 .
	isort --check-only .
	black --diff --check --fast .
```
😄

ricardoV94:
I meant whatever lint was in pre-commit, just defaulted to thinking it was pylint

juanitorduz:
I created a conda env from scratch and it seems to be running in the ci/cd as well. One thing I have noticed is that we are not specifying the revisions (versions) of the lint-packages.

ricardoV94:
Yeah I suspect it's a new behavior

ricardoV94 (approved):

ricardoV94:
We should treat the CI as the ground truth though, otherwise there will be inconsistencies soon enough again

juanitorduz:
I will create an issue

PR #143, opened 2023-02-03
Updates for release

ricardoV94:
Update setup.py details

Changes in dependencies:
* Removes numpyro/jax dependencies which are incompatible with Windows
* Adds explicit pandas and xarray dependencies which are used throughout the codebase


juanitorduz (approved):

juanitorduz:
It would be nice to be able to use the jax samplers when available. I'll open an issue.

PR #146, opened 2023-02-08
Make more extensive usage of the `pypoject.toml` configuration file

tomicapretto:
Lots of information was moved to `pyproject.toml`. This means we remove the following files

* requirements.txt
* lint-requirements.txt
* docs-requirements.txt
* test-requirements.txt
* setup.cfg
* setup.py
* setupegg.py

I want to comment on some changes

* There's a new file `.flake8` there. See the comments in there. TLDR: Flake8 configuration is not compatible with pyproject.toml. I think leaving it under setup.cfg could be a little misleading because setup.cfg can have non-flake8 configuration as well.

The `pyproject.toml` file has the following lines

```python
dynamic = ["version"]
# ...
[tool.setuptools.dynamic]
version = {file = "pymc_marketing/version.txt"}
```

which is basically saying the version should be dynamically determined and indicates where to read the version from.  And that's why I added `pymc_marketing/version.txt` as well as `pymc_marketing/version.py`. The downside is there's a "weird" `version.txt` file, the upside is there's a unique source of truth for the version. 

I considered two other possibilities, but they're worse for the reasons that follow

* Have `__version__` in the `__init__.py` file: This will make `pymc-marketing` (and all its dependencies) part of the `requires` entry in the `build-system` entry. I tried this with Bambi and its a mess, build times increase a lot.
* Hard-code the version in both `__init__.py` and `pyproject.toml`. This is the solution in CausalPy now. But I think we don't want to have two sources of truth (and two files to update).


lucianopaz (approved):
Looks good to me @tomicapretto! The only thing that I noticed was that the sdist includes the tests directory, while the wheel doesn't. Is this intended? I think that the tests don't hurt much, but if this was an accident, we definitely need to control that the docs don't end up in sdists, or the package size will grow too large.

tomicapretto:
> Looks good to me @tomicapretto! The only thing that I noticed was that the sdist includes the tests directory, while the wheel doesn't. Is this intended? I think that the tests don't hurt much, but if this was an accident, we definitely need to control that the docs don't end up in sdists, or the package size will grow too large.

Thanks for the review! 

Including the `tests` is not intended. But I'm not sure how to check that either. I'm checking the `/dist/pymc-marketing*.tar.gz` and I don't see `tests`. Am I checking the right thing?

tomicapretto:
I do think they could have been included before, with the now old `setup.py` file since it uses `find_packages()` from `setuptools`. See

```python
>>> from setuptools import find_packages
>>> find_packages()
```

```
['tests', 'pymc_marketing', 'tests.mmm', 'tests.clv', 'tests.clv.models', 'pymc_marketing.mmm', 'pymc_marketing.clv', 'pymc_marketing.clv.models']
```

But the `pyproject.toml` file does not include the `tests` directory

lucianopaz:
> > Looks good to me @tomicapretto! The only thing that I noticed was that the sdist includes the tests directory, while the wheel doesn't. Is this intended? I think that the tests don't hurt much, but if this was an accident, we definitely need to control that the docs don't end up in sdists, or the package size will grow too large.
> 
> Thanks for the review!
> 
> Including the `tests` is not intended. But I'm not sure how to check that either. I'm checking the `/dist/pymc-marketing*.tar.gz` and I don't see `tests`. Am I checking the right thing?

Weird, it might be that my version of `build` is old… nevermind my comment then

pyproject.toml
```diff
@@ -1,3 +1,70 @@
...
+    "pytest==7.0.1",
+    "pytest-cov==3.0.0"
+]
+
+[tool.setuptools]
+packages = [
+    "pymc_marketing", 
+    "pymc_marketing.mmm", 
+    "pymc_marketing.clv", 
+    "pymc_marketing.clv.models"
```
ricardoV94, line 55: @tomicapretto why was this line `clv.models` needed?

tomicapretto: So it's possible to do `from pymc_marketing.clv.models import ...`. 

See this build that failed https://readthedocs.org/projects/pymc-marketing/builds/19423628/

```
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/146/lib/python3.10/site-packages/sphinx/config.py", line 350, in eval_config_file
    exec(code, namespace)
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/checkouts/146/docs/source/conf.py", line 5, in <module>
    import pymc_marketing  # isort:skip
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/146/lib/python3.10/site-packages/pymc_marketing/__init__.py", line 1, in <module>
    from pymc_marketing import clv, mmm
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/146/lib/python3.10/site-packages/pymc_marketing/clv/__init__.py", line 1, in <module>
    from pymc_marketing.clv.models import (
ModuleNotFoundError: No module named 'pymc_marketing.clv.models'
```

PR #147, opened 2023-02-09
Fix PyPI release workflow and test on pull/push

ricardoV94:


.github/workflows/pipy.yml
```diff
@@ -60,12 +66,12 @@ jobs:
       - name: Test pip install from test.pypi
         run: |
           python -m venv venv-test-pypi
-          venv-test-pypi/bin/python -m pip install --index-url https://test.pypi.org/simple/ pymc-marketing
+          venv-test-pypi/bin/python -m pip install --index-url https://test.pypi.org/simple/pymc-marketing
```
lucianopaz, line 69: This is wrong. The index url shouldn’t have pymc-marketing in it. The index url just says that the package should be looked up in test pypi, and pymc-marketing is just the package that will be installed

lucianopaz (changes requested):

.github/workflows/pipy.yml
ricardoV94: Ah I get it now

lucianopaz (approved):
Lgtm

PR #151, opened 2023-02-14
Delete unused sklearn import in MMM Example NB

cluhmann:
Super small PR to omit an sklearn import that was unused in the MMM example notebook.  However, the change means that this particular notebook no longer depends on sklearn.  Let me know if I a missing something.


juanitorduz (approved):

ricardoV94:
Do we still use sklearn elsewhere or can we remove it as a dependency?

PR #155, opened 2023-02-15
Fix headers MMM Notebook

juanitorduz:
The notebook had many first-level headers 🙈


ricardoV94:
While we are at it, I noticed a typo in the second sentence: " of the MMM presented in thw paper".

ricardoV94 (approved):

PR #160, opened 2023-02-16
Include CLV Quickstart in Readme

ricardoV94:
Using the latest so that the links work already


twiecki (approved):

ricardoV94:
Maybe stable is better, I will just re-release so the docs show the new notebook already

Issue #126, opened 2023-01-10
Add `clv_summary` Utility Function


ColtAllen:
When modeling entire customer populations, the `clv` models require `recency`, `frequency`, and `T` summary variables for each customer. For the Gamma-Gamma model, average spend beyond the first transaction is also required. Formatting input data in this fashion is non-trivial, so a utility function would be highly convenient for end users.

The `summary_data_from_transaction_data()` function from Lifetimes can be brought over for this, but will likely require revision because it currently ignores `SettingwithCopyWarning` in Pandas. The name of the Lifetimes function also seems verbose to me, but if `clv_summary()` is not descriptive enough, we can discuss alternatives.

ricardoV94:
`clv_summary` sounds fine

Issue #100, opened 2022-12-08
Add `BetaGeoModel` example and corresponding tests using the CDNOW dataset


larryshamalama:

ColtAllen:
I think it'd be worthwhile to add a dataset utility function for CDNOW. Practically every research paper I've read on continuous, non-contractual models use CDNOW for benchmarking. Reproducing research results is the best way to proceed with testing in my mind, and there are results based on CDNOW available for many other models.

I can create a PR for this.

ColtAllen:
I've decided against this approach of reproducing research for model testing. My reasoning is provided in the closing comment of https://github.com/pymc-labs/pymc-marketing/pull/104. If there are no objections I'll close this issue.

ColtAllen:
I've added a `pytest` fixture for the CDNOW dataset in https://github.com/pymc-labs/pymc-marketing/pull/164 that can be used for this.

Issue #142, opened 2023-02-03
Pin lint versions in Ci/CD and pre-commit


juanitorduz:
We want to have a deterministic way to run the lint locally and on ht CI/CD so that we avoid PR like https://github.com/pymc-labs/pymc-marketing/pull/141).

wd60622:
Can this be closed @juanitorduz ?

Issue #127, opened 2023-01-10
Add Pareto/NBD Model


ColtAllen:
Now that the [Gaussian Hypergeometric Function](https://github.com/pymc-devs/pytensor/pull/90) is available in PyTensor (thanks @ricardoV94!) the way is paved for adding the Pareto/NBD model. This model was the first `CLV` model introduced and is still one of the most performant.

The Lifetimes implementation contains a numerical error in the likelihood function (which has given me grief when using this model in practice). The fix is described in the very last line of this paper.

http://brucehardie.com/notes/027/bgnbd_num_error.pdf

This model also has an additional predictive method for predicting the probability of a customer being alive X days into the future. To my knowledge, no other CLV models provide this, and can only predict the probability of being alive in the current moment. This method is not available in Lifetimes, but I did add to my [BTYD fork](https://github.com/ColtAllen/btyd/blob/main/btyd/fitters/pareto_nbd_fitter.py#L583).

zwelitunyiswa:
I am excited about the function to predict the probability of a customer being alive X days into the future. CLVtool in R has the PNB but not this function, which I always found a little odd.

ColtAllen:
@ricardoV94 brought this up in https://github.com/pymc-labs/pymc-marketing/pull/131

> The bad news is that NUTS seems completely useless to sample the cdnow dataset (even when taking draws from the RNG directly and with more data). It could be that this set of parameters is specifically bad. I had to disable jittering in the init method, otherwise the sampler would start at some point where the gradients would fail to converge immediately. Even without jittering, it was taking way too long to sample, suggesting very high tree depth or still too many iterations in the gradient methods :/
> 
> I had to sample with `Slice`. The `find_MAP` is also quite slow (maybe 30s-1min to fit)
> 
> Perhaps more informative priors would have helped, but I didn't bother to try.
> 
> If the bottleneck is really the speed of the hyp2F1 derivatives (which I suspect it is) we can try to: 1) compute the multiple derivatives together, and 2) jit it with Numba

CDNOW is a small dataset, so more informative priors would have helped for sure. However, are there any other suggestions or things for me to keep in mind as I proceed with adding this model? The `hyp2F1` derivatives seem like more of a `pytensor` issue.

ricardoV94:
I did try with a sample of 10k using the CDNOW fit as the true parameters and NUTS still couldn't sample. Is that still very small data? I don't think more data will help as the problem was not due to flat posteriors.

[...]

PR #164, opened 2023-02-19
Add `clv_summary` Utility Function

ColtAllen:
This PR resolves https://github.com/pymc-labs/pymc-marketing/issues/126 and adapts `summary_data_from_transaction_data()` from Lifetimes for summarizing raw transaction data. This is the primary data preprocessing step for CLV models. It also adds a `pytest` fixture for the CDNOW dataset that can be applied towards https://github.com/pymc-labs/pymc-marketing/issues/100.

`summary_data_from_transaction_data()` had two bugs which I've fixed in this PR:  A Pandas `SettingwithCopyWarning` and the output dataframe index. Both contributed to unstable behavior when used in Lifetimes models, and resetting the index also provides a Customer ID column. I've also removed some redundant datatype checks, a useless function parameter, and renamed two parameters and edited the docstrings for clarity.

Note that `clv_summary()` calls an internal `_find_first_transactions()` function. These two functions could be consolidated, but `_find_first_transactions()` is used elsewhere in Lifetimes for estimating the cumulative transactions of an individual customer. Until we determine if that functionality is worth adding, I want to keep these two functions separate for now.

Two minor questions I have before merging this:

1. Should this function be importable from the root `clv` module like the models:
```
from pymc_marketing.clv import clv_summary
```
or left in the submodule:
```
from pymc_marketing.clv.utils import clv_summary
```
2. I've always considered many of the function and parameter names in lifetimes to be unnecessarily verbose. For user convenience, should some of the parameter names be shortened? For instance, change `customer_id_col` to `id_col`?


ColtAllen:
Just fixed a lint issue that `pre-commit` didn't catch. Could this be related to https://github.com/pymc-labs/pymc-marketing/issues/142?

conftest.py
```diff
@@ -19,3 +20,17 @@ def pytest_collection_modifyitems(config, items):
...
             item.add_marker(skip_slow)
+
+
+@pytest.fixture(scope="module")
+def cdnow_data() -> pd.DataFrame:
+    """
+    Load CDNOW_sample.txt into a Pandas dataframe for model and utility function testing.
+    """
+    df = pd.read_csv(
+        "tests/clv/datasets/CDNOW_sample.txt",
```
ricardoV94, line 31: Why not save it as a `csv` directly instead of a `txt`?

ricardoV94 (commented):
I think `costumer_id_col` is good just because we are calling it `customer_id` in the CLV models?

About the scope of the function, do as you think is better :)

conftest.py
ColtAllen: Consistency with the source data, mainly. I've modified the fixture to reflect the fact this is transaction data, and also provide a link to the original source file:

https://www.brucehardie.com/datasets/

ColtAllen: Also, is there a reason why `conftest.py` is in the root directory instead of the testing folder?

ricardoV94: I think it's in the root directory because we run pytest from there? Is that unusual?

ricardoV94: I think a CLV is much more readable, is there a license problem that doesn't allow us to add the columns?

ricardoV94 (approved):
I still think a csv would be better, but it's not a blocker on my part.

ColtAllen:
> About the scope of the function, do as you think is better :)

Importing from the root `clv` module is slightly more convenient, but I see some upstream changes in the `__init__.py` module, so I'll just add it later when I work on the [ParetoNBD](https://github.com/pymc-labs/pymc-marketing/issues/127) PR.

> I think it's in the root directory because we run pytest from there? Is that unusual?

I usually see `conftest.py` in the test folder. I can move it there if you want to tidy up the root directory, but functionally there's no difference.

> I think a CLV is much more readable, is there a license problem that doesn't allow us to add the columns?

No licensing issues; just subjective preference.

I'll wait a few more days for feedback before merging. If popular demand prefers `.csv` over `.txt` I'll change the filetype.

twiecki:
Let's stick to csv as we use it in the package elsewhere already.

ColtAllen:
It's a CSV now; I'll go ahead and merge.

PR #179, opened 2023-02-28
Relax matplotlib dependency

### LINKED ISSUE #120: Must unpin numpy 1.24 once seaborn cuts a release greater than 0.12.1
`numpy` version 1.24 changed some signature regarding masked arrays. This change, unearthed a [bug in matplotlib](https://github.com/matplotlib/matplotlib/issues/24106), which was fixed with version 3.6.1. The problem is that seaborn still does not support the newer version of matplotlib with the fix. This means that we need to pin numpy to stay below 1.24 until seaborn cuts a new release. This issue is here to remind us to unpin numpy once seaborn cuts a release greater than 0.12.1

juanitorduz:
Shall we relax matplotlib's dependency? I do not remember wey we have a specific one.

Closes #120

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,12 +13,12 @@ maintainers = [{ name = "PyMC Labs", email = "info@pymc-labs.io" }]
 
 dependencies = [
     "arviz>=0.13.0",
-    "matplotlib>=3.5.1",
+    "matplotlib>=3.5.1,!=3.6.1",
     "numpy>=1.17,!=1.24.0",
     "pandas",
     "pymc>=5.0.0",
     "scikit-learn>=1.1.1",
-    "seaborn>=0.12.1",
+    "seaborn>=0.12.2",
     "xarray"
 ]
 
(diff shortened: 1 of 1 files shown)


ricardoV94:
Could it be because of https://github.com/pymc-labs/pymc-marketing/issues/120 ?

juanitorduz:
> Could it be because of #120 ?

Seems so! Seaborn has cut a release and 0.12.2 is available now

juanitorduz:
In https://github.com/pymc-labs/pymc-marketing/pull/179/commits/43626f419c9cf2bbfe07fa3639a3f85c3c14965c I tried adding consistent restrictions as in https://github.com/mwaskom/seaborn/blob/master/pyproject.toml#L25-#L29. Still, let me know if there is a safer way of proceeding.

ricardoV94:
Seaborn itself seems to take care of the invalid numpy and matplotlib dependencies? 

https://github.com/mwaskom/seaborn/blob/55a328ba4301f429aa454f5824861323caf91cd0/pyproject.toml#L26-L28

Would it be enough to mark seaborn >= 0.12.2 (and remove the specific `!=` matplotlib and numpy ones)?

juanitorduz:
Changed in https://github.com/pymc-labs/pymc-marketing/pull/179/commits/fa1e047a72bf196d2a4a7441d171da66a6f9eab6

ricardoV94 (approved):

Issue #176, opened 2023-02-23
Add `BetaGeoBetaBinomModel`


ColtAllen:
The BG/BB model is to my knowledge the only one out there for Discrete Time, Non-Contractual modeling. However, a lot of people don't realize this model is for a different use case than the others in `lifetimes` (myself included when I first started using the library). Maybe rename the model class to `DiscreteNonContractModel` for clarity instead?

https://www.brucehardie.com/papers/020/fader_et_al_mksc_10.pdf

Equation (14) on page 6 of the research paper is an expression for discounted present value of all expected future transactions, which is an interesting alternative to the current `customer_lifetime_value` implementation.

juanitorduz:
Great idea!

alexwilde01:
@ColtAllen thanks for surfacing this one! Do you have a sense of prioritization on BG/BB & extensions against other enhancements?

ColtAllen:
@alexwilde01 I'm currently working on https://github.com/pymc-labs/pymc-marketing/pull/177 and targeting to have it done by 11-Mar. It has some changes which are relevant to BG/BB, so once that's done I'll start on this next.

ColtAllen:
https://github.com/pymc-labs/pymc-marketing/pull/177 has some additional complexities to work out, and [the aspects impacting BG/BB](https://github.com/pymc-labs/pymc-marketing/issues/182) can be done in a separate PR. I'll start on this tomorrow.

ColtAllen:
Expressions 15, 16 & 17 on pages 6-7 of the paper provide marginal posterior distributions for the transaction and churn rates, and product moment (ie, correlation), respectively. The correlation expression is certainly worth adding, but what about the marginal posteriors?

wd60622:
@ColtAllen is this good to close?

ColtAllen:
Yes (sorry, I thought this would auto-close when the PR was merged).

Issue #182, opened 2023-03-01
Keep `_process_priors()` and `_unload_params()` in `CLVModel`


ColtAllen:
Currently, `_process_priors()` and `_unload_params()` are being defined in every CLV model, but the only differences are the specific prior/param names. If we add both methods to the base `CLVModel` as generic loop operators, and a list attribute in the child model classes containing the param names, it would eliminate a lot of redundant code when adding new models. https://github.com/pymc-labs/pymc-marketing/pull/133 partially resolves `_process_priors()`.

Also, in each child model class every predictive method calls `_unload_params()` and `clv.utils.to_xarray()`. It may be worthwhile to write a decorator in `CLVModel` for predictive methods that handles this automatically, but I need to look into this more.

ColtAllen:
I started working on this, but I'm not sure if these code changes can accommodate covariates, so I'll return to this after progress has been made on https://github.com/pymc-labs/pymc-marketing/issues/134.

ColtAllen:
I'm currently adding support for covariates to `ParetoNBDModel`, which  creates nuances to `_unload_params` that are not agnostic across all models. `_process_priors` is also in the base class now, so I'm closing this.

PR #177, opened 2023-02-24
Add ParetoNBDModel

ColtAllen:
This PR closes https://github.com/pymc-labs/pymc-marketing/issues/127.

### Outstanding Issues

- [x] `mypy` CI errors
- [x] https://github.com/ColtAllen/pymc-marketing/blob/clv_pareto_nbd/pymc_marketing/clv/models/pareto_nbd.py#L172
- [x] https://github.com/ColtAllen/pymc-marketing/blob/clv_pareto_nbd/pymc_marketing/clv/models/pareto_nbd.py#L526
- [x] Consensus on predictive method names
- [x] docs build errors
- [x] Revise docstring references

### UML Diagram
classes

### Important Changes

I've marked this model as experimental because NUTS is having convergence errors due to `pytensor.Tensor.hyp2f1` in the model `logp`. @ricardoV94 has been trying out some improvements for that function.

I renamed the predictive methods in this model to be more concise:
`expected_purchases`
`expected_purchases_new_customer`
`expected_probability_alive`
`expected_purchase_probability`

However, these are not consistent with the equivalent method names in `BetaGeoModel`. These names need to be standardized for plotting functionality, so it's important we're in agreement on the naming conventions.

`xarray-einstats>=0.5.1` is now a required library dependency.

Customer input data arrays are also assigned to class attributes as default predictive arguments in this model, making life easier for end users unless they want to run predictions on new customers.

I also added some additional utility methods which can be used in other CLV models if moved to `CLVModel`, and would go towards resolving https://github.com/pymc-labs/pymc-marketing/issues/182.

Unlike the other CLV models, `ParetoNBDModel` uses a dedicated `ParetoNBD` distribution block for the `logp`. My rationale for this is that the distribution class will append an `observed` column for customer frequency and recency inputs to the returned `arviz.InferenceData` object, and enable additional diagnostic plots like the Bayesian p-value and PPC plots once those methods are added:

https://python.arviz.org/en/stable/examples/index.html#Model%20Checking

I've documented some future additions to this model in https://github.com/pymc-labs/pymc-marketing/issues/183 since this PR already has enough to review.

Issue #134, opened 2023-01-25
Add Time-Invariant Covariates to CLV Models


ColtAllen:
Both the Beta-Geo/NBD and Pareto/NBD models can be extended rather easily to support time-invariant covariates:

https://www.brucehardie.com/notes/019/time_invariant_covariates.pdf

juanitorduz:
In https://juanitorduz.github.io/bg_nbd_pymc/ I tried to do so :)

ColtAllen:
Instead of adding separate covariate models, I think it would save us a lot of redundancy to simply extend the existing `BetaGeoModel` and upcoming [`ParetoNBDModel`](https://github.com/pymc-labs/pymc-marketing/issues/135). The only additional code required seems to just be 2-3 more priors in the `__init__` method and input parameters for the covariates. Everything else is identical and these extra priors could be omitted by a conditional if the covariates are unspecified.

> In https://juanitorduz.github.io/bg_nbd_pymc/ I tried to do so :)

You've added to this since I last read it! You mentioned the BG/NBD covariate model may benefit from an alternative formulation for the likelihood (which is provided in the PDF link) but I had a lot of success with hierarchical pooling of hyperpriors in the BG/NBD implementation I did for `btyd`:

https://github.com/ColtAllen/btyd/blob/main/btyd/models/beta_geo_model.py#L97

That may be worth trying before doing any `logp` modifications, because `btyd` also has a `BetaGeoCovarsFitter()` that uses the same likelihood as `BetaGeoFitter` and gives identical results when `np.ones` are provided for the covariates.

juanitorduz:
Hey! thanks for the feedback! I will definitely take a look at your work on hierarchical models! Your comments make a lot of sense and might be in the right direction! My initial approach was a more "brute force approach".

navish92:
Hi @ColtAllen 
Thanks for the amazing contributions to this & BTYD package. 
Curious if time-invariant covariates for BG/NBD was added or is in the pipeline?

ColtAllen:
Thanks @navish92,

Covariate support for `BetaGeoModel` will be added after https://github.com/pymc-labs/pymc-marketing/issues/527 is resolved. If you have any feedback on that issue I'd be interested, because end-user convenience is a significant factor of it.

In the meantime, do not discount `ParetoNBDModel`, because although it is slower to fit than `BetaGeoModel`, it has greater functionality. If `probability_alive` is important to your use case, it also has superior performance.

[...]

PR #188, opened 2023-03-05
Add `BetaGeoBetaBinomModel`

ColtAllen:
This PR closes https://github.com/pymc-labs/pymc-marketing/issues/176

The `BetaGeoBetaBinomModel` is the only model I'm aware of for discrete, non-contractual modeling. This PR adds a model class as well as a testing dataset and distribution block.

For now this is a draft PR because many of the tests are blocked by incorrect `logp` calculations; for reference please see (5) on p4 of the research paper:
https://www.brucehardie.com/papers/020/fader_et_al_mksc_10.pdf

I don't think I'm using `pytensor.scan()` properly in the `logp` expression, so if someone could help me out with that I'd appreciate.

Some other concerns:

1. I added `purchase_heterogeniety` and `churn_heterogeniety` variables for plotting the population purchase and churn rate posterior distributions, but since they won't be called until after the model is fit, they are failing `flake8`.
2. Docstrings still require revision. How prudent is it to add the LaTeX math equations? I'm not sure how useful it is for users of this library, and a weblink to the research paper referencing the same expressions is already provided.


pymc_marketing/clv/models/beta_geo_beta_binom.py
```diff
@@ -0,0 +1,364 @@
...
+    details.
+
+    Parameters
+    ----------
+    customer_id: array_like
+        Customer labels. Must not repeat.
+    frequency: array_like
+        The number of purchases of customers.
+    recency: array_like
+        The time of the last, i.e. xth, purchase.
```
larryshamalama, line 39: We can perhaps make a point that `recency` is discrete, i.e. `int`-like.

pymc_marketing/clv/models/beta_geo_beta_binom.py
```diff
@@ -0,0 +1,364 @@
...
+
+    References
+    ----------
+    .. [1] Fader, Peter S., Bruce G.S. Hardie, and Jen Shang (2010),
+       "Customer-Base Analysis in a Discrete-Time Noncontractual Setting,"
+       Marketing Science, 29 (6), 1086-1108. https://www.brucehardie.com/papers/020/fader_et_al_mksc_10.pdf
+    """
+
+    _model_name = "BG/BB"  # Beta-Geometric Beta-Binomial Model
+    _params = ["alpha", "beta", "gamma", "delta"]
```
larryshamalama, line 91: This addition is related to #128 IIRC. Is that correct? I believe that the existing models don't have this `_params` structure but it could be a nice way of closing #128 with additional methods to the base `CLVModel` class. Was what you had in mind?

larryshamalama (commented):
Some minor comments to begin with!

pymc_marketing/clv/models/beta_geo_beta_binom.py
ColtAllen: Yeah that docstring still requires revision. Could be a good idea to add a validation check for this as well.

pymc_marketing/clv/models/beta_geo_beta_binom.py
ColtAllen: https://github.com/pymc-labs/pymc-marketing/issues/182 you mean? I wanted to get draft PRs for this and https://github.com/pymc-labs/pymc-marketing/pull/177 up before opening a PR for that one. After considering https://github.com/pymc-labs/pymc-marketing/pull/197, some major test restructurings could be underway, so I think I'll start work on https://github.com/pymc-labs/pymc-marketing/issues/182 next week.

In the meantime, I could use some help with `logp` in this PR.

pymc_marketing/clv/distributions.py
```diff
@@ -2,14 +2,11 @@
...
+from pymc.distributions.dist_math import betaln, check_parameters
+from pytensor import scan
 from pytensor.tensor.random.op import RandomVariable
 
-__all__ = [
-    "ContContract",
-    "ContNonContract",
-    "ParetoNBD",
-]
+__all__ = ["ContContract", "ContNonContract", "ParetoNBD", "BetaGeoBetaBinom"]
```
ricardoV94, line 9: If you add a comma after the last item black will let it stay one item per line, which I think is better for git diffs

pymc_marketing/clv/distributions.py
```diff
@@ -483,3 +480,184 @@ def logp(value, r, alpha, s, beta, T):
...
+        def _B_iter(ix):
+            return betaln(alpha + x, beta + t_x - x + ix) + betaln(
+                gamma + 1, delta + t_x + ix
+            )
+
+        iter_result, _ = scan(
+            fn=_B_iter,
+            # outputs_info=pt.zeros(1),
+            sequences=pt.arange(t_recent),
+        )
```
ricardoV94, line 634: If I understand what you're trying to do, it should be something like:

```suggestion
        def _B_iter(ix, alpha, beta, t_x, x, gamma, delta):
            return betaln(alpha + x, beta + t_x - x + ix) + betaln(
                gamma + 1, delta + t_x + ix
            )

        iter_result, _ = scan(
            fn=_B_iter,
            outputs_info=[None],
            sequences=pt.arange(t_recent),
            non_sequences=[alpha, beta, t_x, x, gamma, delta],
        )
```
`outputs_info=[None]` tells Pytensor it expects a single non-recurrent output
`non_sequences` passes the constants explicitly to the inner function, which is a bit more clean imho

ricardoV94: Note that your scan can be replaced by relying on broadcasting.

```suggestion
        n_dims_expr = max([alpha.ndim, beta.ndim, x.ndim, t_x.ndim, gamma.ndim, delta.ndim, t_recent.ndim])
        extra_dims_needed = (n_dims_expr - t_recent.ndim) + 1
        ixs = pt.shape_padright(pt.arange(t_recent), extra_dims_needed)
        iter_result = (
          betaln(alpha + x, beta + t_x - x + ixs)
          + betaln(gamma + 1, delta + t_x + ixs)
        )
```

pymc_marketing/clv/models/beta_geo_beta_binom.py
```diff
@@ -0,0 +1,364 @@
...
+            delta = pm.Deterministic("delta", (1.0 - churn_pool) * churn_hyper)
+
+            # TODO: This is for plotting the population posterior, but will fail flake8
+            purchase_heterogeniety = pm.Beta(
+                "purchase_heterogeneity", alpha=alpha, beta=beta
+            )
+            # TODO: This is for plotting the population posterior, but will fail flake8
+            churn_heterogeniety = pm.Beta(
+                "churn_heterogeneity", alpha=gamma, beta=delta
+            )
```
ricardoV94, line 163: If these are only needed later you can create them in a dummy model and call `posterior_predictive`. This may be better because NUTS does not need to worry about these variables while sampling the ones of interest.

This is what's done for instance here: https://github.com/pymc-labs/pymc-marketing/blob/9579e1fda9077090fcaeb7d59789095915440449/pymc_marketing/clv/models/gamma_gamma.py#L51

We could have included `nu` in the model that's sampled in `fit` but it's easier for NUTS if we don't. 

When you call `posterior_predictive` with variables that are missing from the trace those will be sampled conditioned on everything else that's on the trace, leading to the same result for variables that don't contribute to the likelihood like these.

pymc_marketing/clv/models/beta_geo_beta_binom.py
larryshamalama: Oops yes #182, sorry. And sounds good.

I see that the `logp` utilizes `scan`, which I'm not super familiar with. I see that @ricardoV94 is giving some good pointers below

pymc_marketing/clv/models/beta_geo_beta_binom.py
ColtAllen: I'll save this for a future PR involving the plotting module.

ColtAllen:
Closing this one for now. Will resume work and open another PR after https://github.com/pymc-labs/pymc-marketing/issues/134, https://github.com/pymc-labs/pymc-marketing/pull/298, and https://github.com/pymc-labs/pymc-marketing/issues/182 are resolved.

PR #197, opened 2023-03-09
Speedup BetaGeoModel tests

ricardoV94:
This PR removes the costly fitting from the setup class. Even running a small test like `test_model_repr` would take around one minute as it would have to wait for the useless fitting (in the context of that test) to be done.

Also simplified the patching of the lifetimes reference model. It's enough to override `params_` after initializing it.

Finally added a `pytest.mark.slow` for the `test_model_convergence` so it doesn't run by default (it does run in our CI here). This way the tests take just ~2 seconds on my slow laptop.

### DIFF
--- a/tests/clv/models/test_beta_geo.py
+++ b/tests/clv/models/test_beta_geo.py
@@ -85,14 +85,15 @@ class TestBetaGeoModel:
 
     @pytest.mark.slow
     @pytest.mark.parametrize(
-        "N, rtol",
+        "N, fit_method, rtol",
         [
-            (500, 0.3),
-            (2000, 0.1),
-            (10000, 0.055),
+            (500, "mcmc", 0.3),
+            (2000, "mcmc", 0.1),
+            (10000, "mcmc", 0.055),
+            (2000, "map", 0.1),
         ],
     )
-    def test_model_convergence(self, N, rtol):
+    def test_model_convergence(self, N, fit_method, rtol):
         rng = np.random.default_rng(146)
         recency, frequency, _, T = self.generate_data(
             self.a_true, self.b_true, self.alpha_true, self.r_true, N, rng=rng
@@ -105,7 +106,8 @@ class TestBetaGeoModel:
             recency=recency,
             T=T,
         )
-        model.fit(chains=1, progressbar=False, random_seed=rng)
+        sample_kwargs = dict(random_seed=rng, chains=2) if fit_method == "mcmc" else {}
+        model.fit(fit_method=fit_method, progressbar=False, **sample_kwargs)
         fit = model.fit_result.posterior
         np.testing.assert_allclose(
             [fit["a"].mean(), fit["b"].mean(), fit["alpha"].mean(), fit["r"].mean()],
--- a/tests/clv/models/test_beta_geo.py
+++ b/tests/clv/models/test_beta_geo.py
@@ -1,89 +1,43 @@
+import arviz as az
 import numpy as np
 import pymc as pm
 import pytest
-
-# shorter name for black code style formatter
-from lifetimes.fitters.beta_geo_fitter import BetaGeoFitter as BGF
+from lifetimes.fitters.beta_geo_fitter import BetaGeoFitter
 
 from pymc_marketing.clv.distributions import continuous_contractual
 from pymc_marketing.clv.models.beta_geo import BetaGeoModel
 
 
 class TestBetaGeoModel:
-    # Hyperparameters
-    a_true = 0.8
-    b_true = 2.5
-    alpha_true = 3
-    r_true = 4
-    rng = np.random.default_rng(34)
-
-    N = 2000
-
-    @classmethod
-    def generate_data(cls, N):
+    @staticmethod
+    def generate_data(a, b, alpha, r, N, rng):
         # Subject level parameters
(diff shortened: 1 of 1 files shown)


ColtAllen:
To speed this up further, should the model fitting in `setup_class()` be done with MAP? `test_model_convergence()` is already covering a MCMC fit.

tests/clv/models/test_beta_geo.py
```diff
@@ -147,7 +105,7 @@ def test_model_convergence(self, N, rtol):
             recency=recency,
             T=T,
         )
-        model.fit(chains=1, progressbar=False, random_seed=self.rng)
+        model.fit(chains=1, progressbar=False, random_seed=rng)
```
ColtAllen, line 108: Should a MAP fit also be added here for the CI?

tests/clv/models/test_beta_geo.py
```diff
@@ -160,63 +118,133 @@ def test_expected_probability_alive(self):
...
         assert est_prob_alive.dims == ("chain", "draw", "customer_id")
 
         np.testing.assert_allclose(
-            true_prob_alive.mean(),
+            alive.mean(),
             est_prob_alive.mean(),
             rtol=0.05,
         )
 
     def test_expected_num_purchases(self):
```
ColtAllen, line 166: What's the justification for combining these two tests? These are two separate methods with different input and output dimensions.

ColtAllen (commented):
`BetaGeoModel` is being redefined in every test. Would it make more sense to define it only once in `setup_class()` as was the case previously? Same with `lifetimes.BetaGeoFitter` as it could be compared against in `test_expected_probability_alive`.  This would also simplify required changes if we decide to implement https://github.com/pymc-labs/pymc-marketing/issues/100.

tests/clv/models/test_beta_geo.py
```diff
@@ -160,63 +118,133 @@ def test_expected_probability_alive(self):
...
-        true_prob_alive = self.alive.mean()  # scalar
-        est_prob_alive = self.fixed_model.expected_probability_alive(
-            self.customer_id,
-            self.frequency,
-            self.recency,
-            self.T,
+        rng = np.random.default_rng(152)
+
+        N = 100
+        # Almost deterministic p = .02
```
larryshamalama, line 124: What does an almost deterministic dropout probability here check?

tests/clv/models/test_beta_geo.py
```diff
@@ -160,63 +118,133 @@ def test_expected_probability_alive(self):
...
+
+        bg_model = BetaGeoModel(
+            customer_id=customer_id,
+            frequency=frequency,
+            recency=recency,
+            T=T,
+            a_prior=pm.Normal.dist(a, 1e-3),
+            b_prior=pm.Normal.dist(b, 1e-3),
+            alpha_prior=pm.Normal.dist(alpha, 1e-3),
+            r_prior=pm.Normal.dist(r, 1e-3),
```
larryshamalama, line 141: Why the informative priors here?

larryshamalama (commented):
Looks great, thanks @ricardoV94! The initial solution of having the model fit without the `pytest.mark.slow` was a little naive and due for a speed up. I just added some questions regarding some parameter choices in the tests.

tests/clv/models/test_beta_geo.py
ricardoV94: That the probability of dropout is stable at a value that should induce a nice `p_alive` ~ 0.5, nothing too low nor too high

tests/clv/models/test_beta_geo.py
ricardoV94: For a fake posterior based on prior_predictive sampling around the true values

tests/clv/models/test_beta_geo.py
ricardoV94: They both require building the same type of values, patching the pymc-makerting and lifetimes model, and they are both tested against the same lifetimes method. There was a TODO before suggesting they be merged already, probably for similar reasoning

ricardoV94:
> `BetaGeoModel` is being redefined in every test. Would it make more sense to define it only once in `setup_class()` as was the case previously? Same with `lifetimes.BetaGeoFitter` as it could be compared against in `test_expected_probability_alive`. This would also simplify required changes if we decide to implement #100.

No, I think we should create it every time, because we usually monkey-patch the fit result and we don't want to be operating on the same stateful object (i.e., erasing a previous fit).

It was a conscious choice not to create a class-level model in the other model tests, like the gamma-gamma.

tests/clv/models/test_beta_geo.py
ricardoV94: Added a comment

tests/clv/models/test_beta_geo.py
ricardoV94: Replaced by direct sampling with numpy when creating the fake fit

tests/clv/models/test_beta_geo.py
ricardoV94: Done

ricardoV94:
I set this as draft as it includes the commit from #198 

Otherwise I think I addressed your review comments @larryshamalama and @ColtAllen. Let me know if there's anything else.

larryshamalama (approved):
Looks good to me :)

PR #199, opened 2023-03-10
Make docs footer smaller and more responsive

### LINKED ISSUE #191: readthedocs: make the footer not massive
Not sure if this is intentional, but the footer is _massive_ and obscures the core content.

If it's possible, could we have it as 'non sticky' so it doesn't occupy any of the screen unless you've scrolled to the bottom of the page?

Safari
Screenshot 2023-03-06 at 10 39 28
Chrome
Screenshot 2023-03-06 at 10 40 53

ricardoV94:
Preview: https://pymc-marketing--199.org.readthedocs.build/en/199/

Closes #191 

The footer now looks like this on my laptop:

image

In addition, I tweaked it so that it looks like this on small screens (subject to fancy CSS support):

image

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -4,15 +4,23 @@
 }
 
 /* Add a separating border line for all but the last item */
+/*
 .footer-item:not(:last-child) {
   border-right: 1px solid var(--pst-color-text-base);
   margin-right: .5em;
   padding-right: .5em;
 }
+*/
 
 .footer-item ul {
   list-style: none;
   display: flex;
+  padding-left: 0;
+  margin-bottom: 0;
+}
+
+.footer-item ul li {
+    padding-right: .5em;
 }
 
 .footer-item li a.nav-link {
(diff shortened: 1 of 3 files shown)


drbenvincent:
Oh, it works if you click the small "view docs" but not the big green "view docs" button
Screenshot 2023-03-10 at 12 34 50

drbenvincent (approved):
This is much better, and enough to close the issue. But for bonus points, it is possible to make the footer non-sticky so it's not always there? It's fine on laptop, but on mobile it occupies a lot of screen real estate.

ricardoV94:
@drbenvincent I made it non-sticky on screens with less than 550px high. Is this good?

ricardoV94:
Actually I think the problem was not the height, but the footer becoming higher when lines wrap-around. I made it non-sticky in those cases. Can you check after it re-renders?

drbenvincent:
I'll have another look when I'm back from an appointment

drbenvincent:
Looks good. No sticky footer on my phone. Smaller sticky footer on my laptop. Thought it's not 'responsive' in that if you make the browser window small on your laptop the sticky footer remains. But I think you've totally owned this issue... ready to merge.

ricardoV94:
> Thought it's not 'responsive' in that if you make the browser window small on your laptop the sticky footer remains.

It doesn't, if you reduce the window so that it's smaller than 550 pixels wide

PR #210, opened 2023-03-21
Adds UML diagrams to give developers a nice overview of the codebase

### LINKED ISSUE #178: UML Diagrams
I created some UML diagrams while working on https://github.com/pymc-labs/pymc-marketing/pull/177, which are shown below. Is it worth adding these to the new-release CI so they are auto-generated each time? Furthermore, are they worth adding to the documentation despite being rather developer-specific?

## MMM
mmm_uml
## CLV
clv_uml

## MMM Tests
tests_mmm_uml

## CLV Tests
tests_clv_uml

Note these diagrams only illustrate internal class methods/attributes and inheritance, and will not display imported function calls like `clv.utils.to_xarray()` and `clv.utils.customer_lifetime_value()`.

To recreate these diagrams, install `graphviz` and `pylint` and run this CLI command:
```!bash
$ pyreverse directory/path -f 'ALL' -o png
``` 
This will also create an additional _packages.png_ file which should be deleted.

drbenvincent:
* Closes #178 
* Adds UML diagrams for packages + classes
* Adds `pylint` as a new dependency under the `lint` section of `[project.optional-dependencies]` in `pyproject.toml`
* Generates diagrams for MMM and CLV seperately
* Decided to _not_ do this for `tests/mmm` and `tests/clv`. I think it's less useful for the tests than the core code.
* Results are shown in `CONTRIBUTING.md`. I assume that this will be an obvious place where [potential] developers will go, so they will easily find it
* Images are stored in `docs/source/uml/`
* Doesn't have to be this way, but new images are triggered with `make lint` ❗️Tell me if this doesn't make sense especially in terms of the remote lint procedure etc ❗️


Makefile
```diff
@@ -7,6 +7,8 @@ lint:
 	pip install .[lint]
 	isort .
 	black .
+	pyreverse pymc_marketing/mmm -d docs/source/uml -f 'ALL' -o png -p mmm
```
juanitorduz, line 10: Shall we add `pyreverse` to the list of dev-packages or is it from `pylint` below ?

drbenvincent: It's from `pylint`

juanitorduz: thanks!

juanitorduz:
I do not know why the docs are failing :(

ColtAllen:
Creating these UML diagrams requires `graphviz` (not a python library). That's the only thing I can think of that could cause the docs build to fail.

drbenvincent:
Maybe a good solution would be to trigger these in the pre commit checks and not to run them remotely?

drbenvincent:
Temporary fix. I've removed these lines from the makefile
```
pyreverse pymc_marketing/mmm -d docs/source/uml -f 'ALL' -o png -p mmm
pyreverse pymc_marketing/clv -d docs/source/uml -f 'ALL' -o png -p clv
```

So if the docs build now, then we are good BUT it would require the UML diagrams to be manually triggered by the user. I propose we accept this for now and I'll create another issue to automate the process.

juanitorduz (approved):
Let's create the issue then :)

PR #221, opened 2023-03-30
Implement vectorized adstock transformations

### LINKED ISSUE #22: Adstock transformation without `for` loop

### LINKED ISSUE #196: MMM transformer(s) fail when parameters are not scalars?
I may be naive in attempting to use existing transformers to do forward sampling (e.g., after sampling from the posterior) (and not wanting to use an inefficient for loop: https://github.com/pymc-labs/pymc-marketing/issues/22).  @lucianopaz seems to be working on something similar here: https://github.com/pymc-labs/pymc-marketing/pull/114, though it seems pretty involved (as shape issues tend to be). And @juanitorduz mentioned that he wrote some extra functions to do the forward pass, but the idea of having nearly duplicate code (e.g., `geometric_adstock()` and `geometric_adstock_forward()`) seems less than ideal.  So take that context into account while reading the following.  If what I have below is overly simplistic and https://github.com/pymc-labs/pymc-marketing/pull/114 would be a better, more robust solution, that's fine.

### Adstock
The following works:
```python
x = rng.random(size=100)
alpha = 1
geometric_adstock(x=x, alpha=alpha).eval()
```
The following fails with a constant_folding error:
```python
x = rng.random(size=100)
alpha = rng.random(size=10)
geometric_adstock(x=x, alpha=alpha).eval()
```
I assume this is a reasonable thing to want to do, but maybe I confused.

Perhaps [this line](https://github.com/pymc-labs/pymc-marketing/blob/e4e05e0abcdb6322850969766ef618f97e0ab7f7/pymc_marketing/mmm/transformers.py#L38) should instead be:

```python
return pt.dot(w.T, x_cycle)
```

I think the 2 versions are equivalent when `w.shape==(l_max,)`, but diverge when `w.shape==(l_max,alpha.shape[0])`?  Hand coding this into a fresh geometric adstock function seems to work in both the cases above.  But again, I may be confused.

### Saturation
The following works:
```python
x = rng.random(size=100)
lam = 1
logistic_saturation(x=x, lam=lam).eval()
```
The following fails:
```python
x = rng.random(size=100)
lam = rng.random(size=10)
logistic_saturation(x=x, lam=lam).eval()
```
with this shape error:
```
ValueError: operands could not be broadcast together with shapes (10,) (100,) 
```
The solution here seem to require some reshaping of `x` and `lam` so that the multiplication (e.g. [here](https://github.com/pymc-labs/pymc-marketing/blob/e4e05e0abcdb6322850969766ef618f97e0ab7f7/pymc_marketing/mmm/transformers.py#L128)) works out correctly (e.g., `x.reshape(-1,1)` and `lam.reshape(1,-1)`).

ricardoV94:
Follow up to #114

Unlike #114 it doesn't try to respect `broadcastable` flags as those are in a limbo in PyTensor at the moment (see https://github.com/pymc-devs/pytensor/pull/149) and are not respected anywhere else.

Closes #22
Closes #196

### DIFF
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -1,7 +1,70 @@
 import pytensor.tensor as pt
+from pytensor.tensor.random.utils import params_broadcast_shapes
 
 
-def geometric_adstock(x, alpha: float = 0.0, l_max: int = 12, normalize: bool = False):
+def batched_convolution(x, w, axis: int = 0):
+    """Apply a 1D convolution in a vectorized way across multiple batch dimensions.
+
+    Parameters
+    ----------
+    x :
+        The array to convolve.
+    w :
+        The weight of the convolution. The last axis of ``w`` determines the number of steps
+        to use in the convolution.
+    axis : int
+        The axis of ``x`` along witch to apply the convolution
+
+    Returns
+    -------
+    y :
+        The result of convolving ``x`` with ``w`` along the desired axis. The shape of the
+        result will match the shape of ``x`` up to broadcasting with ``w``. The convolved
+        axis will show the results of left padding zeros to ``x`` while applying the
+        convolutions.
+    """
+    # We move the axis to the last dimension of the array so that it's easier to
+    # reason about parameter broadcasting. We will move the axis back at the end
+    orig_ndim = x.ndim
+    axis = axis if axis >= 0 else orig_ndim + axis
+    w = pt.as_tensor(w)
+    x = pt.moveaxis(x, axis, -1)
+    l_max = w.type.shape[-1]
+    if l_max is None:
+        try:
+            l_max = w.shape[-1].eval()
+        except Exception:
+            pass
+    # Get the broadcast shapes of x and w but ignoring their last dimension.
+    # The last dimension of x is the "time" axis, which doesn't get broadcast
+    # The last dimension of w is the number of time steps that go into the convolution
+    x_shape, w_shape = params_broadcast_shapes([x.shape, w.shape], [1, 1])
+    x = pt.broadcast_to(x, x_shape)
+    w = pt.broadcast_to(w, w_shape)
+    x_time = x.shape[-1]
+    shape = (*x.shape, w.shape[-1])
+    # Make a tensor with x at the different time lags needed for the convolution
+    padded_x = pt.zeros(shape, dtype=x.dtype)
+    if l_max is not None:
+        for i in range(l_max):
+            padded_x = pt.set_subtensor(
+                padded_x[..., i:x_time, i], x[..., : x_time - i]
+            )
+    else:  # pragma: no cover
+        raise NotImplementedError(
+            "At the moment, convolving with weight arrays that don't have a concrete shape "
+            "at compile time is not supported."
+        )
+    # The convolution is treated as an element-wise product, that then gets reduced
+    # along the dimension that represents the convolution time lags
+    conv = pt.sum(padded_x * w[..., None, :], axis=-1)
+    # Move the "time" axis back to where it was in the original x array
+    return pt.moveaxis(conv, -1, axis + conv.ndim - orig_ndim)
+
+
+def geometric_adstock(
+    x, alpha: float = 0.0, l_max: int = 12, normalize: bool = False, axis: int = 0
+):
     """Geometric adstock transformation.
 
     Adstock with geometric decay assumes advertising effect peaks at the same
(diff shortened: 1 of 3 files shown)


cluhmann:
Ok. Can verify this works as (I) expected in https://github.com/pymc-labs/pymc-marketing/issues/196. Will review now!

pymc_marketing/mmm/transformers.py
```diff
@@ -1,7 +1,61 @@
 import pytensor.tensor as pt
+from pytensor.tensor.random.utils import params_broadcast_shapes
 
 
-def geometric_adstock(x, alpha: float = 0.0, l_max: int = 12, normalize: bool = False):
+def batched_convolution(x, w, axis: int = 0):
```
twiecki, line 5: This looks like it took 5 minutes to write...

ricardoV94: What do you mean?

Btw I didn't write this code, it's picked directly from @lucianopaz PR

pymc_marketing/mmm/transformers.py
```diff
@@ -1,7 +1,61 @@
...
+    axis = axis if axis >= 0 else orig_ndim + axis
+    w = pt.as_tensor(w)
+    x = pt.moveaxis(x, axis, -1)
+    l_max = w.type.shape[-1]
+    if l_max is None:
+        try:
+            l_max = w.shape[-1].eval()
+        except Exception:
+            pass
+    x_shape, w_shape = params_broadcast_shapes([x.shape, w.shape], [1, 1])
```
lucianopaz, line 36:
```suggestion
    # Get the broadcast shapes of x and w but ignoring their last dimension.
    # The last dimension of x is the "time" axis, which doesn't get broadcast
    # The last dimension of w is the number of time steps that go into the convolution
    x_shape, w_shape = params_broadcast_shapes([x.shape, w.shape], [1, 1])
```

pymc_marketing/mmm/transformers.py
```diff
@@ -1,7 +1,61 @@
...
+    x = pt.broadcast_to(x, x_shape)
+    w = pt.broadcast_to(w, w_shape)
+    x_time = x.shape[-1]
+    shape = (*x.shape, w.shape[-1])
+    padded_x = pt.zeros(shape, dtype=x.dtype)
```
lucianopaz, line 41:
```suggestion
    # Make a tensor with x at the different time lags needed for the convolution
    padded_x = pt.zeros(shape, dtype=x.dtype)
```

pymc_marketing/mmm/transformers.py
```diff
@@ -1,7 +1,61 @@
...
+    if l_max is not None:
+        for i in range(l_max):
+            padded_x = pt.set_subtensor(
+                padded_x[..., i:x_time, i], x[..., : x_time - i]
+            )
+    else:  # pragma: no cover
+        raise NotImplementedError(
+            "At the moment, convolving with weight arrays that don't have a concrete shape "
+            "at compile time."
```
lucianopaz, line 50:
```suggestion
            "at compile time is not supported."
```

pymc_marketing/mmm/transformers.py
```diff
@@ -1,7 +1,61 @@
...
+        )
+    conv = pt.sum(padded_x * w[..., None, :], axis=-1)
```
lucianopaz, line 52:
```suggestion
    # The convolution is treated as an element-wise product, that then gets reduced
    # along the dimension that represents the convolution time lags
    conv = pt.sum(padded_x * w[..., None, :], axis=-1)
```

pymc_marketing/mmm/transformers.py
```diff
@@ -1,7 +1,61 @@
...
+    return pt.moveaxis(conv, -1, axis + conv.ndim - orig_ndim)
```
lucianopaz, line 53:
```suggestion
    # Move the "time" axis back to where it was in the original x array
    return pt.moveaxis(conv, -1, axis + conv.ndim - orig_ndim)
```

lucianopaz (changes requested):
I suggest adding a bunch of comments to make the convolution code a bit easier to read

ricardoV94:
Thanks @lucianopaz!

aseyboldt:
@ricardoV94 Didn't this PR change the runtime of the geometric adstock to something slower (from $O(n)$ to $O(n n_\text{lag})$)?
I didn't perform any testing, and maybe on a GPU it doesn't make as much difference, but at least for longer lag times this should be worse?

ricardoV94:
> @ricardoV94 Didn't this PR change the runtime of the geometric adstock to something slower (from O(n) to $O(n n_\text{lag})$)? I didn't perform any testing, and maybe on a GPU it doesn't make as much difference, but at least for longer lag times this should be worse?

Possibly, the concern was vectorization not speed. I didn't write the code, it was actually @lucianopaz, this was a small change from #114

aseyboldt:
Makes sense. I guess we can just try to remember in case we run into performance issues :-)

wd60622:
This came up recently #670

twiecki:
Could we keep both implementations?

PR #226, opened 2023-03-31
Resolve MAP fit failing in `BetaGeoModel`

### LINKED ISSUE #214: MAP fits Failing for `BetaGeoModel`
Should probably fix this before the Official Release:

```
from lifetimes.datasets import load_cdnow_summary
from pymc_marketing.clv import BetaGeoModel


df = load_cdnow_summary()

model = BetaGeoModel(customer_id=df['ID'], 
                     frequency=df['frequency'], 
                     recency=df['recency'],
                     T=df['T'],
                    )

model.fit(fit_method='map')
```

```
SamplingError: Initial evaluation of model at starting point failed!
Starting values:
{'a_log__': array(0.), 'b_log__': array(0.), 'alpha_log__': array(0.), 'r_log__': array(0.)}

Initial evaluation results:
{'a': 0.0, 'b': 0.0, 'alpha': 0.0, 'r': 0.0, 'likelihood': nan}
```

larryshamalama:
Closes #214. The issue was due to the evaluation of $\mathsf{C}_4 = \dfrac{a}{b + x - 1}$ for non-zero frequency values and $b \sim \operatorname{HalfFlat}$ being initialized at 1. Thanks to @ColtAllen for pointing this out.

This could have meant that we need to solely look at the parameter space where $b + x > 1$. With $x$ being an integer, if we only evaluate $\mathsf{C}_4$ for non-zero $x$ values, the log-likelihood expression should be stable.

Two other minor things:
- I also slid in a quick test that checks for `customer_id` to be unique.
- I believe that `x_zero` in the logp expression checks for _non-zero_ values. I changed the name to `x_non_zero`.

https://github.com/pymc-labs/pymc-marketing/blob/eb52537543229d8e8a3316958426e112d01c0c31/pymc_marketing/clv/models/beta_geo.py#L148


ColtAllen:
Can we add a MAP fit example to the notebook to verify the issue is resolved?

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -145,7 +145,7 @@ def logp(t_x, x, a, b, r, alpha, T):
                 The log-likelihood expression here aligns with expression (4) from [3]
                 due to the possible numerical instability of expression (3).
                 """
-                x_zero = pt.where(x > 0, 1, 0)
+                x_non_zero = pt.where(x > 0, 1, 0)
```
ricardoV94, line 148: Isn't this enough?
```suggestion
                x_non_zero = x > 0
```

tests/clv/models/test_beta_geo.py
```diff
@@ -83,14 +83,30 @@ def test_model(self, a_prior, b_prior, alpha_prior, r_prior):
             "r_log__": (),
         }
 
-    def test_numerically_stable_logp(self):
+    def test_customer_id_warning(self):
+        with pytest.raises(ValueError):
```
ricardoV94, line 87: Add a `match=`?
```suggestion
        with pytest.raises(ValueError, match=...):
```

ricardoV94 (commented):
Looks good, I left some small suggestions below.

tests/clv/models/test_beta_geo.py
larryshamalama: Added!

ricardoV94: I don't see it?

larryshamalama: My bad, I messed up a rebase. It should be there now

ColtAllen (approved):

PR #227, opened 2023-03-31
Use MaxAbsScaler instead of MinMaxScaler for Target MMM

juanitorduz:
Use MaxAbsScaler instead of MinMaxScaler for Target MMM. This is important when computing ROAS. It also makes the transformation (and inverse transformations easier)

### DIFF
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -1,12 +1,10 @@
 from typing import Any, Dict, List, Optional
 
-import arviz as az
 import pandas as pd
 import pymc as pm
-from xarray import DataArray
 
 from pymc_marketing.mmm.base import MMM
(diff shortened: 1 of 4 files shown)


cluhmann:
Haven't tested, but it looks like this removes `compute_channel_contribution_original_scale()` entirely.  Was that intentional?  That method is currently used in various places including in the `plot_contribution_curves()` method and directly used in the MMM example notebook.  Doesn't `compute_channel_contribution_original_scale()` just need to be revised to use the new `MaxAbsScaleTarget`? If so, I'm happy to do that.

juanitorduz:
I remove it because it's already implemented in the base class. So the example notebook should still work (would you mind testing just to double check 🤗)

cluhmann:
Yes. I had missed the method in the base class (now I remember you saying this yesterday).

ricardoV94:
Does this impart the MMM example?

juanitorduz:
I tested it and it was ok! I think @cluhmann also tested it right ? (In the example the min values are zero anyway)

ricardoV94 (approved):

Issue #128, opened 2023-01-10
CLV Distribution RVs not Model-Specific


ColtAllen:
I'm considering using the RVs in the CLV Distributions module to generate synthetic data for [testing the Pareto/NBD model](https://github.com/pymc-labs/pymc-marketing/issues/127). However, after looking at the `rng_fn` for both classes, I'm concerned the RVs may not be robust across all model types, and the distribution classes could have similar pathologies.

As currently defined, the `sim_data` method in `rng_fn` is using a binomial RV within a while loop for the dropout probability. This works well for the Modified BG/NBD model, but I do not see a provision for the BG/NBD assumption that all non-repeat customers are alive with probability 1. The Pareto/NBD also does not use a binomial RV at all - instead it uses an exponential RV to predict the dropout time period prior to the while loop. 

The data generation functions in Lifetimes/BTYD are a useful reference:

https://github.com/ColtAllen/btyd/blob/main/btyd/generate_data.py

ricardoV94:
We should add those variants to generate data. Do you want to assign this issue to yourself?

ColtAllen:
I consider this a prerequisite for https://github.com/pymc-labs/pymc-marketing/issues/127, so I'll get started on adding a ParetoNBD distribution sometime next week.

The existing distributions should also be revised to reflect being written with the BG/NBD model in mind. I can also look into vectorization for `sim_data`.

ColtAllen:
I'll be creating a PR for this soon. @larryshamalama I see you made the original commit for the `ContContract` distribution class. Is there a research citation you can provide for me to add to the docstring?

larryshamalama:
> I'll be creating a PR for this soon. @larryshamalama I see you made the original commit for the `ContContract` distribution class. Is there a research citation you can provide for me to add to the docstring?

Sorry, this message slipped by my attention. I did not use any research articles since I did not find any when I was writing out the likelihood... Perhaps there is one out there that I'm unaware of. I can write out the likelihood derivation again if it would help

ColtAllen:
@larryshamalama Let's refactor `ContNonContract` and `ContContract`into `BetaGeoNBD` and `BetaGeoNBDAggregate`, respectively, so we can close this out:

- [ ] Update docstring of `ContContract`/`BetaGeoNBDAggregate`

[...]

PR #230, opened 2023-04-04
Vectorize `rng_fn` in CLV distribution classes

### LINKED ISSUE #37: Vectorize the random method in `ContNonContract`
The `rng_fn` in `ContNonContract` uses a `for` loop, but it can surely be vectorized although it would require some clever math and perhaps some concepts from order statistics.

larryshamalama:
Closes #37.

The `rng_fn` of `ContNonContract` (and `ContContract`) is done via a `while` loop: generate $T_1$, decide if a customer "dies", then generate $T_2$ and so on as long as the purchase times are under some censoring time $T$. However, this data-generating mechanism can be vectorized, but it required some thought. `ContContract` can be vectorized in a similar manner. The tests should and are left untouched.

Also, I remove `T0` as an argument because it is unused. This is in the first commit of this PR.

The bigger overarching question: how should we better construct, name and establish our CLV distribution building blocks? For instance, `ContNonContract` refers to a continuous non-contractual setting, but the dropout variable manifests as a Bernoulli variable (e.g. BG/NBD) rather than another censoring time (e.g. ParetoNBD). This is related to #128.

### Statistical Details

Let $0 < \tilde T_1 < \tilde T_2, \dots$ be the purchase times and $N(t) = \sum_j 1[T_j \leq t]$ be a counting process for the number of purchases in $[0, t]$. Because customers can dropout/die after any purchase, let $D_j \sim \operatorname{Bernouilli}(p)$ be the dropout indicator after the $j$ th purchase. We define these dropouts for every purchase times even if customers are already "dead". More specifically, for BG/NBD, we assume that $N(t)$ is a homogenous Poisson process with hazard $\lambda$, hence $T_j - T_{j-1} \stackrel{\small{\text{i.i.d.}}}{\sim} \operatorname{Exp}(\lambda)$. Formally, $( \tilde T_j, D_j)_{j \geq 1}$ characterizes a marked point process, but that's not really useful here.

Data from a Poisson process can be generated sequentially. That is, generate $T_j - T_{j-1} \sim \operatorname{Exp}(\lambda)$ until $T_j$ exceeds some censoring time $T$. Coupled with the dropout process due to the non-contractual setting, $D_j$ is also generated with every $T_j$ and we stop when $D_j = 1$. This sequential process is more intuitive in its implementation and that's what we have now.

However, we can opt for a more efficient data-generating mechanism by using the fact that the joint distribution of $T_1, \dots, T_x$ is the uniform order statistics over the interval $[0, T]$ conditioned on $N(t) = n$. In other words, we can do: (1) $N(t) \sim \operatorname{Poisson}(\lambda T)$ followed by (2) $T_j \equiv U_{(j)}$ where $U_{(j)}$ is the order statistics of $U_1, \dots, U_x$ with $U_j \stackrel{\small{\text{i.i.d.}}}{\sim} \operatorname{Unif}(0, T)$.

Conditioned on $N(t) = x$, we don't know if a customer will drop out before completing the "x" purchases (according to this data-generating process). To account for the drop out, we can simply retrieve the order statistic $U_{(R)}$ corresponding to when the customer drops out with $R \sim \operatorname{Geometric}(p)$. Given that the order statistics of uniformly distributed variables over $[0, 1]$ follows a [Beta distribution](https://en.wikipedia.org/wiki/Order_statistic#Order_statistics_sampled_from_a_uniform_distribution), $U_{(R)}$ can be sampled from a Beta distribution then scaling it by $T$ to map $[0, 1]$ to $[0, T]$.

All in all, the data-generating procedure in this PR is more efficient and vectorized, and all tests are passing.


pymc_marketing/clv/distributions.py
```diff
@@ -19,59 +19,47 @@ class ContNonContractRV(RandomVariable):
...
-                    if dropout == 1:
-                        break
+            return rng.beta(a, b)
 
-            return np.array([t, n])
-
-        for index in np.ndindex(*size):
-            output[index] = sim_data(lam[index], p[index], T[index], T0[index])
+        t_x = (
+            np.vectorize(modified_rng_beta)(
```
ricardoV94, line 55: This will force a slow loop through `(a, b)` pairs. A better approach would be to call `beta` on the inputs and then override with zeros where needed:

```python
t_x = rng.beta(x, np.maximum(...))
t_x[x == 0] = 0
```

And suppress any warnings if needed

pymc_marketing/clv/distributions.py
```diff
@@ -19,59 +19,47 @@ class ContNonContractRV(RandomVariable):
...
 
-                    if dropout == 1:
-                        break
+            return rng.beta(a, b)
 
-            return np.array([t, n])
-
-        for index in np.ndindex(*size):
-            output[index] = sim_data(lam[index], p[index], T[index], T0[index])
+        t_x = (
```
ricardoV94, line 54: Are you sure this is equivalent? Could you add some comments with the derivation perhaps?

pymc_marketing/clv/distributions.py
```diff
@@ -152,36 +140,33 @@ class ContContractRV(RandomVariable):
...
+            size = np.broadcast_shapes(lam.shape, p.shape, T.shape)
 
         lam = np.broadcast_to(lam, size)
         p = np.broadcast_to(p, size)
         T = np.broadcast_to(T, size)
-        T0 = np.broadcast_to(T0, size)
 
         output = np.zeros(shape=size + (3,))
 
-        def sim_data(lam, p, T, T0):
```
ricardoV94, line 184: When I talked about vectorization in the original issue I was thinking about something slightly different.

The idea is to replace nested loop of this form:
```python
ys = []
# sim_data
for x in xs:
  y = y0
  while not cond(y):
    y = rng.foo(bar(y, x))
  ys.append(y)
return ys
```

By:
```python
ys = full(xs.shape, y0)
unfinished = ones(xs.shape, dtype=bool)
while any(unfinished):
  ys[unfinished] = rng.foo(bar(ys[unfinished], xs[unfinished])
  unfinished[unfinished] = not cond(unfinished[unfinished])
return ys
```

Where the repeated fancy indexing would be only done once.

This can be much faster because the calls to `foo` and `bar` can be numpy vectorized, at the cost of some overhead with fancy indexing, versus slow Python loops. 

It may not apply to all (any) of our distributions, and it's not super critical we investigate.

pymc_marketing/clv/distributions.py
ricardoV94: Oh I saw your explanation on the top message :O really cool! I would suggest to add it as a comment in the code for future devs ;)

pymc_marketing/clv/distributions.py
larryshamalama: `rng.beta(a, b)` throws an error for `a <= 0`. Perhaps the following would work better?
```python
t_x[x > 0] = rng.beta(x, np.maximum(...))
```

pymc_marketing/clv/distributions.py
larryshamalama: Should I add it in the `rng_fn` docstring? What about writing it as a blogpost and linking it somewhere in the class? I think that the math can be a bit terse and pictures would complement well the derivation

ricardoV94: Yeah that docstring would be a good place, regardless of a neat blogpost ;)

pymc_marketing/clv/distributions.py
larryshamalama: Does NumPy some sort of lazy evaluation where `rng.beta(x, ...)` does not need to be evaluated where `x == 0`?

ricardoV94: Not sure. You can index yourself. Are x=0 common?

```python
ret = np.zeros()
ret[x~0] = rng.beta(a[x~0], b[x~0])
```

But if they are not common I wouldn't bother [the two extra slices could be slower than the useless beta). Unless numpy raises ofc

larryshamalama: Depending on the hazard parameter, it is generally uncommon, but important to account for. It characterizes the situation where there is too little time to observe if any purchases can be made.

I'll try what you suggest above later this afternoon

ricardoV94: I meant I wouldn't bother to index before and would just override with zeros after (if zeros don't cause it to crash)

larryshamalama: Zeros still makes this crash... with `size` is just 1D, I can use
```python
nzp = x != 0 # nzp = non-zero purchases

        t_x = np.zeros(shape=size)
        t_x[nzp] = rng.beta(x[nzp], np.maximum(x_1 + 1 - x_2, 1)[nzp])*T
```
but this does not work when `size` is 2D (e.g. `(5, 3)` in tests). I'm not sure how to mitigate this issue...

ricardoV94:
Some failing test but otherwise looks good

pymc_marketing/clv/distributions.py
```diff
@@ -152,62 +139,51 @@ class ContContractRV(RandomVariable):
...
-            while not dropout:
-                wait = rng.exponential(scale=1 / lam)
-                # If we didn't go into the future
-                if (t + wait) < T:
-                    n += 1
-                    t = t + wait
-                    dropout = rng.binomial(n=1, p=p)
-                else:
-                    break
+        if x.shape == ():
```
larryshamalama, line 172: Should this part be more elegant?

larryshamalama (commented):
> Some failing test but otherwise looks good

Should be good now, I made a silly mistake. One code style question

pymc_marketing/clv/distributions.py
ricardoV94: Nah, good enough

ricardoV94 (approved):

PR #235, opened 2023-04-06
Test oldest pymc version

### LINKED ISSUE #193: Run tests on oldest supported version of PyMC
We should run our tests on the oldest supported version of PyMC (from the `pyproject.toml`) as well as the latest (which we do already)

ricardoV94:
Closes #193 

Ignore the (expected) tests. Those are the current required tests for PRs, after this one they would be replaced by the new `all`

----
:books: Documentation preview :books:: https://pymc-marketing--235.org.readthedocs.build/en/235/


ricardoV94:
Nice, this showed that `pymc==5.00` is actually too old already as it relies on a PyTensor version that didn't have the hyp2F1 function yet

PR #240, opened 2023-04-09
Use model builder for MMM classes

michaelraczycki:
Necessary changes of MMM base and delayed_saturated_mmm along with some new functions to allow of full integration of MMM with Model Builder.
Done so far:
created data formatters that prepare necessary configs and dataset for Model Builder, need discussion on whether this will be final design.
@tomicapretto , @twiecki , @ricardoV94 please feel free to suggest any changes that would allow easier integration into other pymc-related repos

----
:books: Documentation preview :books:: https://pymc-marketing--240.org.readthedocs.build/en/240/


pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -36,51 +37,56 @@ def __init__(
...
 
-            intercept = pm.Normal(name="intercept", mu=0, sigma=2)
+            intercept = pm.Normal(
+                name="intercept",
+                mu=model_config.intercept["mu"],
+                sigma=model_config.intercept["sigma"],
+            )
 
             beta_channel = pm.HalfNormal(
-                name="beta_channel", sigma=2, dims="channel"
```
twiecki, line 69: If these are hard-coded in the original model, we don't need to make them configurable here.

michaelraczycki: what if I already added them? The basic functionality will stay the same, and the flexibility will be just in case we'd like to have it, but right now that would actually require work to make it have less flexibility

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -36,51 +37,56 @@ def __init__(
 
     def build_model(
         self,
-        data: pd.DataFrame,
-        adstock_max_lag: int = 4,
+        model_data: dict,
+        model_config: dict,
```
ricardoV94, line 41: Do we need to change the API?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -36,51 +37,56 @@ def __init__(
...
+                sigma=model_config.beta_channel["sigma"],
+                dims=model_config.beta_channel["dims"],
             )  # ? Allow prior depend on channel costs?
 
-            alpha = pm.Beta(name="alpha", alpha=1, beta=3, dims="channel")
+            alpha = pm.Beta(
+                name="alpha",
+                alpha=model_config.aplha["alpha"],
+                beta=model_config.aplha["beta"],
+                dims=model_config.alpha["dims"],
```
ricardoV94, line 72:
```suggestion
                alpha=model_config.alpha["alpha"],
                beta=model_config.alpha["beta"],
                dims=model_config.alpha["dims"],
```

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: If we change that now we'll gain the flexibility to alter the initial params via the api using model_config), so far they were hardcoded (if we're sure that it will never need to be altered by users then indeed maybe it's not necessary)

Issue #81, opened 2022-11-17
Add type checking into pre-commit


lucianopaz:
We would have to add mypy as a dev dependency, a Makefile target and also add it as a pre-commit action

PR #248, opened 2023-04-16
Add mypy support

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/81

Let's aim to have mypy! 😄 

- [x] Add mypy as dev dependency, pre-commit and GitHub workflows.
- [x] Fix hints MMM.
- [x] Fix hints CLV.

----
:books: Documentation preview :books:: https://pymc-marketing--248.org.readthedocs.build/en/248/

### DIFF
--- a/pymc_marketing/clv/models/gamma_gamma.py
+++ b/pymc_marketing/clv/models/gamma_gamma.py
@@ -5,6 +5,7 @@ import pandas as pd
 import pymc as pm
 import pytensor.tensor as pt
 import xarray
+from pymc.util import RandomState
 from pytensor.tensor import TensorVariable
 
 from pymc_marketing.clv.models.basic import CLVModel
@@ -33,7 +34,7 @@ class BaseGammaGammaModel(CLVModel):
         customer_id: Union[np.ndarray, pd.Series],
         mean_transaction_value: Union[np.ndarray, pd.Series, TensorVariable],
         frequency: Union[np.ndarray, pd.Series, TensorVariable],
-        random_seed=None,
+        random_seed: Optional[RandomState] = None,
     ) -> xarray.DataArray:
         """Posterior distribution of transaction value per customer"""
 
@@ -385,7 +386,7 @@ class GammaGammaModelIndividual(BaseGammaGammaModel):
         self,
         customer_id: Union[np.ndarray, pd.Series],
         individual_transaction_value: Union[np.ndarray, pd.Series, TensorVariable],
-        random_seed=None,
+        random_seed: Optional[RandomState] = None,
     ) -> xarray.DataArray:
         """Return distribution of transaction value per customer"""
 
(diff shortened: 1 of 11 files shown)


Garve (approved):
Looks good to me!

ricardoV94:
I hate the `signature incompatible with super type` thing. Do we need/want to care about it?

juanitorduz:
> I hate the `signature incompatible with super type` thing. Do we need/want to care about it?

I do not have a strong opinion and we can ignore this one if we want. I think mypy should be a dev tool and not a blocker ;) Up to us to decide IMO.

ricardoV94:
Let's remove that and check the remaining errors?

pymc_marketing/clv/utils.py
```diff
@@ -244,13 +248,13 @@ def clv_summary(
...
         Column in the transactions DataFrame that denotes the datetime the purchase was made.
-    monetary_value_col: string, optional
+    monetary_value_col: Optional[str], by default None
         Column in the transactions DataFrame that denotes the monetary value of the transaction.
         Optional; only needed for spend estimation models like the Gamma-Gamma model.
-    observation_period_end: datetime, optional
-         A string or datetime to denote the final date of the study.
-         Events after this date are truncated. If not given, defaults to the max 'datetime_col'.
-    datetime_format: string, optional
+    observation_period_end: Optional[Union[str, pd.Period, datetime]], by default None
```
ricardoV94, line 254: In PyMC we don't usually bother to do this, since it's duplicated work and docstrings can become much more verbose. We usually have type-hints on the function call, and readable type descriptions in the docstrings.

```suggestion
    observation_period_end: Optional[Union[str, pd.Period, datetime]], by default None
```

No strong preference, but I like less work rather than more :)

juanitorduz: I think as long we have a convention we can be flexible, shall I undo the doctring changes?

ricardoV94: Yeah

ColtAllen:
I like the changes that have been made to `clv_summary`! However, these revisions aren't covered by the current test parameterizations.

Should the tests be modified now, or in a future PR? I ask because today's meeting highlighted the priority of adding the `calibration_and_holdout_data` function from `lifetimes`.

pymc_marketing/clv/utils.py
juanitorduz: reverted them inhttps://github.com/pymc-labs/pymc-marketing/pull/248/commits/07d923ea717534f3657d49a31bf82e6b9346f33a

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -404,7 +405,7 @@ def expected_customer_spend(
         self,
         customer_id: Union[np.ndarray, pd.Series],
         individual_transaction_value: Union[np.ndarray, pd.Series, TensorVariable],
-        random_seed=None,
+        random_seed=Optional[None],
```
ricardoV94, line 408: Should this be `Optional[RandomState]`?

PR #247, opened 2023-04-16
add contribution curves over time

Garve:

PR #251, opened 2023-04-18
MMM: scale contributions to original scale and allow custom colors

juanitorduz:
Builds upon https://github.com/pymc-labs/pymc-marketing/pull/247

- [x] Scale contributions to the original scale.
- [x] Allows custom colors.

----
:books: Documentation preview :books:: https://pymc-marketing--251.org.readthedocs.build/en/251/

### DIFF
--- a/tests/mmm/test_base.py
+++ b/tests/mmm/test_base.py
@@ -210,7 +210,10 @@ class TestMMM:
             ("plot_grouped_contribution_breakdown_over_time", {}),
             (
                 "plot_grouped_contribution_breakdown_over_time",
-                {"stack_groups": {"controls": ["control_1"]}},
+                {
+                    "stack_groups": {"controls": ["control_1"]},
+                    "area_kwargs": {"alpha": 0.5},
+                },
             ),
         ],
     )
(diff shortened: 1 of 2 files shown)


juanitorduz:
What do you think @Garve ? :)

Garve:
Heya! Should we also introduce a flag for the original scale? Like in the other plot method.

tests/mmm/test_base.py
```diff
@@ -210,7 +210,11 @@ def toy_preprocessing(self, data):
             ("plot_grouped_contribution_breakdown_over_time", {}),
             (
                 "plot_grouped_contribution_breakdown_over_time",
-                {"stack_groups": {"controls": ["control_1"]}},
+                {
```
ricardoV94, line 213: Not for this PR, but I wonder if we should test the actual expected plots with something like https://matplotlib.org/stable/api/testing_api.html#module-matplotlib.testing

At least we will know if they accidentally change. Not super convinced.

ricardoV94 (approved):

tests/mmm/test_base.py
juanitorduz: wow! I did not know about this! I'll take a look and create an issue

PR #257, opened 2023-04-20
Improve `freq` docstring and fix CLV quickstart

### LINKED ISSUE #256: CLV quickstart example uses wrong frequency
First of all thanks for this amazing library and it's great to see many people actively working on this!

I have two questions regarding the example notebook [CLV Quickstart](https://www.pymc-marketing.io/en/stable/notebooks/clv/clv_quickstart.html#estimating-clv).

1. While estimating the CLV, one of the arguments of the `expected_customer_lifetime_value` method is `time`. The comment in this line says you are predicting for 12 months while the units of time in the rest of the notebook are in weeks. In addition, if I check the API documentation of this method, the default value for `freq` is "D" (days). So if you are predicting for months, `freq` should equal "M", right?
2. How did you define "Customer Lifetime Value" in your package because if we speak about the lifetime of a customer, that also includes the time before the prediction. I believe the method is calculating the value a customer creates after time T, where the period up to T is used to fit the model. 
 
An example: Customer X makes its first purchase at 20-04-2022 (DD-MM-YYYY). We fit a model today (20-04-2023) and predict the number of purchases and their average monetary value for one year in the future (up untill 20-04-2024). The CLV is then the value that the customer generated from 20-04-2022 untill 20-04-2024, right? Is this also what we are computing when we use this package? Or do we estimate the value that the customer will generate in the period today untill 20-04-2024 (so for the future only)?

ricardoV94:
Closes #256


pymc_marketing/clv/utils.py
```diff
@@ -56,12 +56,12 @@ def customer_lifetime_value(
...
         The lifetime expected for the user in months. Default: 12
     discount_rate: float, optional
-        The monthly adjusted discount rate. Default: 1
+        The monthly adjusted discount rate. Default: 0.01
     freq: string, optional
-        Frequency of discrete time steps used to estimate the customer lifetime value.
-        Defaults to "D" for daily. Other options are "W" (weekly), "M" (monthly), and "H" (hourly).
-        Smaller time frames estimate better the effects of discounting rate, at the cost of more
-        evaluations.
+        Sampling frequency of purchase history. Defaults to "D" for daily.
```
larryshamalama, line 61: Would units (of time) be more appropriate?

ricardoV94: Can you write the full suggestion?

larryshamalama: Units of purchase frequency. Defaults to "D" for daily.

^How does that sound?

ricardoV94: Pandas calls it "the frequency of the time series". I think the `freq` kwarg is borrowed directly from them 

https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.freq.html.

So maybe just borrow that directly?

ricardoV94, line 61:
```suggestion
        Sampling frequency of the purchase history. Defaults to "D" for daily.
```

larryshamalama: Yeah I think that sounds good!

larryshamalama (approved):

larryshamalama:
Looks good! Thanks @ricardoV94 :)

PR #283, opened 2023-05-09
Run lint action calls pre-commit to get reproducible env

### LINKED ISSUE #282: Lint action should run pre-commit instead of Makefile target
While working on #281, I ran into a linting problem which was only visible in the CI workflow. This happened because the version of the packages used for linting is not pinned in the development requirements, which are installed via `make check_lint`. To prevent this, we should do as [we did in `causalpy`](https://github.com/pymc-labs/CausalPy/commit/b5afc1910f913556753b3cf5f47635e24ef06e73) and make the lint action call pre-commit instead. The `pre-commit` configuration file pins all the versions of the linting packages and yields a reproducible test environment.

lucianopaz:
Closes #282 

To be honest, I don't know why pymc was being installed in the lint job so I removed those steps :shrug:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,14 +22,8 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install oldest version of PyMC
-        if: ${{ matrix.oldest-pymc }}
-        run: pip install pymc==${{ env.OLDEST_PYMC_VERSION }}
       - name: Run lint
-        run: make check_lint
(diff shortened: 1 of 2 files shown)


lucianopaz:
This ended up being a bigger change than what I had imagined. The pre-commit configuration took the `black`, `isort`, `flake8` and `mypy` executables from the system version that was installed. It didn't create a venv with a pinned version, which was the entire point of switching from `make check_lint` to `pre-commit run`. I had to change the configuration and chose some arbitrary versions to use for pins.

juanitorduz (approved):
👍 LGTM

PR #285, opened 2023-05-15
initial refactoring to use model_builder in MMM

michaelraczycki:
Adaptations to implement changes introduced in BayesianEstimator, preparing the class to be sklearn-compatible.
Main changes include:
Breaking down create_sample_input into class method generate_model_data, and 2 property functions: 
default_model_config and default_sampler_config. Implementation of the above in baseMMM as abstract classes and actual implementation in DelayedSaturatedMMM. 

----
:books: Documentation preview :books:: https://pymc-marketing--285.org.readthedocs.build/en/285/


pymc_marketing/mmm/base.py
```diff
@@ -41,24 +42,39 @@ def __init__(
...
             self.validate(self.data)
         self.preprocessed_data = self.preprocess(self.data.copy())
+        super().__init__()
 
         self.build_model(
             data=self.preprocessed_data,
             **kwargs,
         )
 
+    @property
```
twiecki, line 61: Don't we inherit this form the base class?

twiecki:
@juanitorduz wdyt?

mypy.ini
```diff
@@ -17,3 +17,9 @@ ignore_missing_imports = True
 
 [mypy-scipy.*]
 ignore_missing_imports = True
+
+[mypy-pandas.*]
+ignore_missing_imports = True
```
juanitorduz, line 22: Do we really need this? We can add the stubs (we did not have pandas typing issues before so I would not remove them)?

pymc_marketing/mmm/base.py
```diff
@@ -30,36 +29,36 @@
...
+class BaseMMM(ModelBuilder):
     model: pm.Model
 
     def __init__(
         self,
-        data: pd.DataFrame,
         target_column: str,
         date_column: str,
         channel_columns: Union[List[str], Tuple[str]],
+        data: Optional[pd.DataFrame] = None,
```
juanitorduz, line 40: can we leave `data` as a the first argument?

pymc_marketing/mmm/base.py
```diff
@@ -30,36 +29,36 @@
...
         channel_prior: Optional[TensorVariable] = None,
         validate_data: bool = True,
+        model_config: Optional[Dict] = None,
+        sampler_config: Optional[Dict] = None,
```
juanitorduz, line 44: It would be nice to add docstrings to this class since it is unclear what these arguments mean if one is unfamiliar with the `ModelBuilder` class.

pymc_marketing/mmm/base.py
juanitorduz: Also, why can this be none? We should always have input data right?

pymc_marketing/mmm/base.py
```diff
@@ -167,37 +149,39 @@ def plot_prior_predictive(
         )["likelihood"]
 
         fig, ax = plt.subplots(**plt_kwargs)
+        if self.data is not None:
```
juanitorduz, line 152: Here and in all methods below? Why do we allow `data` to be `None`?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -19,10 +19,12 @@
 class BaseDelayedSaturatedMMM(MMM):
     def __init__(
         self,
-        data: pd.DataFrame,
         target_column: str,
         date_column: str,
         channel_columns: List[str],
+        data: Optional[pd.DataFrame] = None,
```
juanitorduz, line 25: As above, why `data` can be `None`?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -63,80 +65,170 @@ def __init__(
...
+                name="intercept",
+                mu=model_config["intercept"]["mu"],
+                sigma=model_config["intercept"]["sigma"],
+            )
 
+            """beta_channel = pm.HalfNormal(
+                name="beta_channel",
+                sigma=model_config["beta_channel"]["sigma"],
+                dims=model_config["beta_channel"]["dims"],
+            )  # ? Allow prior depend on channel costs?"""
```
juanitorduz, line 202: why is this commented?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -63,80 +65,170 @@ def __init__(
...
 
             channel_adstock = pm.Deterministic(
                 name="channel_adstock",
                 var=geometric_adstock(
                     x=channel_data_,
                     alpha=alpha,
-                    l_max=adstock_max_lag,
+                    l_max=kwargs["adstock_max_lag"]
+                    if "adstock_max_lag" in kwargs
+                    else 4,
```
juanitorduz, line 231: I would not set a default as I think users should know with `adstock_max_lag` they need.

juanitorduz (commented):
In this first review, I mostly asked questions :) . Ideally, when reviewing this PR we need to make sure we can run the mmm notebook end-to-end.

pymc_marketing/mmm/base.py
michaelraczycki: Not requiring the data as an argument makes the class more scikit-learn like, where the data needs to be available only for fit, also an idea I've discussed with Thomas is to utilise the generate_model_data, which this PR adds to allow sandbox-like experience, where the user can just initialise the class with minimal parameters to learn about the expected data structure and functionalities. But overall I'd say that making data optional is a step towards making the class more intuitive for scikit-learn users

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: This is actually something I wanted to ask about, but forgot. After you've added the line:
            beta_channel = self.model.register_rv(
                rv_var=channel_prior, name="beta_channel", dims="channel"
            )
under the commented part, tests are failing with:
ValueError: Variable name beta_channel already exist
it does indeed seem like we're trying to register a rv with exactly the same name as the pm.HalfNormal from previous beta_channel, I'm not sure if it's a bug and it should be allowed to do it like that, or it's a typo and it was supposed to be named for example "beta_channel_" and the _ was lost on the way

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: as above, this was just my attempt of creating the sandbox for new pymc users, but if it's something you're decided on not having I can make it as required parameter, and remove the default values

mypy.ini
michaelraczycki: it might be a versioning thing, but locally it was complaining about the missing stubs. if having the new dependency is fine with you I can add it to the project, and remove this exemption

pymc_marketing/mmm/base.py
michaelraczycki: It's a mypy thing, since in this PR I was going with the idea of accepting None as data (as explained above, to make usage more scikit-learn like) mypy does just static analysis, and will throw errors if it finds a property that possibly can be None and the check in code is not made.

pymc_marketing/mmm/base.py
michaelraczycki: so in general, with the sandbox approach I'm basically implementing the generate_model_data in a way that it will return the example data used in the tests, in case no data would be provided. The users still keep the option of:
a) As actually the default example dataset should allow for not specifying the date_column and target_column, as in the default scenario it can be specified up front. In this case we don't have any non-default parameters and the data can be provided as a first argument, keeping the previous api unchanged. -> This requires creation of a wrapper function, read below
c)can provide the data directly to fit. In this case, the data needs to be passed in a X,y format to make it compatible with sklearn.
As I've realised now, If we still want to allow the users to pass the entire dataset as it's being done now, I'll need to add a wrapper function that will call fit under the hood while passing the data in a correct format, so far as the users would be forced to call fit by themselves, passing data to the class constructor makes very little sense, as they would anyway need to specify which columns of the dataset are the predictors. So in this case I need your opinion on how much we want to keep the old api. The easiest solution would be to just require users to pass the data in sklearn format to fit

pymc_marketing/mmm/base.py
twiecki: Rather, it's because it's optional as you can pass it into `__init__` or `fit()`, the latter to keep API compat with sklearn.

michaelraczycki:
Already while adjusting the tests to match the new model_data structure (with value, type and dims as fields of nested dict) I noticed that in my opinion it just looks messy, anyway to make the discussion possible I've decided to push it all. Do you think we'd have any benefit from having the type/dims in the model dataset, or for style purpose it's better to remove it?

michaelraczycki:
#309 implements the full solution from this initial concept

Issue #271, opened 2023-05-03
Allow prior for MMMs


ricardoV94:
Allow custom priors for the channel contribution in the `BaseDelayedSaturatedMMM` model.

PR #286, opened 2023-05-15
Allow custom priors of the `BaseDelayedSaturatedMMM` model

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/271

----
:books: Documentation preview :books:: https://pymc-marketing--286.org.readthedocs.build/en/286/

### DIFF
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -4,6 +4,7 @@ import numpy as np
 import numpy.typing as npt
 import pandas as pd
 import pymc as pm
+from pytensor.tensor import TensorVariable
 
 from pymc_marketing.mmm.base import MMM
 from pymc_marketing.mmm.preprocessing import MaxAbsScaleChannels, MaxAbsScaleTarget
@@ -21,6 +22,7 @@ class BaseDelayedSaturatedMMM(MMM):
         target_column: str,
         date_column: str,
         channel_columns: List[str],
+        channel_prior: Optional[TensorVariable] = None,
         validate_data: bool = True,
         control_columns: Optional[List[str]] = None,
         adstock_max_lag: int = 4,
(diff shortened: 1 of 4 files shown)


pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -60,9 +69,26 @@ def __init__(
...
+        if self.channel_prior is not None and self.channel_prior.shape != (
+            len(self.channel_columns),
+        ):
+            raise ValueError(
+                f"Channel prior shape {self.channel_prior.shape} does not match the number of channels {len(self.channel_columns)}"
+            )
+
+    def _preprocess_channel_prior(self) -> TensorVariable:
+        return (
+            pm.HalfNormal.dist(sigma=2, shape=len(self.channel_columns))
```
juanitorduz, line 88: @ricardoV94 Is there a nice way the user could simply pass `channel_prior=pm.HalfNormal.dist(sigma=2)` and add the shape internally? That is, we modify the user input and set `pm.HalfNormal.dist(sigma=2, shape=len(self.channel_columns)` as we currently have? Otherwise, the information seems redundant right?

twiecki:
This functionality is also added through https://github.com/pymc-labs/pymc-marketing/pull/285.

juanitorduz:
> This functionality is also added through #285.

Ok! Maybe this one could serve as a  placeholder (interim solution) as we do not have an ETA for the model builder integration. Custom priors were driving huge interest (one of the biggest strengths of the Bayesian approach).  An alternative is to wait for (or even support!) https://github.com/pymc-labs/pymc-marketing/pull/285 and close this one. No strong option :)

pymc_marketing/mmm/delayed_saturated_mmm.py
ricardoV94: Yes with `change_dist_size`, it's used inside `pm.Censored` if you want to see it being used.

juanitorduz: Amazing! (this is a hidden gem if you ask me ;)

ricardoV94: There are some gotchas, like you cant resize an intermediate variable after a dependency was created (another variable already depends on it), so we don't advertise it too much.

tests/mmm/test_base.py
```diff
@@ -131,10 +132,32 @@ def build_model(self, data, **kwargs):
     return mmm
 
 
+class TestBaseMMM:
+    def test_bad_inheritance(self) -> None:
```
ricardoV94, line 136: Nitpick: this seems unnecessary. It's testing what the `abstractmethod` decorator is supposed to do. 
https://docs.python.org/3/library/abc.html#abc.abstractmethod

ricardoV94 (approved):
LGTM, expect the new test that is testing a basic Python functionality?

juanitorduz:
Thanks @ricardoV94 ! It's about wrong incentives 🙈. I added this silly test to keep up the code coverage #shameshame 😂

Shall I remove it anyway ?

ricardoV94:
Ahaha, well the raise NotImplementedError isn't even needed (the abstractmethod will do that automatically). So you can get the coverage by removing it?

Issue #259, opened 2023-04-29
Is contribution curve (response curve) correct?


thipokKub:
From [here](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/base.py#L379), the plot seems to suggest 1:1 mapping between input, and output. However it doesn't account for adstock effect (lagging). If I were to correct this in toy example, it would give a different curve, like so

![image](https://user-images.githubusercontent.com/21255489/235313314-f4a0470f-938f-420a-bda8-17930bea16a9.png)

Where black line is an input, blue line is adstock transformed, and green line is adstock + saturation transformed. Now for the last plot, I'm not sure which response curve is correct

Reproduction code

```python
import numpy as np
import pandas as pd
import matplotlib.cm as mplcm
import matplotlib.pyplot as plt
import matplotlib.colors as colors
from scipy.optimize import curve_fit
from sklearn.preprocessing import MaxAbsScaler

def geometric_adstock(x, alpha: float = 0.0, l_max: int = 12):
    """Geometric adstock transformation."""
    # l_max = min(l_max, len(x))
    cycles = [
        np.concatenate(
            [np.zeros(i), x[: x.shape[0] - i]]
        )
        for i in range(l_max)
    ]
    x_cycle = np.stack(cycles)
    w = np.array([np.power(alpha, i) for i in range(l_max)])
    
    discounted = x.reshape(-1, 1) @ w.reshape(1, -1)
    contrib = np.zeros((len(discounted), len(discounted)))
    for i in range(len(discounted)):
        lim = min(len(discounted), i + l_max)
        contrib[i:lim, i] = discounted[i, :lim - i] 
    
    return np.dot(w, x_cycle), contrib

def saturation(x, k, s):
    return 1/(1 + (x/k)**(-s))

def best_fit(x, y, n_points: int = 100):
    xScaler, yScaler = MaxAbsScaler(), MaxAbsScaler()
    xScaler.fit(x.reshape(-1, 1)), yScaler.fit(y.reshape(-1, 1))
    x_scaled = xScaler.transform(x.reshape(-1, 1)).reshape(-1)
    y_scaled = yScaler.transform(y.reshape(-1, 1)).reshape(-1)
    
    xs = np.linspace(x_scaled.min(), x_scaled.max(), n_points)
    soln, _ = curve_fit(saturation, x_scaled, y_scaled, bounds=[[0.1, 0.1], [10, 10]])
[...]

PR #287, opened 2023-05-18
Creating budget allocation and optimal point estimations

carlangastr:
Hello, pymc team.

I made some changes that I believe can enhance the library, generating new insights for those who use it.

1. I added a function called '**elbow_function**': Its objective is to receive two parameters (x, y) and calculate the inflection point within a curve. I have attached a reference image.
2. I created the '**budget_allocator**' function: The objective is to take the estimated contribution curves for each channel, generated by your model, and maximize the contribution given a specific budget.

To do this, the function needs the estimation of the plateau point (where even if we increase the budget, the contribution does not increase) and the 25th percentile, which will be used as the maximum and minimum budget for each channel within the constraints in the 'budget_allocator' function.

I modified the '**plot_contribution_curves**(estimators = True)' function and added the 'estimators' parameter. Now, the function returns a tuple of a figure and a dataframe. When the new parameter is true, the dataframe will have information about the optimal point (estimated by 'elbow_function') and the plateau point.

I have attached two reference images to show how the new graph and respective dataframe would look like.

ps: Please don´t consider the typo on the image where is written ´plato´ and not ´plateau´. Sometimes Spanish and English words mixed in my head. Additionally, open to hearing new name propositions for the functions and estimated points.

image

image

image

----
:books: Documentation preview :books:: https://pymc-marketing--287.org.readthedocs.build/en/287/


carlangastr:
Still is not complete, I´ll validate it works properly during the day.

juanitorduz (changes requested):

juanitorduz:
Thank you for this PR. The optimization component is crucial and this is a great start! I just added some comments regarding the style. In addition, we should add some unit-test to ensure we are getting the expected behavior, especially regarding the `find_elbo` function.

Now a more conceptual comment regarding the discussion on https://github.com/pymc-labs/pymc-marketing/issues/259. The current channel contributions are describing the direct effect and do not account for the carryover effect (as described in the issue). The reason is that is very hard to account for it as information from the past can leak into the aggregated estimation.  One possibility is to rename the method/plot to `direct_cost_curves` and actually perform the optimization on the contribution of the total time frame when multiplying it by the scaling factor, see https://github.com/pymc-labs/pymc-marketing/issues/259#issuecomment-1529118848. I think is better (or also, in addition to this direct optimization approach), to optimize these curves as they do include the carryover effect. I would love to hear your thoughts :) 

FYI: @cluhmann ;)

pymc_marketing/mmm/base.py
```diff
@@ -374,11 +378,13 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
             coords=channel_contribution.coords,
         )
 
-    def plot_contribution_curves(self) -> plt.Figure:
+    def plot_contribution_curves(self, estimators = True) -> tuple:
```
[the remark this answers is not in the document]
carlangastr: Done!

pymc_marketing/mmm/base.py
```diff
@@ -390,24 +396,181 @@ def plot_contribution_curves(self) -> plt.Figure:
...
                 },
                 ax=ax,
             )
-            ax.legend(loc="upper left")
-            ax.set(title=f"{channel}", xlabel="total_cost_eur")
-
-        fig.suptitle("Contribution Plots", fontsize=16)
-        return fig
+            
+            if estimators:
```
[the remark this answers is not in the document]
carlangastr: Lo habia pensado pero no lo hice 😅 I did a new function in `utils` which now I use on `base` as well. At same time, I add a unit-test on the `test_utils` file, please help me checking. I'm not so familiarize with unit-test 🙂

pymc_marketing/mmm/base.py
```diff
@@ -390,24 +396,181 @@ def plot_contribution_curves(self) -> plt.Figure:
...
+                    title=f"{channel} Legend",
+                    fontsize='small'
+                )
+            
+        
+        axes.set(xlabel="Spent in Euros", ylabel="Contribution")
+        fig.suptitle("Response Curves", fontsize=16)
+        return fig, df_estimations
+    
+    def budget_allocator(self, total_budget = None, df_estimations = None, budget_bounds = None) -> pd.DataFrame:
```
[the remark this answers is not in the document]
carlangastr: Done!

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +34,34 @@ def generate_fourier_modes(
             for func in ("sin", "cos")
         }
     )
+
+def find_elbow(x, y):
```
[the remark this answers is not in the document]
carlangastr: Done!

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +34,34 @@ def generate_fourier_modes(
...
+    """
```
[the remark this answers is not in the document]
carlangastr: Done!

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +34,95 @@ def generate_fourier_modes(
             for func in ("sin", "cos")
         }
     )
+
+def find_elbow(x: np.array, y: np.array) -> int:
```
juanitorduz, line 38: I think we should use the hints as  in https://github.com/pymc-labs/pymc-marketing/actions/runs/5017993802/jobs/8997616662?pr=287. In this case `x:  npt.NDArray[np.float_]`. Otherwise `mypy` complains 
```python
pymc_marketing/mmm/utils.py:39: error: Function "numpy.core.multiarray.array" is not valid as a type  [valid-type]
```

juanitorduz: Same for all methods 😅

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +34,95 @@ def generate_fourier_modes(
...
+
+    # Define x spaces
+    x_space_actual = np.linspace(x.min(), x.max(), 100)
+    x_space_projected = np.linspace(min(x.min(), min(roots)), max(x.max(), max(roots)), 100)
+
+    # Calculate y spaces
+    y_space_actual = polynomial(x_space_actual)
+    y_space_projected = polynomial(x_space_projected)
+
+    return polynomial, x_space_actual, y_space_actual, x_space_projected, y_space_projected, roots
```
juanitorduz, line 128: This function returns too many things. Maybe is worth using a `CurveResult` class ?

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +34,95 @@ def generate_fourier_modes(
...
+    # Calculate the y-values of the line
+    line_y = np.poly1d(line_coeffs)(x)
+
+    # Calculate the distances from the points of the curve to the line
+    distances = distance.cdist(np.column_stack((x, y)), np.column_stack((x, line_y)), 'euclidean')
+
+    # Return the index of the point with the maximum distance to the line
+    return np.argmax(distances.diagonal())
+
+def calculate_curve(x: np.array, y: np.array) -> tuple[np.poly1d, np.ndarray, np.ndarray, np.ndarray, np.ndarray, list]:
```
juanitorduz, line 72: use `Tuple` instead to `tuple` as we want to support python3.8

pymc_marketing/mmm/base.py
```diff
@@ -390,24 +396,168 @@ def plot_contribution_curves(self) -> plt.Figure:
...
+                # Call the new function
+                polynomial, _, _, _, _, _ = calculate_curve(x, y)
+        
+                # Calculate the collaboration for this channel based on its quadratic curve
+                total_collaboration -= polynomial(allocation)
+            return total_collaboration
+
+        def budget_constraint(budget_allocations):
+            total_budget_allocated = np.sum(budget_allocations)
+            return total_budget_allocated - total_budget  # this should be zero
```
juanitorduz, line 528: Maybe we can extract these functions to a module e.g. `pymc_marketing/mmm/budget_optimizer`?

juanitorduz (changes requested):

juanitorduz:
Thanks! It is looking much nicer. We recommend installing the `pre-commit` hooks to have the style set up before pushing, see https://pre-commit.com/. 

- You can run it locally as `pre-commit run --all-files`.
- To run `pre-commit` as part of git workflow, use `pre-commit install`.

I see `mypy` is complaining about they types (see comments above) and this is also making tests fail (see for example https://github.com/pymc-labs/pymc-marketing/actions/runs/5017993802/jobs/8997617126?pr=287 which ran in your latest commit ;) )

carlangastr:
Hey! @juanitorduz 

First of all, really thank you for the feedback! I already apply the changes and test the code here. Everything looks in place but please take a look [here](https://colab.research.google.com/drive/1Q9_5tUczYjH2XQ5OsnesmGssyuu0qS7P#scrollTo=Fukgd-kWDvkT).

**Regarding the second point**, I also noted that point. Currently, as we were trying to bypass it, it was setting in the model configuration `adstock_max_lag=0` but applying the lag of the variable inside the dataframe, so the variable would be lagged from the fitting. Therefore, the value of the "immediate contribution curves" would be considering the lag. I'm not sure if it's a good option by the way, but it's what we've been testing. 

Perhaps we could create an additional function that retrains the model, receiving the lag parameter (This parameter would create a lag in the target variable) and after re-training with this lag information, it would return the curves only. I'm not sure if it's the best option but I think it could solve the problem in the short term, being flexible to use for those who want to take the time.

I was testing the code that you mention in the code with one of the models that I trained, in order to compare the results of these new curves with my results and I think that it is not working as in your example, for which I am a bit confused. Do you think we can see check it?

juanitorduz:
Regarding your comments:
> Regarding the second point, I also noted that point. Currently, as we were trying to bypass it, it was setting in the model configuration adstock_max_lag=0 but applying the lag of the variable inside the dataframe, so the variable would be lagged from the fitting. Therefore, the value of the "immediate contribution curves" would be considering the lag. I'm not sure if it's a good option by the way, but it's what we've been testing.
I think the whole point is that the carryover transformation depends on the lags scaled by the decay parameter alpha. Adding the lags is a different approach. I think a "better" way to add the lags is through the `delayed_adstock` (see https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/transformers.py#L103) where the `theta` parameter encodes the lag.

> I was testing the code that you mention in the https://github.com/pymc-labs/pymc-marketing/issues/259#issuecomment-1529118848 with one of the models that I trained, in order to compare the results of these new curves with my results and I think that it is not working as in your example, for which I am a bit confused. Do you think we can see check it?
Is there an error in the code (it might be, it was a draft)? If so can you share the message? Or is it that the results do not look good?  Maybe you add this into the colab notebook if you have not already ? (I will check it in the up coming days)

pymc_marketing/mmm/base.py
```diff
@@ -374,11 +377,17 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
             coords=channel_contribution.coords,
         )
 
-    def plot_contribution_curves(self) -> plt.Figure:
+    def plot_contribution_curves(
+        self, estimators: Optional[bool] = True
```
juanitorduz, line 381: Can you please add some doctrings to this method? I think users might find it useful (specially the meaning of the `estimators` argument)

pymc_marketing/mmm/base.py
```diff
@@ -390,24 +399,198 @@ def plot_contribution_curves(self) -> plt.Figure:
...
+                facecolor="white",
+                title=f"{channel} Legend",
+                fontsize="small",
+            )
+
+            ax.set(xlabel="Spent", ylabel="Contribution")
+        fig.suptitle("Immediate response curves", fontsize=16)
+        return fig, df_estimations
+
+    def budget_allocator(
```
juanitorduz, line 497: I suggest we move this out to an independent module `pymc_marketing/mmm/budget_optimizers`

pymc_marketing/mmm/base.py
```diff
@@ -390,24 +399,198 @@ def plot_contribution_curves(self) -> plt.Figure:
...
+        self,
+        total_budget: Optional[float] = None,
+        df_estimations: Optional[pd.DataFrame] = None,
+        budget_bounds: Optional[dict] = None,
```
juanitorduz, line 501: Use upper case type hints (Dict) since we want to support python3.8

pymc_marketing/mmm/base.py
```diff
@@ -390,24 +399,198 @@ def plot_contribution_curves(self) -> plt.Figure:
...
+        )
+
+        # Add the new column to the dataframe
+        df["Optimal Budget"] = np.round(result.x)
+
+        print("The maximum total contribution with these allocations is:")
+        print(-result.fun)
+
+        print("The proposed budget is:")
+        print(np.round(result.x.sum()))
```
juanitorduz, line 590: I am not super keen on having print statements ... shall we output a dataframe ?

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +34,131 @@ def generate_fourier_modes(
...
+        return self.y_space_actual
+
+    def get_x_space_projected(self):
+        return self.x_space_projected
+
+    def get_y_space_projected(self):
+        return self.y_space_projected
+
+    def get_roots(self):
+        return self.real_roots
```
juanitorduz, line 164: I do not think these are necessary as we can simply call the attribute right?

juanitorduz (changes requested):

carlangastr:
Added requested changes 🟢

1. Waiting for feedback regarding `budget_optimizers` module. 🟡
Screenshot 2023-05-19 at 22 16 34

pymc_marketing/mmm/base.py
```diff
@@ -511,7 +511,7 @@ def budget_allocator(
         self,
         total_budget: Optional[float] = None,
         df_estimations: Optional[pd.DataFrame] = None,
-        budget_bounds: Optional[Dict] = None,
+        budget_bounds: Optional[Dict] = {},
```
juanitorduz, line 520: **Small warning ;) :** it is not good practice to use mutable objects as defaults, see https://www.30secondsofcode.org/python/s/mutable-default-arguments/

You can keep it `None` and add inside the function:
```python
if budget_bounds is None:
    budget_bounds = {}
```

pymc_marketing/mmm/base.py
```diff
@@ -16,11 +16,17 @@
 import pymc as pm
 import seaborn as sns
 from pymc.util import RandomState
+
+from scipy.optimize import Bounds, minimize
```
juanitorduz, line 20: The only part of the linter is failing is `isort` (see ci message):
image
`from scipy.optimize import Bounds, minimize` should be below `from pytensor.tensor import TensorVariable` (I am curious why the pre-commit does not fail)

carlangastr:
Yes, really strange why pre-commit is working for me.

Example:
image

carlangastr:
Hey Juan! I did all the necessary adjustments to see everything is working.

I'm still missing your input here:
> > @juanitorduz Question, Do you mean by moving the whole function to a module but still keeping the output as a method on the class?
> 
> Just ask because if you observe the code, it relies on self for accessing attributes like `self.compute_channel_contribution_original_scale()`, and `self.data`. I could move it, but in other to get the information would end by the end of the day as a method under **BaseMMM**.

Based on that, I'll proceed.

cluhmann:
I am late to this (obviously), but looking over `budget_optimizer.py`, I am not entirely sure what the function is generating.  In large part, I think this is probably due to the use of the term "collaboration" (and associated variable names such as `total_collaboration`).  Any hints as to what this means?

carlangastr:
> I am late to this (obviously), but looking over `budget_optimizer.py`, I am not entirely sure what the function is generating. In large part, I think this is probably due to the use of the term "collaboration" (and associated variable names such as `total_collaboration`). Any hints as to what this means?

Hey @cluhmann You are right, the term `collaboration` is not the most appropriate. It should be related to `contribution` since the function basically seeks to find the maximum contribution possible from a channel given maximum advertising spending (cost) under certain constraints. While considering the potential growth and ranges of each one.

cluhmann:
Makes much more sense now!

Next question: what exactly is `find_elbow()` calculating?  I understand the mechanics, but I think it's going to be confusing to just tell users that this "optimizes" the spend because that implies that you are minimizing or maximizing some quantity.  If I understand correctly, `find_elbow()` returns the maximum spend subject to the constraint that the ROAS is equal to what one would expect if the contribution function were linear between its minimum and maximum.  One thing that sticks out to me is that this linear interpolation is made between the arbitrary (?) `[x[0], y[0]]` and `[x[-1], y[-1]]`.  So what exactly can we guarantee about the returned "elbow"?  I guess I am looking for more conceptual description rather than a mathematical description.

carlangastr:
> Makes much more sense now!
> 
> Next question: what exactly is `find_elbow()` calculating? I understand the mechanics, but I think it's going to be confusing to just tell users that this "optimizes" the spend because that implies that you are minimizing or maximizing some quantity. If I understand correctly, `find_elbow()` returns the maximum spend subject to the constraint that the ROAS is equal to what one would expect if the contribution function were linear between its minimum and maximum. One thing that sticks out to me is that this linear interpolation is made between the arbitrary (?) `[x[0], y[0]]` and `[x[-1], y[-1]]`. So what exactly can we guarantee about the returned "elbow"? I guess I am looking for more conceptual description rather than a mathematical description.

Hey @cluhmann, I'll split the response by points.

**Regarding the first point**, the `find_elbow()` function is essentially trying to identify the point on a curve (in the 2D space, defined by your x and y coordinates) where the curve seems to change its direction the most - referred to as the "elbow" of the curve. 

This point usually divides the curve into two sections: one where we can still achieve results (response) at a high cost, and one where we can achieve cost-effective (cheap) results but in smaller volumes. That's what I show in the first image of the description. **How is calculated?** The function looks at the first and the last points and creates a straight line between them. The `elbow` point is the point that is the furthest away from this line. So it represents the point where the actual curve (Given by the MMM model) deviates the most from a straight line drawn between the start and end points of the curve. Again, try to follow the picture in the first comment.

**Second point**, how the optimization is happening? `budget_allocator` (Inside base.py) function is designed to optimally allocate a given total budget across different channels. It uses information about the channels, including estimated contribution and budget constraints. It tries to find the set of budget allocations that will maximize the total contribution, subject to certain constraints, such as not exceeding the total budget and not going outside the bounds of each channel (Which can be set by the user or by default as the plateau point as maximum and the quantile 25th as minimum spent). In a nutshell, it aims to distribute the budget in a way that maximizes some measure of 'contribution'. What did I mean by contribution? The possible estimate results by the model to each channel.

In the implementation, I use the `scipy.optimize.minimize` function, which is a general-purpose function for solving optimization problems. In simple words, it tries to find the input values that will make the output of a given function as small as possible. In the context of `budget_allocator` function, it's being used to find the budget allocations that will make the 'cost' as small as possible (or equivalently, make the 'contribution' as large as possible). It does this by iteratively adjusting the input values, computing the function output, and then making further adjustments based on the results.

Hope this adds more clarity! Let me know 🚀

carlangastr:
Solving error on unit test, basically the function `plot_contribution_curves()` is returning a tuple of (plt.Figure, pd.DataFrame) but your assertion is checking if the return type is a single `plt.Figure` object. This is why the test was failing.

Screenshot 2023-06-03 at 21 17 28

The new code will now check if the result is a tuple and if its first element is a `plt.Figure` and the second is a `pd.DataFrame`, only when the function is `plot_contribution_curves()`. For all other functions, it will check if the result is a `plt.Figure` object.

pymc_marketing/mmm/base.py
```diff
@@ -393,24 +415,206 @@ def plot_contribution_curves(self) -> plt.Figure:
...
+        Allocate the budget optimally among different channels based on estimations and budget constraints.
+
+        Parameters
+        ----------
+        total_budget : float, optional
+            The total budget available for allocation.
+        df_estimations : pd.DataFrame, optional
+            A DataFrame containing estimations and information about different channels.
+        budget_bounds : dict, optional
+            A dictionary specifying the budget bounds for each channel.
```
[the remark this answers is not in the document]
juanitorduz: My hunch is that they should not be allowed to be `None` right?

pymc_marketing/mmm/base.py
```diff
@@ -393,24 +415,206 @@ def plot_contribution_curves(self) -> plt.Figure:
...
+
+        contributions = pd.DataFrame(
+            {
+                "estimated_total_budget": np.round(result.x.sum()),
+                "estimated_max_contribution": (-result.fun),
+            },
+            index=[0],
+        )
+
+        result_dict = {"allocation": df, "contribution": contributions}
```
juanitorduz, line 614: you can simply `return {"allocation": df, "contribution": contributions}` so that we do not have unnecessary temporal variables ;)

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Returns
+    -------
+    float
+        The total collaboration of the budget allocations.
+
+    Notes
+    -----
+    This function calculates the total collaboration of budget allocations for each channel based on a quadratic curve fitted to the channel data.
+    The total collaboration is computed as the sum of the polynomial evaluation at the allocation for each channel.
```
juanitorduz, line 39: You can add these remarks at the top of the doc strings and not as notes.

pymc_marketing/mmm/base.py
```diff
@@ -393,24 +415,206 @@ def plot_contribution_curves(self) -> plt.Figure:
...
+
+        Notes
+        -----
+        - The function allocates the budget optimally among different channels based on the provided estimations and budget constraints.
+        - It performs validation checks on the input parameters to ensure they are of the correct type and not None.
+        - The channel contributions are computed using the `compute_channel_contribution_original_scale` method.
+        - The lower and upper bounds for the budget allocation are determined based on the provided `budget_bounds` or default values.
+        - Optimization is performed to find the optimal budget allocation that maximizes contribution while satisfying the budget constraint.
+        - The results are stored in a dictionary containing the allocated budget and contribution information.
+        - The updated DataFrame with the allocated budget and contribution information is returned as part of the dictionary.
```
juanitorduz, line 549: You can add these remarks at the top of the doc strings and not as notes.

pymc_marketing/mmm/budget_optimizer.py
juanitorduz: The word collaboration is confusing. Can you change it please? Maybe "contribution"?

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import List
+
+import numpy as np
+from pandas import DataFrame
+from xarray import DataArray
+
+from pymc_marketing.mmm.utils import CurveCalculator
+
+
+def cost_function(
```
juanitorduz, line 11: As we are using the class `CurveCalculator` to have a polynomial fit. Can we add a more descriptive name to this method? Something like `compute_total_cost_after_budget_optimization` or something like that (be creative ;) )

juanitorduz: also, can we add unit-tests for this function as well :)

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,79 @@
...
+    float
+        The difference between the total allocated budget and the total available budget.
+
+    Notes
+    -----
+    This function calculates the difference between the total allocated budget and the total available budget.
+    This difference should be zero in the optimization problem.
+    """
+    total_budget_allocated = np.sum(budget_allocations)
+    return total_budget_allocated - total_budget  # this should be zero
```
juanitorduz, line 79: Can you remove the comment `  # this should be zero` as its confusing :)

juanitorduz: You can add these remarks at the top of the doc strings and not as notes.

carlangastr:
## **Last Updates:**
Adding the first round of modifications based on the feedback provided by Juan.

1. Creating the function `_plot_estimations` to calculate the plot of plateau & elbow point independently from the contribution plot.
2. Adding a parameter on `plot_contribution_curves`, now if `show_estimations` is `True` then `_plot_estimations` will append the figure to the contribution plot, otherwise remain the same.
3. New method was created `compute_channel_estimate_points_original_scale` which returns a dataframe with information about plateau & elbow points.
3. The `CostFunction` on the budget optimizer was renamed by `compute_optimal_contribution_based_budget_allocation`

PS: The notes on the function are deleted as well and added on the top of the docstring.

## **Next Steps:**
1. Create a unit test for all the changes made it.
2. Test the code to check all results are consistent and correct.
3. Solve conflicting errors with the main branch.

ricardoV94:
@carlangastr sounds like a lot of (valuable) work! Let us know if you need any help or have it all figured out.

carlangastr:
Hey @ricardoV94 really thank you!

There are still things we can check, I think we can talk about them at the marketing meeting? The one @juanitorduz was mentioning. 

If so, I would have time to add some changes to some functions during this week, which I would like to show you. Because this week a wasn't able to add the promised for the this week (July 10th).

PR #290, opened 2023-05-24
MMM Prior Specification to MMM Notebook Example

juanitorduz:
Follow-up of https://github.com/pymc-labs/pymc-marketing/pull/286. 

On this PR we add the prior specification capabilities into the MMM example. We follow a simple heuristic argument similar to Google Lightweight MMM, see https://github.com/google/lightweight_mmm/blob/main/lightweight_mmm/models.py#L350

----
:books: Documentation preview :books:: https://pymc-marketing--290.org.readthedocs.build/en/290/


ricardoV94 (approved):

juanitorduz:
@carlangastr Thanks for your question! We have not included yet any example notebook on calibration with experiments as this heavily depends on the context and the data. Still, it is our roadmap.

Regarding the custom distributions, it should be possible to add the gamma prior with vector parameters as you suggest using the dist API as in the example. If for whatever reason it does not work please raise an issue :)

Issue #294, opened 2023-06-05
Encapsulate beta prior in a dedicated method.


ferrine:
https://github.com/pymc-labs/pymc-marketing/blob/d4feeeafad6f0299b115dac3582ad6d98bd8241f/pymc_marketing/mmm/delayed_saturated_mmm.py#L123-L126

Right now it is mixed logic of creating RV and registering it in the model. In principle that should be the only one method that returns already registered RV. That will allow a user to override prior specification

juanitorduz:
Thanks for the feedback! 👍 This was a change introduced in https://github.com/pymc-labs/pymc-marketing/pull/286 as an attempt to make the API somehow easy for the user. We will probably change it in https://github.com/pymc-labs/pymc-marketing/pull/295

michaelraczycki:
The channel prior was completely removed in #295 , as model_config from model_builder allows for customisation of every parameter

wd60622:
Closed with #397

PR #295, opened 2023-06-05
MMM sklearn approach

michaelraczycki:
This PR contains complete rework of MMM-based classes to adopt sklearn-based api, with full integration with ModelBuilder. Some tests might not be passing yet, as ModelBuilder was also slightly modified, because of that pymc-experimental needs to get a release. Locally all the tests were passing, still ToDo are:
Adaptation of notebooks, adding docstrings, more test coverage. Please share your thoughts on solutions included in this PR

@twiecki @ricardoV94 @juanitorduz 

----
:books: Documentation preview :books:: https://pymc-marketing--295.org.readthedocs.build/en/295/

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -17,3 +17,6 @@ ignore_missing_imports = True
 
 [mypy-scipy.*]
 ignore_missing_imports = True
+
+[mypy-pymc_experimental.*]
+ignore_missing_imports = True
--- a/pymc_marketing/mmm/base.py
+++ b/pymc_marketing/mmm/base.py
@@ -1,4 +1,3 @@
-from abc import abstractmethod
 from inspect import (
     getattr_static,
     isdatadescriptor,
@@ -15,10 +14,10 @@ import numpy as np
 import pandas as pd
 import pymc as pm
 import seaborn as sns
-from pymc.util import RandomState
+from pymc_experimental.model_builder import ModelBuilder
 from sklearn.pipeline import Pipeline
 from sklearn.preprocessing import FunctionTransformer
-from xarray import DataArray
+from xarray import DataArray, Dataset
 
 from pymc_marketing.mmm.validating import (
     ValidateChannelColumns,
@@ -29,35 +28,27 @@ from pymc_marketing.mmm.validating import (
 __all__ = ("BaseMMM", "MMM")
 
 
-class BaseMMM:
+class BaseMMM(ModelBuilder):
     model: pm.Model
 
     def __init__(
         self,
-        data: pd.DataFrame,
-        target_column: str,
         date_column: str,
         channel_columns: Union[List[str], Tuple[str]],
+        data: pd.DataFrame = None,
         validate_data: bool = True,
+        model_config: Optional[Dict] = None,
+        sampler_config: Optional[Dict] = None,
         **kwargs,
     ) -> None:
-        self.data: pd.DataFrame = data
-        self.target_column: str = target_column
+        self.X: Optional[pd.DataFrame] = None
+        self.y: Optional[pd.Series] = None
         self.date_column: str = date_column
         self.channel_columns: Union[List[str], Tuple[str]] = channel_columns
-        self.n_obs: int = data.shape[0]
         self.n_channel: int = len(channel_columns)
         self._fit_result: Optional[az.InferenceData] = None
         self._posterior_predictive: Optional[az.InferenceData] = None
-
-        if validate_data:
-            self.validate(self.data)
-        self.preprocessed_data = self.preprocess(self.data.copy())
-
-        self.build_model(
-            data=self.preprocessed_data,
-            **kwargs,
-        )
+        super().__init__(model_config=model_config, sampler_config=sampler_config)
 
     @property
     def methods(self) -> List[Any]:
@@ -75,22 +66,151 @@ class BaseMMM:
         ]
 
     @property
-    def validation_methods(self) -> List[Callable[["BaseMMM", pd.DataFrame], None]]:
-        return [
-            method
-            for method in self.methods
-            if getattr(method, "_tags", {}).get("validation", False)
-        ]
+    def validation_methods(
+        self,
+    ) -> Tuple[
+        List[Callable[["BaseMMM", Union[pd.DataFrame, pd.Series]], None]],
+        List[Callable[["BaseMMM", Union[pd.DataFrame, pd.Series]], None]],
+    ]:
+        """
+        A property that provides validation methods for features ("X") and the target variable ("y").
+
+        This property scans the methods of the object and returns those marked for validation.
+        The methods are marked by having a _tags dictionary attribute, with either "validation_X" or "validation_y" set to True.
+        The "validation_X" tag indicates a method used for validating features, and "validation_y" indicates a method used for validating the target variable.
+
+        Returns
+        -------
+        tuple of list of Callable[["BaseMMM", pd.DataFrame], None]
+            A tuple where the first element is a list of methods for "X" validation, and the second element is a list of methods for "y" validation.
+
+        """
+        return (
+            [
+                method
+                for method in self.methods
+                if getattr(method, "_tags", {}).get("validation_X", False)
+            ],
+            [
+                method
+                for method in self.methods
+                if getattr(method, "_tags", {}).get("validation_y", False)
+            ],
+        )
+
+    def validate(self, target: str, data: Union[pd.DataFrame, pd.Series]) -> None:
+        """
+        Validates the input data based on the specified target type.
+
+        This function loops over the validation methods specified for
+        the target type and applies them to the input data.
+
+        Parameters
+        ----------
+        target : str
+            The type of target to be validated.
+            Expected values are "X" for features and "y" for the target variable.
+        data : Union[pd.DataFrame, pd.Series]
+            The input data to be validated.
+
+        Raises
+        ------
+        ValueError
+            If the target type is not "X" or "y", a ValueError will be raised.
+        """
+        if target not in ["X", "y"]:
+            raise ValueError("Target must be either 'X' or 'y'")
+        if target == "X":
+            validation_methods = self.validation_methods[0]
+        elif target == "y":
+            validation_methods = self.validation_methods[1]
+
+        for method in validation_methods:
+            method(self, data)
 
     @property
     def preprocessing_methods(
         self,
-    ) -> List[Callable[["BaseMMM", pd.DataFrame], pd.DataFrame]]:
-        return [
-            method
-            for method in self.methods
-            if getattr(method, "_tags", {}).get("preprocessing", False)
-        ]
+    ) -> Tuple[
+        List[
+            Callable[
+                ["BaseMMM", Union[pd.DataFrame, pd.Series]],
+                Union[pd.DataFrame, pd.Series],
+            ]
+        ],
+        List[
+            Callable[
+                ["BaseMMM", Union[pd.DataFrame, pd.Series]],
+                Union[pd.DataFrame, pd.Series],
+            ]
+        ],
+    ]:
+        """
+        A property that provides preprocessing methods for features ("X") and the target variable ("y").
+
+        This property scans the methods of the object and returns those marked for preprocessing.
+        The methods are marked by having a _tags dictionary attribute, with either "preprocessing_X" or "preprocessing_y" set to True.
+        The "preprocessing_X" tag indicates a method used for preprocessing features, and "preprocessing_y" indicates a method used for preprocessing the target variable.
+
+        Returns
+        -------
+        tuple of list of Callable[["BaseMMM", pd.DataFrame], pd.DataFrame]
+            A tuple where the first element is a list of methods for "X" preprocessing, and the second element is a list of methods for "y" preprocessing.
+        """
+        return (
+            [
+                method
+                for method in self.methods
+                if getattr(method, "_tags", {}).get("preprocessing_X", False)
+            ],
+            [
+                method
+                for method in self.methods
+                if getattr(method, "_tags", {}).get("preprocessing_y", False)
+            ],
+        )
+
+    def preprocess(
+        self, target: str, data: Union[pd.DataFrame, pd.Series]
+    ) -> Union[pd.DataFrame, pd.Series]:
+        """
+        Preprocess the provided data according to the specified target.
+
+        This method applies preprocessing methods to the data ("X" or "y"), which are specified in the preprocessing_methods property of this object.
+        It iteratively applies each method in the appropriate list (either for "X" or "y") to the data.
+
+        Parameters
+        ----------
+        target : str
+            Indicates whether the data represents features ("X") or the target variable ("y").
+
+        data : pd.DataFrame
+            The data to be preprocessed.
+
+        Returns
+        -------
+        Union[pd.DataFrame, pd.Series]
+            The preprocessed data.
+
+        Raises
+        ------
+        ValueError
+            If the target is neither "X" nor "y".
+
+        Example
+        -------
+        >>> data = pd.DataFrame({"x1": [1, 2, 3], "y": [4, 5, 6]})
+        >>> self.preprocess("X", data)
+        """
+        if target == "X":
+            for method in self.preprocessing_methods[0]:
+                data = method(self, data)
+        elif target == "y":
+            for method in self.preprocessing_methods[1]:
+                data = method(self, data)
+        else:
+            raise ValueError("Target must be either 'X' or 'y'")
+        return data
 
     def get_target_transformer(self) -> Pipeline:
         try:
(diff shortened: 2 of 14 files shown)


pymc_marketing/mmm/preprocessing.py
```diff
@@ -5,40 +5,45 @@
...
+
+
+def preprocessing_method_y(method: Callable) -> Callable:
+    if not hasattr(method, "_tags"):
+        method._tags = {}  # type: ignore
+    method._tags["preprocessing_y"] = True  # type: ignore
     return method
 
 
 class MaxAbsScaleTarget:
```
twiecki, line 30: Is there a path to replace these with sklearn scalers?

twiecki:
I think this approach looks much clearer already. Curious what @lucianopaz and @juanitorduz think of this direction of going full sklearn API.

michaelraczycki:
The property "output_var" is needed for load function, as in this way it doesn't require the class to be initialised.

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -63,80 +63,154 @@ def __init__(
...
 
             channel_adstock = pm.Deterministic(
                 name="channel_adstock",
                 var=geometric_adstock(
                     x=channel_data_,
                     alpha=alpha,
-                    l_max=adstock_max_lag,
+                    l_max=kwargs["adstock_max_lag"]
+                    if "adstock_max_lag" in kwargs
+                    else 4,
```
juanitorduz, line 213: Do not set a default value. Use must specify it!

pymc_marketing/mmm/base.py
```diff
@@ -30,37 +29,27 @@
...
     def __init__(
         self,
-        data: pd.DataFrame,
-        target_column: str,
         date_column: str,
         channel_columns: Union[List[str], Tuple[str]],
         channel_prior: Optional[TensorVariable] = None,
-        validate_data: bool = True,
+        model_config: Optional[Dict] = None,
+        sampler_config: Optional[Dict] = None,
```
juanitorduz, line 41: Over all the PR looks consistent except for this part. Why we should not specify the target name but we should specify the date column and channel columns?

juanitorduz (commented):
Here is my first take (please bear in mind I am not really experienced in designing user-facing APIs). In the class construction: Why should we not specify the target name but we should specify the date column and channel columns? It feels more natural to leave the target name as input IMO even though the data `y` is passed after. I just want to make sure the users feel the API as natural as possible. 

A possibility is to have a `MMMConfig` class which carries all these channel names and yearly seasonality params so that we simply have 

```python
config = MMMConfig(channel_names=["ch1", "ch2"])
my_mmm = MMM(config=config)
my_mmm.fit(X, y)
```

It would be very neat if we could use [`pydantic`](https://docs.pydantic.dev/latest/) for the `MMMConfig `class as it offers data validation ;) 


Besides this question and comment (where feedback is would be much appreciated!) we should make sure
- [ ] All test and lint passes.
- [ ] We also make sure we can reproduce all the results of the mmm example notebook and add the corresponding changes.

pymc_marketing/mmm/base.py
michaelraczycki: now target name needs to be specified on the class level. The main issue is that load() from model builder is a static function, so can't access the user-passed target_column. @property function output_var was created to address this issue, as with it I can split assign the data from fit_data (part of idata) into self.X and self.y

michaelraczycki: also, correct me if I'm wrong but shouldn't y be a single column df, or a pd.Series? In that case it's not needed to specify which column is responsible for target, if there's only one

michaelraczycki: In my opinion, because the parameters that MMMconfig would be holding are simply strings, it just adds an extra step for the user without real benefit in exchange. Parameter encapsulation might be a good idea, if gathering of those params would be a complex process with some preprocessing etc. spanning across multiple functions.

michaelraczycki: and regarding target_column being removed from params: the fact that we require date_column and channel_columns is based on the real need, as otherwise we'd make very rigid class that hardcodes a lot of information. X has multiple columns and users should be allowed to call them inside their dataset as they wish. With y there's no such need, and including a parameter that has no real use inside the class would be just strange.

juanitorduz: Thanks for the clarification! The mmm config approach would be indeed useful if we do not simply use strings. 

In this case I'll suggest with the two open points (tests and mmm notebook) to see how it would look at the end. At the moment I do not see any blockers. Still I would like to hear from other devs with more experience with user-facing API design as my experience is rather limited.

pymc_marketing/mmm/base.py
```diff
@@ -78,132 +67,245 @@ def methods(self) -> List[Any]:
...
-        )["likelihood"]
+        likelihood_hdi_94: DataArray = az.hdi(ary=prior_predictive_data, hdi_prob=0.94)[
+            "likelihood"
+        ]
+        likelihood_hdi_50: DataArray = az.hdi(ary=prior_predictive_data, hdi_prob=0.50)[
+            "likelihood"
+        ]
 
         fig, ax = plt.subplots(**plt_kwargs)
+        if self.X is not None and self.y is not None:
```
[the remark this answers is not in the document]
michaelraczycki: it is used exactly 2 times, I could add it but I actually did it on purpose, as the solution has exactly the same number of codeines, as it removes.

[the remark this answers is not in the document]
juanitorduz: Ok! I thought I saw more 🙈

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: max_adstock_lag is now a mandatory input parameter

mypy.ini
```diff
@@ -17,3 +17,9 @@ ignore_missing_imports = True
 
 [mypy-scipy.*]
 ignore_missing_imports = True
+
+[mypy-pandas.*]
+ignore_missing_imports = True
```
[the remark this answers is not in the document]
michaelraczycki: done

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -63,80 +63,147 @@ def __init__(
...
+            )
 
+            """beta_channel = pm.HalfNormal(
+                name="beta_channel",
+                sigma=model_config["beta_channel"]["sigma"],
+                dims=model_config["beta_channel"]["dims"],
+            )  # ? Allow prior depend on channel costs?"""
             channel_prior = self._preprocess_channel_prior()
             beta_channel = self.model.register_rv(
                 rv_var=channel_prior, name="beta_channel", dims="channel"
```
michaelraczycki, line 182: @juanitorduz I need your input on this one, as the change was introduced by you. "beta)channel" when uncommented throws an error that the name is already registered. How should I modify it so it works as intended and doesn't crash?

pymc_marketing/mmm/preprocessing.py
juanitorduz: I think for now we do not need to worry about the scalers and we should take this discussion to another PR :) (it is an important topic!)

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: The whole point of lines 180 - 182 is to add custom priors (see unit test and notebook example). As far as I understand the model builder approach already supports custom priors right? In this case, you can simply delete the lines added in  https://github.com/pymc-labs/pymc-marketing/pull/286 (see also feedback from https://github.com/pymc-labs/pymc-marketing/issues/294)

pymc_marketing/mmm/base.py
```diff
@@ -78,132 +67,241 @@ def methods(self) -> List[Any]:
...
+        The 'validation_X' tag indicates a method used for validating features, and 'validation_y' indicates a method used for validating the target variable.
+
+        Returns
+        -------
+        tuple of list of Callable[["BaseMMM", pd.DataFrame], None]
+            A tuple where the first element is a list of methods for 'X' validation, and the second element is a list of methods for 'y' validation.
+
+        Example
+        -------
+        >>> self.validation_methods
```
juanitorduz, line 90: This is not very informative. I suggest removing it.

pymc_marketing/mmm/base.py
```diff
@@ -78,132 +67,241 @@ def methods(self) -> List[Any]:
...
+                if getattr(method, "_tags", {}).get("validation_X", False)
+            ],
+            [
+                method
+                for method in self.methods
+                if getattr(method, "_tags", {}).get("validation_y", False)
+            ],
+        )
+
+    def validate(self, target: str, data: Union[pd.DataFrame, pd.Series]):
```
juanitorduz, line 105: can you add a `None` return as a type hint ;)

pymc_marketing/mmm/base.py
```diff
@@ -78,132 +67,241 @@ def methods(self) -> List[Any]:
...
+        tuple of list of Callable[["BaseMMM", pd.DataFrame], pd.DataFrame]
+            A tuple where the first element is a list of methods for 'X' preprocessing, and the second element is a list of methods for 'y' preprocessing.
 
-    def preprocess(self, data: pd.DataFrame) -> pd.DataFrame:
-        for method in self.preprocessing_methods:
-            data = method(self, data)
-        return data
+        Example
+        -------
+        >>> self.preprocessing_methods
```
juanitorduz, line 166: Again, I would remove all of these examples which bring no value.

pymc_marketing/mmm/base.py
```diff
@@ -78,132 +67,241 @@ def methods(self) -> List[Any]:
...
-    @abstractmethod
-    def build_model(self, *args, **kwargs) -> None:
-        raise NotImplementedError()
+        data : pd.DataFrame
+            The data to be preprocessed.
 
-    def get_prior_predictive_data(self, *args, **kwargs) -> az.InferenceData:
+        Returns
+        -------
+        pd.DataFrame
```
juanitorduz, line 200: Return doctring does not match retunr type `Union[pd.DataFrame, pd.Series]`

pymc_marketing/mmm/base.py
```diff
@@ -78,132 +67,241 @@ def methods(self) -> List[Any]:
...
+
+        Raises
+        ------
+        ValueError
+            If the target is neither 'X' nor 'y'.
+
+        Example
+        -------
+        >>> data = pd.DataFrame({"x1": [1, 2, 3], "y": [4, 5, 6]})
+        >>> self.preprocess('X', data)
```
juanitorduz, line 211: Can we use " for strings? Should be `self.preprocess ("X", data)`

pymc_marketing/mmm/base.py
```diff
@@ -78,132 +67,241 @@ def methods(self) -> List[Any]:
...
+    @property
+    def prior_predictive(self) -> az.InferenceData:
+        if self.idata is None or "prior_predictive" not in self.idata:
+            raise RuntimeError("The model hasn't been fit yet, call .fit() first")
+        return self.idata["prior_predictive"]
 
     @property
-    def fit_result(self) -> az.InferenceData:
-        if self._fit_result is None:
+    # requesting discussion, is it something that we want to keep? Should I put on the deprecation warning to phase it out in later versions?
```
juanitorduz, line 237: Just for my understanding ... why should we remove it?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -210,16 +313,84 @@ def _get_fourier_models_data(self) -> pd.DataFrame:
...
+        ------
+        RuntimeError
+            If the data for the channel is not provided in `X`.
+        TypeError
+            If `X` is not a pandas DataFrame or a numpy array, or
+            if `y` is not a pandas Series or a numpy array and is not None.
+
+        Returns
+        -------
+        None
```
juanitorduz, line 363: I think we can also remove these None returns

juanitorduz (changes requested):
Added small comments regarding style! I will test the changes vis the notebook.

juanitorduz:
@michaelraczycki  When I run the model in the notebook it works as expected! 😄 !

Nevertheless, there is a bug on the y-scaler. When I run 

```python
mmm.plot_posterior_predictive(original_scale=True);
```
I get 

image

We need to make sure we do not overwrite the target data with its scaled version. This is also reflected in the contributions over time and ROAS (from the notebook)

<details>
<summary>Click me to see the code</summary>

```python
channels_contribution_original_scale = mmm.compute_channel_contribution_original_scale()
channels_contribution_original_scale_hdi = az.hdi(ary=channels_contribution_original_scale)

fig, ax = plt.subplots(
    nrows=2, figsize=(15, 8), ncols=1, sharex=True, sharey=False, layout="constrained"
)

for i, x in enumerate(["x1", "x2"]):
    # Estimate true contribution in the original scale from the data generating process
    sns.lineplot(
        x=df["date_week"],
        y=amplitude * betas[i] * df[f"{x}_adstock_saturated"],
        color="black",
        label=f"{x} true contribution",
        ax=ax[i],
    )
    # HDI estimated contribution in the original scale
    ax[i].fill_between(
        x=df["date_week"],
        y1=channels_contribution_original_scale_hdi.sel(channel=x)["x"][:, 0],
        y2=channels_contribution_original_scale_hdi.sel(channel=x)["x"][:, 1],
        color=f"C{i}",
        label=f"{x} $94%$ HDI contribution",
        alpha=0.4,
    )
    # Mean estimated contribution in the original scale
    sns.lineplot(
        x=df["date_week"],
        y=get_mean_contributions_over_time_df[x].to_numpy(),
        color=f"C{i}",
        label=f"{x} posterior mean contribution",
        alpha=0.8,
        ax=ax[i],
    )
    ax[i].legend(loc="center left", bbox_to_anchor=(1, 0.5))
    ax[i].set(title=f"Channel {x}")
```

image


</details>

<details>
<summary>Click me to see the code</summary>

```python
channel_contribution_original_scale = mmm.compute_channel_contribution_original_scale()

roas_samples = (
    channel_contribution_original_scale.stack(sample=("chain", "draw")).sum("date")
    / data[["x1", "x2"]].sum().to_numpy()[..., None]
)

fig, ax = plt.subplots(figsize=(10, 6))
sns.histplot(
    roas_samples.sel(channel="x1").to_numpy(), binwidth=0.05, alpha=0.3, kde=True, ax=ax
)
sns.histplot(
    roas_samples.sel(channel="x2").to_numpy(), binwidth=0.05, alpha=0.3, kde=True, ax=ax
)
ax.axvline(x=roas_1, color="C0", linestyle="--", label=r"true ROAS $x_{1}$")
ax.axvline(x=roas_2, color="C1", linestyle="--", label=r"true ROAS $x_{2}$")
ax.legend(loc="center left", bbox_to_anchor=(1, 0.5))
ax.set(title="Posterior ROAS distribution", xlabel="ROAS");
```

</details>

image

pymc_marketing/mmm/base.py
michaelraczycki: sure, I'll try to pay attention to that from now on, I'm 'energy efficient' so when I need to press only one button I choose ' over " intuitively :)

pymc_marketing/mmm/base.py
michaelraczycki: honestly I didn't see much difference between calling baseobject.idata.posterior and baseobject.fit_result, so that's why I put the comment to make sure that it's something that we still want there.

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: Most of the docstrings are auto-generated by chatGPT, I'm doing an experiment where I try to see how much of the boring stuff can I outsource to an AI. Obviously it needs some supervision, but as I didn't see this stage of PR as the final one I didn't review every single one of them

pymc_marketing/mmm/base.py
michaelraczycki: I agree, removed.

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: ok, beta_priors were removed and the build_model lines creating beta_channel based on model_config were uncommented

pymc_marketing/mmm/base.py
michaelraczycki: I've replaced all ' for " (where it was possible)

michaelraczycki:
<p>If we do that, 2 cells above will stop working, as there won't be model available to pass to model_to_graphviz, or to check its type</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/295/discussion/#comment-1587933983'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

michaelraczycki:
> @michaelraczycki When I run the model in the notebook it works as expected! 😄 !
> 
> Nevertheless, there is a bug on the y-scaler. When I run
> 
> ```python
> mmm.plot_posterior_predictive(original_scale=True);
> ```
> 
> I get
> 
> image
> 
> We need to make sure we do not overwrite the target data with its scaled version. This is also reflected in the contributions over time and ROAS (from the notebook)
> 
> Click me to see the code
> image
> 
> Click me to see the code
> image
> 
> **The rest of the notebook results look fine!** 💪

Could you share your system and env configuration? it doesn't work for me which makes it hard to debug the scaler problem

michaelraczycki:
<p>do you mean only the line 3 or all sigmas in the model are supposed to always be equal to the sigma? They all have sigma = 2, so I wasn't sure if it's a coincidence, or they have always the same sigma</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/295/discussion/#comment-1591846838'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>I mean the <code>sigma</code> fot the <code>beta_channel</code> variable (see https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_example.html#id1)  </p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/295/discussion/#comment-1591846838'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
@michaelraczycki One question. In the current notebook you have 

```python
model_config= {
            "intercept": {"mu": 0, "sigma": 2},
            "beta_channel": {"sigma": model_config['sigma']['sigma'], "dims": ("channel",)},
            "alpha": {"alpha": 1, "beta": 3, "dims": ("channel",)},
            "lam": {"alpha": 3, "beta": 1, "dims": ("channel",)},
            "sigma": {"sigma": 2},
            "gamma_control": {
                "mu": 0,
                "sigma": 2,
                "dims": ("control",),
            },
            "mu": {"dims": ("date",)},
            "likelihood": {"dims": ("date",)},
            "gamma_fourier": {"mu": 0, "b": 1, "dims": "fourier_mode"},
        }
```
But this looks a bit of circular with respect to `model_config` as it is used in its definition. What about simply having:

```python
model_config= {
            "intercept": {"mu": 0, "sigma": 2},
            "beta_channel": {"sigma": prior_sigma, "dims": ("channel",)},
            "alpha": {"alpha": 1, "beta": 3, "dims": ("channel",)},
            "lam": {"alpha": 3, "beta": 1, "dims": ("channel",)},
            "sigma": {"sigma": 2},
            "gamma_control": {
                "mu": 0,
                "sigma": 2,
                "dims": ("control",),
            },
            "mu": {"dims": ("date",)},
            "likelihood": {"dims": ("date",)},
            "gamma_fourier": {"mu": 0, "b": 1, "dims": "fourier_mode"},
        }
```
Also (more like a question), is it necessary to set all parameters, or can we update a couple and let the others by default? Something like 

```python
model_config= {
            "beta_channel": {"sigma": prior_sigma, "dims": ("channel",)},
        }
```

michaelraczycki:
@juanitorduz that's actually a great idea. I agree that in later versions (at least after it's all merged) it should be made possible to change just single values from model_config (so far we either use the default one or user-provided one, but no function that would determine which fields are missing from user-provided one and fill it out with default values is available. 
For now I'm adding more intuitive way of defining the model_config sigma prior, and adding an extra cell showing how we can see the required structure of model_config in mmm to help users define their own.

juanitorduz (approved):
Great! The notebook looks nice :) Curious what other devs think!

mypy.ini
```diff
@@ -17,3 +17,6 @@ ignore_missing_imports = True
 
 [mypy-scipy.*]
 ignore_missing_imports = True
+
+[mypy-pymc_experimental.*]
+ignore_missing_imports = True
```
ricardoV94, line 22: Why?

pymc_marketing/mmm/base.py
ricardoV94: I don't have a strong opinion, but it makes more sense as a Github issue than as a comment in the code

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -63,80 +55,128 @@ def __init__(
...
-        return (
-            pm.HalfNormal.dist(sigma=2, shape=len(self.channel_columns))
-            if self.channel_prior is None
-            else change_dist_size(
-                dist=self.channel_prior, new_size=len(self.channel_columns)
-            )
-        )
+    @property
+    def default_sampler_config(self) -> Dict:
+        return {"progressbar": True, "random_seed": 1234}
```
ricardoV94, line 71: I don't think we should provide a fixed `random_seed` by default
```suggestion
        return {"progressbar": True}
```

The progressbar also seems unnecessary, since that's the default anyway?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -63,80 +55,128 @@ def __init__(
...
             coords["fourier_mode"] = fourier_features.columns.to_numpy()
+            X_data = pd.concat([X_data, fourier_features], axis=1)
+
+        self.model_coords = coords
+        if self.validate_data:
+            self.validate("X", X_data)
+            self.validate("y", y)
+        self.preprocessed_data: Dict[str, Union[pd.DataFrame, pd.Series]] = {
+            "X": self.preprocess("X", X_data.copy()),
+            "y": self.preprocess("y", y.copy()),
```
ricardoV94, line 121: Why the copy?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -210,21 +292,89 @@ def _get_fourier_models_data(self) -> pd.DataFrame:
...
+
 
 class DelayedSaturatedMMM(
     MaxAbsScaleTarget,
     MaxAbsScaleChannels,
     ValidateControlColumns,
     BaseDelayedSaturatedMMM,
 ):
-    ...
+    ...
```
ricardoV94, line 380: Newline missing at the end of file

pyproject.toml
```diff
@@ -17,10 +17,11 @@ dependencies = [
     "numpy>=1.17",
     "pandas",
     # NOTE: Keep minimum pymc version in sync with ci.yml `OLDEST_PYMC_VERSION`
-    "pymc>=5.3.0",
+    "pymc>=5.5.0",
```
ricardoV94, line 20: You need to update `OLDEST_PYMC_VERSION` in `ci.yml` as mentioned in the comment above

ricardoV94 (changes requested):
Just some small comments and issues

mypy.ini
michaelraczycki: without it it was throwing errors about missing stubs (which as I know aren't available for pymc-experimental). I also opened before the issue in this repo about linters and checks in pymc-experimental being different than pymc-experimental to the point that what passes linters in one crashes the linters in the other, but it didn't get much response

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: yes, each of the child classes need to implement it's own default sampler_config. In model builder it's an abstract function. If you don't think here any settings should be passed I can make it into an empty dict

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: the original implementation was working on a copy of the data, so I kept it

pymc_marketing/mmm/delayed_saturated_mmm.py
ricardoV94: Sounds like it doesn't need to be an `abstract` property in the base class, in that case? 
Usually we can use PyMC sampler defaults just fine

pymc_marketing/mmm/delayed_saturated_mmm.py
ricardoV94: Okay, seems wasteful, but if it's not your change, no need to discuss it here

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: I'd say that I'm open to all cosmetics being changed in the future, but as a patch to model builder. Now I'd focus on all things that would be deal-breakers to finally merge it. Then we can focus on tuning and small updates

michaelraczycki:
<p>it's neither (at least from my perspective), it's just regular text in markdown</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/295/discussion/#comment-1597055021'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: I checked with tests and everything still works as expected without the copies, so I removed it.

pymc_marketing/mmm/delayed_saturated_mmm.py
michaelraczycki: now default_sampler_config = {}

pyproject.toml
michaelraczycki: updated

michaelraczycki: updated

michaelraczycki:
@ricardoV94 Do you think we can merge it for now? The change requests are now addressed

michaelraczycki:
with #309 merged this one can be closed (#309 didn't show the merge conflicts appearing here, despite being a checked out of this branch)

PR #298, opened 2023-06-12
Allow saving and loading CLV models via ModelBuilder

michaelraczycki:
This is an initial integration of basic.py with ModelBuilder,
Please let me know if there's something here you find unacceptable, or something that is crucial is missing. It's most likely not the final stage, but the initial one, nevertheless because of number of classes dependent I prefer to have this one shaped in advance, so I don't need to propagate the changes to all child classes

----
:books: Documentation preview :books:: https://pymc-marketing--298.org.readthedocs.build/en/298/


michaelraczycki:
I just realized I can't add reviewers @twiecki @juanitorduz @ricardoV94

michaelraczycki:
@juanitorduz @twiecki @ColtAllen @ricardoV94 
The integration of model_builder and its core functionalities (saving and loading models) is done, the tests are passing. 
There is 1 exception, but I highly doubt that it's because of the integration: 
TestGammaGammaModel::test_model_convergence result doesn't fit within relative tolerance limit, the individual one passes. 
Please review the changes to see if everything is as expected, and please share your thoughts on the subject

pymc_marketing/clv/models/basic.py
```diff
@@ -49,21 +43,46 @@ def fit(self, fit_method="mcmc", **kwargs):
...
-        return res
+        self.idata.add_groups(fit_data=self.data.to_xarray())  # type: ignore
+
+        return self.idata
 
-    def _fit_mcmc(self, **kwargs):
-        """Draw samples from model posterior using MCMC sampling"""
-        with self.model:
-            return pm.sample(**kwargs)
+    def _fit_mcmc(self, **kwargs) -> None:
```
ricardoV94, line 61: Return signature is wrong?

pymc_marketing/clv/models/basic.py
```diff
@@ -78,19 +97,125 @@ def _fit_MAP(self, **kwargs):
         map_strace.record(map_res)
         map_strace.close()
         trace = MultiTrace([map_strace])
-        return pm.to_inference_data(trace, model=model)
+        idata = pm.to_inference_data(trace, model=model)
+        self.set_idata_attrs(idata)
+        return idata
```
ricardoV94, line 102: Why doesn't this save to self.idata and mcmc does?

pymc_marketing/clv/models/basic.py
```diff
@@ -78,19 +97,125 @@ def _fit_MAP(self, **kwargs):
...
+
+    @staticmethod
+    def _check_prior_ndim(prior, ndim=0):
+        if prior.ndim != ndim:
+            raise ValueError(
+                f"Prior variable {prior} must be have {ndim} ndims, but it has {prior.ndim} ndims."
+            )
+
+    def create_distribution_from_prior(self, name: str, **kwargs) -> TensorVariable:
+        if name == "gamma":
```
ricardoV94, line 159: Why not just `getattr("pymc", name)`? This seems unnecessary and more convoluted

pymc_marketing/clv/models/basic.py
```diff
@@ -78,19 +97,125 @@ def _fit_MAP(self, **kwargs):
...
-    def fit_result(self):
-        if self._fit_result is None:
+    def prior_predictive(self) -> az.InferenceData:
+        if self.idata is None or "prior_predictive" not in self.idata:
             raise RuntimeError("The model hasn't been fit yet, call .fit() first")
-        return self._fit_result
+        return self.idata["prior_predictive"]
+
+    @property
+    # if we include fit_result then this should be added for consistancy
```
ricardoV94, line 201: I can't quite tell what this comment means.

pymc_marketing/clv/models/basic.py
```diff
@@ -103,6 +228,3 @@ def fit_summary(self, **kwargs):
             return res["mean"].rename("value")
         else:
             return az.summary(self.fit_result, **kwargs)
-
-    def __repr__(self):
```
ricardoV94, line 107: Does the base class have a `__repr__`? This method should actually have been `__str__` btw

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -101,44 +90,86 @@ class BetaGeoModel(CLVModel):
...
             raise ValueError(
                 "The BetaGeoModel expects exactly one entry per customer. More than"
                 " one entry is currently provided per customer id."
             )
+        self.coords = {"customer_id": self.customer_id}
+
+    @property
+    def default_model_config(self) -> Dict[str, Dict]:
+        return {
+            "a_prior": {"dist": "halfflat", "kwargs": {}},
```
ricardoV94, line 150: We should use the same capitalization convention as pymc distributions

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -174,35 +205,22 @@ def logp(t_x, x, a, b, r, alpha, T):
...
-            r_prior = pm.HalfFlat.dist()
-        else:
-            self._check_prior_ndim(r_prior)
-
-        return super()._process_priors(a_prior, b_prior, alpha_prior, r_prior)
-
     def _unload_params(self):
-        trace = self.fit_result.posterior
-
+        try:
```
ricardoV94, line 220: Why this try except?

pymc_marketing/clv/models/shifted_beta_geo.py
```diff
@@ -77,59 +76,81 @@ class ShiftedBetaGeoModelIndividual(CLVModel):
...
+            "alpha_prior": {"dist": "halfflat", "kwargs": {}},
+            "beta_prior": {"dist": "halfflat", "kwargs": {}},
+        }
 
-        coords = {"customer_id": np.asarray(customer_id)}
-        with pm.Model(coords=coords) as self.model:
-            alpha = self.model.register_rv(alpha_prior, name="alpha")
-            beta = self.model.register_rv(beta_prior, name="beta")
+    @property
+    def _serializable_model_config(self) -> Dict:
```
ricardoV94, line 128: Why isn't this in the base CLVModel class?

pymc_marketing/clv/models/shifted_beta_geo.py
```diff
@@ -77,59 +76,81 @@ class ShiftedBetaGeoModelIndividual(CLVModel):
...
+        return self.model_config
+
+    @property
+    def default_sampler_config(self) -> Dict:
```
ricardoV94, line 132: Why isn't this in the base CLV model class?

pymc_marketing/clv/models/shifted_beta_geo.py
```diff
@@ -77,59 +76,81 @@ class ShiftedBetaGeoModelIndividual(CLVModel):
...
 
             churn_raw = pm.Geometric.dist(theta)
-
             pm.Censored(
                 "churn_censored",
                 churn_raw,
                 lower=None,
-                upper=T,
-                observed=t_churn,
+                upper=self.T[0],
```
ricardoV94, line 149: Why is this `T[0]`?

tests/clv/models/test_basic.py
```diff
@@ -1,80 +1,41 @@
...
-    def test_fit_override(self):
-        with warnings.catch_warnings():
-            warnings.simplefilter("error")
-
-            model = CLVModelTest()
-            model.fit_result = 1
-            with pytest.warns(UserWarning, match="Overriding pre-existing fit_result"):
-                model.fit_result = 2
+
+    # The rest of the tests need to be moved to subclasses, as basic.py is not a standalone class
```
ricardoV94, line 38: Why not create a test subclass that implements the same kind of dummy model as was before? Then the shared methods can be tested here just like it was done so far

tests/clv/models/test_beta_geo.py
```diff
@@ -40,37 +45,47 @@ def setup_class(cls):
...
         )
         assert isinstance(
             model.model["alpha"].owner.op,
-            pm.HalfFlat if alpha_prior is None else type(alpha_prior.owner.op),
+            type(pm.HalfCauchy.dist(beta=2).owner.op),
         )
         assert isinstance(
             model.model["r"].owner.op,
-            pm.HalfFlat if r_prior is None else type(r_prior.owner.op),
+            type(pm.Gamma.dist(alpha=1, beta=1).owner.op),
```
ricardoV94, line 88: Why isn't this testing the default anymore?

tests/clv/models/test_beta_geo.py
```diff
@@ -172,15 +212,21 @@ def test_expected_probability_alive(self):
...
         bg_model = BetaGeoModel(
-            customer_id=customer_id,
-            frequency=frequency,
-            recency=recency,
-            T=T,
+            data=data,
         )
-
+        bg_model.build_model()
+        bg_model.fit(fit_method="map")
```
ricardoV94, line 229: Why is this fit call needed?

tests/clv/models/test_beta_geo.py
```diff
@@ -207,20 +253,41 @@ def test_expected_probability_alive(self):
...
+                "frequency": test_frequency,
+                "recency": test_recency,
+                "T": test_T,
+            }
         )
-        bg_model._fit_result = az.from_dict(
+
+        bg_model = BetaGeoModel(data=data)
+        bg_model.build_model()
+        bg_model.fit("map")
```
ricardoV94, line 289: Why is this needed?

tests/clv/models/test_beta_geo.py
```diff
@@ -267,14 +334,18 @@ def test_expected_num_purchases_new_customer(self):
...
+                "customer_id": customer_id,
+                "frequency": test_frequency,
+                "recency": test_recency,
+                "T": test_T,
+            }
         )
-        bg_model._fit_result = az.from_dict(
+        bg_model = BetaGeoModel(data=data)
+        bg_model.build_model()
+        bg_model.fit("map")
```
ricardoV94, line 347: Same question

tests/clv/models/test_beta_geo.py
```diff
@@ -324,3 +398,32 @@ def test_model_repr(self):
...
+        assert isinstance(model, BetaGeoModel)
+
+        # Load data from the file to cross verify
+        filepath = Path(str(temp))
+        idata = az.from_netcdf(filepath)
+        dataset = idata.fit_data.to_dataframe()
+        # Check if the loaded data matches with the model data
+        assert np.array_equal(model2.customer_id.values, dataset.customer_id.values)
+        assert np.array_equal(model2.frequency.values, dataset.frequency.values)
+        assert np.array_equal(model2.T.values, dataset["T"])
```
ricardoV94, line 425: Should use np.testing utilities, as they provide much better message when it fails

tests/clv/models/test_gamma_gamma.py
```diff
@@ -179,26 +218,22 @@ def test_spend(self, distribution):
...
+            "v_prior": {"dist": "normal", "kwargs": {"mu": v_mean, "sigma": 0.01}},
+        }
+        model = GammaGammaModel(
+            data=data,
+            model_config=custom_model_config,
         )
-        fake_fit.add_groups(dict(posterior=fake_fit.prior))
-        model._fit_result = fake_fit
+        model.build_model()
+        model.fit(chains=1, progressbar=False, random_seed=self.rng)
```
ricardoV94, line 236: No need to call mcmc to get a fake fit, why was this changed?

ricardoV94 (commented):
Looks good @michaelraczycki 

I left some comments below. It would be easier to review if the changes to the tests were kept minimal, specially stuff like the fake fits. Is there a fundamental restriction that one can't set the fit manually as we were doing before?

pymc_marketing/clv/models/basic.py
ricardoV94: Ah, yes prior predictive is less of a thing in CLV models in general

tests/clv/models/test_basic.py
michaelraczycki: I'm not a fan of tests where everything is mocked, they don't bring much security and are just confusing.

pymc_marketing/clv/models/basic.py
ricardoV94: But we can definitely add some property for it

tests/clv/models/test_basic.py
ricardoV94: If the point is to test methods implement in the base class it seems reasonable to implement the simplest subclass of the baseclass to test them

ricardoV94: It's fine to change how tests work but that's probably better suitable in a standalone PR than in a large refactor like this

tests/clv/models/test_beta_geo.py
michaelraczycki: fit_result property takes value from class.idata['posterior'], before the fit() is performed class.idata doesn't exist, so trying to overwrite it would fail

tests/clv/models/test_shifted_beta_geo.py
```diff
@@ -76,71 +121,61 @@ def test_model(self, alpha_prior, beta_prior):
...
+        dataset["T"] = [8, 9, 10]
         with pytest.raises(ValueError, match=match_msg):
             ShiftedBetaGeoModelIndividual(
-                customer_id=customer_id,
-                t_churn=[-1, 8, 9],
-                T=[8, 9, 10],
+                data=pd.DataFrame(dataset),
             )
 
-    def test_model_repr(self):
```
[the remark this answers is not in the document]
michaelraczycki: it wasn't removed, as I explained with the __repr__ in basic.py, I was taught the convention of putting __functions on top of the file, so I feel strange having functions testing exactly that in other place, it's still in the file, but here marked as 'added' on the top of the file

tests/clv/models/test_beta_geo.py
michaelraczycki: the priors aren't passed as a standalone parameters, but as a model_config. the default model config in every class doesn't allow the priors to be None, as it has them always with halfflat dist. So no None case test is needed

pymc_marketing/clv/models/shifted_beta_geo.py
michaelraczycki: as previously the T was passed as a standalone value it was used as an int. Now as I'm passing dataset of input params, it's value is propagated as a value of each row of parameters, with just self.T the pm.Censored upper limit was defined as <constant> in the str representation

pymc_marketing/clv/models/shifted_beta_geo.py
michaelraczycki: because technically every single model type can have it's own unserialisable parts of prior that would need to be converted to their serialisable versions here. If we're sure that we don't need it I can move it to basic.py

tests/clv/models/test_beta_geo.py
ricardoV94: Sure but we should still test the defaults somewhere even if in a separate test

tests/clv/models/test_beta_geo.py
ricardoV94: So can we set class.idata manually?

pymc_marketing/clv/models/shifted_beta_geo.py
ricardoV94: I think T is allowed to be a vector per observation, although I guess we didn't test it if it didn't show up as a failure.

ricardoV94: Yeah the docstrings say that explicitly. Need to fix. 

If passing a single value is no longer possible because everything must be a DataFrame, the tests and docstrings must be updated.

tests/clv/models/test_shifted_beta_geo.py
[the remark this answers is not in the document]
ricardoV94: Stylistic conventions should be enforced by pre-commit hooks like black, otherwise every developer will have their own opinion and change the same code around without need. This makes git changes harder to review both in this PR and for future developers.

pymc_marketing/clv/models/basic.py
BBDS-Colt: I think we should add the prior predictive property, because I have an idea on how it can be added for the `ParetoNBD` and `BetaGeoBetaBinom` models and even be used for data drift monitoring after this PR is merged and save/load functionality is supported.

BBDS-Colt:
> `TestGammaGammaModel::test_model_convergence` result doesn't fit within relative tolerance limit, the individual one passes.

This is probably due to the test dataset not being sufficiently large enough. `GammaGammaModel` can only be used for `frequency` >= 1, which will cut down the number of viable observations by at least a third.

pymc_marketing/clv/models/beta_geo.py
michaelraczycki: true, I realised that as I was already halfway through the implementation. I've decided to leave it to 'aftercare' as I was sure that it won't be the only thing needing some tweaks, and this way I could deliver it faster

pymc_marketing/clv/models/beta_geo.py
michaelraczycki: It's a weird thing with the MAP sampling and creating the inferencedata out of it, for reason I couldn't find it creates the posterior with one more nested 'posterior' inside. So depending on the selected sampling once the .posterior postfix was needed and once it wasn't. It's something that for sure should be fixed, but I wanted to get more reviews faster, and not delay it to fix this minor thing

pymc_marketing/clv/models/basic.py
michaelraczycki: ok, So I guess the discussion on this one can be closed, I'll remove the comments and keep the prior_predictive and posterior_predictive properties to have the APIs unified

ricardoV94: sounds good

pymc_marketing/clv/models/basic.py
michaelraczycki: fixed.

pymc_marketing/clv/models/shifted_beta_geo.py
michaelraczycki: moved to basic.py

tests/clv/models/test_shifted_beta_geo.py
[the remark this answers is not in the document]
michaelraczycki: I won't change anything style-wise from now on here, promise!

tests/clv/models/test_beta_geo.py
michaelraczycki: fixed, added extra check in @fit_result.setter and now if self.idata doesn't exist yet it just assigns it, instead of complaining

michaelraczycki: so right now the fit is no longer needed (for the sake of tests of course)

tests/clv/models/test_beta_geo.py
michaelraczycki: no longer using fit in this test

tests/clv/models/test_beta_geo.py
michaelraczycki: no longer using fit here

pymc_marketing/clv/models/basic.py
michaelraczycki: fixed, now they both act the same.

tests/clv/models/test_gamma_gamma.py
michaelraczycki: fit_result rewriting altered and now fit() could be removed for testing

tests/clv/test_utils.py
```diff
@@ -50,49 +50,64 @@ def test_summary_data() -> pd.DataFrame:
...
+        "b_prior": {"dist": "diracdelta", "kwargs": {"c": 3.18105431}},
+        "r_prior": {"dist": "diracdelta", "kwargs": {"c": 0.16385072}},
+    }
+    model = BetaGeoModel(
+        data=data,
+        model_config=model_config,
     )
-
+    model.build_model()
+    model.fit()
```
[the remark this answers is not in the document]
michaelraczycki: Even with the suggested change, the test fails when trying to get 5 decimal match, changed to 3 decimal to pass

tests/clv/models/test_beta_geo.py
michaelraczycki: all test_model are now testing for both default and custom config dict scenarios

pymc_marketing/clv/models/shifted_beta_geo.py
michaelraczycki: I've adapted the docstrings regarding expected input parameters, please check again

pymc_marketing/clv/models/basic.py
michaelraczycki: I changed it to use the getattr() way

tests/clv/models/test_beta_geo.py
michaelraczycki: replaced

tests/clv/test_utils.py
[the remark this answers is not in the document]
michaelraczycki: the fit is no longer used, but the precision from fake_fit doesn't pass the required precision

pymc_marketing/clv/models/basic.py
```diff
@@ -49,21 +43,80 @@ def fit(self, fit_method="mcmc", **kwargs):
...
         elif fit_method == "map":
-            res = self._fit_MAP(**kwargs)
+            self._fit_MAP(**kwargs)
         else:
             raise ValueError(
                 f"Fit method options are ['mcmc', 'map'], got: {fit_method}"
             )
-        self.fit_result = res
-        return res
+        self.idata.add_groups(fit_data=self.data.to_xarray())  # type: ignore
```
ricardoV94, line 57: Can you selectively suppress the warning that arviz issues about `fit_data` not being a default group?

CC @OriolAbril is there any thing we should pay attention (why is there a warning like that)?

pymc_marketing/clv/models/basic.py
```diff
@@ -78,31 +131,120 @@ def _fit_MAP(self, **kwargs):
...
+    @property
+    def default_sampler_config(self) -> Dict:
+        return {}
 
     @property
-    def fit_result(self):
-        if self._fit_result is None:
+    def prior_predictive(self) -> az.InferenceData:
+        if self.idata is None or "prior_predictive" not in self.idata:
             raise RuntimeError("The model hasn't been fit yet, call .fit() first")
```
ricardoV94, line 216: Model doesn't need to be fit in order to do prior_predictive

ricardoV94: Also make sure to test at least in a dummy subclass

pymc_marketing/clv/models/basic.py
```diff
@@ -78,31 +131,120 @@ def _fit_MAP(self, **kwargs):
...
     @fit_result.setter
-    def fit_result(self, res):
-        if self._fit_result is not None:
+    def fit_result(self, res: az.InferenceData) -> None:
+        if self.idata is None:
+            self.idata = res
+        elif "posterior" in self.idata:
             warnings.warn("Overriding pre-existing fit_result")
-        self._fit_result = res
+            self.idata.posterior = res
```
ricardoV94, line 231: This will fail if a user called `prior_predictive` already, in which case there should be an `idata` already. Related to my comment above

pymc_marketing/clv/models/basic.py
```diff
@@ -78,31 +131,120 @@ def _fit_MAP(self, **kwargs):
...
+
+    # if we include fit_result then this should be added for consistancy
```
ricardoV94, line 233: Remove comment

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -101,44 +99,82 @@ class BetaGeoModel(CLVModel):
...
+        return {
+            "a_prior": {"dist": "HalfFlat", "kwargs": {}},
+            "b_prior": {"dist": "HalfFlat", "kwargs": {}},
+            "alpha_prior": {"dist": "HalfFlat", "kwargs": {}},
+            "r_prior": {"dist": "HalfFlat", "kwargs": {}},
+        }
+
+    @property
+    def _serializable_model_config(self) -> Dict:
+        return self.model_config
```
ricardoV94, line 167: Move to base class and only redefine when something specific is needed

pymc_marketing/clv/models/shifted_beta_geo.py
ricardoV94: We should only define it when needed. The default seems sensible enough

pymc_marketing/clv/models/shifted_beta_geo.py
ricardoV94: Sounds like you should pass the whole column? There is no constraint that all `T` must be the same. Before we allowed broadcasting if a user passed a scalar but they could also have passed a vector with different values.

tests/clv/models/test_basic.py
```diff
@@ -1,81 +1,60 @@
...
     def test_repr(self):
         model = CLVModelTest()
+        model.build_model()
         assert (
-            model.__repr__() == "CLVModelTest\nx ~ HalfNormal(0, 100)\ny ~ Normal(x, 1)"
+            model.__repr__()
+            == "CLVModelTest\na ~ Normal(0, 1)\nb ~ Normal(0, 1)\ny ~ Normal(f(a, b), 1)"
         )
 
-    def test_fit_override(self):
```
ricardoV94, line 70: This test seems to still be relevant?

tests/clv/models/test_beta_geo.py
```diff
@@ -189,7 +280,7 @@ def test_expected_probability_alive(self):
                 "r": rng.normal(r, 1e-3, size=(2, 25)),
             }
         )
-        bg_model._fit_result = fake_fit
+        bg_model.fit_result = fake_fit
```
ricardoV94, line 283: I think the equivalent to the old test approach is to override `self.idata`, directly which won't trigger any of the checks or warnings in the `setter` method.

This would apply to other fake fit tests

tests/clv/models/test_beta_geo.py
```diff
@@ -324,3 +441,34 @@ def test_model_repr(self):
             "\nr~HalfFlat()"
             "\nlikelihood~Potential(f(r,alpha,b,a))"
         )
+
+    def test_save_load_beta_geo(self, data):
```
ricardoV94, line 445: Can you make a generic test util to test save_load for different CLV models? I imagine most of the boilerplate can be reused?

tests/clv/test_utils.py
```diff
@@ -137,7 +150,7 @@ def test_customer_lifetime_value_with_known_values(test_summary_data, fitted_bg)
         time=1,
         discount_rate=0.0,
     ).mean(("chain", "draw"))
-    np.testing.assert_almost_equal(clv_d0, expected, decimal=5)
+    np.testing.assert_almost_equal(clv_d0, expected, decimal=3)
```
ricardoV94, line 153: I think this test was completely deterministic (given the `rng`). In that case the values should be exactly the same as before your changes. This could mean somewhere we're not propagating some seed or something, which would be important to fix. 

The first thing to check is whether the test was indeed deterministic, meaning `clv_d0` and `clv_d1` was always the same.

tests/clv/models/test_gamma_gamma.py
```diff
@@ -219,16 +275,14 @@ def test_new_customer_spend(self, distribution):
             preds = model.expected_new_customer_spend()
             assert preds.shape == (1, 1000)
             np.testing.assert_allclose(
-                preds.mean(("draw", "chain")), expected_preds_mean, rtol=0.05
+                preds.mean(("draw", "chain")), expected_preds_mean, rtol=0.1
```
ricardoV94, line 278: Instead of using `self.rng` for the prior predictive let's pass a custom `rng` to this test. It should make it order invariant and hopefully the old `rtol` can be kept. I usually tried a couple of seeds and set on an `rtol` that seemed to work okay. Otherwise we would need to investigate if the change in precision is because of some subtle change in behavior of the model.

tests/clv/models/test_beta_geo.py
```diff
@@ -132,25 +210,33 @@ def test_numerically_stable_logp(self, frequency, recency, logp_value):
             (500, "mcmc", 0.3),
             (2000, "mcmc", 0.1),
             (10000, "mcmc", 0.055),
-            (2000, "map", 0.1),
+            (2000, "map", 0.11),
```
ricardoV94, line 213: Same here, let's try to pass a custom seed, so the seeding is not test-order dependent, and see if the old map expected result can stay the same

ricardoV94 (changes requested):
Looks like we are on the final runway @michaelraczycki, great work!

Left some more comments

tests/clv/models/test_beta_geo.py
ricardoV94: This can be left as an open issue if you don't want to do it in this PR

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -63,13 +52,22 @@ class BetaGeoModel(CLVModel):
...
+            sampler_config={
+                "draws": 1000,
+                "tune": 1000,
+                "chains": 2,
+                "cores": 2,
+                "nuts_kwargs": {"target_accept": 0.95},
+            },
         )
-
+        model.build_model()
```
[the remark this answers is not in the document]
michaelraczycki: do you mean instead just passing data = data to leave the whole data definition:
data=pd.DataFrame({
            'frequency':[4, 0, 6, 3, ...],
             'recency':[30.73, 1.72, 0., 0., ...],
})
?

pymc_marketing/clv/models/basic.py
michaelraczycki: in this case the implementation wasn't because of mathematical requirements for obtaining prior_predictive samples, but because the mcmc fit called 'sample_model' function, which did all 3 samplings, prior_predictive, posterior and posterior_predictive. So in case that the self.idata was not set, or the prior_predictive was not found in idata was only because none of the samplings were done ( at least from the API perspective). I can change it as anyway now sampling from prior predictive will need to be called separately

michaelraczycki: and it's already tested, all added properties and functions are 100% test covered

tests/clv/models/test_beta_geo.py
michaelraczycki: I'll open an issue for that, for now let's get them finally merged

pymc_marketing/clv/models/beta_geo.py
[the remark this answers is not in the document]
ricardoV94: Yes

pymc_marketing/clv/models/basic.py
ricardoV94: The codecov bot here on github showed this one as not being tested, that's why I asked. Not sure how reliable it is though. Skimming through the tests, I didn't see a call to `sample_prior_predictive` directly

michaelraczycki: that's because prior to re-defining the sample_model again for CLV classes (as you said , it shouldn't call by default sample_prior_predictive here) it was always done with fit, I've added dedicated sample_prior_predictive method in base.py and it's corresponding test inside test_basic.py (coverage back to 100%)

pymc_marketing/clv/models/basic.py
michaelraczycki: added a case that doesn't need 'posterior' to be in the idata, covered with test

pymc_marketing/clv/models/basic.py
michaelraczycki: removed

tests/clv/models/test_beta_geo.py
michaelraczycki: replaced with model.idata = fake_fit

tests/clv/models/test_gamma_gamma.py
michaelraczycki: passing new seed allowed to get back to the old rtol

tests/clv/models/test_beta_geo.py
michaelraczycki: here it's already creating new seed with each test iteration, I'm not sure how should I make it even more per test

pymc_marketing/clv/models/shifted_beta_geo.py
michaelraczycki: it doesn't make the model fail, but the str. representation will need to be adapted to churn_censored ~Censored(Geometric(theta), -inf,<constant>) as I assume that PyMC also assumes that in case of an array the number doesn't need to be the same

michaelraczycki: I'll adapt the test and make it again entire T

pymc_marketing/clv/models/basic.py
michaelraczycki: It's something that would need to be done on Arviz side, I've asked @OriolAbril about that before as it was messing with tests in pymc-experimental, he said that it wasn't anything to be worried about (back then)

pymc_marketing/clv/models/beta_geo.py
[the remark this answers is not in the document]
michaelraczycki: data examples added

tests/clv/models/test_beta_geo.py
ricardoV94: `self.rng` will be order dependent. You can create a seed inside the test (it's fine if the different parametrizations use the same seed)

michaelraczycki: but it's already done like that:
    @pytest.mark.slow
    @pytest.mark.parametrize(
        "N, fit_method, rtol, seed",
        [
            (
                500,
                "mcmc",
                0.3,
            ),
            (2000, "mcmc", 0.1),
            (10000, "mcmc", 0.055),
            (2000, "map", 0.1),
        ],
    )
    def test_model_convergence(self, N, fit_method, rtol, seed, model_config):
        rng = np.random.default_rng(146)
        recency, frequency, _, T = self.generate_data(
            self.a_true, self.b_true, self.alpha_true, self.r_true, N, rng=rng
        )
        data = pd.DataFrame(
            {
                "customer_id": list(range(len(frequency))),
                "frequency": frequency,
                "recency": recency,
                "T": T,
            }
        )
        # b parameter has the largest mismatch of the four parameters
        model = BetaGeoModel(
            data=data,
            model_config=model_config,
        )
        model.build_model()

        sample_kwargs = dict(random_seed=rng, chains=2) if fit_method == "mcmc" else {}
        model.fit(fit_method=fit_method, progressbar=False, **sample_kwargs)

        fit = model.fit_result
        np.testing.assert_allclose(
            [fit["a"].mean(), fit["b"].mean(), fit["alpha"].mean(), fit["r"].mean()],
            [self.a_true, self.b_true, self.alpha_true, self.r_true],
            rtol=rtol,
        )

tests/clv/test_utils.py
michaelraczycki: I've went through the model class changes and I didn't notice any seeds being passed in the class itself. also the customer_lifetime_value or subsequently called functions don't include one. If the test was completely deterministic, why use assert_almost_equal, instead assert_equal?

ricardoV94: because for float comparisons you never want to compare strict equality. Here the results are computer by us and lifetimes so they could differ slightly (even if deterministically)

pymc_marketing/clv/models/basic.py
OriolAbril: A PR to add a flag/boolear arg to add_groups and/or inferencedata init would also be welcome. It is generally useful to have that because using groups that are not used by any arviz function is quite uncommon, as opposed to making typos with the group names for example (the original reason for the warning IIRC). There is nothing wrong with using different group names though, the inferencedata schema is even explicit about that, so not being able to turn these off is anoying for cases like this.

pymc_marketing/clv/models/basic.py
```diff
@@ -49,21 +43,87 @@ def fit(self, fit_method="mcmc", **kwargs):
...
+        if self.model is None:
+            raise RuntimeError(
+                "The model hasn't been built yet, call .build_model() first or call .fit() instead."
+            )
+
         with self.model:
-            return pm.sample(**kwargs)
+            sampler_args = {**self.sampler_config, **kwargs}
+            idata = pm.sample(**sampler_args)
+            idata.extend(pm.sample_posterior_predictive(idata))
```
ricardoV94, line 123: We can't call `sample_posterior_predictive` by default, for the same reason we couldn't call `sample_prior_predictive`. Not all CLV models have a standard likelihood that can be sampled from.

Is this `sample` something that in the `ModelBuilder` class calls all three sample methods in succession? If so we should just raise NotImplementedError in the base CLV class

pymc_marketing/clv/models/basic.py
```diff
@@ -78,31 +138,152 @@ def _fit_MAP(self, **kwargs):
...
             warnings.warn("Overriding pre-existing fit_result")
-        self._fit_result = res
+            self.idata.posterior = res
+        else:
+            self.idata.posterior = res
+
+    @property
+    def posterior_predictive(self) -> Dataset:
+        if self.idata is None or "posterior_predictive" not in self.idata:
+            raise RuntimeError("The model hasn't been fit yet, call .fit() first")
```
ricardoV94, line 276: This is wrong because `fit` doesn't trigger `sample_posterior_predictive` AFAICT. 

On a second thought I think we should remove the `prior/posterior` predictive properties and methods for now, as CLV models are a bit special in that they generally marginalize variables in a way that can't be sampled forward, or if it can, it's not really useful at all. That's why most CLV models have special models to make specific kind of posterior predictions or summaries.

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -197,14 +197,25 @@ class GammaGammaModel(BaseGammaGammaModel):
...
-                v_prior=pm.HalfNormal.dist(10),
+                data=pd.DataFrame({
+                    "customer_id": [0, 1, 2, 3, ...],
+                    "mean_transactionn_value" :[23.5, 19.3, 11.2, 100.5, ...],
+                    "frequency": [6, 8, 2, 1, ...],
+                }),
+                model_config={
+                    "p_prior": {dist: 'HalfNorm', kwargs: {}},
+                    "q_prior": {dist: 'HalfStudentT', kwargs: {}},
+                    "v_prior": {dist: 'HalfCauchy', kwargs: {}},
```
ricardoV94, line 208: This is invalid, as the HalfStudentT for sure requires at least `df`. I suggest to just keep the original example with `HalfNormal(sigma=10)` for everything

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -309,13 +336,27 @@ class GammaGammaModelIndividual(BaseGammaGammaModel):
...
+
+            model = GammaGammaModelIndividual(
+                data=pd.DataFrame({
+                    "customer_id": [0, 0, 0, 1, 1, 2, ...],
+                    "individual_transaction_value": [5.3. 5.7, 6.9, 13.5, 0.3, 19.2 ...],
+                }),
+                model_config={
+                    "p_prior": {dist: 'HalfNorm', kwargs: {}},
+                    "q_prior": {dist: 'HalfStudentT', kwargs: {}},
+                    "v_prior": {dist: 'HalfCauchy', kwargs: {}},
```
ricardoV94, line 349: Same issue

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -13,21 +13,27 @@
...
+        )
+        self.q_prior = self.create_distribution_from_prior(
+            self.model_config["q_prior"]["dist"],
+            **self.model_config["q_prior"]["kwargs"],
+        )
+        self.v_prior = self.create_distribution_from_prior(
+            self.model_config["v_prior"]["dist"],
+            **self.model_config["v_prior"]["kwargs"],
+        )
+        self._process_priors(self.p_prior, self.q_prior, self.v_prior)
```
ricardoV94, line 36: We should be still be calling `_check_prior_ndim`, to make sure users are not trying to create multivariate priors or priors with the wrong shape, such as `{"p_prior": {"dist": "Normal", kwargs: {"mu": [0, 0, 0]}}`.

By the way, how does serialization handle numpy objects in the kwargs?

ricardoV94 (changes requested):
Some question

pymc_marketing/clv/models/basic.py
michaelraczycki: we're anyway already re-defining the sample function inside basic.py so I'd say let's remove the sample_posterior_predictive from default sampling

pymc_marketing/clv/models/gamma_gamma.py
michaelraczycki: so far it would need to be added manually, that's where the serialisable model config comes in. I'll create an users guide to that initially, with time I'd say we should aim for having a module listing most common non-serialisable structures and providing their serializable equivalents for the purpose of saving

pymc_marketing/clv/models/basic.py
michaelraczycki: I've suppressed the warning with:
        with warnings.catch_warnings():
            warnings.filterwarnings(
                "ignore",
                category=UserWarning,
            )
            self.idata.add_groups(fit_data=self.data.to_xarray())  # type: ignore
for now should be good enough as the users won't see the warning

pymc_marketing/clv/models/basic.py
michaelraczycki: @ColtAllen do you also agree on completely removing the prior/posterior? I remember you mentioning that you had some plans considering it

pymc_marketing/clv/models/gamma_gamma.py
michaelraczycki: I've replaced it empty kwargs dict with {"nu": 4, "sigma": 10} as in other examples

pymc_marketing/clv/models/gamma_gamma.py
michaelraczycki: I've replaced it empty kwargs dict with {"nu": 4, "sigma": 10} as in other examples

pymc_marketing/clv/models/gamma_gamma.py
michaelraczycki: I could add a function like that:
`def make_serializable(obj):
    """
    Make an object json serializable by converting non-serializable fields to strings.
    """
    # Base case: if obj is serializable, we're done.
    try:
        json.dumps(obj)
        return obj
    except TypeError:
        pass

    # If obj is a dictionary, recursively make its fields serializable.
    if isinstance(obj, dict):
        return {k: make_serializable(v) for k, v in obj.items()}

    # If obj is a list, recursively make its elements serializable.
    elif isinstance(obj, list):
        return [make_serializable(elem) for elem in obj]

    # If obj is a numpy array, convert it to a list.
    elif isinstance(obj, np.ndarray):
        return obj.tolist()

    # If obj is a datetime, convert it to a string.
    elif isinstance(obj, datetime.datetime):
        return obj.isoformat()

    # If obj is a non-serializable object, convert it to a string.
    else:
        return str(obj)`

and then our serialisation would look like that:
I would add a setter for the _serializable_model_config so users can manually add cases that are not yet covered by the make_serializable function
`@property
def _serializable_model_config(self):
    return self.make_serializable(self.model_config)
`
what do you think about it?

michaelraczycki: > We should be still be calling `_check_prior_ndim` (in all models), to make sure users are not trying to create multivariate priors or priors with the wrong shape, such as `{"p_prior": {"dist": "Normal", kwargs: {"mu": [0, 0, 0]}}`.
> 
> By the way, how does serialization handle numpy objects in the kwargs?

I've included _check_prior_ndim inside create_distribution_for_prior, this way current code can stay as-is (maybe with small tweak regarding verbosity as you requested) and the check will be done

pymc_marketing/clv/models/basic.py
```diff
@@ -78,31 +137,151 @@ def _fit_MAP(self, **kwargs):
...
+    def _check_prior_ndim(prior, ndim=0):
+        if prior.ndim != ndim:
+            raise ValueError(
+                f"Prior variable {prior} must be have {ndim} ndims, but it has {prior.ndim} ndims."
+            )
+
+    def create_distribution_from_prior(self, name: str, **kwargs) -> TensorVariable:
+        try:
+            prior = getattr(pm, name).dist(**kwargs)
+            CLVModel._check_prior_ndim(prior)
```
ricardoV94, line 197: You need to allow passing `ndim`, to `_check_prior_ndim`

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -101,44 +99,82 @@ class BetaGeoModel(CLVModel):
...
+        super().__init__(
+            model_config=model_config,
+            sampler_config=sampler_config,
         )
-
+        self.data = data
+        self.a_prior = self.create_distribution_from_prior(
+            self.model_config["a_prior"]["dist"],
+            **self.model_config["a_prior"]["kwargs"],
+        )
```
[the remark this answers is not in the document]
ricardoV94: I still don't get why not just pass `self.model_config["a_prior"]`

pymc_marketing/clv/models/basic.py
```diff
@@ -78,31 +137,151 @@ def _fit_MAP(self, **kwargs):
...
+        return model
+
+    @staticmethod
+    def _check_prior_ndim(prior, ndim=0):
+        if prior.ndim != ndim:
+            raise ValueError(
+                f"Prior variable {prior} must be have {ndim} ndims, but it has {prior.ndim} ndims."
+            )
+
+    def create_distribution_from_prior(self, name: str, **kwargs) -> TensorVariable:
```
ricardoV94, line 194: This name doesn't make much sense. Perhaps `create_distribution_from_dict`? Also I think it can be private with a leading underscore. Users are unlikely to care about this.

ricardoV94: Also should be a staticmethod AFAICT

ricardoV94 (changes requested):

pymc_marketing/clv/models/basic.py
michaelraczycki: create_distribution_from_dict makes even less sense as we're not passing a dict to the function, maybe simply create_distribution? I'm not really seeing a reason why this should be static, it's not meant to be used outside of init, after the super() are initialised, so it will be available in child classes, and once again the basic shouldn't be initialised and used as a standalone class

pymc_marketing/clv/models/gamma_gamma.py
michaelraczycki: @ricardoV94 the prior creation was adapted and renamed as you requested, please in a free moment provide feedback on the serialisation concept proposed above

pymc_marketing/clv/models/basic.py
michaelraczycki: provided

pymc_marketing/clv/models/basic.py
```diff
@@ -78,31 +137,153 @@ def _fit_MAP(self, **kwargs):
...
+
+    @staticmethod
+    def _check_prior_ndim(prior, ndim=0):
+        if prior.ndim != ndim:
+            raise ValueError(
+                f"Prior variable {prior} must be have {ndim} ndims, but it has {prior.ndim} ndims."
+            )
+
+    def create_distribution_and_check_dim(
+        self, prior: Dict, dim: Optional[int] = 0
```
ricardoV94, line 195:
```suggestion
        self, prior: Dict, ndim: Optional[int] = 0
```

pymc_marketing/clv/models/basic.py
```diff
@@ -78,31 +137,153 @@ def _fit_MAP(self, **kwargs):
...
+    ) -> TensorVariable:
```
ricardoV94, line 196:
```suggestion
    def create_distribution_from_dict(
        self, dist: Dict, ndim: Optional[int] = 0
    ) -> TensorVariable:
```

The fact that you pass a dict as input agrees with the name I think.

ricardoV94: Make sure to update `dim` to `ndim` inside

ricardoV94: If you don't use `self`, it should be a classmethod/staticmethod. Those will still work from `self` in instantiated subclasses

tests/clv/models/test_basic.py
ricardoV94: @michaelraczycki can you check this?

ricardoV94: Seems to be covered in a test below

ricardoV94 (approved):

tests/clv/test_utils.py
```diff
@@ -137,8 +150,7 @@ def test_customer_lifetime_value_with_known_values(test_summary_data, fitted_bg)
         time=1,
         discount_rate=0.0,
     ).mean(("chain", "draw"))
-    np.testing.assert_almost_equal(clv_d0, expected, decimal=5)
-
+    np.testing.assert_almost_equal(clv_d0, expected, decimal=3)
```
ricardoV94, line 153: I am going to have to take a look at this one. I couldn't see why the results should have changed on a quick look just now.

ricardoV94 (changes requested):

pymc_marketing/clv/models/basic.py
michaelraczycki: renamed, it's called now _create_distribution, a static method with the ndim param

tests/clv/test_utils.py
michaelraczycki: any ideas what might have changed it? I went quickly through the changes and I can't see any seed changes that could affect that

ricardoV94: The easiest is to just print all the intermediate quantities before and after the refactoring, and see what changed where

michaelraczycki: I found it, the test was failing with the precision because I have accidentally overwritten the Dirac's delta c param with the c from alpha.

michaelraczycki: but it was within the test itself, while creating the model_config for it

ricardoV94 (approved):

tests/clv/test_utils.py
ricardoV94: That's great!

ricardoV94:
We can squash and merge once the tests pass! Great work @michaelraczycki

Issue #301, opened 2023-06-17
Add posterior population distributions for `BetaGeoModel`


wd60622:
Think one of the cool attributes on the `BetaGeoModel` assumptions are the population distributions. 

Think they are useful for understanding the fit high level but also might be interesting to see population shift the time varying parameter version or other variations down the line

Seems like it can be implemented pretty straight forward and wrapped in a function like some of the [GammaGamma Models methods](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/models/gamma_gamma.py)

```python 
with pm.Model(): 
    # Dummy for the sampling
    a = pm.HalfFlat("a")
    b = pm.HalfFlat("b")
    alpha = pm.HalfFlat("alpha")
    r = pm.HalfFlat("r")

    population_dropout = pm.Beta("population_dropout", alpha=a, beta=b)
    # Need to double check this parameterization
    population_purchase_rate = pm.Gamma("population_purchase_rate", alpha=alpha, beta=r)
    
    pp = pm.sample_posterior_predictive(
        self.fit_results, 
        var_names=["population_dropout", "population_purchase_rate"], 
        ... # would be function args like in gamma gamma
    ).posterior_predictive
```

Any thoughts on if this would be a valuable addition? Would be willing to implement and add to the docs 😄

ricardoV94:
@wd60622 would that be equivalent to `distribution_new_customer_spend` in the `GammaGamma` model?

If so, I think it would be a good addition.

wd60622:
Think that is closely related, goes with two of the assumptions of the generation process. 

From these samples, we could create these figures [(taken from here)](https://towardsdatascience.com/modeling-customer-lifetime-value-with-lifetimes-71171a35f654#:~:text=CLV%20estimation%20using%20BG%2DNBD&text=This%20estimate%20relies%20on%20the,values%20in%20the%20observation%20period.) or make statistics about the population

<img width="500" alt="beta-geo-pop-params" src="https://github.com/pymc-labs/pymc-marketing/assets/57733339/3d2703e2-c306-4980-a5c7-0631fe6a755c">

PR #303, opened 2023-06-19
Sample BetaGeo population parameters

wd60622:
Addresses the [issue here](https://github.com/pymc-labs/pymc-marketing/issues/301)

~Currently only address models fit with mcmc~

----
:books: Documentation preview :books:: https://pymc-marketing--303.org.readthedocs.build/en/303/


wd60622:
Fixes the docs [here](https://pymc-marketing--303.org.readthedocs.build/en/303/api/generated/pymc_marketing.clv.models.beta_geo.BetaGeoModel.html#pymc_marketing.clv.models.beta_geo.BetaGeoModel)

Docs for the new method are [here](https://pymc-marketing--303.org.readthedocs.build/en/303/api/generated/pymc_marketing.clv.models.beta_geo.BetaGeoModel.sample_population_distributions.html#pymc_marketing.clv.models.beta_geo.BetaGeoModel.sample_population_distributions)

wd60622:
@ricardoV94

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -344,3 +346,67 @@ def expected_num_purchases_new_customer(
         return (left_term * right_term).transpose(
             "chain", "draw", "t", missing_dims="ignore"
         )
+
+    def sample_population_distributions(
```
ricardoV94, line 350: Maybe keep the name of the methods more similar to those in the gamma-gamma? Something like `distribution_new_customer_dropout` and alike. Internally they can use this method which should be pretty cheap anyway.

I think that will make them more discoverable

tests/clv/models/test_beta_geo.py
```diff
@@ -324,3 +325,16 @@ def test_model_repr(self):
...
+        model = BetaGeoModel(
+            customer_id=self.customer_id,
+            frequency=self.frequency,
+            recency=self.recency,
+            T=self.T,
+        )
+
+        model.fit(fit_method=fit_method)
+        samples = model.sample_population_distributions()
+        assert isinstance(samples, xr.Dataset)
```
ricardoV94, line 340: Can we assert something more concrete, like the mean/std is close enough to what would be expected from the fit?

Also let's not do an actual fit which would be a relatively slow test. You can borrow the logic in the other tests of doing a fake fit from prior_predictive

pymc_marketing/clv/models/beta_geo.py
ColtAllen: Do you think this method could be moved into `BaseModel` and standardized across all models?

tests/clv/models/test_beta_geo.py
wd60622: Yes. Good idea

pymc_marketing/clv/models/beta_geo.py
wd60622: Good thinking. 
I view this as the closest solution

https://github.com/pymc-labs/pymc-marketing/blob/d4feeeafad6f0299b115dac3582ad6d98bd8241f/pymc_marketing/clv/models/shifted_beta_geo.py#L158-L201

wd60622:
> Looks good so far. Only suggestion I have is to add a few more sentences to the docstrings explaining how to interpret the distributions. The Gamma distribution is simple as it describes the time between transactions for the customer population, but the Beta is a little trickier:
> 
> It illustrates the probability of a customer permanently dropping out after making a repeat purchase. Transactions for the BG/NBD model are assumed independent and identically distributed, and if a customer has only made a single purchase, then they are assumed 100% still active.

Yes, I will add to the docstrings. Good suggestion. 

I will just use the verbiage from [Haddie's paper](http://www.brucehardie.com/papers/bgnbd_2004-04-20.pdf) because, imo, the inclusion or exclusion of a "zero point" in the data will require the "repeat". 
Let me know if you have a different take

wd60622:
Method summary
Screenshot 2023-06-25 at 17 04 24

distribution_new_customer_dropout
Screenshot 2023-06-25 at 17 04 55

distribution_new_customer_purchase_rate

Screenshot 2023-06-25 at 17 05 09

ricardoV94 (approved):

ricardoV94:
Looks great @wd60622, we might just wait for the large refactor PR of the CLV models to be merged first before updating this and merging it. Easier to solve conflicts here than there.

wd60622:
> Looks great @wd60622, we might just wait for the large refactor PR of the CLV models to be merged first before updating this and merging it. Easier to solve conflicts here than there.

Perfect. Sounds good. Thanks for the feedback

ricardoV94:
@wd60622 we merged the big CLV refactor PR #298. Do you want to update this one so we can merge it?

wd60622:
> @wd60622 we merged the big CLV refactor PR #298. Do you want to update this one so we can merge it?

Sounds good. I will update it

ricardoV94 (approved):

ricardoV94:
Some small bug, as I think the property in the new model already gives you the posterior? Should be easy to fix

wd60622:
Tests adjusted to use the `idata` instead of `fit_results` after the refactor. 
Tests are passing locally

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -351,3 +353,77 @@ def expected_num_purchases_new_customer(
...
+        var_names: Sequence[str] = ("population_dropout", "population_purchase_rate"),
+    ) -> xr.Dataset:
+        with pm.Model():
+            a = pm.HalfFlat("a")
+            b = pm.HalfFlat("b")
+            alpha = pm.HalfFlat("alpha")
+            r = pm.HalfFlat("r")
+
+            # This is the shape with fit_method="map"
+            if self.idata.posterior.dims == {"chain": 1, "draw": 1}:
```
ricardoV94, line 369: We should use the property because it will raise the informative error in case the model was not fit yet

```suggestion
            if self.fit_result.dims == {"chain": 1, "draw": 1}:
```

ricardoV94 (changes requested):

ricardoV94 (approved):

ricardoV94:
Thanks @wd60622

wd60622:
> Thanks @wd60622!

Thanks for the feedback

PR #309, opened 2023-06-26
Solving mmm conflicts

michaelraczycki:
----
:books: Documentation preview :books:: https://pymc-marketing--309.org.readthedocs.build/en/309/

PR #291, opened 2023-05-25
Improve MMM contribution plots

juanitorduz:
Address https://github.com/pymc-labs/pymc-marketing/issues/259

Here we are more precise about the contribution plots and distinguish between direct and global contributions. The MMM notebook has been updated.

----
:books: Documentation preview :books:: https://pymc-marketing--291.org.readthedocs.build/en/291/

PR #313, opened 2023-06-29
Improve MMM contribution plots

juanitorduz:
Port of https://github.com/pymc-labs/pymc-marketing/pull/291 so that it is compatible with https://github.com/pymc-labs/pymc-marketing/pull/309

- [x] Add implementation and tests.
- [x] Update example notebook.

----
:books: Documentation preview :books:: https://pymc-marketing--313.org.readthedocs.build/en/313/

### DIFF
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -154,7 +154,6 @@ def logistic_saturation(x, lam: Union[npt.NDArray[np.float_], float] = 0.5):
     ----------
     x : tensor
         Input tensor.
-    lam : float, optional, by default 0.5
     lam : float or array-like, optional, by default 0.5
         Saturation parameter.
 
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -1,3 +1,7 @@
+from typing import Union
+
+import numpy as np
+import numpy.typing as npt
 import pytensor.tensor as pt
 from pytensor.tensor.random.utils import params_broadcast_shapes
 
@@ -144,14 +148,14 @@ def delayed_adstock(
     return batched_convolution(x, w, axis=axis)
 
 
-def logistic_saturation(x, lam: float = 0.5):
+def logistic_saturation(x, lam: Union[npt.NDArray[np.float_], float] = 0.5):
     """Logistic saturation transformation.
-
     Parameters
     ----------
     x : tensor
         Input tensor.
     lam : float, optional, by default 0.5
+    lam : float or array-like, optional, by default 0.5
         Saturation parameter.
 
     Returns
--- a/tests/mmm/test_delayed_saturated_mmm.py
+++ b/tests/mmm/test_delayed_saturated_mmm.py
@@ -275,6 +275,9 @@ class TestDelayedSaturatedMMM:
             channel_contributions_forward_pass_mean.shape
             == channel_contributions_mean.shape
         )
+        # The forward pass results should be in the original scale of the target variable.
+        # The trace fits the model with scaled data, so when scaling back, they should match.
+        # Since we are using a `MaxAbsScaler`, the scaling factor is the maximum absolute, i.e y.max()
         np.testing.assert_array_almost_equal(
             x=channel_contributions_forward_pass_mean / channel_contributions_mean,
             y=mmm_fitted.y.max(),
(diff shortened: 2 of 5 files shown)


juanitorduz:
@ricardoV94 I completed adapting the code from the already approved PR ✅ !

Regarding your comment:

> 1. If you are doing a single call, and compilation takes a considerably chunk, you can consider setting the mode to FAST_COMPILE and/or disable C compilation via config.cxx="".
> 2. If a user is likely to call this method many times in a single workflow, it may make sense to compile a generic function and cache it. In subsequent calls one could just pass the new inputs to the function without having to recompile it every time.

I am still not sure how many times a user is expected to call this function. Still, I definitely think it is worth trying as the compilation time does take a bit. I will create an issue and tackle it in a different PR.

pymc_marketing/mmm/transformers.py
```diff
@@ -144,14 +148,14 @@ def delayed_adstock(
...
 
-def logistic_saturation(x, lam: float = 0.5):
+def logistic_saturation(x, lam: Union[npt.NDArray[np.float_], float] = 0.5):
     """Logistic saturation transformation.
-
     Parameters
     ----------
     x : tensor
         Input tensor.
     lam : float, optional, by default 0.5
```
ricardoV94, line 157: duplicated
```suggestion
```

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -254,6 +258,49 @@ def test_get_fourier_models_data(
...
+        contributions = mmm_fitted.get_channel_contributions_forward_pass_grid(
+            start=0, stop=1.5, num=grid_size
+        )
+        assert contributions.shape == (
+            grid_size,
+            chains,
+            draws,
+            data_range,
+            n_channels,
+        )
```
ricardoV94, line 278: Could we test some simple numerical cases?

ricardoV94: Because this is an extremely complicated model, I would suggest providing a fake_fit, and pre-computed results from the forward pass. Just make sure those seem correct! This will be a regression test and not for the correctness of the output, but that's already something.

The reason for suggesting the extra-work is that this is the sort of place where it's easy for bugs to creep in without us realizing for a long time.

juanitorduz: Yes! You are right! I will improve this test! Thanks for the suggestions!

juanitorduz:
<p>Fixed in https://github.com/pymc-labs/pymc-marketing/pull/313/commits/54776de98b31ffc64c944fde4da166b9c8684918</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/313/discussion/#comment-1619828088'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tests/mmm/test_delayed_saturated_mmm.py
juanitorduz: @ricardoV94 I added more detailed tests that test the expected output of the forward pass. Let me know what you think about it!

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -254,6 +258,93 @@ def test_get_fourier_models_data(
...
+        channel_contributions_forward_pass_grid = (
+            mmm_fitted.get_channel_contributions_forward_pass_grid(
+                start=0, stop=2, num=3
+            )
+        )
+        assert channel_contributions_forward_pass_grid[0].sum().item() == 0
+        assert (
+            channel_contributions_forward_pass_grid[1].to_numpy()
+            == channel_contributions_forward_pass
+        ).all()
```
ricardoV94, line 299: In general I like the `np.testing.assert_allclose` more because it gives a nice output when it fails. I see you used above, so it's not really an advice :)

No need to change at all

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -254,6 +258,93 @@ def test_get_fourier_models_data(
...
+            "channel_contributions"
+        ].mean(dim=["draw", "chain"])
+        assert (
+            channel_contributions_forward_pass_mean.shape
+            == channel_contributions_mean.shape
+        )
+        np.testing.assert_array_almost_equal(
+            x=channel_contributions_forward_pass_mean / channel_contributions_mean,
+            y=mmm_fitted.y.max(),
+        )
```
ricardoV94, line 281: Interesting equivalence. Is this generally true or just because of the specific toy MMM used.

Out of curiosity (although it could make a good comment in the test), why is it true?

ricardoV94 (approved):

tests/mmm/test_delayed_saturated_mmm.py
juanitorduz: The logic of the test is: the forward pass should be in the original scale. The trace fits the model in with scaled data, so when scaling back, they should match. We are using a `MaxAbsScaler` (see https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/delayed_saturated_mmm.py#L379), so it should work for all models (not this specific toy dataset) :) 

Do you think it is worth adding this comment to the test?

ricardoV94: > Do you think it is worth adding this comment to the test?

Definitely, documenting the logic of tests is very valuable for future devs. Not only for debugging but also during refactoring, to understand if a test is still relevant and not.

juanitorduz: I added them as it is not trivial from the test. (https://github.com/pymc-labs/pymc-marketing/pull/313/commits/c48adc055c0f8471fecdc6babc3498859cf34d16)

tests/mmm/test_delayed_saturated_mmm.py
juanitorduz: Actually, your suggestion is nicer and more consistent. Changed in https://github.com/pymc-labs/pymc-marketing/pull/313/commits/c48adc055c0f8471fecdc6babc3498859cf34d16

ricardoV94:
Looks great! The comments on the grid test are even more useful. Thanks! 

Good on my side when the tests pass

Issue #319, opened 2023-07-08
Cannot Specify Other Samplers


ColtAllen:
When attempting to use a sampler other than the default (NUTS), a `ValueError` is raised because the model variables are being assigned twice. The following example is for the `ParetoNBDModel`, but it's the same deal with `BetaGeoModel`:

```
import pandas as pd
from pymc_marketing.clv.models import ParetoNBDModel

# load data
url_cdnow_rfm = "https://raw.githubusercontent.com/pymc-labs/pymc-marketing/main/datasets/clv_quickstart.csv"
df = pd.read_csv(url_cdnow_rfm)
df['customer_id'] = df.index

# init and fit model
pnbd = ParetoNBDModel(df)
pnbd.build_model()

with pnbd.model:
    
    sample_config = {
        "step": pm.Slice(), 
        "draws": 2000, 
        "tune": 1000
    }

    pnbd.fit(**sample_config)
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[51], line 18
     10 with pnbd.model:
     12     sample_config = {
     13         "step": pm.Slice(), 
     14         "draws": 2000, 
     15         "tune": 1000
     16     }
---> 18     pnbd.fit(**sample_config)

File ~/Projects/pymc-marketing/pymc_marketing/clv/models/pareto_nbd.py:319, in ParetoNBDModel.fit(self, fit_method, **kwargs)
    313 with warnings.catch_warnings():
    314     warnings.filterwarnings(
    315         message="Optimization Warning: The Op hyp2f1 does not provide a C implementation. As well as being potentially slow, this also disables loop fusion.",
    316         action="ignore",
    317         category=UserWarning,
    318     )
--> 319     super().fit(fit_method, **kwargs)

File ~/Projects/pymc-marketing/pymc_marketing/clv/models/basic.py:47, in CLVModel.fit(self, fit_method, **kwargs)
     30 def fit(
     31     self,
     32     fit_method: str = "mcmc",
     33     **kwargs,
     34 ) -> az.InferenceData:
     35     """Infer model posterior
     36 
     37     Parameters
   (...)
     44         Other keyword arguments passed to the underlying PyMC routines
     45     """
---> 47     self.build_model()
     49     if fit_method == "mcmc":
     50         self._fit_mcmc(**kwargs)

[...]

PR #320, opened 2023-07-11
enabling passing pymc step function reference in sampler_config

michaelraczycki:
Addressing issue reported by @ColtAllen  in #319. Because of context requirement the step function should be passed as a reference in the sampler_config.

----
:books: Documentation preview :books:: https://pymc-marketing--320.org.readthedocs.build/en/320/

### DIFF
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -116,12 +116,16 @@ class CLVModel(ModelBuilder):
             raise RuntimeError(
                 "The model hasn't been built yet, call .build_model() first or call .fit() instead."
             )
-
         with self.model:
             sampler_args = {**self.sampler_config, **kwargs}
             if "step" in sampler_args:
-                sampler_args["step"] = sampler_args["step"]()
-            idata = pm.sample(**sampler_args)
+                step_function_name = sampler_args["step"]
+                step_function = getattr(pm, step_function_name)
+                sampler_args["step"] = step_function()
+                idata = pm.sample(**sampler_args)
+                sampler_args["step"] = step_function_name
+            else:
+                idata = pm.sample(**sampler_args)
 
         self.set_idata_attrs(idata)
         return idata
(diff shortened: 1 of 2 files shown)


michaelraczycki:
passing function reference causes problems with json serialization, as ABCMeta objects (which the pymc step function objects inherit from) aren't JSON-serializable, therefore we need to proceed with the same way as we did with priors. So the 'step' needs to contain string representation of the desired pymc step function:
instead of passing pm.Slice() -> 'Slice'

ricardoV94:
> instead of passing pm.Slice() -> 'Slice'

Why do the sampler args need to be serializable?

ricardoV94:
Maybe it makes sense to subclass `fit` in the pareto model to allow `fit_method="mcmc_slice"`? I don't think we need to worry about custom step samplers in other models.

michaelraczycki:
The main point of model builder is the ability to completely reproduce the model, without previous knowledge about it (let's assume some devops deploying a model to AWS EC2) In that case he shouldn't be forced to know which step method should be applied on the fit step, especially if originally the models developer included it in the sampler_config, which is meant to be preserved.

michaelraczycki:
We're not, because it's not part of neither model nor sampler config. We're saving only idata, which attrs contain model_config, sampler_config and input parameters in case of MMM models (the CLV models don't have any input params outside of dataset, which is saved within the idata). This and model version + id

ricardoV94:
If you're not saving `fit_method="mcmc"` vs `fit_method="map"`, I don't see how that's different to also not save `fit_method="mcmc_slice"`? Your argument against `"mcmc_slice"` would also apply to `"map"`

michaelraczycki:
I'm getting confused now, you @'ed me under issue reporting this, and when I asked if it's something that we'd like to have everywhere you didn't object. For the future reference if it's not something that we want to address on a global level let's state it directly in the issue, so I know that I shouldn't spend time for investigation and implementation.
We could add another handler for covering the mcmc/MAP choice and keep it managed completely through the sampler_config, or just abandon the idea, but seeing it now actually it looks like something I'd like to have on the model_builder level, then it's up to users to either keep it when they re-implement the fit() function to match their project needs

ricardoV94:
If the idea of ModelBuilder is that the sampler settings are also saved/loaded then I agree it makes sense in the base class.

Issue #172, opened 2023-02-22
TestBetaGeo `setup_class` is too slow


ricardoV94:
The `setup_class` of TestBetaGeo is too slow as it performs MCMC sampling on a synthetic dataset. We should make this a fixture of tests that actually need it (if there's more than one), and possibly mark them with `@pytest.mark.slow` so that they are not run locally by default.

https://github.com/pymc-labs/pymc-marketing/blob/7ebc19465ad6daf62fbcb56d1b91fc40d09da7a0/tests/clv/models/test_beta_geo.py#L50

For other tests I have faked posterior draws by doing prior sampling with narrow priors around the "expected" values, and setting that as the posterior dataset. This allows us to check the summary methods work as expected without doing slow mcmc sampling.

CC @larryshamalama

ColtAllen:
I'm working on the [ParetoNBD PR](https://github.com/pymc-labs/pymc-marketing/issues/127) right now, and have added `pytest` fixtures for `CDNOW_sample.csv` and `CDNOW_master.csv` which could potentially resolve this issue. These CSVs contain 2,357 and 23,570 rows respectively, and any tests using them should probably be marked as `@pytest.mark.slow`.

My opinion on using CDNOW for testing has flip-flopped in recent weeks because even `CDNOW_master` is much smaller than many datasets encountered in practice, but these are real-world benchmarks used in many research papers, and will be useful for testing against lifetimes `MLE` convergence and reproducing research results.

ColtAllen:
Resolved in https://github.com/pymc-labs/pymc-marketing/pull/709

Issue #326, opened 2023-07-13
Adapt Evaluation Plots from `lifetimes`


ColtAllen:
`lifetimes` contains a variety of plotting functions for model evaluation:

- [ ] `plot_period_transactions`
- [ ] `plot_calibration_purchases_vs_holdout_purchases`
- [x] `plot_frequency_recency_matrix`
- [x] `plot_probability_alive_matrix`
- [ ] `plot_expected_repeat_purchases`
- [ ] `plot_history_alive`
- [ ] `plot_cumulative_transactions`
- [ ] `plot_incremental_transactions`
- [ ] `plot_transaction_rate_heterogeneity`
- [ ] `plot_dropout_rate_heterogeneity`

[This notebook](https://github.com/ColtAllen/marketing-case-study/blob/main/case-study.ipynb) provides some examples of their use.

@larryshamalama already added the matrix plots, but the others would need to be modified for posterior confidence intervals. Some of them also require utility functions that I'll create PRs for soon. I don't consider `plot_history_alive` to be all that useful, but if there's interest it's something to consider.

It's also worth reviewing the [research papers](https://github.com/pymc-labs/pymc-marketing/issues/91) for additional ideas of plots to add.

zwelitunyiswa:
I have also found tracking plots incredibly useful for explaining to
stakeholders. CLVTools has some very nice implementations:

https://www.clvtools.com/articles/CLVTools.html (scroll down)

ColtAllen:
> I have also found tracking plots incredibly useful for explaining to
stakeholders. CLVTools has some very nice implementations:

I think `plot_incremental_transactions` is the equivalent in `lifetimes`. Here's an example:

![image](https://github.com/pymc-labs/pymc-marketing/assets/10178857/847cd387-4dd8-4394-9d37-7d87dc431b4a)

ColtAllen:
`plot_period_transactions` is essentially plotting a posterior predictive check for customer purchase frequency:

![image](https://github.com/pymc-labs/pymc-marketing/assets/10178857/41075fa0-b037-427e-8879-71890fee44b4)

 Unless I'm mistaken, this means we can't use a `Potential` to define the likelihood. `GammaGammaModel` is fine because this plotting method doesn't apply, but `BetaGeoModel` would need its own distribution block per https://github.com/pymc-labs/pymc-marketing/issues/128.

zwelitunyiswa:
[...]

PR #327, opened 2023-07-13
`ParetoNBDModel` population distribution sampling and save/load test

ColtAllen:
This PR is a prerequisite to use the plotting functions in https://github.com/pymc-labs/pymc-marketing/issues/326 for the `ParetoNBDModel`. I also added a test for the save/load functionality.

This PR is still a draft until I figure out why `pytest` can't detect the `customer_id` column in the test dataset. More importantly, the `rtol` is failing for the distribution tests. Could this have something to do with HalfFlat priors being used for posterior predictive sampling? `ParetoNBDModel` uses Weibull priors by default.

Also, the naming conventions here are slightly different compared to the equivalent methods in other models. I'm not sure how I feel about including 'new_customer' in the name because these distributions were estimated from existing customers.


ricardoV94:
> Could this have something to do with HalfFlat priors being used for posterior predictive sampling? 

Those are just dummy variables in Posterior predictive, that are replaced by the draws from the trace. Some explanations in this blogpost: https://www.pymc-labs.io/blog-posts/out-of-model-predictions-with-pymc/  Look for `Flat` uses in it.

ricardoV94:
>  More importantly, the rtol is failing for the distribution tests.

Isn't your `rtol` just too restrictive. If you sample two identical `Gamma` with those parameters and sample size doesn't it fail as well?

ColtAllen:
Tests pass when N is increased to 10,000. If there are any objections to setting it that high let me know, but after a visual inspection, heavy-tailed distributions are pretty common with this model. I'd prefer a large sample size so users don't run into similar issues.

ricardoV94:
10k seems pretty extreme. If users do the normal fit with 4k draws, that's what they get back.

For tests purposes it's fine to have a higher `rtol`

ColtAllen:
After adding some plots to the dev notebook, I noticed the X-axis for these PDFs didn't quite look right. I think `pm.Gamma` needs to be parametrized with `mu` and `sigma` instead of `alpha` and `beta`, in which case https://github.com/pymc-labs/pymc-marketing/pull/303 will also need to be revised.

Using `mu` and `sigma` requires dramatically larger tolerances. I had to set `rtol=0.5, N=4000` just to get these tests to pass.

pymc_marketing/clv/models/pareto_nbd.py
```diff
@@ -618,3 +615,79 @@ def _log_B_three(i):
...
+            beta = pm.HalfFlat("beta")
+
+            # This is the shape if using fit_method="map"
+            if self.fit_result.dims == {"chain": 1, "draw": 1}:
+                shape_kwargs = {"shape": 4000}
+            else:
+                shape_kwargs = {}
+
+            pm.Gamma("population_purchase_rate", mu=r, sigma=alpha, **shape_kwargs)
+            pm.Gamma("population_dropout", mu=s, sigma=beta, **shape_kwargs)
```
ricardoV94, line 641: Are you sure those parameters correspond to `mu` and `sigma`. Statistical analysis of the `Gamma` usually work with `rate` and `shape`, so I would expect that's what the Pareto marginalized likelihood is defined in terms of

ricardoV94:
> I'm not sure how I feel about including 'new_customer' in the name because these distributions were estimated from existing customers.

The posterior population distribution is the prior distribution for a new customer after inference. The distribution for the observed customers should actually be different and are unlikely to still be a Gamma. You don't have access to them because they were marginalized during sampling. I think the `new_customer` suffix makes sense in this light.

pymc_marketing/clv/models/pareto_nbd.py
ColtAllen: The research papers all state the `scale` parameter is used instead of `rate`, but I don't see `scale` among the arguments for `pm.Gamma`.

Using the `rate` parameter results in distributions like this, which makes no sense for a dataset where most transactions are days apart:

image

ricardoV94: `scale = 1 / rate`. Does that seem to work? We should confirm carefully though.

ColtAllen: It seems to, after running it through this PDF app:

https://homepage.divms.uiowa.edu/~mbognar/applets/gamma.html

ricardoV94: I meant to confirm this is the parametrization that the paper uses. But given the raw outputs I am pretty convinced it is

ColtAllen:
> The population distribution you are sampling is the posterior predictive distribution for a new customer after doing inference on your current customers. That is, your updated prior, for what a new customer might be like.

After considering the distributions for individual customers, this makes sense. I'm about to leave for the weekend, but I'll change the names after I get back. Adding dedicated plotting functions for these distributions would require standardizing method names across all models, but I'm sure how advantageous this would be over just using something like `arviz.plot_dist`.

I'll also try using `arviz.plot_dist` with these methods instead of `seaborn`.

> I don't know if the ParetoNBDModel has closed form solutions that allow you to recover the individual posteriors, but those would belong to a separate method.

It does, and it's on my list of things to add:

https://github.com/pymc-labs/pymc-marketing/pull/327#issuecomment-1636122278

pymc_marketing/clv/models/pareto_nbd.py
ColtAllen: From https://www.brucehardie.com/notes/009/pareto_nbd_derivations_2005-11-05.pdf:

Screenshot 2023-07-16 at 12 42 37 PM

ColtAllen: No DIV/0 warnings came up when I tested this with NUTS in the notebook, so I don't see a need to suppress those warnings for now.

ColtAllen:
Method names have been changed and tests are passing with `N=4000` and `rtol=0.05` with the conversion from scale to rate. I've also added a variety of plotting options to the dev notebook.

I noticed this warning for the save/load functionality which is also impacting `BetaGeoModel`:

```
tests/clv/models/test_pareto_nbd.py::TestParetoNBDModel::test_save_load_pareto_nbd
  /python3.8/site-packages/arviz/data/inference_data.py:152: UserWarning: fit_data group is not defined in the InferenceData scheme
    warnings.warn(
```

ColtAllen:
I realized yesterday the `ParetoNBD` distribution block can also be included in `_distribution_new_customers` to simulate customer transactions! I'll get this added to the PR.

I'm also wondering if `_distribution_new_customers` should be extended to support `sample_prior_predictive`. There is a property for this in the base class, but I haven't found a way to use it for these purchase and dropout distributions.

ricardoV94:
> I'm also wondering if _distribution_new_customers should be extended to support sample_prior_predictive

Sounds like a good idea

ColtAllen:
> > I'm also wondering if _distribution_new_customers should be extended to support sample_prior_predictive
> 
> Sounds like a good idea

Does this mean [`CLVModel.sample_prior_predictive`](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/models/basic.py#L228) and [`CLVModel.prior_predictive`](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/models/basic.py#L228) can both be removed after `_distribution_new_customers` supports this for all models? If so I'll do that in a future PR.

ricardoV94:
I am not sure about removing prior predictive. What does it do now?

ColtAllen:
Reduced `N` to 1k and widened the `rtol` a bit so these functions sample faster for end users. The notebook is a mess of ideas I need to clean up before merging.

Please look over my latest commit: It's a work-in-progress refactoring for prior predictive sampling and eliminating `_distribution_new_customers`. How important is it that the returned object be an `xarray.Dataset`?

ricardoV94:
> It could also permit prior sampling from the purchase and dropout rate distributions if those variables are added to ParetoNBDModel.model, but sampling would be much slower in that scenario, so it's better to leave them in _distribution_new_customers.

We don't have to use the model in `build_model` for prior / posterior predictive checks. There is little reason to do so in these CLV models with marginalized variables. Still it then raises the question of what should `sample_prior_predictive` do. If it's too ambiguous for users / there are many things they would want, then using the specialized methods like `distribution_new_customers` sounds like a better approach.

Note that you can use `pm.sample_posterior_predictive` to do prior predictive as well. Just pass an empty trace and you'll get prior predictive draws. That way you may be able to reuse the same code more easily? 

**However**, prior_predictive will fail with default `HalfFalt` priors because those can't be taken draws. Not a problem for this model, but something to keep in mind for others.

> How important is it that the returned object be an xarray.Dataset?

Quite important for consistency and visualization with arviz

ColtAllen:
> Note that you can use `pm.sample_posterior_predictive` to do prior predictive as well. Just pass an empty trace and you'll get prior predictive draws. That way you may be able to reuse the same code more easily?

I'd prefer this approach over attempting to reinvent the wheel to save a few lines of code. What does passing an empty trace look like? I tried this, It seems the trace argument is required:

```
with self.model:
    self.idata.extend(pm.sample_posterior_predictive())
```

ricardoV94:
Oh sorry I meant calling `pm.sample_prior_predictive()` on the same model where you would call `sample_posterior_predictive` to get the same kind of draws

ColtAllen:
> Oh sorry I meant calling `pm.sample_prior_predictive()` on the same model where you would call `sample_posterior_predictive` to get the same kind of draws

Would this still work properly with the dummy `HalfFlat` priors though? I thought the priors would need to match that of `base_model` for a prior predictive check.

>  Still it then raises the question of what should sample_prior_predictive do. If it's too ambiguous for users / there are many things they would want,

I've been thinking about this, and honestly don't see it coming up a lot in practice. We can already do both prior and post predictive checks with the `ParetoNBD` block, which is far more important because it pertains to the frequency of purchases and will be used for model evaluations. I'd rather not hold up this PR trying to get prior checks to work with the rate distributions, so I'll archive the code I wrote in that last commit in case anyone requests this in the future.

ColtAllen:
I've cleaned up the dev notebook and decided to go with the existing code convention that is also used in other models.

I'm not sure how useful a prior predictive check is for the customer purchase and dropout distributions since there is no observable data to compare against. There's an example in the dev notebook on how to do this plot, but adding this functionality to the CLV module is enough work to warrant a separate PR.

I'll save the customer purchase sampling/simulation for another PR as well to keep this one more focused. There's a lot of code relevant to the customer purchase distribution in the notebook already, so I'll work on that PR next.

Lastly, the `shape_kwargs = {}` line in the if/else block isn't being covered by the tests (it requires an `arviz.idata` object with chain and draw > 1) but I worked with a fitted MCMC model in the notebook, so it's not causing any problems. If that's cool, I think we're good to merge.

ColtAllen:
We can go ahead and merge this if you want, but I've decided to append the prior predictives for the purchase and dropout rate distributions to the returned `xarray.Dataset` containing the posterior predictives for those respective methods.

Also gonna add the population `ParetoNBD` class to `_distribution_new_customers` and give it its own similar method with prior and posterior predictives (it could call `BaseModel.sample_posterior_predictive` under the hood).

Lastly, I think `_distribution_new_customers` should be expanded and renamed to what it basically is: `_sample_posterior_predictive`, with an arg to extend to `self.idata`.

ColtAllen:
That failing test was for the work we decided to move to a future PR. I've backed it up in a new branch and removed it from this one.

tests/clv/models/test_pareto_nbd.py
```diff
@@ -331,3 +334,97 @@ def test_expected_purchase_probability(self, test_n, test_t):
...
+        )
+        np.testing.assert_allclose(
+            customer_dropout.mean(), pm.draw(mu.mean(), random_seed=rng), rtol=rtol
+        )
+        np.testing.assert_allclose(
+            customer_dropout.var(), pm.draw(mu.var(), random_seed=rng), rtol=rtol
+        )
+
+    def test_save_load_pareto_nbd(self):
+        # TODO: Why isn't self.data working here?
```
ricardoV94, line 402: Unclear TODO, can we remove it?
```suggestion
        # TODO: Why isn't self.data working here?
```

tests/clv/models/test_pareto_nbd.py
```diff
@@ -331,3 +334,97 @@ def test_expected_purchase_probability(self, test_n, test_t):
             est_purchases_new_customer.mean(("chain", "draw")),
             rtol=0.001,
         )
+
+    def test_dropout_purchase_distributions(self) -> None:
+        # TODO: Why isn't self.data working here?
```
ricardoV94, line 339: Unclear TODO
```suggestion
        # TODO: Why isn't self.data working here?
```

ricardoV94 (commented):
Looks good, just a question about the TODO comments. Are they still relevant? If yes, what do they mean?

ricardoV94 (approved):

tests/clv/models/test_pareto_nbd.py
ColtAllen: Test data is currently being loaded in the `setup_class` method as a class attribute, but is not being recognized by these tests for some reason. I've made a note to create a `pytest` fixture for the test dataset in a future PR, and modified these TODOs accordingly.

tests/clv/models/test_pareto_nbd.py
ColtAllen: See previous comment.

A `pytest` fixture for the test dataset could also be used for testing the `BetaGeoModel` and probably even resolve https://github.com/pymc-labs/pymc-marketing/issues/172, so it may be better to leave it for a future PR.

PR #329, opened 2023-07-25
Budget allocation and optimal point estimations

cetagostini:
Hello team!

# Context
As marketers, we're perpetually hunting for ways to amp up the effectiveness of the campaigns we churn out regularly.

Hence, possessing reliable methods to allocate our budget efficiently—wherein we maximize our outcomes—is paramount in ensuring the success of our marketing investments.

## Competitors
Other tools like [Robyn](https://github.com/facebookexperimental/Robyn) and [Lightweight](https://github.com/google/lightweight_mmm) MMM offer features to create budget allocation aimed at maximizing the target. This highlights the importance it holds within the community.

# Goal
Integrating a budget allocator function within PyMC-Marketing that allows users to conveniently select a budget and allocate their resources to those channels that maximize the value of their target variable.

## Secondary goals (Value proposition)
1. Including optional variables that allow the use of business rules within the budget optimizer, so users can tinker more freely and optimize the variable based on their own conceptions (Priors).
2. Incorporating additional information that allows users to better understand where they should not be overspending, and where their spend start to diminish.
3. Add uncertainty around the curves.

# Hypotesis
As we are dealing with diminishing return curves, which have a sigmoidal shape, meaning, as `x` becomes excessively positive (approaching infinity), the sigmoid function approaches its upper limit. **Based on my interpretation of the `logistic_saturation` function on `mmm/transformers.py`.**

We should find different sections on the curve:
1. One where our curve increase and gets more results, even if those results are just a few on absolute numbers.
2. Other where our curve increases very slowly and gets a lot of results on absolute numbers, but at a higher cost.

These two sections of the curve would be divided by the `elbow`, basically where our curve changes direction and starts to be flatted.

image

# Solution
If we know the function that defines the curve of our data, we can project it to discern which channels have the most potential for development according to our model and which do not. This empowers the budget allocator to distribute our resources non-linearly, based on the data.

Example output:
Screenshot 2023-07-25 at 23 15 05

Screenshot 2023-07-25 at 23 09 14

# Work.
To achieve our goal, the first step was to find a function that could interpret the points in a coherent manner. Using the original function was challenging due to its scaled nature, and it operates with values and parameters between 0 and 1, which are difficult to retrieve after the model is trained.

I chose the function of Michael Mentes, which is expressed as follows:
`y = L * x / K + x`
​
where:
- L is the maximum value the curve approaches as x goes to infinity (akin to the saturation level).
- k is the x value when the function reaches half its maximum value. This can be thought of as a measure of the steepness of the curve or the position of the "elbow".

## How look like?
- When `k=0.5`, the curve saturates quickly and then plateaus. This is because the "elbow" is closer to the y-axis.
- As k increases, the curve starts to look more like an linear curve, with the elbow moving closer to the x-axis.
**ps: This given the same value for `L`**

image

## How do these changes affect the current curve fitting?
Since we now have a quadratic function, at times the model tends to believe that after passing the maximum point, any additional input will yield negative returns. This is indeed true when discussing marginal returns, but in this case, our curve is dealing with absolute values.

**Example 1: Current fit**
Screenshot 2023-07-25 at 23 35 19

**Example 2: Modify fit (Michael Mentes)**
Screenshot 2023-07-25 at 23 36 02

Great to hear your opinions here: @ricardoV94 @juanitorduz 

# Questions
-  First of all, I would like to get your feedback on the reasoning and logic behind, it and if makes sense to you to move in this direction with PyMC-Marketing. Based on my experience, makes a lot of sense but I could be biased.
- Even when we use our model to fit the data into a logistic (sigmoid) function, within the response curve plots, we make use of a quadratic (polynomial) fit. Given by default on Seaborn. Why? 

# Help
- **Unit tests,** I already build some but I'm not so familiarized. It would be amazing if you could check the functions built and validate are working as expected.

# Important
This PR is a synthesis of the one previously opened [here](https://github.com/pymc-labs/pymc-marketing/pull/287). The idea is to take all the feedback received earlier to create this cleaner and more succinct draft for new contributors. I hope you find it appropriate. I'm also open to suggestions.

_I still need to run this on some notebooks to ensure that the results are coherent. Although everything has worked as custom functions in my test notebook, I must try installing this version of the PR._

## Code snippet
``` python
#Estimate your curve (Michaelis Menten)
parameters = mmm.compute_channel_estimate_points_original_scale()

#Budget allocation based on the estimations
mmm.budget_allocation(
    total_budget=5,
    parameters = parameters,
    budget_bounds = {'x1':[1,30],
                     'x2':[1,60]
                     }
)
```

``` python
#Check the curve
mmm.plot_direct_contribution_curves(show_estimations=True)
```

## Example notebook
[Google Colab: #329 PR](https://colab.research.google.com/drive/1Q9_5tUczYjH2XQ5OsnesmGssyuu0qS7P?authuser=1#scrollTo=EVdcCGNT7KdB)

cc: @juanitorduz @ricardoV94 @twiecki @cluhmann 

----
:books: Documentation preview :books:: https://pymc-marketing--329.org.readthedocs.build/en/329/


pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,43 @@ def generate_fourier_modes(
             for func in ("sin", "cos")
         }
     )
+
+
+def michaelis_menten(x, L, k) -> float:
```
ferrine, line 41: This is also useful as a dedicated saturation function in mmm.transformers

cetagostini: I like this idea, checking Google MMM lightweight they had included several model configurations (Hill, Carryover, Adstock) which you could choose to define Saturation and Lagging.
**Example:**
Screenshot 2023-08-01 at 14 49 09

They refer to the following wiki where the saturation mentioned it is very similar to the `Michaelis Menten`. I could imagine we can implement something similar here, in order to be more extensive to how different existing data sets can fit better to different saturation and lagging functions depending on their own conditions.

Screenshot 2023-08-01 at 14 46 00

pymc_marketing/mmm/base.py
```diff
@@ -472,8 +474,105 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
...
+    def _plot_estimations(self, x: np.ndarray, channel: str, i: int) -> plt.Figure:
+
+        channel_contributions = self.compute_channel_contribution_original_scale().mean(
+            ["chain", "draw"]
+        )
+
+        fig_estimations, ax_estimations = plt.subplots(figsize=(8, 6))
+
+        L, k = estimate_menten_parameters(channel, self.X, channel_contributions)
+        plateau_x = k * (0.99 * L / (L * 0.01))
```
juanitorduz, line 486: Why do you use these 0.99 and 0.01 ?

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,111 @@
...
+    for channel, budget in zip(channels, x):
+        L, k = parameters[channel]
+        sum_contributions += michaelis_menten(budget, L, k)
+
+    return -1 * sum_contributions
+
+
+def optimize_budget_distribution(total_budget, budget_ranges, parameters, channels):
+    """
+    Calculate the optimal budget distribution that minimizes the objective function.
```
juanitorduz, line 61: Here we could also benefit by what you mean by `objective function`

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+
+    return {channel: budget for channel, budget in zip(channels, result.x)}
+
+
+def budget_allocator(
+    total_budget: int = 1000,
+    channels: Union[List[str], Tuple[str]] = [],
+    parameters: Optional[Dict[str, Tuple[float, float]]] = {},
+    budget_ranges: Optional[Dict[str, Tuple[float, float]]] = {},
```
juanitorduz, line 97: Same comment about `Optional` as above. Also we **should never use mutable structures as defaults**, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments 😈

juanitorduz (changes requested):
This was very nice and easy o review thanks! I left some small comments about code style. 

I will continue testing the budget allocator in the mmm example notebook :)

pymc_marketing/mmm/budget_optimizer.py
juanitorduz: Can you please add docstrings for this function? This is very important from a user perspective.

juanitorduz:
@cetagostini Would you mind sharing an example code of how this will be used in our example? You can assume the `mmm` object is already fitted. Thank would help the review. 

I want to run:

```pyhton
mmm.budget_allocation(
    total_budget=1000,
)
```

and I get the (expected error)

```python
TypeError                                 Traceback (most recent call last)
Cell In[43], line 1
----> 1 mmm.budget_allocation(
      2     total_budget=1000,
      3 )

TypeError: BaseMMM.budget_allocation() missing 2 required positional arguments: 'parameters' and 'budget_bounds'
```

So a reproducible code snipped will help the review :)

cetagostini:
@juanitorduz Again, thank you very much for your review!

Before sharing, I did a small verification and found a couple of issues to be corrected. It took me a bit longer, but here is the snippet and a small Colab to check the results, so you can play with it.

## Code snippet
``` python
#Estimate your curve (Michaelis Menten)
parameters = mmm.compute_channel_estimate_points_original_scale()

#Budget allocation based on the estimations
mmm.budget_allocation(
    total_budget=5,
    parameters = parameters,
    budget_bounds = {'x1':[1,30],
                     'x2':[1,60]
                     }
)
```

``` python
#Check the curve
mmm.plot_direct_contribution_curves(show_estimations=True)
```

## Example notebook
[Google Colab: #329 PR](https://colab.research.google.com/drive/1Q9_5tUczYjH2XQ5OsnesmGssyuu0qS7P?authuser=1#scrollTo=EVdcCGNT7KdB)

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,111 @@
...
+import numpy as np
+from pandas import DataFrame
+from scipy.optimize import minimize
+
+from pymc_marketing.mmm.utils import michaelis_menten
+
+
+def calculate_expected_contribution(parameters, optimal_budget):
+    """
+    Calculate the total expected contribution of budget allocations across various channels.
```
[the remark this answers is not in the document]
cetagostini: Done, I add a larger docstring trying to describe the process and trying to be specific.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Returns
+    -------
+    dict
```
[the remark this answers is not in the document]
cetagostini: Copy that!

pymc_marketing/mmm/budget_optimizer.py
cetagostini: Basically, the negative of the total expected contribution for the given budget allocation.

pymc_marketing/mmm/budget_optimizer.py
cetagostini: I think I fixed it, and removed all possible mutable objects inside my function, but let me know if you see any. The docstring is done as well.

cetagostini:
All changes already done!

juanitorduz (changes requested):

pymc_marketing/mmm/base.py
```diff
@@ -472,8 +474,136 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
...
+        that maximizes the total expected contribution across all channels.
+
+        Parameters
+        ----------
+        total_budget : int, requiere
+            The total budget to be distributed across channels.
+        parameters : dict, requiere
+            A dictionary where keys are channel names and values are tuples (L, k) representing the
+            Michaelis-Menten parameters for each channel.
+        budget_bounds : dict, optional
```
juanitorduz, line 550: Here do you mean optional as in there is a default value (because there is no at the moment)? Recall that the type hint `Optional` refers to the fact that it can take `None` values. Not that is really optional.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+
+    return {channel: budget for channel, budget in zip(channels, result.x)}
+
+
+def budget_allocator(
+    total_budget: int,
+    channels: List[str],
+    parameters: Dict[str, Tuple[float, float]],
+    budget_ranges: Optional[Dict[str, Tuple[float, float]]],
```
juanitorduz, line 157: Again we should remove the `Optional` hint if `budget_ranges` can not be `None`.

pymc_marketing/mmm/base.py
```diff
@@ -472,8 +474,136 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
...
+            )
+
+        return budget_allocator(
+            total_budget=total_budget,
+            channels=list(self.channel_columns),
+            parameters=parameters,
+            budget_ranges=budget_bounds,
+        )
+
+    def compute_channel_estimate_points_original_scale(self) -> Dict:
```
juanitorduz, line 584: I would rename this method as `compute_channel_plateat_points_original_scale`

pymc_marketing/mmm/base.py
```diff
@@ -472,8 +474,136 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
...
+            ["chain", "draw"]
+        )
+
+        return {
+            channel: estimate_menten_parameters(channel, self.X, channel_contributions)
+            for channel in self.channel_columns
+        }
+
+    def plot_direct_contribution_curves(
+        self, show_estimations: bool = False, x_stop=None
```
juanitorduz, line 603: can re rename the parameter `show_estimations` to `show_michaelis_menten_fit`

juanitorduz: By default we should keep the original range. In the example notebook when I run 

```python
fig = mmm.plot_direct_contribution_curves(show_estimations=True)
[ax.set(xlabel="x") for ax in fig.axes]
```
I get:

image

juanitorduz: I would rename `x_stop` to `xlim_max`

juanitorduz: Otherwise, using `x_stop = 1.5` I get 
image

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -0,0 +1,171 @@
...
+    optimal_budget = optimize_budget_distribution(
+        total_budget, budget_ranges, parameters, channels
+    )
+
+    return DataFrame(
+        {
+            "estimated_contribution": calculate_expected_contribution(
+                parameters, optimal_budget
+            ),
+            "optimal_budget": optimal_budget,
```
juanitorduz, line 169: From the example notebook I get 
image
Should we add the expected total value instead of `NaN`?

juanitorduz (changes requested):
Looks very nice and works on the example notebook. I left minor comments regarding naming and type hints. After that I think we can merge this an add an expermiental warning to get feedback from users 💪

juanitorduz:
@cetagostini I also suggest opening a followup PR where we load the mmm model from the example (we can use the model builder here) to run an explain this optimization procedure.

cetagostini:
Hey @juanitorduz 

I apply all the changes you mention, and additionally create a new [PR with the documentation here](https://github.com/pymc-labs/pymc-marketing/pull/370).

# **Relevant changes**
## 1 - Adding custom saturation method:
I made some important changes, the first and most important is that I added a new function inside the PR. The idea is that now users can choose the method to fit their curve, not only `michaelis-menten` but also `sigmoid`. For this use the native function of the model with a little adjustment.

$$
\frac{\alpha - \alpha \cdot \exp(-\lambda x)}{1 + \exp(-\lambda x)}
$$

**Where:**
* α (Alpha): Represents the Asymptotic Maximum or Ceiling Value. It is the value the function approaches as the input x becomes large.
* λ (Lambda): Affects the steepness of the curve. A higher value of λ makes the curve steeper, while a lower value makes it more gradual.

**Why modify the function and add the alpha parameter?**
Right now, since most of the process requires understanding the maximum points of the curve and the inflection points, adding it made me more scalable, without requiring so many changes, because now both functions will be estimated with two parameters. Additionally, I think that having a cealing parameter also makes the function somewhat more interpretable, personal opinion.

By the way, and I think we can debate, doing tests I think it would even be good to estimate it directly with the base model. We could have the model with two saturation functions, which could be estimated and chosen by the user. Each one with its pros and cons (But this is the subject of another PR).

**Why not use the internal functions of PyMC?** 
I was playing around with this idea for a while but I didn't get a satisfactory fit. Even passing the base parameters of the model. Also as I mentioned previously, even trying to move forward with this idea, use the function due to the difference in parameters would have required more code lines.

``` python
lam = float(mmm.fit_result['lam'].sel(channel='tiktok_cost_android').mean())
beta = float(mmm.fit_result['beta_channel'].sel(channel='tiktok_cost_android').mean())
alpha = float(mmm.fit_result['alpha'].sel(channel='tiktok_cost_android').mean())

print(beta,lam, alpha)

def max_abs_scaler_factor(x):
    return np.max(np.abs(x))

x = df['tiktok_cost_android'].to_numpy()
y = channel_contributions.sel(channel='tiktok_cost_android').to_numpy()

factor_x = max_abs_scaler_factor(x)
factor_y = max_abs_scaler_factor(y)

x_range = np.linspace(min(x), max(x), 100)

y_saturated = logistic_saturation(
    x=(x_range / factor_x),
    lam=lam
).eval()

plt.scatter(x, y,)
plt.plot(x_range , y_adstock_saturated * factor_y, #re-scaling based on the `y`
linestyle='-')
plt.xlabel('x')
plt.ylabel('y')
plt.title('Fitting Curve')
plt.show()
```
Screenshot 2023-08-29 at 00 04 39

## 2 - Adding plot method to compare budget recommendations:
Basically, based on the recommended budget, create a method to be able to compare the initial scenario with respect to the new ones proposed, allowing you to see the differences.

Here is an small snippet but you can see more on the other PR inside of the [example notebook](https://github.com/pymc-labs/pymc-marketing/pull/370).

``` python
figure_ = mmm.plot_budget_scenearios(base_data=initial_scenario, method='sigmoid', scenarios_data=[result_sigmoid])
```
image

As you can see there are some shaded sections inside the plot, basically I'm looking to add the intervals of the curve using the 0.05 and 0.95 quantiles of our posterior. Meaning, the shaded areas represent the maximum expected contribution, the minimum, and the expected average. 

## 3 - Adding intervals on the method plot_direct_contribution_curves:
Now we can see the intervals, again based on the quantiles and the selected fit (Sigmoid or Menten) in addition to the estimate of the alpha (Ceiling) parameter for both functions.
``` python
sigmoid_response_curve_fig = mmm.plot_direct_contribution_curves(
show_fit = True, method='michaelis-menten'
)
```
image

I would think that with this we are closer to leaving. However, you may be leaving something in the dark regarding hints or unit tests. For which I ask you for help if you consider a function that we should test correctly and you do not see it there in the tests.

I hope for your always appreciated feedback! 🚀

pymc_marketing/mmm/base.py
```diff
@@ -472,8 +474,136 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
...
+        ----------
+        x : np.ndarray
+            The x-axis data, usually representing the amount of input (e.g., substrate concentration in enzymology terms).
+        ax : plt.Axes
+            The matplotlib axes object where the plot should be drawn.
+        channel : str
+            The name of the channel for which the curve fit is being plotted.
+        color_index : int
+            An index used for color selection to ensure distinct colors for multiple plots.
+
```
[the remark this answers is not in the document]
cetagostini: Added

pymc_marketing/mmm/base.py
cetagostini: Correct, should be optional (None allow). If the user doesn't use it and the value is None then some bounds are used by default.

pymc_marketing/mmm/budget_optimizer.py
cetagostini: Yes, can be optional. If the user doesn't use it and the value is None then some bounds are used by default.

pymc_marketing/mmm/base.py
cetagostini: I switch to `compute_channel_curve_parameters_original_scale` given the fact we can use a differents curve fit now. What do you think?

pymc_marketing/mmm/budget_optimizer.py
cetagostini: Correct, already working as expected.

Screenshot 2023-08-29 at 00 40 31

cetagostini:
# Small changes
1. New docstrings added to secondary support functions. 
2. The error in the unit test was solved (Apparently). Still, I believe the unit test could be basic and need a more sophisticated approach to cover the majority of the cases, a hand here would be nice.
3. Adjust the plot to avoid warning of redundant parameters.

# To be defined
1. Scaled X data: Right now I'm using the parameter `self.X` to perform the fit according to the determined function, the same parameter is used later inside the budget allocation functions, this means that the user would have to pass the scaled budget bounds, and the results of the optimal budget would also be scaled each one to its respective factor according to the channel.
- Easy fix: Replace `self.X` with the unprocessed X data.
- Less easy: Use the internal scaler parameter to unscale the data and return the results to the original scale.

PS: I'm not sure if this was always the case, I think in earlier versions the direct response curves were shown in the original scale. I haven't checked with previous versions, but we should keep an eye on it.

pymc_marketing/mmm/base.py
juanitorduz: Thanks! Do you want to add a shorter summary to the doc strings? :)

pymc_marketing/mmm/base.py
```diff
@@ -472,19 +481,491 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
...
+        """
+        channel_contributions_quantiles = (
+            self.compute_channel_contribution_original_scale().quantile(
+                q=[0.05, 0.95], dim=["chain", "draw"]
+            )
+        )
+
+        # Estimate parameters based on the method
+        if method == "sigmoid":
+
```
juanitorduz, line 520: Here is some duplicated code. You can use

```python
if method == "michaelis-menten":
    func = estimate_menter_parameters
elif method == "sigmoid":
    func = estimate_sigmoid_parameters

alpha_limit_upper, lam_constant_upper = func(
                channel, self.X, channel_contributions_quantiles.sel(quantile=0.95)
            )
```

pymc_marketing/mmm/base.py
```diff
@@ -472,19 +481,491 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
...
+            fit_function = michaelis_menten
+        else:
+            raise ValueError("`method` must be either 'michaelis-menten' or 'sigmoid'.")
+
+        y_fit_lower = fit_function(budget, alpha_limit_lower, lam_constant_lower)
+        y_fit_upper = fit_function(budget, alpha_limit_upper, lam_constant_upper)
+
+        return y_fit_lower, y_fit_upper
+
+    def _standardize_scenarios_dict_keys(self, d: Dict, keywords: list):
```
juanitorduz, line 547: IMO, this should be in a utils module as it does not depend on the MMM class.

- The type hint should be `keywords: List[str]`

pymc_marketing/mmm/base.py
```diff
@@ -472,19 +481,491 @@ def compute_channel_contribution_original_scale(self) -> DataArray:
...
+
+        return budget_allocator(
+            method=method,
+            total_budget=total_budget,
+            channels=list(self.channel_columns),
+            parameters=parameters,
+            budget_ranges=budget_bounds,
+        )
+
+    def compute_channel_curve_parameters_original_scale(self, method: str) -> Dict:
```
juanitorduz, line 909: I will rename this to `compute_channel_curve_optimization_parameters_original_scale` so that people do not think these are model parameters (adstock or saturation)

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
             for func in ("sin", "cos")
         }
     )
+
+
+def michaelis_menten(x, alpha, lam) -> float:
```
juanitorduz, line 41: type hints missing

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+    Returns
+    -------
+    float
+        The value of the Michaelis-Menten function given the parameters.
+    """
+
+    return alpha * x / (lam + x)
+
+
+def extense_sigmoid(x, alpha, lam) -> float:
```
juanitorduz, line 63: type hints missing

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+    """
+    Calculate the Michaelis-Menten function value.
+
+    Parameters
+    ----------
+    x : float
+        The spent on a channel.
+    alpha (Limit/Vmax) : float
+        The maximum contribution a channel can make (also known as the plateau point).
+    lam (k) : float
```
juanitorduz, line 51: what is k? (maybe you add the formula in the doctrings ?)

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+    if alpha <= 0 or lam <= 0:
+        raise ValueError("alpha and lam must be greater than 0")
+
+    return (alpha - alpha * np.exp(-lam * x)) / (1 + np.exp(-lam * x))
+
+
+def estimate_menten_parameters(
+    channel: str,
+    original_dataframe,
+    contributions,
```
juanitorduz, line 83: type hints

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+    of the extended sigmoid function. The parameters are estimated by minimizing the
+    least squares error between the observed data and the values predicted by the
+    extended sigmoid function.
+
+    Parameters
+    ----------
+    x : array-like
+        The input data for which the parameters are to be estimated.
+    y : array-like
+        The observed data for which the parameters are to be estimated.
```
juanitorduz, line 98: These are not the same as the signature function

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+    initial_guess = [max(y), 0.001]
+    # Curve fitting
+    popt, pcov = curve_fit(michaelis_menten, x, y, p0=initial_guess)
+
+    # Save the parameters
+    return popt
+
+
+def estimate_sigmoid_parameters(
+    channel: str, original_dataframe, contributions, **kwargs
```
juanitorduz, line 118: type hints

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+    of the extended sigmoid function. The parameters are estimated by minimizing the
+    least squares error between the observed data and the values predicted by the
+    extended sigmoid function.
+
+    Parameters
+    ----------
+    x : array-like
+        The input data for which the parameters are to be estimated.
+    y : array-like
+        The observed data for which the parameters are to be estimated.
```
juanitorduz, line 133: These names do not correspond to the signature function

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+        x,
+        y,
+        p0=[alpha_initial_estimate, 0.001],
+        bounds=parameter_bounds_modified,
+    )
+
+    return popt
+
+
+def compute_sigmoid_second_derivative(x, alpha, lam) -> float:
```
juanitorduz, line 157: type hints

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+        The asymptotic maximum or ceiling value of the sigmoid function.
+    lam : float
+        The parameter that affects how quickly the function approaches its upper and lower asymptotes.
+
+    Returns
+    -------
+    float
+        The second derivative of the sigmoid function at the input value.
+    """
+    # Compute the second derivative
```
juanitorduz, line 179: We can remove this comment

pymc_marketing/mmm/utils.py
```diff
@@ -33,3 +36,184 @@ def generate_fourier_modes(
...
+    return (
+        -alpha
+        * lam**2
+        * np.exp(-lam * x)
+        * (1 - np.exp(-lam * x) - 2 * lam * x * np.exp(-lam * x))
+        / (1 + np.exp(-lam * x)) ** 3
+    )
+
+
+def find_sigmoid_inflection_point(alpha, lam) -> Tuple[Any, float]:
```
juanitorduz, line 189: type hints

juanitorduz (changes requested):
Looking good! Here are some minor style suggestions

cetagostini:
Unit test from utils working. Still, debugging is needed on the budget optimizer unit test.
Screenshot 2023-09-07 at 01 36 50

juanitorduz:
The lint problem has nothing to do with your code. Is probably and update in mypy, see https://github.com/pymc-labs/pymc-marketing/pull/376/commits/dba627882c70ae5c4c3535f7057073d5ff8734a4

cetagostini:
Budget optimizer unit test:
Screenshot 2023-09-09 at 14 28 14

juanitorduz:
Tests and lint is 🟢 ! Yay! I will take a look in the upcoming days :)

twiecki:
Can we merge?

cetagostini:
> Can we merge?

I think we are set. Not sure if the team has other questions @twiecki 

From my side:
1. We have already all checks except `codecov`. Still, I added a few new unit tests to improve and increase the target.
2. The branch is updated up to September 19th.
3. I ran a test on the current notebook using my branch and everything works properly ([Test here](https://colab.research.google.com/drive/1duJewYhl_lHmv_dO8yavqjYowfqVt0Uj#scrollTo=gvz-1QmP9eXu)).
4. I ran a test on the new example notebook for budget allocation and it works as well ([Test here](https://colab.research.google.com/drive/1r6vryO7h79eM17iTHRsm45ybKybvjWpJ#scrollTo=74q2NZd03SMr))

I added some flexibility to the `curve_fit` functions to give users the possibility to use them outside of the mmm class. This also gives the chance to handle non-fit errors given by the optimizer.

The `pre-commit` looks correct when I test it locally:
Screenshot 2023-09-19 at 20 16 15

The error given now by `docs/readthedocs.org:pymc-marketing` is related to the Numpy new version, but I assume this should be solved from the GitHub config checks. Right?

Wait for your feedback!

cc: @ricardoV94 @juanitorduz

juanitorduz:
Can we add a warning saying that this feature is experimental? What do you think ?

cetagostini:
> Can we add a warning saying that this feature is experimental? What do you think ?

Sure, where do you think can we added? Notebook? @juanitorduz

juanitorduz:
> > Can we add a warning saying that this feature is experimental? What do you think ?
> 
> Sure, where do you think can we added? Notebook? @juanitorduz

Sorry for the late reply 🙈 ! What about the public methods regarding optimization in `pymc_marketing/mmm/base.py`?

cetagostini:
I added an alert to the docstrings located under `base.py`. It is okay? @juanitorduz 

Screenshot 2023-09-30 at 14 23 04

juanitorduz:
> I added an alert to the docstrings located under `base.py`. For each new function, It is okay? @juanitorduz
> 
> Screenshot 2023-09-30 at 14 23 04

Thank you @cetagostini ! In addition, we can add a warning in the code itself as:

```python
import warnings

...

warnings.warn("This budget allocator method is experimental", UserWarning)
```

After that we can merge from my side :)

juanitorduz (approved):
@cetagostini Thank for for this contribution! IMO let us ship this as experimental and get feedback from users and iterate! 

WDYT @ricardoV94 ? @twiecki ?

twiecki:
Yes, we should merge this.

twiecki:
Incredible contribution @cetagostini!

cetagostini:
Hey guys! Didn't have the time to reply properly a few days ago, just wan to highlight a few things.

1. Really thank you for the support, I'm thrilled to see the merge finally happening 🙌🏻 I'm quite inspired by this momentum, and I'll surely try to help with the other issues that are now active, since this one is closed.
2. A great mention to @juanitorduz without all the time he dedicated and helped me get this PR forward, I would surely have stayed halfway. A great source of inspiration in the community, as well a great person to debate and share ideas reliably in the community.

ps: I'll probably be doing a little self-promotion this week to invite users to test.

Go PyMC-Marketing 🚀

PR #331, opened 2023-07-29
Drop support for python 3.8

michaelraczycki:
Since pymc-marketing doesn't support python 3.8 anymore, we need to update the CI's to reflect that, otherwise they will keep failing, blocking other current CI's

----
:books: Documentation preview :books:: https://pymc-marketing--331.org.readthedocs.build/en/331/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 env:
-  OLDEST_PYMC_VERSION: "5.5.0"
+  OLDEST_PYMC_VERSION: "5.6.1"
 
 jobs:
   lint:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.11"]
+        python-version: ["3.9", "3.11"]
         oldest-pymc: [false, true]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -29,23 +29,24 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        config: [ {python-version: "3.8", oldest-pymc: false}, {python-version: "3.11", oldest-pymc: true}]
+        python-version: ["3.9", "3.11"]
```
ricardoV94, line 32: This is gonna double the number of jobs, we were doing oldest pymc and oldest python and most recent pymc and most recent python. I think that's enough

ricardoV94:
Need to update this line as well: https://github.com/pymc-labs/pymc-marketing/blob/main/pyproject.toml#L8

.github/workflows/ci.yml
michaelraczycki: oh, good catch, actually I didn't think of that result.

ricardoV94:
Looks like you need to update PYMC oldest support version because the pymc-experimental requires 5.6.1

michaelraczycki:
@ricardoV94, can I have thumbs up on that one?

ricardoV94:
Did you mean to bump pymc-experimental as well?

michaelraczycki:
here the pymc-experimental is still kept at 0.0.8, (bump to 0.0.9 is in the notebook PR). When it comes to the 5.6.1 req, I think its something you've added

ricardoV94:
I don't understand the point of this PR then? Why is Python 3.8 not supported in pymc-marketing anymore? Which PR changed that?

ricardoV94 (approved):

michaelraczycki:
@ricardoV94 any idea how to solve that? Locally I can't reproduce this fail of Pareto_nbd test

ricardoV94:
> @ricardoV94 any idea how to solve that? Locally I can't reproduce this fail of Pareto_nbd test

It detected the just released `pymc==5.7.0`. Can you reproduce if you update pymc?

PR #341, opened 2023-08-05
working image for pymc-marketing pre installed env

michaelraczycki:
This PR includes a Dockerfile to the repo, which allows users to build an image with working jupyter lab environment, pre-installed with all necessary libraries in order to run the tutorial notebooks. Because of that no local environment creation is needed.

----
:books: Documentation preview :books:: https://pymc-marketing--341.org.readthedocs.build/en/341/


michaelraczycki:
now users can spin up the container using :
docker run -u root -p 8888:8888 -v .:/home/jovyan/work tutorial_env
command, after building the image locally.

maresb:
I think you only need two lines in the Dockerfile:

```Dockerfile
FROM jupyter/minimal-notebook:latest
RUN mamba install pymc-marketing
```

And you don't need `-u root` in the Docker command.

michaelraczycki:
oh, actually it does work with the 2 liner, now thinking about it I don't know why I was creating virtualenv for installation within the container :)

maresb:
One thing to note here is that this won't give you any sort of reproducibility. You might want to consider building the Docker image in CI. You might also consider some CI to ensure that the notebooks all execute without error. (As dependencies update, things sometimes break.) Then at least you'd know when things stop working.

michaelraczycki:
Do you mean like creating a CI that would build and push the image to the dockerhub and then get the users to pull and spin that up?

maresb:
Ya. I'd probably recommend GitHub Container Registry over DockerHub though.

michaelraczycki (commented):
@maresb did you have something like this in mind?

maresb:
Yes, that looks about right, at least for the tricky half of it. Great work!

The other half which I'd suggest implementing is to push to GHCR (GitHub Container Registry) after the "Check for errors" step succeeds.

I'd recommend adding an `on`-trigger for pull requests and also a schedule so that it runs periodically (like maybe once per day) so that you stay on top of breakage due to dependency updates. For these two cases you should skip the "Push to GHCR" step. (Note: "on push branches main" means "when you either directly push to `main` or when you merge a PR into `main`.)

Finally, regarding the scheduled tests, I believe that when they fail they will do so silently, so you will probably want to look into some notification mechanism, e.g. open an issue when the workflow fails.

juanitorduz:
I just want to point out that this PR is an excellent initiative 💪 !

maresb:
Woah @BBDS-Colt, that's some huge type!

I totally agree, and no worries, I've got you covered. Check out the code I wrote a few years ago (before the whole Aesara/PyTensor fork) [here](https://github.com/conda-forge/pytensor-suite-feedstock/blob/main/recipe/check-for-warnings.py).

maresb:
> you should install all the dependencies, then uninstall the pytensor-base and install it again using conda or mamba install

@michaelraczycki, this seems like a bug. Would you be able to file a report, perhaps at PyTensor's conda-forge feedstock?

maresb:
Didn't you also have some crazy recipe on Discord for the BLAS setup? Could you also create an issue for that? I want to build it into the feedstock when I get the chance.

michaelraczycki:
> Didn't you also have some crazy recipe on Discourse for the BLAS setup? Could you also create an issue for that? I want to build it into the feedstock when I get the chance.

Actually I don't remember the BLAS case, I can try to find it but I think it was also connected to the same pytensor issue. I can open the issue at Conda-forge pytensor, but first I'd try to reproduce the error again, it's been few months since I've last time seen it

michaelraczycki:
> Ah, on second thought it's not quite that easy for Apple Silicon... GitHub doesn't provide ARM runners, so if you want to run tests you'd have to emulate with QEMU. It's doable but a little more complicated.

was this meant as a follow up to your conversation with Colt, or to our one about the Git CI for checking the notebooks?

ColtAllen:
> More directly it was a followup with Colt, but we'd probably implement it in the CI.
> 
> There's quite a lot to do here, so perhaps it makes sense to merge this as a start and break up the other suggestions into a bunch of enhancements.

I agree; let's keep this one focused and leave the BLAS contingencies for a future PR. If you want me to post any issues for what's been discussed, let me know.

Also, this library runs flawlessly with `pip install pymc-marketing` on my 2023 M2 Pro, but pretty much any prod deployment in industry will be container-based.

maresb:
Ya, the BLAS issues are anyways specific to PyTensor and don't have anything directly to do with this.

twiecki:
Looks useful, I say let's merge it.

ricardoV94:
I don't want a new CI to test 2 lines of code if that's okay

twiecki:
> I don't want a new CI to test 2 lines of code if that's okay

That's okay ;).

maresb:
There's a lot happening in this PR, so probably best to regroup under more targeted efforts.

Potentially:

1. Build and test a pymc-marketing image (would probably make sense to reopen this after adding a push step to the CI)
2. Improve the way we test various environments and versions in general, but this is probably best tackled first at the PyTensor/PyMC level

PR #354, opened 2023-08-10
Update beta_geo.py

konkinit:
Correcting the expected alive probability theoretical formula


larryshamalama (approved):
Thanks for noticing and correcting this! :)

PR #358, opened 2023-08-17
Add optimization notebook

cluhmann:
Initial version of the optimization notebook. I have not yet added all the "bells and whistles" that are possible, because I want to get feedback on the core first.  Then we can make it pretty.

One thing that I think can immediately be improved is to replace much of the manual contribution/ROAS calculation code with things like `mmm.channel_contributions_forward_pass()` etc. These helper functions didn't exist when the notebook was first created.

Also, I had planned on omitting all the MMM example notebook code, but realized I need to serialize the model, the inferenceData object, and the input data (and possibly more?). So I just kept the code but stripped it down to the essentials. Keeping all these cells also prevents "syncing" issues between this notebook and the MMM example notebook.

Todos:
- [ ] Replace objective function code with pymc-marketing functionality as possible
- [ ] Make objective function a compiled pytensor function to avoid `eval()`s
- [ ] Add additional exploration of optimization results (e.g., risk attitudes, diversification penalty, etc.)

----
:books: Documentation preview :books:: https://pymc-marketing--358.org.readthedocs.build/en/358/


wd60622:
This seems stale. Shall we close and turn into smaller issues? 
Thoughts @cluhmann @juanitorduz

juanitorduz:
This notebook has many great ideas that I would like to explore further. Can we leave it open while I try to get some time? Maybe we can add a special label?

juanitorduz:
Or actually, as it is a dev notebook, can we merge it into the folder (as the clv dev ones)?

cluhmann:
Happy to help hammer this into something (more) useful.

wd60622:
How about merge in since it is dev notebook and then create an issue or two?

cluhmann:
Fine with me.

wd60622:
Did the latest optimization notebooks kill the need for this?

cluhmann:
The one thing that I will want out of this PR is a notebook in which we compare one model/data set and various approaches to optimization, including:

- optimization using point estimates (i.e., the old method of optimization)
- optimization using full posterior but neutral utility (e.g., mean/expected response)
- optimization using full posterior but utility functions reflecting various risk profiles

I think this that seeing how these different approaches yield different optimized spends would really drive home the value of this new approach to optimization and introduce the Bayesian decision theory stuff at the same time.  I talked to @carlosagostini about this when he was first working on the optimization.

cetagostini:
I'll close here, because we address all comments here. Current optimizer works with all posterior, and allow risk profile management.

PR #249, opened 2023-04-18
Add ref by by Dr. Robert Kübler

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/243

PR #359, opened 2023-08-18
Add bolt's journey with pymc-marketing

SangamSwadiK:
Add Bolt's journey with PyMC-marketing on pymc-marketing.io

----
:books: Documentation preview :books:: https://pymc-marketing--359.org.readthedocs.build/en/359/


ricardoV94:
I think we should put this in `mmm #resources` like in this PR? https://github.com/pymc-labs/pymc-marketing/pull/249

ricardoV94:
> I think we should do both.

If we want to add it on the index page, we should add a new section? Maybe something along the lines of `Who is using PyMC-Marketing`?

twiecki:
@SangamSwadiK Can you make those changes?

twiecki:
thanks!

Issue #375, opened 2023-09-06
MMM Channel Scaler do not working as expected


juanitorduz:
After running the model from https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_example.html I ran

```python
mmm.channel_transformer["scaler"].scale_
```
and I always get `array([1., 1.])` no matter if I scale the channel variables in `X`.

juanitorduz:
I think the problem is that we are doing the preprocessing twice:
- https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/model_builder.py#L463
- https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/delayed_saturated_mmm.py#L176

juanitorduz:
One should also be sure not to have `pymc-experimental` installed in the python env otherwise the `ModelBuilder` gets messed up ...

juanitorduz:
FYI @cetagostini

ricardoV94:
> One should also be sure not to have `pymc-experimental` installed in the python env otherwise the `ModelBuilder` gets messed up ...

Why? It shouldn't matter

juanitorduz:
I have no idea but when I was listing locals() that was one issue. After I recreated a new environment it worked as expected. I have no idea 🫠

PR #376, opened 2023-09-06
improve pre-processing MMM

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/375

The changes are small, but he big portion is just the notebook update.

- Make sure we use copies not to modify data in-place.
- This means `mmm.X` has the original channel data (not scaled!)
-  Update the mmm example notebook to make sure it is compatible with https://github.com/pymc-labs/pymc-marketing/pull/365
- Small `mypy` improvements 🤦 

----
:books: Documentation preview :books:: https://pymc-marketing--376.org.readthedocs.build/en/376/

### DIFF
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -173,6 +173,7 @@ class BaseDelayedSaturatedMMM(MMM):
             The PyMC model object containing all the defined stochastic and deterministic variables.
         """
         model_config = self.model_config
+        self._generate_and_preprocess_model_data(X, y)
         with pm.Model(coords=self.model_coords) as self.model:
             channel_data_ = pm.MutableData(
                 name="channel_data",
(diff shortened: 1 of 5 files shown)


pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -173,7 +173,6 @@ def build_model(
             The PyMC model object containing all the defined stochastic and deterministic variables.
         """
         model_config = self.model_config
-        self._generate_and_preprocess_model_data(X, y)
```
juanitorduz, line 176: @michaelraczycki There are some tests failing, and I think is because of this line. 

I think the problem is that we are doing the preprocessing twice:

- https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/model_builder.py#L463

- https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/delayed_saturated_mmm.py#L176

Is the design to preprocess the data in the build_model or fit? (at the moment is happening at both in `main` branch)

juanitorduz: IMO, it should be in the `build_model`; otherwise, when we run prior_predictive, we will not have the transformed data.

michaelraczycki (approved):
LGTM, good catch on the processed data not being copied!

PR #379, opened 2023-09-15
Ameynen branch: Mistake in betageo documentation

AMeynen:
Change to beta_geo model doc, where mode_config had incorrect dictionary keys (r instead of r_prior etc.)


ricardoV94 (approved):
Thanks 🙏!

Issue #383, opened 2023-09-23
Avoid PyTensor evals whenever possible


cluhmann:
In `BaseDelayedSaturatedMMM`, the `channel_contributions_forward_pass()` method calls `eval()` on the pytensor value to be returned ([here](https://github.com/pymc-labs/pymc-marketing/blob/3a9db517d1d3c5a930eeb80ad24c540382d1bc33/pymc_marketing/mmm/delayed_saturated_mmm.py#L386)). This causes problems if you expect the tensor returned by the various transformation methods to be passed along as-is.  For example, if you try to compile a pytensor function that incorporates a call to `channel_contributions_forward_pass()`, pytensor will complain and break.

Can we avoid calls to `eval()` whenever possible?  Obviously, handing back tensors may be a bit confusing to some users, but having the `eval()` calls scattered throughout the codebase in unsystematic ways is going to lead to problems.

cluhmann:
Thinking about this more, it probably makes sense to make any "internal" functions propagate results as tensors and any "user-facing" functions call `eval()`. Because the `eval()` mentioned above is in `BaseDelayedSaturatedMMM`, it should be considered "internal" and thus not call `eval()`.  Seem fair?

cluhmann:
We may also want to change the names of the base class methods so that they are not overridden by the child class methods. The `super.channel_contributions_forward_pass()` method may be useful even when you have a `BaseDelayedSaturatedMMM` object (i.e., the contribution forward pass without the target variable transformation).  It would be easy enough to rename the base method to `_channel_contributions_forward_pass()` or something.

cluhmann:
This is also relevant to #314

Issue #154, opened 2023-02-15
sklearn pipelines to handle data transformations?


juanitorduz:
The current MMM implementation relies on certain scalers to transform the data -> fit the model and then re-scale back. Ideally, users should not need to carry these scalers and they will be handled internally. The natural question is: **do we need sklearn to do this? why not simply carry the scaling factors?** . This is a fair question! I personally like sklearn pipelines as we can potentially allow log-transforms to allow for multiplicative models. No strong opinion though. I just open this issue to track the discussion :)

PR #386, opened 2023-10-01
Remove unnecessary calls to PyTensor eval() from user-facing methods

cluhmann:
Fixes https://github.com/pymc-labs/pymc-marketing/issues/383. The calls to `eval()` have been moved out of the `BaseDelayedSaturatedMMM.channel_contributions_forward_pass()` method and to user-facing methods in `DelayedSaturatedMMM` .  It is not entirely clear to me whether `get_channel_contributions_forward_pass_grid()` is supposed to user-facing, but I assumed it is.

What do we think about the possibility of more clearly distinguishing between the user-facing methods that return numpy/xarray objects  and the internal methods that are working with tensors?  The user-facing methods could then just be light wrappers converting/`eval()`ing the tensors to numpy/xarray arrays.

----
:books: Documentation preview :books:: https://pymc-marketing--386.org.readthedocs.build/en/386/

### DIFF
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -383,7 +383,7 @@ class BaseDelayedSaturatedMMM(MMM):
         channel_contribution_forward_pass = (
             beta_channel_posterior_expanded * logistic_saturation_posterior
         )
-        return channel_contribution_forward_pass.eval()
+        return channel_contribution_forward_pass
 
     @property
     def _serializable_model_config(self) -> Dict[str, Any]:
@@ -564,7 +564,7 @@ class DelayedSaturatedMMM(
             excluded=[1, 2],
             signature="(m, n) -> (m, n)",
         )
-        return target_transformed_vectorized(channel_contribution_forward_pass)
+        return target_transformed_vectorized(channel_contribution_forward_pass).eval()
 
     def get_channel_contributions_forward_pass_grid(
         self, start: float, stop: float, num: int
@@ -596,7 +596,7 @@ class DelayedSaturatedMMM(
             channel_contribution_forward_pass = self.channel_contributions_forward_pass(
                 channel_data=channel_data
             )
-            channel_contributions.append(channel_contribution_forward_pass)
+            channel_contributions.append(channel_contribution_forward_pass.eval())
         return DataArray(
             data=np.array(channel_contributions),
             dims=("delta", "chain", "draw", "date", "channel"),
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -9,8 +9,9 @@ import numpy.typing as npt
 import pandas as pd
 import pymc as pm
 import seaborn as sns
-from xarray import DataArray
 from pytensor.tensor.variable import TensorVariable
+from xarray import DataArray
+
 
 from pymc_marketing.mmm.base import MMM
 from pymc_marketing.mmm.preprocessing import MaxAbsScaleChannels, MaxAbsScaleTarget
@@ -557,8 +558,10 @@ class DelayedSaturatedMMM(
         array-like
             Transformed channel data.
         """
-        channel_contribution_forward_pass = super().channel_contributions_forward_pass_untransformed(
-            channel_data=channel_data
+        channel_contribution_forward_pass = (
+            super().channel_contributions_forward_pass_untransformed(
+                channel_data=channel_data
+            )
         )
         target_transformed_vectorized = np.vectorize(
             self.target_transformer.inverse_transform,
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -10,6 +10,7 @@ import pandas as pd
 import pymc as pm
 import seaborn as sns
 from xarray import DataArray
+from pytensor.tensor.variable import TensorVariable
 
 from pymc_marketing.mmm.base import MMM
 from pymc_marketing.mmm.preprocessing import MaxAbsScaleChannels, MaxAbsScaleTarget
@@ -344,9 +345,9 @@ class BaseDelayedSaturatedMMM(MMM):
             n_order=self.yearly_seasonality,
         )
 
-    def channel_contributions_forward_pass(
+    def channel_contributions_forward_pass_untransformed(
         self, channel_data: npt.NDArray[np.float_]
-    ) -> npt.NDArray[np.float_]:
+    ) -> TensorVariable:
         """Evaluate the channel contribution for a given channel data and a fitted model, ie. the forward pass.
         Parameters
         ----------
@@ -556,7 +557,7 @@ class DelayedSaturatedMMM(
         array-like
             Transformed channel data.
         """
-        channel_contribution_forward_pass = super().channel_contributions_forward_pass(
+        channel_contribution_forward_pass = super().channel_contributions_forward_pass_untransformed(
             channel_data=channel_data
         )
         target_transformed_vectorized = np.vectorize(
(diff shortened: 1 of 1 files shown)


pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -564,7 +564,7 @@ def channel_contributions_forward_pass(
             excluded=[1, 2],
             signature="(m, n) -> (m, n)",
         )
-        return target_transformed_vectorized(channel_contribution_forward_pass)
+        return target_transformed_vectorized(channel_contribution_forward_pass).eval()
```
ricardoV94, line 567: I think you need to put the eval inside, since `target_transformed_vectorized` is pure numpy code

```suggestion
        return target_transformed_vectorized(channel_contribution_forward_pass.eval())
```

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -383,7 +383,7 @@ def channel_contributions_forward_pass(
         channel_contribution_forward_pass = (
             beta_channel_posterior_expanded * logistic_saturation_posterior
         )
-        return channel_contribution_forward_pass.eval()
+        return channel_contribution_forward_pass
```
ricardoV94, line 386: The type signature has to be updated. However the subclass shouldn't change the meaning of the method (by returning a numpy value).

pymc_marketing/mmm/delayed_saturated_mmm.py
ricardoV94: @juanitorduz can we use PyTensor functions for the `inverse_transform`s? That would simplify this PR quite a lot. We could use `pytensor.graph.vectorize` instead of `numpy.vectorize`

pymc_marketing/mmm/delayed_saturated_mmm.py
ricardoV94: In the meantime, should we call the base-class method "x_untransformed", so that the sub-classes can call it without being ambiguous about the meaning?

cluhmann:
Note to self: we decided that the only difference (for now) between the parent class version and child class version of `channel_contributions_forward_pass` is the transformation. So naming the parent class method something like `channel_contribution_forward_pass_untransformed()` should be ok.

cluhmann:
Ok. So this is at the point where I can either modify the tests to reflect the new return value/method name or try to convert the target transformation method to be a native pytensor function.  The latter can obviously be a separate PR, but it seems like the tests would have to be rewritten again after converting the method to a pytensor function, so I figured it would make sense to ask for feedback.

We had a bit of discussion about the adoption of sklearn in https://github.com/pymc-labs/pymc-marketing/issues/154, but there was not definitive resolution.

ricardoV94:
@cluhmann completely up to you, whether you want to do the extra work or leave it for later

wd60622:
Personally, I am for sklearn especially with transformation of dataframe X in order to support new data.  The transformation that are currently used carry a state (max value or mean, std of data) which we would have to mimic in some form even if we migrate away.
The `fit`, `fit_transform`, `transform` API is become a common place but sklearn functionality for transformation on certain dataframe columns, preserving index, and other pandas support which has caused some of the recent bugs (`reset_index` etc).

I am working on a PR which shows how much sklearn can help reduce the transformation code we have to maintain while supporting the new data use-case.

Yes, for just `y` it seems a bit overkill but other locations, I think it is beyond helpful. 

Are there concerns of it as a dependency?

ricardoV94:
> Are there concerns of it as a dependency?

The issue is that we can't integrate this with other PyMC/PyTensor methods, e.g., the optimizer stuff that @cluhmann has been working on. It's a silly limitation to just do some scale/max transforms which we can do just fine in PyTensor.

wd60622:
> > Are there concerns of it as a dependency?
> 
> The issue is that we can't integrate sklearn stuff with other PyMC/PyTensor methods, e.g., the optimizer stuff that @cluhmann has been working on. It's a silly limitation since all we are using them for is to do some scale/max transforms which we can do just fine in PyTensor.

My concern is for new data. The machine learning world keep track of the mean, std of training / fit data and use those for transformations of the leave out set.

For instance, I would want to avoid cases like this: 
```python
with pm.Model() as model: 
    y = pm.MutableData("y", [1, 2, 3])
    y_normalized = y / y.max()

y_normalized.eval() # array([0.33333333, 0.66666667, 1.        ])

with model: 
    pm.set_data({"y": [2, 3, 4]})
    
y_normalized.eval() # array([0.5 , 0.75, 1.  ])
```

I few this as a data leakage

ricardoV94:
> I few this as a data leakage

That's not leakage, it's how the model was written. You use constants if you don't want them to change

wd60622:
Obviously that is the intended behavior of pytensor. However, I am talking within the context of a MMM. if `y` was `cost` then the max cost that is used to determine the model parameters is based on the data fit, not the data in a holdout set. 

Making the transformations on constants like you said. I just claim using sklearn will be most managable way of doing that before putting transformed data into `MutableData` containers

ricardoV94:
> Making the transformations on constants like you said. I just claim using sklearn will be most managable way of doing that before putting transformed data into `MutableData` containers

The issue is when doing optimization we want to have PyTensor expressions that depend on those transformations (so that we can auto-diff through them). At that point the sklearn becomes an issue more than it helps, because it is not composable with PyTensor expressions. Basically we want to write those sklearn transforms in PyTensor, so they cache the max/mean/std whatever of the original data as a constant and provide a back/forward method that can be called with new PyTensor objects.

cluhmann:
To me the question is how pervasive we expect the sklearn components to be through the pymc-marketing codebase. Right now, it would be easy enough (famous last words) to replace the sklearn transformation (with all the robustness that @wd60622 notes) because the sklearn ingredients are not used all that often. But if we foresee lots of sklearn functionality being adopted within the project, then rewriting everything in pytensor becomes less appealing. But I don't have a good sense of what the future looks like in regard to sklearn usage.

ricardoV94:
@cluhmann I agree, and my hunch is that eventually someone will want these operations to be part of the model itself and not just data pre-processing (perhaps downstream of MutableData). As soon as that happens you can no longer use sklearn or numpy routines.

wd60622:
I am not attached to sklearn. However, I am for exchangeability of transformation methods. Having them all as mixins makes it almost very hard to change without having to rewrite much of the code, leaving room for errors, etc.

I raised #407 to get away from mixins currently used for transformation. My draft PR has `x_transformer` which mirrors the `target_transformer` attribute. Doing so, provides the forward pass on new data X via a single `transform` call which was incorrectly implemented (#405) but is open (with some restriction) to changing the transformation happening.

I just praise sklearn for the fit / fit_transform and transform separation which is important especially with parameters being estimated with artifacts of the "training" data (mean, std, max, etc). And since different transformation are happening to different subsets of DataFrame instances, the [`ColumnTransformer`](https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html) is made just for that. 
I am all for having transformation on the data in pytensor if these details aren't missed.

To illustrate the difficulty of the current code, I tried to use a `MaxAbsScaler` for control columns as well. But to do so requires overriding mixins or completely redefining the class and copying much of the code.

ricardoV94:
I agree that the current models are very hard to customize, I just don't think sklearn plays any role in that.

juanitorduz:
I do not wave a strong opinion. Any volunteer(s) to draft a PR removing these transformers? We can all provide feedback 🙌

wd60622:
I have [this PR](https://github.com/pymc-labs/pymc-marketing/pull/405/files) up which removes the mixins for preprocessing and uses a single transformer (ColumnTransformer) to handle the transformations of new data. 

It addresses a lot but the original purpose was to support new data which need to be transformed only in the `_data_setter`

juanitorduz:
🙌 I'll take a look at it in detail once I'm back (🏖️🌴)

wd60622:
Also a stale one. Can we close and reframe as some issues ?

cluhmann:
What do you mean? @juanitorduz said he would take a look (in 2023 :joy: ).  Yeah, I think we can close this. If it's still an issue, it will crop up (again) as we work on https://github.com/pymc-labs/pymc-marketing/pull/358

juanitorduz:
OMG! 2023 was a thought year 😄🙈

PR #393, opened 2023-10-12
Add method `thin_fit_result` to CLV models

### LINKED ISSUE #374: `clv.utils.customer_lifetime_value` generates one row for every chain/draw/customer_id permutation instead of 1 row per customer_id
As the title says, `clv.utils.customer_lifetime_value` returns an unmanageably large dataset for me. For a sample dataset with 25k users, our final dataset results in 100 million rows.

Below are the key steps I ran, where `clv_df_all` is my full dataset and `clv_df_freq1` filters that dataset down to users with frequency>0:
```python
import pandas as pd
import pymc_marketing.clv as clv
bgm = clv.BetaGeoModel(data = clv_df_freq1)
bgm.build_model()
bgm.fit()
user_ltvs = clv.utils.customer_lifetime_value(
    transaction_model = bgm,
    customer_id = clv_df_all['customer_id'], 
    frequency = clv_df_all['frequency'], 
    recency = clv_df_all['recency'], 
    T = clv_df_all['T'], 
    monetary_value = clv_df_all['monetary_value'], 
    time = 12, 
    discount_rate = 0.01, 
    freq = 'D'
)
``` 
Is this expected behavior (i.e. am I missing a post-processing step)? Or is this unintended behavior?

ricardoV94:
CLV summary statistic models can be quite memory intensive and fail with the standard 4k draws. Because these methods are all built assuming there is a `self.idata` lying around (called in multiple places), the two hacky solutions would be to pass a `thin` argument to every method or to modify idata inplace.

This PR instead adds a method that returns a copy of the CLV model with a thinned dataset. A user can then use the methods in this thinned model to obtain summary stats. 

Closes #374 

This PR also cleans up several internals in the CLVModel base-class.

# TODO
- [x] Add test using `customer_lifetime_value`

----
:books: Documentation preview :books:: https://pymc-marketing--393.org.readthedocs.build/en/393/

### DIFF
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Dict, Optional, Tuple, cast
 
 import arviz as az
+import pandas as pd
 import pymc as pm
 from pymc import Model, str_for_dist
 from pymc.backends import NDArray
@@ -20,10 +21,13 @@ class CLVModel(ModelBuilder):
 
     def __init__(
         self,
+        data: Optional[pd.DataFrame] = None,
+        *,
         model_config: Optional[Dict] = None,
         sampler_config: Optional[Dict] = None,
     ):
         super().__init__(model_config, sampler_config)
+        self.data = data
 
     def __repr__(self):
         return f"{self._model_type}\n{self.model.str_repr()}"
@@ -138,24 +142,52 @@ class CLVModel(ModelBuilder):
         """
         filepath = Path(str(fname))
         idata = az.from_netcdf(filepath)
-        dataset = idata.fit_data.to_dataframe()
+        return cls._build_with_idata(idata)
 
+    @classmethod
+    def _build_with_idata(cls, idata: az.InferenceData):
+        dataset = idata.fit_data.to_dataframe()
         model = cls(
             dataset,
             model_config=json.loads(idata.attrs["model_config"]),  # type: ignore
             sampler_config=json.loads(idata.attrs["sampler_config"]),
         )
         model.idata = idata
-
         model.build_model()  # type: ignore
-
         if model.id != idata.attrs["id"]:
-            raise ValueError(
-                f"The file '{fname}' does not contain an inference data of the same model or configuration as '{cls._model_type}'"
-            )
-        # All previously used data is in idata.
-
+            raise ValueError(f"Inference data not compatible with {cls._model_type}")
         return model
+
+    def thin_fit_result(self, keep_every: int):
+        """Return a copy of the model with a thinned fit result.
+
+        This is useful when computing summary statistics that may require too much memory per posterior draw.
+
+        Examples
+        --------
+
+        .. code-block:: python
+
+            fitted_gg = ...
+            fitted bg = ...
+
+            fitted_gg_thinned = fitted_gg.thin_fit_result(keep_every=10)
+            fitted_bg_thinned = fitted_bg.thin_fit_result(keep_every=10)
+
+            clv_thinned = fitted_gg_thinned.expected_customer_lifetime_value(
+                transaction_model=fitted_bg_thinned,
+                customer_id=t.index,
+                frequency=t["frequency"],
+                recency=t["recency"],
+                T=t["T"],
+                mean_transaction_value=t["monetary_value"],
+            )
+
+        """
+        self.fit_result  # Raise Error if fit didn't happen yet
+        assert self.idata is not None
+        new_idata = self.idata.isel(draw=slice(None, None, keep_every)).copy()
+        return type(self)._build_with_idata(new_idata)
 
     @staticmethod
     def _check_prior_ndim(prior, ndim: int = 0):
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -32,6 +32,16 @@ class CLVModel(ModelBuilder):
     def __repr__(self):
         return f"{self._model_type}\n{self.model.str_repr()}"
 
+    def _add_fit_data_group(self, data: pd.DataFrame) -> None:
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore",
+                category=UserWarning,
+                message="The group fit_data is not defined in the InferenceData scheme",
+            )
+            assert self.idata is not None
+            self.idata.add_groups(fit_data=data.to_xarray())
+
     def fit(  # type: ignore
         self,
         fit_method: str = "mcmc",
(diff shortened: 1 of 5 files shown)


pymc_marketing/clv/models/basic.py
```diff
@@ -22,14 +21,27 @@ class CLVModel(ModelBuilder):
...
         return f"{self._model_type}\n{self.model.str_repr()}"
 
+    def _add_fit_data_group(self, data: pd.DataFrame) -> None:
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore",
+                category=UserWarning,
+                message="The group fit_data is not defined in the InferenceData scheme",
+            )
+            assert self.idata is not None
```
ColtAllen, line 42: This `assert` statement will be skipped if python is ran in optimized mode. An IF statement raising a `RunTimeError` instead will prevent this.

pymc_marketing/clv/models/basic.py
```diff
@@ -173,25 +146,28 @@ def load(cls, fname: str):
...
-                f"The file '{fname}' does not contain an inference data of the same model or configuration as '{cls._model_type}'"
-            )
-        # All previously used data is in idata.
-
+            raise ValueError(f"Inference data not compatible with {cls._model_type}")
         return model
 
+    def thin_fit_result(self, keep_every: int):
+        self.fit_result  # Raise Error if fit didn't happen yet
+        assert self.idata is not None
```
ColtAllen, line 167: Same as above, replace this `assert` with an IF statement raising a `RunTimeError`.

ColtAllen (changes requested):

pymc_marketing/clv/models/basic.py
ricardoV94: This is just for mypy

pymc_marketing/clv/models/basic.py
ricardoV94: Same

pymc_marketing/clv/models/basic.py
ricardoV94: Also did you ever see anyone running python in optimized mode :) ?

pymc_marketing/clv/models/basic.py
```diff
@@ -228,39 +200,6 @@ def default_sampler_config(self) -> Dict:
...
-        return prior_predictive_samples
-
-    @property
-    def prior_predictive(self) -> az.InferenceData:
-        if self.idata is None or "prior_predictive" not in self.idata:
-            raise RuntimeError(
-                "No prior predictive samples available, call sample_prior_predictive() first"
-            )
-        return self.idata["prior_predictive"]
-
```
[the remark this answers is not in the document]
ricardoV94: Isn't it part of the base ModelBuilder class now?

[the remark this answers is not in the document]
ColtAllen: Ah, you're right:

https://github.com/pymc-devs/pymc-experimental/blob/main/pymc_experimental/model_builder.py

pymc_marketing/clv/models/basic.py
ColtAllen: `mypy` requires assert statements some of the typing (non)-issues it raises?

I could see a junior developer hearing that python has an optimized mode, getting excited & running a script without realizing literally all it does is skip `assert` lines, then getting stuck with an issue that could be quite hard to debug. Just a little extra guarantee for reliability.

ColtAllen:
The `assert` stuff is probably just splitting hairs, but maybe raise a `UserWarning` in `fit()` to suggest using this function when dealing with huge datasets and/or draw sizes? Say, draw size > 10k and customer_ids > 50k. That way users know to try this when encountering memory crashes.

ricardoV94:
> The `assert` stuff is probably just splitting hairs, but maybe raise a `UserWarning` in `fit()` to suggest using this function when dealing with huge datasets and/or draw sizes? Say, draw size > 10k and customer_ids > 50k. That way users know to try this when encountering memory crashes.

That could be useful. I wonder if we should do it in fit though? Or in the heavy functions like clv? Easier in fit but potentially more annoying as well.

ColtAllen:
> > The `assert` stuff is probably just splitting hairs, but maybe raise a `UserWarning` in `fit()` to suggest using this function when dealing with huge datasets and/or draw sizes? Say, draw size > 10k and customer_ids > 50k. That way users know to try this when encountering memory crashes.
> 
> That could be useful. I wonder if we should do it in fit though? Or in the heavy functions like clv? Easier in fit but potentially more annoying as well.

Where's it most likely to crash? I only suggested `fit` because draw size can be obtained from `sampler_config`.

ricardoV94:
> > > The `assert` stuff is probably just splitting hairs, but maybe raise a `UserWarning` in `fit()` to suggest using this function when dealing with huge datasets and/or draw sizes? Say, draw size > 10k and customer_ids > 50k. That way users know to try this when encountering memory crashes.
> > 
> > 
> > That could be useful. I wonder if we should do it in fit though? Or in the heavy functions like clv? Easier in fit but potentially more annoying as well.
> 
> Where's it most likely to crash? I only suggested `fit` because draw size can be obtained from `sampler_config`.

We always know the size of the samples from the idata

pymc_marketing/clv/models/basic.py
ricardoV94: Yes, when mypy doesn't know about type information, you can convince him that all is right by adding an assert there. In here the issue in self.model is always None at first, and mypy doesn't know that we only call this method once it is no longer None. 

It's really just for mypy not for users sake that we have the assert

ricardoV94:
> Where's it most likely to crash? I only suggested fit because draw size can be obtained from sampler_config.

It shouldn't be in fit, because there are not so many parameters in the model. Any problems will appear in functions that compute summary statistics per observations (most predictive methods, including the `customer_lifetime_value` one). It's also not model specific. I would keep it simple for now, and we revisit the warning idea if people keep being puzzled / don't find out naturally about the `thin_fit_result` method.

ColtAllen:
> > Where's it most likely to crash? I only suggested fit because draw size can be obtained from sampler_config.
> 
> It shouldn't be in fit, because there are not so many parameters in the model. Any problems will appear in functions that compute summary statistics per observations (most predictive methods, including the `customer_lifetime_value` one). It's also not model specific. I would keep it simple for now, and we revisit the warning idea if people keep being puzzled / don't find out naturally about the `thin_fit_result` method.

Sounds good. Maybe add a comment about using `thin_fit_result` in the docstrings so it shows up in the docs, and/or the CLV Quickstart? The latter can be its own PR as there are plenty of other ways the Quickstart can be improved (ie an overview of the CLV modeling domains, use of the `clv_summary` function for data prep, etc.)

ColtAllen (commented):
Aside from the documentation suggestions, the only other remarks I have concern the reductions in testing coverage.

Are `fit_result` and `expected_customer_lifetime_value`fully covered when the `--runslow` tests are ran?

Also, are there plans to expand the placeholder methods  `output_var`, `_generate_and_preprocess_model_data`, and `_data_setter`? I'm not sure what the need is for those methods at the moment.

If you're chasing 100% coverage in `beta_geo.py`, you can adapt the following code:
https://github.com/pymc-labs/pymc-marketing/blob/main/tests/clv/models/test_pareto_nbd.py#L302

ricardoV94:
I think the coverage is just because the tests are not passing yet. 

Regarding the weird methods, they are abstract methods that the base ModelBuilder class demands but are not at all relevant for CLV

ColtAllen (approved):
Did CodeCov not update for the new `customer_lifetime_value` test? I also made a note in https://github.com/pymc-labs/pymc-marketing/issues/448 that it is not resolved by this PR, nor does it have to be.

As for the placeholder methods, I have some ideas on how to use them, but it would be CLV model-specific and probably outside the scope of this PR.

I think this is good to merge. Thanks for cleaning up the CLV base class!

tomthepeach:
Awesome thanks guys!

PR #395, opened 2023-10-13
Correcting typo and adding feedback.

cetagostini:
Hey team!

I received initial feedback from a friend regarding the optimizer function they used for their custom MMM. 

They pointed out a small typo and some bugs in the default parameters when the optimizer bounds are set to `None`. Although these were minor issues, I have made the corrections that were needed.

----
:books: Documentation preview :books:: https://pymc-marketing--395.org.readthedocs.build/en/395/

### DIFF
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -52,14 +52,14 @@ def calculate_expected_contribution(
     total_expected_contribution = 0.0
     contributions = {}
 
-    for channel, channe_budget in budget.items():
+    for channel, channel_budget in budget.items():
         if method == "michaelis-menten":
             L, k = parameters[channel]
-            contributions[channel] = michaelis_menten(channe_budget, L, k)
+            contributions[channel] = michaelis_menten(channel_budget, L, k)
 
         elif method == "sigmoid":
             alpha, lam = parameters[channel]
-            contributions[channel] = extense_sigmoid(channe_budget, alpha, lam)
+            contributions[channel] = extense_sigmoid(channel_budget, alpha, lam)
 
         else:
             raise ValueError("`method` must be either 'michaelis-menten' or 'sigmoid'.")
@@ -164,7 +164,15 @@ def optimize_budget_distribution(
 
     if budget_ranges is None:
         budget_ranges = {
-            channel: (0, min(total_budget, parameters[channel][0]))
+            channel: (
+                0,
+                min(
+                    total_budget,
+                    9 * parameters[channel][1]
+                    if method == "michaelis-menten"
+                    else -np.log(0.1 / 1.9) / parameters[channel][1],
+                ),
+            )
             for channel in channels
         }
 
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -166,12 +166,7 @@ def optimize_budget_distribution(
         budget_ranges = {
             channel: (
                 0,
-                min(
-                    total_budget,
-                    9 * parameters[channel][1]
-                    if method == "michaelis-menten"
-                    else -np.log(0.1 / 1.9) / parameters[channel][1],
-                ),
+                total_budget,
             )
             for channel in channels
         }
--- a/tests/mmm/test_budget_optimizer.py
+++ b/tests/mmm/test_budget_optimizer.py
@@ -72,7 +72,7 @@ def test_objective_distribution_invalid_method():
 
 # Testing optimize_budget_distribution with valid inputs
 @pytest.mark.parametrize(
-    "method,total_budget,budget_ranges,parameters,channels,expected_sum",
+    "method,total_budget,budget_ranges,parameters,channels,expected_result",
     [
         (
             "michaelis-menten",
@@ -80,7 +80,7 @@ def test_objective_distribution_invalid_method():
             {"channel1": (0, 50), "channel2": (0, 50)},
             {"channel1": (10, 5), "channel2": (20, 10)},
             ["channel1", "channel2"],
-            100,
+            {'channel1': 50.0, 'channel2': 50.0},
         ),
         (
             "sigmoid",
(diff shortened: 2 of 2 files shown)


cetagostini:
PS: Just to let you know, I would like to test the functionality before merging. I will notify you as soon as I have done so.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -164,7 +164,15 @@ def optimize_budget_distribution(
 
     if budget_ranges is None:
         budget_ranges = {
-            channel: (0, min(total_budget, parameters[channel][0]))
+            channel: (
+                0,
+                min(
+                    total_budget,
+                    9 * parameters[channel][1]
```
wd60622, line 171: where does the `9` come from?

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -164,7 +164,15 @@ def optimize_budget_distribution(
...
+                    if method == "michaelis-menten"
+                    else -np.log(0.1 / 1.9) / parameters[channel][1],
```
wd60622, line 173: similar here. 

If they have some meaning, it might be best to extract and put into a variable with name 😄

wd60622: else here implies `sigmoid`?

pymc_marketing/mmm/budget_optimizer.py
cetagostini: Hey Will, the initial idea was to use the maximum budget or the point where saturation reaches 90% as the default optimization limit. However, I was passing this saturation value to 90% in Y, not X.

Why 9? If you consider Michaelis mente functions is `Y = alpha*X/lam+X` then if you want to find Y at 90% (0.9*Y) after simplifying the function you get X = 9 * lam. I left a small code snippet you can use to reproduce the results, and test if you want.

``` python
# Define parameters for Vmax and Km
alpha = 10
lam = 5

# Generate substrate concentrations (S) from 0 to 100
X = np.linspace(0, 100, 500)
y = michaelis_menten(X, alpha, lam)

# Calculate the substrate concentration at which the reaction rate is 90% of Vmax (Y = 9)
X_90_percent_example = 9 * lam

# Calculate the reaction rate at this substrate concentration
Y_90_percent_example = michaelis_menten(X_90_percent_example, alpha, lam)

# Plotting
plt.figure(figsize=(10, 6))
plt.plot(X, y, label='Michaelis-Menten Curve', linewidth=2)
plt.scatter([X_90_percent_example], [Y_90_percent_example], color='red', zorder=5, label=f'90% of alpha (X={X_90_percent_example}, y={Y_90_percent_example})')
plt.axhline(y=alpha, color='g', linestyle='--', label=f'alpha = {alpha}')
plt.axhline(y=Y_90_percent_example, color='r', linestyle='--', label=f'90% of alpha = {Y_90_percent_example}')
plt.xlabel('Cost Units (X)')
plt.ylabel('Contribution (y)')
plt.title('Michaelis-Menten')
plt.legend()
plt.grid(True)
plt.show()
```
This image uses the typical labels of Michaelis-Menten on Kinetics:
image

pymc_marketing/mmm/budget_optimizer.py
cetagostini: Correct, implies sigmoid.

Regarding, why `-np.log(0.1 / 1.9)` because based on the function `(alpha - alpha * np.exp(-lam * x)) / (1 + np.exp(-lam * x))` the way to get the point on X when Y its at 90% is with this implementation. Example code below:
``` python
# Generate random parameters for alpha and lambda
np.random.seed(42)  # for reproducibility
alpha_random = np.random.uniform(5, 20)  # Random alpha between 5 and 20
lam_random = np.random.uniform(0.1, 1)   # Random lambda between 0.1 and 1

# Generate x values and corresponding function values
x_values = np.linspace(0, 50, 500)
y_values = custom_function(x_values, alpha_random, lam_random)

# Calculate the x value at which the function value is 90% of alpha
x_90_percent = -np.log(0.1 / 1.9) / lam_random

# Calculate the function value at this x value
y_90_percent = sigmoid(x_90_percent, alpha_random, lam_random)

# Plotting
plt.figure(figsize=(10, 6))
plt.plot(x_values, y_values, label='Custom Function Curve', linewidth=2)
plt.scatter([x_90_percent], [y_90_percent], color='red', zorder=5, label=f'90% of alpha (x={x_90_percent:.2f}, y={y_90_percent:.2f})')
plt.axhline(y=alpha_random, color='g', linestyle='--', label=f'alpha = {alpha_random:.2f}')
plt.axhline(y=y_90_percent, color='r', linestyle='--', label=f'90% of alpha = {y_90_percent:.2f}')
plt.xlabel('x')
plt.ylabel('f(x)')
plt.title('Custom Function')
plt.legend()
plt.grid(True)
plt.show()
```
image

cetagostini:
In the afternoon I made the adjustment quickly, but after thinking twice, I will simplify the function and set the default maximum budget for each channel as the total budget, if the user wants to use the 90% saturation point at X, in the [example notebook](https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_budget_allocation_example.html) an introduction was given on how to do it.

pymc_marketing/mmm/budget_optimizer.py
wd60622: Nice, that's awesome. Thanks for the explanation. 
I'm for the allowing the user to set the budgets like in the most recent change. 
Maybe it is good to add reference in the docstring? 

Another idea (if there is some closed form of these functions) would be have some helper functions: 
- `get_michaelis_menten_quantile_budget(q=.9, lam=lam) -> 9 * lam`
- `get_sigmoid_quantile_budget(q=0.9, lam=lam) -> -np.log(0.1 / 1.9) / lam_random`

Then, defining them would might be clear to the user

ricardoV94:
> I will simplify the function and set the default maximum budget for each channel as the total budget

@cetagostini have you done this, or is the PR currently in draft format (if so, can you press the button "convert to draft"?)

Issue #346, opened 2023-08-05
Data validation errors in MMM due to pandas index


lewi0332:
Hi All, 

Just found a very small issue trying to run a sample on the latest release `0.2.0`

If I pass a pandas df that has been truncated from the original length, the `ValidateDateColumn` class throws an error telling me I have duplicate date columns.  

My test script starts with a larger data sample. I have filtered the data to within a date range. This leaves the original pandas index value.  ie... original df has 200 rows, model data now has 100 rows, but the index starts at `99`

I printed a sample in that moment of validation and see that the series has the original length of the DataFrame and an `NAT` for all empty index values.

Issue was easily solved by resetting the index of the Dataframe before starting the .fit() process. This is an outlier example as I would assume most would be loading only the data they intend to sample on.

michaelraczycki:
could you either update here a way to reproduce it, or the error you've been shown? Would it be possible for you to do the data cleaning before passing the dataset into the fit function?

geirrlod:
Hi, 

Just wanted to say that I ran into the exact same issue. If the index starts with anything else than 1 it told me that I have dublicate date columns.  But very easy fix with reseting the index (or using .loc in the first place to avoid the issue)

PhilClarkPhD:
Writing to say I had the same issue. Fixed by resetting the index.

ricardoV94:
Can someone provide a reproducible example?

MarkusSagen:
Here is a reproduced example using the example MMM notebook:

```python
import pandas as pd

# Define the data
data = {
    'date_week': ['2018-04-02', '2018-04-09', '2018-04-16', '2018-04-23', '2018-04-30'],
    'y': [3.984662, 3.762872, 4.466967, 3.864219, 4.441625],
    'x1': [0.318580, 0.112388, 0.292400, 0.071399, 0.386745],
    'x2': [0.0, 0.0, 0.812466, 0.0, 0.2164220],
}
df = pd.DataFrame(data)

df['date_week'] = pd.to_datetime(df['date_week'])

# Add 2 to the index to artificially cause the issue
df.index += 2

X = data.drop("y", axis=1)
y = data["y"]

mmm = DelayedSaturatedMMM(
    date_column="date_week",
    channel_columns=['x1', 'x2'],
    adstock_max_lag=8,
    yearly_seasonality=2,
)

mmm.fit(X=X, y=y, target_accept=0.95, chains=4, random_seed=rng)
```

Reproduces the error
[...]

PR #402, opened 2023-10-21
MMM: improve validation check for dataframes with correct indices

MarkusSagen:
Fixes the issue discussed in #346 Data validation errors in MMM due to pandas index

Since seasonality seem to assume that the index is in order, I've opted to return the error to the user and be explicit about what needs to be fixed

----
:books: Documentation preview :books:: https://pymc-marketing--402.org.readthedocs.build/en/402/

### DIFF
--- a/pymc_marketing/mmm/validating.py
+++ b/pymc_marketing/mmm/validating.py
@@ -12,14 +12,6 @@ __all__ = [
 ]
 
 
-def _has_valid_indices(df: pd.DataFrame) -> bool:
-    return (
-        isinstance(df.index, pd.RangeIndex)
-        and df.index.start == 0
-        and df.index.stop == len(df)
-    )
-
-
 def validation_method_y(method: Callable) -> Callable:
     if not hasattr(method, "_tags"):
         method._tags = {}  # type: ignore
@@ -49,7 +41,11 @@ class ValidateDateColumn:
         if self.date_column not in data.columns:
             raise ValueError(f"date_col {self.date_column} not in data")
 
-        if not _has_valid_indices(data):
+        if (
+            not isinstance(data.index, pd.RangeIndex)
+            or data.index.start != 0
+            or data.index.stop != len(data)
+        ):
             raise ValueError(
                 "X or y has incorrect indices. Try to reset with `data.reset_index(inplace=True)`"
             )
--- a/pymc_marketing/mmm/validating.py
+++ b/pymc_marketing/mmm/validating.py
@@ -12,6 +12,14 @@ __all__ = [
 ]
 
 
+def _has_valid_indices(df: pd.DataFrame) -> bool:
+    return (
+        isinstance(df.index, pd.RangeIndex)
+        and df.index.start == 0
+        and df.index.stop == len(df)
+    )
+
+
 def validation_method_y(method: Callable) -> Callable:
     if not hasattr(method, "_tags"):
         method._tags = {}  # type: ignore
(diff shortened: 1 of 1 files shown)


wd60622:
I think it would be great if the index could be kept. Does keeping the original index in the frames used in the various pd.concat also provide a solution? Here is the method 
https://github.com/pymc-labs/pymc-marketing/blob/29a69f67b9ae10171d1a79c027c7282abfd1c2a7/pymc_marketing/mmm/delayed_saturated_mmm.py#L111

For instance 

```python 
# data with non RangeIndex
dates = pd.date_range("2023-01-01", periods=4, freq="D")
frame = pd.DataFrame({"date_col": dates}, index=["indices", "can", "cause", "issues"])

# _generate_and_preprocess_model_data logic
frame_in_method = pd.concat([
    frame["date_col"], 
    pd.Series([1, 2, 3, 4]).rename("different_index_col")
], axis=1)
frame_in_method["date_col"].is_unique # False 
frame_in_method
```

```text
          date_col  different_index_col
indices 2023-01-01                  NaN
can     2023-01-02                  NaN
cause   2023-01-03                  NaN
issues  2023-01-04                  NaN
0              NaT                  1.0
1              NaT                  2.0
2              NaT                  3.0
3              NaT                  4.0

**Comment by MarkusSagen** [comment]
I think it would. The thing that would be a bit confusing for a user I imagine is when some part of their row values or indices are changed and parts of the MMM class appends performs operation on a dataframe under the assumption that is valid.

If we want it to be more flexible, I can try that. In most cases, I would think these things to occur unintentionally on the users part and instead of guessing the user intent, might be better to fix it like this and return for the user how to resolve the issue? And if there is an explicit use-case for it, maybe solve it then?

I'm flexible and depends on what you think would be the best approach for this 👍

**Comment by wd60622** [comment]
> I think it would. The thing that would be a bit confusing for a user I imagine is when some part of their row values or indices are changed and parts of the MMM class appends performs operation on a dataframe under the assumption that is valid.
> 
> If we want it to be more flexible, I can try that. In most cases, I would think these things to occur unintentionally on the users part and instead of guessing the user intent, might be better to fix it like this and return for the user how to resolve the issue? And if there is an explicit use-case for it, maybe solve it then?
> 
> I'm flexible and depends on what you think would be the best approach for this 👍

I personally think it that it would be the best behavior and is a convenience to the user. Especially since index of `X` should align with `y`

If it can be solved upstream without user intervention, that'd be great. Thanks for your work!

**Review by ricardoV94 — COMMENTED** [review]
Can you add a short test that triggers this error?

**Comment by ricardoV94** [comment]
Ah you have decided to change the behavior in the discussion? In that case this PR isn't ready right?

We can either fix it here, or merge the error message for the time being.

**Comment by MarkusSagen** [comment]
@ricardoV94 I'll test if the proposed change allows the model to train later today. 
Otherwise, will leave it as is

File: pymc_marketing/mmm/validating.py [hunk 1]
```diff
@@ -40,6 +48,12 @@ class ValidateDateColumn:
     def validate_date_col(self, data: pd.DataFrame) -> None:
         if self.date_column not in data.columns:
             raise ValueError(f"date_col {self.date_column} not in data")
+
+        if not _has_valid_indices(data):
+            raise ValueError(
+                "X or y has incorrect indices. Try to reset with `data.reset_index(inplace=True)`"
+            )
```
ricardoV94, line 55: nitpick: this is small enough that i would just include the code here instead of in the hidden function.

MarkusSagen: Fixed

ricardoV94 (commented):
Can you add a test?

juanitorduz:
+1 on the test :)

PR #405, opened 2023-10-21
add datasetter for date and fourier

wd60622:
Addresses #400 

----
:books: Documentation preview :books:: https://pymc-marketing--405.org.readthedocs.build/en/405/


wd60622:
WIP. Will add some tests for the different scenarios that were described

wd60622 (commented):
Just some questions about the intended behavior

Also, there is some heavy overlap between `_generate_and_preprocess_model_data` and this method now and I think much could be combined

pymc_marketing/mmm/preprocessing.py
```diff
@@ -32,7 +32,7 @@ class MaxAbsScaleTarget:
 
     @preprocessing_method_y
     def max_abs_scale_target_data(self, data: pd.Series) -> pd.Series:
-        target_vector = data.reshape(-1, 1)
```
wd60622, line 35: Based on the deprecation #406 

However, is this the correct signature of this function. Could this data also be a `numpy.array`?

wd60622:
Also think that this is still missing the `preprocess` call

wd60622:
Still working through the tests as removing the mixins requires a lot of changes 😢 However, the new interface should be somewhat stable.

I can keep the mixins and just use the transformers that are within to support new data if that is more desirable. I'd appreciate any feedback and requests

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -94,44 +117,45 @@ def _generate_and_preprocess_model_data(  # type: ignore
...
-        X : Union[pd.DataFrame, pd.Series], shape (n_obs, n_features)
+        X : pd.DataFrame, shape (n_obs, n_features)
         y : Union[pd.Series, np.ndarray], shape (n_obs,)
         """
-        date_data = X[self.date_column]
-        channel_data = X[self.channel_columns]
-        coords: Dict[str, Any] = {
+        if self.validate_data:
+            self.validate("X", X)
+            self.validate("y", y)
```
wd60622, line 125: New order because the `x_transformer.transform(X)` will not have `date_col`

wd60622:
Will continue this off of the branch that @cetagostini is working on

PR #397, opened 2023-10-17
Custom priors & Dist to MMM DelayedSaturated

cetagostini:
Hey guys, I'm applying the same configuration of the models on CLV to the MMM models, the idea is to be able to pass priors and custom distributions based on the user needs.

Everything was working properly on the mmm example notebook, except for the `fit()` method 😅 Here is the traceback, looks like is not directly related to the changes:

```
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-26-e8614295d821> in <cell line: 1>()
----> 1 mmm.fit(X=X, y=y, target_accept=0.95, chains=4, random_seed=rng)

10 frames
/content/pymc-marketing/pymc_marketing/model_builder.py in fit(self, X, y, progressbar, predictor_names, random_seed, **kwargs)
    475             with self.model:
    476                 sampler_args = {**self.sampler_config, **kwargs}
--> 477                 self.idata = pm.sample(**sampler_args)
    478 
    479         X_df = pd.DataFrame(X, columns=X.columns)

/usr/local/lib/python3.10/dist-packages/pymc/sampling/mcmc.py in sample(draws, tune, chains, cores, random_seed, progressbar, step, nuts_sampler, initvals, init, jitter_max_retries, n_init, trace, discard_tuned_samples, compute_convergence_checks, keep_warning_stat, return_inferencedata, idata_kwargs, nuts_sampler_kwargs, callback, mp_ctx, model, **kwargs)
    649 
    650     initial_points = None
--> 651     step = assign_step_methods(model, step, methods=pm.STEP_METHODS, step_kwargs=kwargs)
    652 
    653     if nuts_sampler != "pymc":

/usr/local/lib/python3.10/dist-packages/pymc/sampling/mcmc.py in assign_step_methods(model, step, methods, step_kwargs)
    209     methods_list: List[Type[BlockedStep]] = list(methods or pm.STEP_METHODS)
    210     selected_steps: Dict[Type[BlockedStep], List] = {}
--> 211     model_logp = model.logp()
    212 
    213     for var in model.value_vars:

/usr/local/lib/python3.10/dist-packages/pymc/model/core.py in logp(self, vars, jacobian, sum)
    738             return logp_factors
    739 
--> 740         logp_scalar = pt.sum([pt.sum(factor) for factor in logp_factors])
    741         logp_scalar_name = "__logp" if jacobian else "__logp_nojac"
    742         if self.name:

[...]

PR #408, opened 2023-10-23
Refactor mmm allow user defined priors

MarkusSagen:
This PR intends to add custom priors to the DelayedSaturation MMM model

There is an issue however and would love for some feedback and help on how to fix it.
The DelayedSaturation MMM uses priors with dims, but the CLV model I've copied from does not. I'm currious how to add that for the priors to make the code and tests pass?

----
:books: Documentation preview :books:: https://pymc-marketing--408.org.readthedocs.build/en/408/

### DIFF
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -57,6 +57,49 @@ class BaseDelayedSaturatedMMM(MMM):
         yearly_seasonality : Optional[int], optional
             Number of Fourier modes to model yearly seasonality, by default None.
 
+        Examples
+        --------
+            DelayedSaturatedMMM
+
+            .. code-block:: python
+
+                import pymc as pm
+                from pymc_marketing.mmm import DelayedSaturatedMMM
+
+                data_url = "https://raw.githubusercontent.com/pymc-labs/pymc-marketing/main/datasets/mmm_example.csv"
+                data = pd.read_csv(data_url, parse_dates=['date_week'])
+
+                model = DelayedSaturatedMMM(
+                    date_column="date_week",
+                    channel_columns=["x1", "x2"],
+                    control_columns=[
+                        "event_1",
+                        "event_2",
+                        "t",
+                    ],
+                    adstock_max_lag=8,
+                    yearly_seasonality=2,
+                    model_config={
+                        # priors
+                        "intercept": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 2}},
+                        "beta_channel": {"dist": "HalfNormal", "kwargs": {"sigma": 2}, "dims": ("channel",)},
+                        "alpha": {"dist": "Beta", "kwargs": {"alpha": 1, "beta": 3}, "dims": ("channel",)},
+                        "lam": {"dist": "Gamma", "kwargs": {"alpha": 3, "beta": 1}, "dims": ("channel",)},
+                        "sigma": {"dist": "HalfNormal", "kwargs": {"sigma": 2}},
+                        "gamma_control": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 2}, "dims": ("control",)},
+                        "gamma_fourier": {"dist": "Laplace", "kwargs": {"mu": 0, "b": 1}, "dims": "fourier_mode"},
+                        # params
+                        "mu": {"dims": ("date",)},
+                        "likelihood": {"dims": ("date",)},
+                    },
+                )
+
+                X = data.drop('y',axis=1)
+                y = data['y']
+
+                model.fit(X,y)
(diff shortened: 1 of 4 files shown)


twiecki:
This is neat.

I wonder, rather than specifying dists as names so that we can serialize them, couldn't we just use `pm.Distribution.dist(**kwargs)` and when we serialize, get `str(Distribution)` and `kwargs`, and then when unserializing we look up the right dist? Not sure that's possible @ricardoV94. I guess it's more of a PyMC issue to make `pm.Distribution.dist` serializable.

Anyway, I don't think we need to chase this one down all the way and focus on this implementation instead.

ricardoV94:
@MarkusSagen is this a duplicate of #397?

MarkusSagen:
@ricardoV94 It seems like it, yes! 
It is very similar

ricardoV94:
Can you synchronize with @cetagostini to avoid duplicate work?

cetagostini:
@MarkusSagen I was running the notebook and saw some issues with the dims. Look like they are not being used and the model is estimating a single distribution, rather than one per channel or control.

Screenshot 2023-10-23 at 22 18 56

I could take a look in the next couple of days, and we can work on this together if that's okay? Otherwise, I can help you with whatever you need 🙌🏻

MarkusSagen:
@cetagostini That sounds like a good idea! Leaving them fixed prevents users from having unintended consequences. 
Yes, I started with the tests first and was then going to test the notebook example.

The thing I got stuck on was that `register_rv` does not seem to have a way to pass in dimensions

cetagostini:
@MarkusSagen great, then if we are align. What do you think if we have a session during the week? Are you available to talk maybe 45 minutes, and have a session of programming where we can compile the best of the two PR and merge it on one where each of us can modify?

Regarding adding dims, you can modify this in your code, adding the dimensions if you follow the `register_rv` [documentation](https://www.pymc.io/projects/docs/en/v5.5.0/api/generated/classmethods/pymc.Model.register_rv.html) in PyMC.

``` python
self.model.register_rv(
                self.beta_channel, name="beta_channel", dims=("channel",)
            )
```

or how Juan [apply it here](https://github.com/pymc-labs/pymc-marketing/pull/286/files).

cetagostini:
@MarkusSagen I was able to modify everything and make it run smoothly [here](https://github.com/pymc-labs/pymc-marketing/pull/397).

If you want you can take a look, and suggest some changes if needed 🙌🏻

twiecki:
Have you guys seen https://github.com/pymc-labs/pymc-marketing/pull/415? Any way to consolidate the efforts?

MarkusSagen:
Nice! If that would work with incorporating time-varying coefficients, it seems more future proof.
Might make sense though to do one change after the other (First #397 then #415) for tracking and ensuring that all tests still pass?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -75,6 +118,28 @@ def __init__(
...
+        )
+        self._process_priors(
+            self.intercept,
+            self.beta_channel,
+            self.alpha,
+            self.lam,
+            self.sigma,
+            self.gamma_control,
+            self.gamma_fourier,
+        )
```
ricardoV94, line 141: This should only be done in `build_model`

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -187,33 +253,16 @@ def build_model(
...
-                beta=model_config["alpha"]["beta"],
-                dims=model_config["alpha"]["dims"],
-            )
-
-            lam = pm.Gamma(
-                name="lam",
-                alpha=model_config["lam"]["alpha"],
-                beta=model_config["lam"]["beta"],
-                dims=model_config["lam"]["dims"],
+            # FIXME: Need to add the correct dims to `beta_channel`, `alpha`, `lam`,
```
ricardoV94, line 256: The problem here is that `dims` no longer affect the shape of the rv by this point.
Instead of using the approach in CLVModel which works fine because we are always building scalar distributions, use a `_create_model_distribution` that looks like this:

```python
    @staticmethod
    def _create_model_distribution(name: str, dist: Dict, ndim: int = 0, dims = None, model: Optional[Model]=None) -> TensorVariable:
        model = pm.modelcontext(model)
        try:
            with model:
                prior_distribution = getattr(pm, dist["dist"])(name, **dist["kwargs"], dims=dims)
        except AttributeError:
            raise ValueError(f"Distribution {dist['dist']} does not exist in PyMC")
        return prior_distribution
```

Then you don't need to call `model.register_rv`, as you are already creating a model variable

MarkusSagen:
Closing. Fixed in https://github.com/pymc-labs/pymc-marketing/pull/397

PR #411, opened 2023-10-24
Adding a data generator class based on the demo notebook

nikisix:
----
:books: Documentation preview :books:: https://pymc-marketing--411.org.readthedocs.build/en/411/

### DIFF
--- a/pymc_marketing/mmm/data_generator.py
+++ b/pymc_marketing/mmm/data_generator.py
@@ -1,0 +1,521 @@
+from pymc_marketing.mmm.transformers import (
+    delayed_adstock,
+    geometric_adstock,
+    logistic_saturation,
+    tanh_saturation)
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import seaborn as sns
+import warnings
+
+warnings.filterwarnings('ignore', category=FutureWarning)
+
+
+class DataGenerator:
+    ''' # Data Generator
+    Adapted from PYMC-Marketing Demo
+    ## Models a Time-Series Equation:
+        y_t = α + Σ_m β_m*f(x_m_t) + Σ_c γ_c*z_c_t + ε_t
+    Where:
+        y_t = sales or KPI at time t
+        α = baseline intercept and trend of dataset
+        β_m = coef for channel m
+        x_c_t = spend on channel m at time t
+        f = function of spend, often lagged and/or saturated (see below).
+        γ_c = coef for control columns c
+        z_c_t = control column c at time t. (Ex. Year, month, event-spike)
+        ε_t = noise at time t
+
+    ## Generate a dataset while being able to toggle the following effects:
+        - Lagging interest due to ad spend (common approaches include: adstock, carryover, convolution models)
+        - Lagged effects (α and maybe θ)
+        - Saturation effects (λ)
+        - Overall trend
+        - Event effects
+        - Yearly seasonality
+        - Intercept
+    Determine if the model is flexible enough to correctly identify which effects are or are not present.
+    And how close to the true coefficients are recovered.
+
+    ## Media Data and Channel Betas (X & β)
+    Media data can be singular -- spend per channel _only_ --
+    OR dual -- spend per channel and impressions per channel --.
+    This generator allows the creation of dual-media-datasets, and dual sets drawn from a
+    multi-variate normal distribution to model the so called "Halo Effect" or covariance
+    between channel impressions in response to spend in other channels.
+
+    ## Response Variable (y)
+    The response variable can be singular (total sales over all channels (∑sales)) or multiplicative (∑sales, ∑impressions, ∑KPI_3).
+    Basically y can be one or more KPIs.
+    Some models will be able to incorporate the additional information.
+    ''';
+
+    def __init__(
+            self
+            , min_date:str="2018-04-01"
+            , max_date:str="2021-09-01"
+            , num_channels=2
+            , beta_scale=3
+        ):
+        seed: int = sum(map(ord, "six-mmm"))
+        self.rng: np.random.Generator = np.random.default_rng(seed=seed)
+        # create dataframe and channel columns
+        self.add_daterange(min_date, max_date)
+        self.add_channel_spends(num_channels, beta_scale)
+        self.adstock_fns = {'geometric':geometric_adstock, 'delayed':delayed_adstock}
+        # Switch these on if/when added to the signal
+        self.is_lagged = False
+        self.is_saturated = False
+        self.is_trended_and_seasonal = False
+        self.has_events = False
+        self.has_intercept = False
+        print(f"Number of observations: {self.n}")
+
+    def add_daterange(self, min_date, max_date):
+        ''' Set a time range for our data. '''
+        self.min_date = pd.to_datetime(min_date)
+        self.max_date = pd.to_datetime(max_date)
+        self.df = pd.DataFrame(
+                data={"date_week": pd.date_range(start=min_date, end=max_date, freq="W-MON")}
+            ).assign(
+                year=lambda x: x["date_week"].dt.year,
+                month=lambda x: x["date_week"].dt.month,
+                dayofyear=lambda x: x["date_week"].dt.dayofyear,
+            )
+        self.n = self.df.shape[0]
+
+    def add_channel_spends(self, num_channels=2, beta_scale=3):
+        '''2. Media Costs Data
+        - Generate synthetic data from channel columns. We refer to
+          it as the raw signal as it is going to be the input at the modeling phase. We
+          expect the contribution of each channel to be different, based on the
+          carryover and saturation parameters.
+        Parameters
+        ----------
+        num_channels: int
+            Number of media channels present in the dataset.
+        beta_scale: float
+            Mean (λ) of the exponential distribution the channel spend params are sampled from.
+        '''
+        # Save these for benchmarking
+        self.num_channels = num_channels
+        self.betas = np.random.exponential(beta_scale, size=num_channels)
+        self.channel_spend_probs = np.random.uniform(.5, 1.0, num_channels)
+
+        for c in range(num_channels):
+            x = self.rng.uniform(low=0.0, high=1.0, size=self.n)
+            self.df[f"x{c}"] = np.where(x > self.channel_spend_probs[c], x, 0)
+
+
+    def plot_channel_spends(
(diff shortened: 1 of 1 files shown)


pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,521 @@
...
+        l_max : int, by default 12
+            Maximum duration of carryover effect.
+        normalize : bool, by default False
+            Whether to normalize the weights.
+        ''';
+        assert fn in self.adstock_fns.keys()
+        df = self.df
+        self.is_lagged = True # signals that adstock has been applied
+        self.lag_fn = fn
+        self.alphas = np.random.beta(.5, .5, self.num_channels)
```
wd60622, line 163: Might be good to use or pass another random seed here

wd60622 (commented):
Still taking a look at this. You might want to run `make lint` locally in order to format

pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,521 @@
...
+    def plot_channel_spends(
+            self, suffix:str='', fig=None, ax=None, alpha:float=1.0
+            , title:str="Media Costs Data", show:bool=True):
+        '''
+        Parameters
+        suffix: str
+            Channel suffix. I.e. x1_adstock "_adstock" would be the
+            channel suffix.
+        show: bool
+            Can be turned off for overlays.'''
```
wd60622, line 120: Stick with numpy docstrings

pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,521 @@
...
+                fn = geometric_adstock
+            else:
+                fn = delayed_adstock
+                args['theta']=self.thetas[c]
+            df[f"x{c}_adstock"] = fn(**args).eval().flatten()
+
+
+    def plot_adstock_effect(self):
+        '''Plots the raw channels compared to when their adstock effects are applied.'''
+        assert self.is_lagged
```
wd60622, line 188: Some more informative error messages and error types could be helpfuk

pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,521 @@
...
+        assert self.is_saturated
+        if not self.is_lagged:
+            fig, ax = self.plot_channel_spends(show=False)
+            self.plot_channel_spends(
+                    suffix="_saturated", title="Saturation Effect", fig=fig, ax=ax, alpha=.5)
+        else:
+            fig, ax = self.plot_channel_spends(suffix="_adstock", show=False)
+            self.plot_channel_spends(
+                    suffix="_adstock_saturated", alpha=.5, title="Saturation Effect",
+                    fig=fig, ax=ax)
```
wd60622, line 227: These can be consolidate and by using if block to define different kwargs

pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,521 @@
...
+        self.trend = np.random.exponential(trend_scale)
+        df["trend"] = np.linspace(start=0.0, stop=self.trend, num=self.n)
+
+        # TODO Generalize first order seasonality to Forier degree n
+        (self.sin_coef, self.cos_coef) = np.random.exponential(seasonality_scale, size=2)
+        num_periods = len(df)
+        freq = 52 # weeks per year
+        df["cs"] = -self.sin_coef * np.sin(num_periods * 2 * np.pi * df.index / freq)
+        df["cc"] =  self.cos_coef * np.cos(num_periods * 2 * np.pi * df.index / freq)
+        df["seasonality"] = 0.5 * (df["cs"] + df["cc"])
```
wd60622, line 252: This already exists in the mmm/utils.py 
Could that be utilized?

wd60622:
Can you post a picture of some of the different tplots

pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,521 @@
...
+    geometric_adstock,
+    logistic_saturation,
+    tanh_saturation)
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import seaborn as sns
+import warnings
+
+warnings.filterwarnings('ignore', category=FutureWarning)
```
wd60622, line 12: What are these coming from?

nikisix:
```
python3.11/site-packages/seaborn/_core.py:1218: FutureWarning: is_categorical_dtype is deprecated and will be removed in a future version. Use isinstance(dtype, CategoricalDtype) instead
  if pd.api.types.is_categorical_dtype(vector):
```

moved it back to matplotlib

nikisix: Nevermind, too many plots in other areas, just going to keep the warnings statement unless it's a big deal.

pymc_marketing/mmm/data_generator.py
nikisix: That native approach of creating 2N columns and dotting with the sin and cos coefs would make the df unwieldy to pass around. I prefer keeping them compressed into a single 'seasonality' column.

I did however generalize this to N degrees of Fourier series.

pymc_marketing/mmm/data_generator.py
nikisix: good call

nikisix:
A gist with the plots:
https://gist.github.com/nikisix/df755ade12c6a05f879f9241c7c422d4

nikisix:
Everything should be addressed now. Let me know if there's anything else.

nikisix:
Also, looks like a new version just dropped. Let me know if you'd like me to remake this PR on top of a fresh branch or not.

pymc_marketing/mmm/data_generator.py
nikisix: They're coming from seaborn^

pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,611 @@
...
+import pandas as pd
+import seaborn as sns
+
+from pymc_marketing.mmm.transformers import (
+    delayed_adstock,
+    geometric_adstock,
+    logistic_saturation,
+)
+
+warnings.filterwarnings("ignore", category=FutureWarning)
```
ricardoV94, line 14: We shouldn't have a blank warning filter like this. You can suppress specific warnings around specific function calls instead with a local `with warnings.catch_warnings()`

pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,611 @@
...
+FIGSIZE = (15, 8)
```
ricardoV94, line 15: Matplotlib / Seaborn / Arviz use a global variable (something rc) that controls things like this. We should let those be used by default, and users can tweak them that way as well.

pymc_marketing/mmm/data_generator.py
```diff
@@ -0,0 +1,611 @@
...
+
+    def __init__(
+        self,
+        min_date: str = "2020-02-02",
+        max_date: str = "2022-01-20",
+        num_channels=2,
+        beta_scale=3,
+    ):
+        seed: int = sum(map(ord, "six-mmm"))
+        self.rng: np.random.Generator = np.random.default_rng(seed=seed)
```
ricardoV94, line 69: Allow user to provide an optional seed?

ricardoV94:
This functionality is fine, although I fear a bit hard to maintain as we change how MMM works. Ideally the MMM models could be used to generate synthetic data, but that's another story.

We should add a couple of tests to make sure basic functionality is not fundamentally broken.

wd60622:
Hi @nikisix 
Are you still interested in this? 

It might be good to separate the data generation into: 
- X data generation
- y data generation

The y data generation like @ricardoV94 mentioned can be do with the model itself. We have some examples of that being done [here](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_tvp_example.html#simulate-sales) and [here](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_lift_test.html#generate-correlated-spends-and-model-target) with the `pm.do` operator.

Separating these would allow us to not duplicate the data generation process defined by the model!

nikisix:
Hi @wd60622,
 Thanks for reaching out, responses inline.

> Hi @nikisix Are you still interested in this?
> 
A bit, but honestly I don't have a ton of time for this right now.

> It might be good to separate the data generation into:
> 
> * X data generation
> * y data generation
> 
> The y data generation like @ricardoV94 mentioned can be do with the model itself. We have some examples of that being done [here](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_tvp_example.html#simulate-sales) and [here](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_lift_test.html#generate-correlated-spends-and-model-target) with the `pm.do` operator.
> 
> Separating these would allow us to not duplicate the data generation process defined by the model!

Doesn't that seem a bit circular though? As I implied to @ricardoV94, my intuition would be to go the opposite direction and remove all model/transform dependencies. This has 2 benefits:
1. Model selection can run on generated data in an unbiased way. Ex. if generating data from an adstock hill model for instance then adstock hill model types would have an unfair advantage in model comparison.
2. Decouples the data generator from model code, making the code-base more robust.

Am I missing something perhaps?

However, I did write a test file with graphs (as requested). So if that's the delta between this PR and acceptance I can commit it, and y'all can take it from there.

nikisix:
@wd60622 on second thought I agree with you guys about leveraging the models directly. Mainly because I can't think of a way to implement an independent data generation process. So you might as well couple this to a model type, give up on using it for model selection, but have a cleaner approach to parameter recovery. Point taken. Also the `do` operator looks slick.
Like I said, kind of swamped right now. Would you be ok accepting this (+test) as a first round, and then we can update the core generator to use `pymc.do` as a second pass?

wd60622:
> @wd60622 on second thought I agree with you guys about leveraging the models directly. Mainly because I can't think of a way to implement an independent data generation process. So you might as well couple this to a model type, give up on using it for model selection, but have a cleaner approach to parameter recovery. Point taken. Also the `do` operator looks slick. Like I said, kind of swamped right now. Would you be ok accepting this (+test) as a first round, and then we can update the core generator to use `pymc.do` as a second pass?

Hi @nikisix 
I think it'd be good to just have a generation process for 
- dates
- controls / events
- spends

Then just remove the target generation process which can be saved for later. I.e. remove intercept, trend, seasonality, etc because that is handled by the model.

Maybe @juanitorduz has some thoughts too

Issue #328, opened 2023-07-16
Wrong Parameter Specification in `BetaGeoModel._distribution_new_customers`


ColtAllen:
The new customer purchase rate distribution requires `alpha` to be converted into a rate parameter. This is a very easy fix and can done via `beta=1/alpha` in this line of code:

https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/models/beta_geo.py#L375

ricardoV94:
CC @wd60622

ricardoV94:
Good question maybe @larryshamalama can clear out the confusion? 

@ColtAllen can you provide more details on why you think alpha should be converted to a rate parameter in that method?

ColtAllen:
When the method is plotted in its current form, the X-axis scale is not correct; it should be proportional to `recency` values.

https://github.com/pymc-lpabs/pymc-marketing/blob/main/docs/source/notebooks/clv/dev/pareto_nbd.ipynb contains plotted examples for the `ParetoNBDModel` equivalent, which uses a rate parameter.

wd60622:
Sounds good. Yeah, its a easy mistake to have. I kept double checking the hadie paper when I was implementing elsewhere. 
I can look into our parameterization soon, including the tests

ColtAllen:
I've realized there's no need for changes, and taking the reciprocal of this function's output is the correct way to yield results on the time scale.

This is a good cue to review the data simulation function in `ParetoNBD` - I based it on the corresponding function in `lifetimes`; I wonder why the parameter conversion was performed.

ColtAllen:
I've realized there's no need for changes, and taking the reciprocal of this function's output is the correct way to yield results on the time scale.

This is a good cue to review the data simulation function in `ParetoNBD` - I based it on the corresponding function in `lifetimes`; I wonder why the parameter conversion was performed.

wd60622:
> This is a good cue to review the data simulation function in `ParetoNBD` - I based it on the corresponding function in `lifetimes`; I wonder why the parameter conversion was performed.

If you need a second pair of eyes, let me know

PR #430, opened 2023-11-11
Fix parameterization in `_distribution_new_customers`

wd60622:
Address #328 

Wondering why there is discrepancy in the generative process in the top of the test class and what is required here. Also, this test doesn't confirm the validity 

TODO: 
- Add test to confirm @ColtAllen 's claim. Compare the rate to the scale of the observed data
- Is it good to add an example and a plot of this to one of the notebooks? visual confirmation of above

----
:books: Documentation preview :books:: https://pymc-marketing--430.org.readthedocs.build/en/430/

### DIFF
--- a/pymc_marketing/clv/models/beta_geo.py
+++ b/pymc_marketing/clv/models/beta_geo.py
@@ -372,7 +372,9 @@ class BetaGeoModel(CLVModel):
                 shape_kwargs = {}
 
             pm.Beta("population_dropout", alpha=a, beta=b, **shape_kwargs)
-            pm.Gamma("population_purchase_rate", alpha=r, beta=alpha, **shape_kwargs)
+            pm.Gamma(
+                "population_purchase_rate", alpha=r, beta=1 / alpha, **shape_kwargs
+            )
 
             return pm.sample_posterior_predictive(
                 self.fit_result,
--- a/tests/clv/models/test_beta_geo.py
+++ b/tests/clv/models/test_beta_geo.py
@@ -471,7 +471,7 @@ class TestBetaGeoModel:
 
         N = 1000
         p = pm.Beta.dist(self.a_true, self.b_true, size=N)
-        lam = pm.Gamma.dist(self.r_true, self.alpha_true, size=N)
+        lam = pm.Gamma.dist(self.r_true, 1 / self.alpha_true, size=N)
 
         rtol = 0.05
         np.testing.assert_allclose(


wd60622:
I still think the parameterization was correct before. For instance, the distribution for new population_purchase_rate is in the same scale of naive empirical estimates (obviously better as dropout is possible). 
On the otherhand, the 1 / alpha parameterization is way larger. 
I had that population_purchase_rate is the rate / time unit so the previous parametrization makes sense to me
Screenshot 2023-11-11 at 08 25 40
Screenshot 2023-11-11 at 08 26 45

ColtAllen:
Frequency/recency would make for a better emperical comparison, as recency denotes the time period of the customer's last purchase. However, I've been thinking of this in the reciprocal (recency/frequency) which would be the average number of days between purchases.

The best way to test this in practice (if not for this PR) would be to do a posterior predictive check with these purchase and dropout rate distributions as parametrized:



However, PPCs aren't supported in `BetaGeoModel`. Doing so would require resolving this issue:

https://github.com/pymc-labs/pymc-marketing/issues/128

wd60622:
Sure, that is better approximation, but I would argue the previous parameterization makes more sense with even that value. 

For instance, the scale of the previous parameterization goes from 0 to 4 purchases / unit time while alive whereas the suggested parameterization has that goes from 0 to 30+ / unit time. Both the empirical frequency / recency and frequency / T distributions are in that range. 

If we had the model where dropout doesn't happen and keep poisson likelihood with gamma population prior, one of those would be our MLEs so the range of the previous parameterization makes most sense to me.

Screenshot 2023-11-11 at 13 04 33

Pardon any typos in the graphs 🥲

ColtAllen:
Thinking more on this, I've realized you're right - this model is a poisson process for frequency over time, and flipping the rate parameter won't yield time per frequency.

The parameter can remain as previously defined; no need for changes. I do think time between purchases is more interpretable to the user though, so I'll make a note to take the reciprocal of this function's output for plotting in a future PR.

One last nit-pick - the tests in beta geo use arbitrary parameters rather than generating them from a test dataset, and `r=4` is unrealistic. In every dataset I've worked with that parameter hovers around 0.5.

wd60622:
All good. I was wrapping my head around it too. Between reusing greek letters and gamma parameterizations, its hard to keep it straight. 
Thanks for taking a look!

wd60622:
> One last nit-pick - the tests in beta geo use arbitrary parameters rather than generating them from a test dataset, and `r=4` is unrealistic. In every dataset I've worked with that parameter hovers around 0.5.

Might be good for a new issue?

Issue #352, opened 2023-08-09
Add `log-likelihood` Attribute to fitted `InferenceData` object


ColtAllen:
The Widely Available Information Criterion (WAIC) is a useful model comparison metric:

https://python.arviz.org/en/stable/api/generated/arviz.waic.html

However, the `InferenceData` object returned from a model fit does not contain the required `log-likelihood` attribute (the error trace below is for `ParetoNBDModel`):

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[165], line 1
----> 1 az.waic(mcmc_fit.idata)

File ~/miniconda3/envs/pymc-dev/lib/python3.8/site-packages/arviz/stats/stats.py:1613, in waic(data, pointwise, var_name, scale, dask_kwargs)
   1546 """Compute the widely applicable information criterion.
   1547 
   1548 Estimates the expected log pointwise predictive density (elpd) using WAIC. Also calculates the
   (...)
   1610        ...: data_waic.waic_i
   1611 """
   1612 inference_data = convert_to_inference_data(data)
-> 1613 log_likelihood = _get_log_likelihood(inference_data, var_name=var_name)
   1614 scale = rcParams["stats.ic_scale"] if scale is None else scale.lower()
   1615 pointwise = rcParams["stats.ic_pointwise"] if pointwise is None else pointwise

File ~/miniconda3/envs/pymc-dev/lib/python3.8/site-packages/arviz/stats/stats_utils.py:425, in get_log_likelihood(idata, var_name)
    423     return idata.sample_stats.log_likelihood
    424 if not hasattr(idata, "log_likelihood"):
--> 425     raise TypeError("log likelihood not found in inference data object")
    426 if var_name is None:
    427     var_names = list(idata.log_likelihood.data_vars)

TypeError: log likelihood not found in inference data object
```

Adding `log-likelihood` may have other benefits beyond this, so if you can think of any please share them.

ColtAllen:
For MCMC fits this can be done via `sample_config`:

```
mcmc_fit = ParetoNBDModel(df,sampler_config = {"idata_kwargs": {"log_likelihood": True}})
```

Not sure what to do for MAP fits though.

ricardoV94:
We can call `pm.compute_log_likelihood` for methods that need this quantity when it's missing?

https://www.pymc.io/projects/docs/en/latest/api/generated/pymc.compute_log_likelihood.html#pymc.compute_log_likelihood

[...]

PR #431, opened 2023-11-11
Add `BetaGeoBetaBinom` Distribution Block

ColtAllen:
This PR resurrects https://github.com/pymc-labs/pymc-marketing/pull/188  to partially address https://github.com/pymc-labs/pymc-marketing/issues/176. 

The Beta-Geometic/Beta Binomial model is a CLV model for discrete, non-contractual use cases - a good example would be sporting events. This PR adds a distribution block to be used for posterior predictive checks and the `logp` in the full model which will be added in a separate PR.

For now this is a draft PR because I can't get the `logp` to return the correct values; for reference please see (5) on p4 of the research paper:

https://www.brucehardie.com/papers/020/fader_et_al_mksc_10.pdf

I could use some help with the `pytensor` functions this requires.

----
:books: Documentation preview :books:: https://pymc-marketing--431.org.readthedocs.build/en/431/


pymc_marketing/clv/distributions.py
```diff
@@ -460,3 +458,185 @@ def logp(value, r, alpha, s, beta, T):
...
+        # TODO: Resolve this scan loop and logp will be ready
+        def _B_iter(ix):
+            return betaln(alpha + x, beta + t_x - x + ix) + betaln(
+                gamma + 1, delta + t_x + ix
+            )
+
+        iter_result, _ = scan(
+            fn=_B_iter,
+            # outputs_info=pt.zeros(1),
+            sequences=pt.arange(t_recent),
```
ricardoV94, line 612: Is t recent a vector meaning you want to run a scan for different lengths, or a scalar? 

If it's a scalar, how large can it get in reasonable use cases?

pymc_marketing/clv/distributions.py
```diff
@@ -460,3 +458,185 @@ def logp(value, r, alpha, s, beta, T):
...
+
+        # TODO: Resolve this scan loop and logp will be ready
+        def _B_iter(ix):
+            return betaln(alpha + x, beta + t_x - x + ix) + betaln(
+                gamma + 1, delta + t_x + ix
+            )
+
+        iter_result, _ = scan(
+            fn=_B_iter,
+            # outputs_info=pt.zeros(1),
```
ricardoV94, line 611: When an output doesn't carey over into the inner function, you specify with `None`

```suggestion
            outputs_info=[None], 
```

ColtAllen: I added the suggestion you made in the previous PR to eliminate `scan`:
https://github.com/pymc-labs/pymc-marketing/pull/188#discussion_r1132133861

Tests are still failing though.

ricardoV94:
I'll try to check this week @ColtAllen

pymc_marketing/clv/distributions.py
```diff
@@ -460,3 +457,189 @@ def logp(value, r, alpha, s, beta, T):
...
+                purchase = rng.binomial(1, purchase_prob)
+                if active and purchase:
+                    n += 1
+
+            return np.array(
+                [
+                    t,
+                    n,
+                ],
+            )
```
ricardoV94, line 524: Is `t` and `n` supposed to be `t_x` and `x` from the paper? If so, it looks like `t` should be the time of the last observation and not the time of death (which is supposed to be unknown)?

Otherwise the draws from this RV don't match with the logp method

ricardoV94: Also I think we can avoid the while loop. Untested code:

```python
t_death = pt.minimum(rng.geometric(churn_p), T)   # May need an offset on the geometric
purchases = rng.binomial(t_death, purchase_prob)
x = np.sum(purchases)
t_x = T - np.argmax(purchases[::-1])
```

ricardoV94:
I pushed a version of the logp that works with Scan. I think we can make it work without a Scan, by evaluating on the maximum T-tx range. I just don't know if this results in much wasted computation in general use cases (for instance if there's one datapoint with a very large range, but most other datapoints have smaller ranges)

pymc_marketing/clv/distributions.py
```diff
@@ -460,3 +459,221 @@ def logp(value, r, alpha, s, beta, T):
...
+            t_x, alpha, beta, gamma, delta, T
+        )
+
+        def logp_customer_died(t_x_i, x_i, alpha_i, beta_i, gamma_i, delta_i, T_i):
+            i = pt.scalar("i", dtype=int)
+            died = pt.lt(t_x_i + i, T_i)
+
+            unnorm_logprob_customer_died_at_tx_plus_i = betaln(
+                alpha_i + x_i, beta_i + t_x_i - x_i + i
+            ) + betaln(gamma_i + died, delta_i + t_x_i + i)
```
ricardoV94, line 616: The only difference between the series in the sum and the ones outside, in the paper and lifetimes), is that +1 after the gamma. By making it a boolean variable we can handle everything in the same sum.

ColtAllen:
I added the likelihood expression to the docstrings, along with a dev notebook. Not sure if the LaTeX rendered properly in the docstrings though, because the docs build is failing due to a `pytensor` issue:

```
ImportError: cannot import name 'vectorize' from 'pytensor.graph' 
(/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/431/lib/python3.10/site-packages/pytensor/graph/__init__.py)
```

ColtAllen:
> I pushed a version of the logp that works with Scan. I think we can make it work without a Scan, by evaluating on the maximum T-tx range. I just don't know if this results in much wasted computation in general use cases (for instance if there's one datapoint with a very large range, but most other datapoints have smaller ranges)

In general I don't see very large values of `T` being a common use case with this model. Is Scan the most performant approach? If so I'm fine leaving it as-is.

ricardoV94:
> I added the likelihood expression to the docstrings, along with a dev notebook. Not sure if the LaTeX rendered properly in the docstrings though, because the docs build is failing due to a `pytensor` issue:
> 
> ```
> ImportError: cannot import name 'vectorize' from 'pytensor.graph' 
> (/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/431/lib/python3.10/site-packages/pytensor/graph/__init__.py)
> ```

It's now called `vectorize_graph`

ricardoV94:
> Is Scan the most performant approach? If so I'm fine leaving it as-is.

Scan is the simplest approach

ColtAllen:
Updated my dev environment to use `vectorize_graph`. Do the package dependencies also need to be updated?

ColtAllen:
Tests are still failing with `vectorize_graph`. Do the library dependencies require updating?

ricardoV94:
Yes you need to bump the oldest supported pymc version to the most recent

ricardoV94:
In here and in the ci wofkflow file: https://github.com/pymc-labs/pymc-marketing/blob/main/pyproject.toml#L20

pymc_marketing/clv/distributions.py
ricardoV94: @ColtAllen did you see this comment :arrow_up: ?

tests/clv/test_distributions.py
```diff
@@ -253,3 +261,158 @@ def test_pareto_nbd_sample_prior(
...
+                alpha=alpha,
+                beta=beta,
+                gamma=gamma,
+                delta=delta,
+                T=T,
+                size=beta_geo_beta_binom_size,
+            )
+            prior = pm.sample_prior_predictive(samples=10)
+
+        assert prior["prior"]["beta_geo_beta_binom"][0].shape == (10,) + expected_size
```
ricardoV94, line 418: Can we test something about the expected values, not just their shape?

pymc_marketing/clv/distributions.py
```diff
@@ -460,3 +459,206 @@ def logp(value, r, alpha, s, beta, T):
...
+                active = rng.binomial(1, churn_prob)
+                purchase = rng.binomial(1, purchase_prob)
+                if active and purchase:
+                    x += 1
+
+            return np.array(
+                [
+                    t_x,
+                    x,
+                ],
```
ricardoV94, line 525:
```suggestion
            recency = 0
            
            while t_x <= T and active:
                t_x += 1
                active = rng.binomial(1, churn_prob)
                purchase = rng.binomial(1, purchase_prob)
                if active and purchase:
                    recency = t_x
                    x += 1

            return np.array(
                [
                    recency,
                    x,
                ],
```

ColtAllen:
I had to increase `rtol` because we're comparing against the data simulation function in `lifetimes`, and it lacks a random seed.  If the `sim_data` function looks good and I'm indexing properly in [this test](https://github.com/ColtAllen/pymc-marketing/blob/bgbb_dist/tests/clv/test_distributions.py#L438), this should be good to go.

If https://github.com/pymc-labs/pymc-marketing/issues/352 gets resolved in the future, adding `logp` and `sim_data` directly to the model class may be more performant than a separate distribution.

pymc_marketing/clv/distributions.py
```diff
@@ -476,3 +475,207 @@ def logp(value, r, alpha, s, beta, T):
...
+
+        # Maximum prevents invalid T - t_x values from crashing logp
+        # max_range = pt.maximum(pt.max(T - t_x), 0)
+        # i_vec = pt.arange(max_range + 1)
+        # unnorm_logp_died_at_tx_plus_i_vec = vectorize(
+        #     unnorm_logp_died_at_tx_plus_i,
+        #     replace={i: i_vec},
+        # )
+
+        # unnorm_logp = pt.logsumexp(unnorm_logp_died_at_tx_plus_i_vec, axis=0)
```
juanitorduz, line 655: Can we remove this commented code and create an issue instead 🙏 ?

juanitorduz (changes requested):
@ColtAllen this looks good! I wanna merge this one and iterate (I can help with that). I have just two requests :D : 

- Remove commented code
- Fix pre-commit (I think Ruff is complaining)

Otherwise LGTM.

pymc_marketing/clv/distributions.py
```diff
@@ -476,3 +475,207 @@ def logp(value, r, alpha, s, beta, T):
...
+                    )
+
+        # Broadcast all the parameters so they are sequences.
+        # Potentially inefficient, but otherwise ugly logic needed to unpack arguments in the scan function,
+        # since sequences always precede non-sequences.
+        _, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
+            t_x, alpha, beta, gamma, delta, T
+        )
+
+        def logp_customer_died(t_x_i, x_i, alpha_i, beta_i, gamma_i, delta_i, T_i):
```
juanitorduz, line 612: Can we add a reference to the equation this formal uses in https://www.brucehardie.com/papers/020/fader_et_al_mksc_10.pdf?  (is it the log of equation (5) ?)

ColtAllen:
> @ColtAllen this looks good! I wanna merge this one and iterate (I can help with that). I have just two requests :D :
> 
> * Remove commented code
> * Fix pre-commit (I think Ruff is complaining)
> 
> Otherwise LGTM.

First point is done, but Ruff is complaining about a LaTeX expression in the docstring that is too long. I'm not sure if there's any way to shorten it that doesn't prevent the docstring from rendering properly.

wd60622:
> First point is done, but Ruff is complaining about a LaTeX expression in the docstring that is too long. I'm not sure if there's any way to shorten it that doesn't prevent the docstring from rendering properly.

Doesn't seem to be possible at the moment based on [this ruff issue](https://github.com/astral-sh/ruff/issues/3711). Maybe you can define the string outside of the class then use f-string. Could reduce 8 something characters on a line

juanitorduz:
I think we can simply do it as in https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/distributions.py#L407

pymc_marketing/clv/distributions.py
```diff
@@ -476,3 +475,184 @@ def logp(value, r, alpha, s, beta, T):
...
+    Mean      :math:`\mathbb{E}[X(n) | \alpha, \beta, \gamma, \delta] =  (\frac{\alpha}{\alpha+\beta})(\frac{\delta}{\gamma-1}) \cdot{1-\frac{\Gamma(\gamma+\delta)}{\Gamma(\gamma+\delta+n)}\frac{\Gamma(1+\delta+n)}{\Gamma(1+ \delta)}}`
+    ========  ===============================================
+
+    References
+    ----------
+    .. [1] Fader, Peter S., Bruce G.S. Hardie, and Jen Shang (2010),
+       "Customer-Base Analysis in a Discrete-Time Noncontractual Setting,"
+       Marketing Science, 29 (6), 1086-1108. https://www.brucehardie.com/papers/020/fader_et_al_mksc_10.pdf
+
+    """
```
juanitorduz, line 581:
```suggestion
    """  # noqa: E501
```

juanitorduz (approved):

juanitorduz:
Thanks @ColtAllen ! Let's now work on the model 🎉 !

PR #444, opened 2023-11-24
Support New Data for MMM model

wd60622:
Barebone implementation to support new data instead of the PR that removes the mixins

Mixins transformations from child classes can still be used if they exist down the line. 

This PR introduces : 
- mutable coords for the date in order to support new data
- transformations of channel, control, and date features (if exists) for new data
- raise error in data setter method if not pandas DataFrame since columns various columns need to be accessed 
- tests for the `sample_posterior_predictive` method and check for failure of `predict_posterior` method as design matrix has date type. To enable this, there would have to be relaxing of ModelBuilder validation or override the method

----
:books: Documentation preview :books:: https://pymc-marketing--444.org.readthedocs.build/en/444/


pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -484,7 +488,12 @@ def _data_setter(
         -------
         None
         """
+        if not isinstance(X, pd.DataFrame):
+            raise TypeError(
+                "X must be a pandas DataFrame in order to access the columns"
+            )
```
wd60622, line 494: I'm viewing this as required in order to access the columns. Changing the type hint makes mypy mad

If there are any other suggestions, let me know

wd60622:
Failing test for the change in behavior of _data_setter. No longer supports np.ndarray as input with different

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -484,19 +488,41 @@ def _data_setter(
...
-            except KeyError as e:
-                raise RuntimeError("New data must contain channel_data!", e)
-        elif isinstance(X, np.ndarray):
-            new_channel_data = X
-        else:
-            raise TypeError("X must be either a pandas DataFrame or a numpy array")
+        try:
+            new_channel_data = X[self.channel_columns].to_numpy()
+        except KeyError as e:
+            raise RuntimeError("New data must contain channel_data!", e)
```
wd60622, line 501: Should this be the error raises by any missing keys of the dataframe? For instance, date_column or control_columns

wd60622:
Will rework the tests to include:
- old dates only
- old and new dates
- new dates only

I believe there will be an issue if the new data doesn't have more than `max_adstock` rows too

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -506,3 +537,55 @@ def mock_property(self):
...
+                "other_column_1": rng.integers(low=0, high=100, size=n),
+                "other_column_2": rng.normal(loc=0, scale=1, size=n),
+            }
+        )
+
+        with pytest.raises(
+            TypeError,
+            match=r"The DType <class 'numpy.dtype\[datetime64\]'> could not be promoted by",
+        ):
+            mmm.predict_posterior(X_pred=new_X)
```
wd60622, line 583: as noted in the comment, the ModelBuilder method will need an override if this is to work with dates in the input DataFrame

wd60622:
> Hey mate, do you have a code snippet about how should work? What data should the data frame to predict? Are you planning to add the example to the example notebook? (If you need, I can give you a hand with it) 🙌🏻

Good thinking. I will add some to the mmm_example.ipynb

wd60622:
Had to do a force push with the rebases 🙏

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +843,53 @@ def plot_channel_contributions_grid(
...
         return fig
+
+    def predict_posterior(
+        self,
+        X_pred: Union[np.ndarray, pd.DataFrame, pd.Series],
+        extend_idata: bool = True,
+        combined: bool = True,
+        include_last_observations: bool = False,
+        **kwargs,
+    ) -> DataArray:
```
wd60622, line 854: My interpretation and first pass from our conversation the other week 
@ricardoV94 @cetagostini 

The last_observations can optionally be used in the `predict_posterior` method. 
Begs the question, should this be supported the other predict-like methods, changing their parameters too? 
Would like to hear thoughts on this

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: I think 99% of the cases people using this package use pandas, so is ok for me.

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: I think so! I suggest that, for this iteration, we keep this strict and then wait for feedback.

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: I think ideally yes to have consistent signatures.

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +843,53 @@ def plot_channel_contributions_grid(
...
+        Returns
+        -------
+        y_pred : DataArray, shape (n_pred, chains * draws) if combined is True, otherwise (chains, draws, n_pred)
+            Posterior predictive samples for each input X_pred
+        """
+        if not isinstance(X_pred, pd.DataFrame):
+            raise ValueError("X_pred must be a pandas DataFrame")
+
+        if include_last_observations:
+            X_pred = pd.concat([self.X.iloc[-self.adstock_max_lag :], X_pred])
```
juanitorduz, line 879: Nice!

juanitorduz (commented):
@wd60622 This looks great! I left some comments regarding your questions! Just ping me when you would like a more detailed review.

This PR solves https://github.com/pymc-labs/pymc-marketing/issues/450 right?

Also, will you add this extension to the MMM example notebook as part of this PR?

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: okay. Let me work towards that

wd60622:
> @wd60622 This looks great! I left some comments regarding your questions! Just ping me when you would like a more detailed review.
> 
> This PR solves #450 right?
> 
> Also, will you add this extension to the MMM example notebook as part of this PR?

My current goal is to get all the predict methods sorted out, but I'm currently in rebase hell from the last PR

If it clears up, I will add a notebook too. But I'm thinking I will address that sometime in the future. Is that an issue if they come separately?

ricardoV94:
> but I'm currently in rebase hell from the last PR

I would squash those 15 commits first

juanitorduz:
Thanks @wd60622 ! If its easier for you you can create another PR with a clean tree :)

Regarding the notebook: sure, let's make it another different PR. I will test it anyway during my review 💪

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -515,6 +535,61 @@ def mock_property(self):
...
+            {
+                "date": new_dates,
+                "channel_1": rng.integers(low=0, high=400, size=n),
+                "channel_2": rng.integers(low=0, high=50, size=n),
+                "control_1": rng.gamma(shape=1000, scale=500, size=n),
+                "control_2": rng.gamma(shape=100, scale=5, size=n),
+                "other_column_1": rng.integers(low=0, high=100, size=n),
+                "other_column_2": rng.normal(loc=0, scale=1, size=n),
+            }
+        )
```
juanitorduz, line 573: We could have this as a fixture and split this tests as it has many `assert`s checking different functions?

wd60622:
I will start a new branch based using the feedback. 

@juanitorduz would you be able to share the mentioned method to bring the target back to the original scale for xarray DataArray. Is [this](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/delayed_saturated_mmm.py#L686-L694) what you had in mind?

juanitorduz:
This is the method https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/base.py#L478

juanitorduz:
I have an alternative way using an array method, see https://juanitorduz.github.io/flax_numpyro/ I have no preference 😄

wd60622:
> I have an alternative way using an array method, see https://juanitorduz.github.io/flax_numpyro/ I have no preference 😄

Perfect! Thank you for sending that over. 

I am closing this PR in favor of a cleaner branch. 
Already add the original scale back in 😄 This was super helpful 

New PR:

PR #454, opened 2023-12-04
Add convolution modes

abdalazizrashid:
Add different modes to model interactions (build-up, decay, symmetric build-up decay)
Those modes correspond to different offsets for the convolution:
- Before (build-up)
image

- After (decay)
image

- Overlap (build-up and decay)
  * With overlap we have two cases when the kernel is even we drop the excess from the end
  image
  
  * When the kernel is odd we will have symmetric build-up and decay:
  image

### DIFF
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -25,6 +25,8 @@ def batched_convolution(x, w, axis: int = 0, mode: ConvMode = ConvMode.Before):
         to use in the convolution.
     axis : int
         The axis of ``x`` along witch to apply the convolution
+    mode : (ConvMode, optional): The mode of the convolution. Determines how the convolution is applied at the boundaries of x. Defaults to ConvMode.Before.
+
 
     Returns
     -------
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -76,7 +76,8 @@ def batched_convolution(x, w, axis: int = 0, mode: ConvMode = ConvMode.Before):
     elif mode == ConvMode.Before:
         window = slice(None, -l_max + 1)
     elif mode == ConvMode.Overlap:
-        window = slice(l_max // 2, -(l_max // 2))
+        # Handle even and odd l_max differently if l_max is odd then we can split evenly otherwise we drop from the end
+        window = slice((l_max // 2) - (1 if l_max % 2 == 0 else 0), -(l_max // 2))
     else:
         raise ValueError("Wrong Mode")
 
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from typing import Union
 
 import numpy as np
@@ -6,7 +7,13 @@ import pytensor.tensor as pt
 from pytensor.tensor.random.utils import params_broadcast_shapes
 
 
-def batched_convolution(x, w, axis: int = 0):
+class ConvMode(Enum):
+    After = "After"
+    Before = "Before"
+    Overlap = "Overlap"
+
+
+def batched_convolution(x, w, axis: int = 0, mode: ConvMode = ConvMode.Before):
     """Apply a 1D convolution in a vectorized way across multiple batch dimensions.
 
     Parameters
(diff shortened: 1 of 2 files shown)


pymc_marketing/mmm/transformers.py
```diff
@@ -43,22 +52,40 @@ def batched_convolution(x, w, axis: int = 0):
...
+        window = slice(None)
+    elif mode == ConvMode.After:
+        window = slice(l_max - 1, None)
+    elif mode == ConvMode.Before:
+        window = slice(None, -l_max + 1)
+    elif mode == ConvMode.Overlap:
+        # Handle even and odd l_max differently if l_max is odd then we can split evenly otherwise we drop from the end
+        window = slice((l_max // 2) - (1 if l_max % 2 == 0 else 0), -(l_max // 2))
+    else:
+        raise ValueError("Wrong Mode")
```
ferrine, line 82:
```suggestion
        raise ValueError(f"Wrong Mode: {mode}, expected of ConvMode")
```

ferrine (changes requested):
Looks good, added a small nitpick

pymc_marketing/mmm/transformers.py
```diff
@@ -18,6 +25,8 @@ def batched_convolution(x, w, axis: int = 0):
         to use in the convolution.
     axis : int
         The axis of ``x`` along witch to apply the convolution
+    mode : (ConvMode, optional): The mode of the convolution. Determines how the convolution is applied at the boundaries of x. Defaults to ConvMode.Before.
```
ferrine, line 28: Could you please expand on the cases where a specific mode might be needed?

like 
* After: "Adstock" effect
* Before: "Excitement", or "Last minute" effect
* Overlap: "Pull-Forward" and "Pull-Backward" effects

tests/mmm/test_transformers.py
```diff
@@ -65,8 +67,29 @@ def test_batched_convolution(convolution_inputs, convolution_axis):
...
+            np.allclose(y_val[-1], x_val[-1]),
+            np.allclose(y_val[:-1], x_val[1:] + x_val[:-1]),
+        ),
+        ConvMode.Overlap: lambda: (
+            np.allclose(y_val[0], x_val[0]),
+            np.allclose(y_val[1:-1], x_val[1:-1] + x_val[:-2]),
+        ),
+    }
+
+    assert all(mode_assertions[mode]())
```
ricardoV94, line 85: Nitpick, use if else instead of dict and use `np.testing.assert_allclose` which has a nicer error message when the two don't match.

```suggestion
    mode_assertions = {
        ConvMode.Before: lambda: (
            np.allclose(y_val[0], x_val[0]),
            np.allclose(y_val[1:], x_val[1:] + x_val[:-1]),
        ),
        ConvMode.After: lambda: (
            np.allclose(y_val[-1], x_val[-1]),
            np.allclose(y_val[:-1], x_val[1:] + x_val[:-1]),
        ),
        ConvMode.Overlap: lambda: (
            np.allclose(y_val[0], x_val[0]),
            np.allclose(y_val[1:-1], x_val[1:-1] + x_val[:-2]),
        ),
    }

    assert all(mode_assertions[mode]())
```

pymc_marketing/mmm/transformers.py
```diff
@@ -43,22 +56,40 @@ def batched_convolution(x, w, axis: int = 0):
...
+
     x = pt.broadcast_to(x, x_shape)
     w = pt.broadcast_to(w, w_shape)
     x_time = x.shape[-1]
-    shape = (*x.shape, w.shape[-1])
     # Make a tensor with x at the different time lags needed for the convolution
+    x_shape = list(x.shape)
+    # Add the size of the kernel to the time axis
+    x_shape[-1] = x_shape[-1] + w.shape[-1] - 1
+    shape = [*x_shape, w.shape[-1]]
```
ricardoV94, line 67: More/less readable?
```suggestion
    shape = (*x_shape[:-1], x_shape[-1] + w.shape[-1] - 1, w.shape[-1]]
```

ferrine:
@ricardoV94 any other comments, otherwise looks good

ricardoV94 (approved):
Looks great!

pymc_marketing/mmm/transformers.py
ricardoV94: My issue was with the intermediate variable `x_shape` that is changed in place and no longer holds the shape of x. 
If someone, like me, starts by reading the line `shape = [...]` there's a second of confusion.

Issue #412, opened 2023-10-26
Plot specific channels in contribution_curves


wd60622:
I think it would be convenient to be able to optionally pass a list of channels to plot. Not passing would be the default and would create the current figure 

https://github.com/pymc-labs/pymc-marketing/blob/916dce42b39c67ee3a04188ba492ac53f5e6504d/pymc_marketing/mmm/base.py#L938-L940

PR #455, opened 2023-12-06
add support for select channels and overlay charts

wd60622:
Creating two new arguments
- `channels` : Addressing #412 where `channels` is now an argument that can limit which channels are being plotted
- `same_axes` : if true, all channels selected will be on the same axes. (I've found this helpful to comparing channels)

This can lead to figures like: 

Screenshot 2023-12-05 at 19 51 42

Compared to the visual before which is still the default behavior

----
:books: Documentation preview :books:: https://pymc-marketing--455.org.readthedocs.build/en/455/

### DIFF
--- a/pymc_marketing/mmm/base.py
+++ b/pymc_marketing/mmm/base.py
@@ -1,3 +1,4 @@
+from itertools import repeat
 import warnings
 from inspect import (
     getattr_static,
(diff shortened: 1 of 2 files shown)


juanitorduz (approved):
Thank you @wd60622 !

Issue #467, opened 2023-12-27
possible optimization when discount_rate == 0


vincent-grosbois:
Hi !

in customer_lifetime_value method , I see that we do several predictions, to have one prediction per period.
As far as I can see the only purpose of doing several predictions, is to be able to discount each prediction separately (as they all correspond to a specific period).
It seems that in the case of discount_rate == 0.0, we can simply do a single prediction for the whole period and use the result directly. 
Am I correct ? could the code be optimized in that case ? If so I can do a PR

thanks !

vincent-grosbois:
actually it can be done simply by forcing a single time step when no discount rate is applied

here's a commit for this :
https://github.com/pymc-labs/pymc-marketing/pull/468

Note that I haven't checked, but I really suspect that at each begining of the loop when we call  : 
```
  transaction_model.expected_num_purchases(
            customer_id=customer_id,
            frequency=frequency,
            recency=recency,
            T=T,
            t=0,
)
```
it's always 0 (because of t = 0)

PR #468, opened 2023-12-27
Optimisation in customer_lifetime_value when discount_rate == 0

vincent-grosbois:
Optimisation in customer_lifetime_value when discount_rate == 0

closes https://github.com/pymc-labs/pymc-marketing/issues/467

----
📚 Documentation preview 📚: https://pymc-marketing--468.org.readthedocs.build/en/468/

### DIFF
--- a/pymc_marketing/clv/utils.py
+++ b/pymc_marketing/clv/utils.py
@@ -79,7 +79,12 @@ def customer_lifetime_value(
             x = x.squeeze(dims_to_squeeze)
         return x
 
-    steps = np.arange(1, time + 1)
+    if discount_rate == 0.0:
+        # no discount rate: just compute a single time step from 0 to `time`
+        steps = np.arange(time, time + 1)
+    else:
+        steps = np.arange(1, time + 1)
+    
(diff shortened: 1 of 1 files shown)


vincent-grosbois (commented):
whitespace

ricardoV94 (commented):
Hi @vincent-grosbois, thanks for the PR. If it's this simple, that's great.

Can you add a test? Perhaps you can compute the quantity with discount rate == 0, before the PR change, and then hardcode the results and ensure it indeed comes out the same after the changes.

vincent-grosbois:
I think there are some tests here :
https://github.com/pymc-labs/pymc-marketing/blob/2e826a9a97159c032081fac764591b0d04bd3929/tests/clv/test_utils.py#L150
which shows that nothing should have changed (because it's compared against some evaluations when discount_rate == 1, and this one didn't change)

ricardoV94:
Great we already tested the zero case so we're good

ricardoV94 (approved):

ricardoV94:
Thanks for the contribution!

Issue #450, opened 2023-12-01
sample_posterior_predictive doesn't scale X data


kylejcaron:
In the docs example for MMMs, the line of code to sample the posterior predictive is 

```python
mmm.sample_posterior_predictive(X, extend_idata=True, combined=True)
```

When `X` is on the scale of dollars, the data wont get automatically scaled from `mmm.sample_posterior_predictive`, so it causes the results to be off. The example doesn't run into this problem because channel spend is simulated as `X ~ U(0,1)`. I personally got pretty confused adapting the code. to a real dataset and seeing the posterior predictive plot look so off and had to dig through the codebase to figure out what was going on.

I think the following is a more general approach that would lead to less confusion for now in the docs 

```python
mmm.sample_posterior_predictive(mmm.preprocessed_data['X'], extend_idata=True, combined=True)
```

An alternative to changing the docs could be to scale the X data in the `mmm.sample_posterior_predictive` method

wd60622:
Yes, I am addressing this in PR #444

The Fourier features and control channel transformations will be handled appropriately after that PR as well

That PR will be merged in after the custom prior and time varying parameters get added

geirrlod:
Hi,

Also find this a bit confusing as well when I work on real non scaled spend. Do you think PR444 will be pushed out in the next couple of weeks, or should we try to find a way around the problem?

kylejcaron:
> ```python
> ```python
> mmm.sample_posterior_predictive(mmm.preprocessed_data['X'], extend_idata=True, combined=True)
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```

@geirrlod I made this quick change to get it working for me on the proper scale

```python
mmm.sample_posterior_predictive(mmm.preprocessed_data['X'], extend_idata=True, combined=True)
```

also there's this event tomorrow from pymc-labs that looks pretty interesting, maybe they'll talk about time varying covariate

Issue #396, opened 2023-10-13
Value Error on DelayedSaturatedMMM predict() for test data


kmostdiek-k:
https://github.com/pymc-labs/pymc-marketing/blob/9b8ee1078a6a2bec838d2d56697006ed1f44c433/pymc_marketing/model_builder.py#L492

I fit a DelayedSaturatedMMM model on a training set and then attempt to predict on a test set (which is a different size than the training set) and get the following error:
`ValueError: The 'channel_data' variable already had 46 coord values defined for its date dimension. With the new values this dimension changes to length 45, so new coord values for the date dimension are required.`

My code is the following:
```
mmm_cv = DelayedSaturatedMMM(
    model_config = my_model_config,
    sampler_config = sampler_config,
    data = df,
    date_column="spend_week",
    target_column = "closed_won_mrr",
    channel_columns=["amazon", "bing", "capterra", "dv360", "meta", "google", "linkedin", "mntn", "reddit", "retaildive", "youtube"],
    control_columns=[
        "conference",
        "reviews",
        "cdp",
        "pricing",
        "enterprise",
        "powderhorn",
        "t",
        "holiday",
    ],
    adstock_max_lag=12, 
    yearly_seasonality=5, 
)
mmm_cv.fit(train_x,train_y, target_accept=0.95, chains=4, random_seed=rng)

mmm_cv.predict(test_x)
```
And here's a screenshot of the full error message.
<img width="790" alt="Screenshot 2023-10-13 at 10 35 49 AM" src="https://github.com/pymc-labs/pymc-marketing/assets/92057622/8d092db1-39b8-4416-aa61-2063ce23e7b4">

wd60622:
Hello @kmostdiek-k, 
There is a similar error happening in the discussion here: https://github.com/pymc-labs/pymc-marketing/discussions/392

I present a work around solution there that might work

wd60622:
This should be resolved with the latest release. Feel free to open another issue if you encounter any more issues!

Issue #407, opened 2023-10-23
MMM mixins as single sklearn transformer


williambdean:
The sklearn transformers that require state (MaxAbsScaler and StandardScaler) are currently part of the mixins which make it a bit difficult to support new data. 

So the mixins will have to be change in order to help the new data use case.

Instead of just incorporating into the current mixins, I suggest relying on sklearn to handle the transformation pass directly as I think it is much clearer about what is happening to the data. The current implementation that relies on mixins and searching methods based on tags which is confusing and, imo, hard to maintain.

Sklearn provides fit and transform API already which is easy to adapt for the current flow. For instance, the X transform

```python
from sklearn.preprocessing import MaxAbsScaler, StandardScaler
from sklearn.compose import ColumnTransformer

channel_cols = ["Facebook", "Google"]
control_cols = ["event_1", "event_2"]
date_col = "date_week"

class FourierTransformer(BaseEstimator, TransformerMixin):
    # To create based on fourier transformations
    def fit(self, X, y=None) -> Self: 
        ...
    def transform(self, X) -> pd.DataFrame: 
        ...

# This pipeline can be built based on the channel_cols, control_cols, and date_col provided
pipeline = ColumnTransformer([
    ("channel_scaling", MaxAbsScaler(), channel_cols),
    ("control_scaling", StandardScaler(), control_cols),
    ("date_modes", FourierTransformer(n_order=n_order), date_col)
]).set_output(transform="pandas")


# On fit during self.preprocess
pipeline.fit_transform(X_train)

# Should be used on new data. i.e. _data_setter for predict_posterior, etc
pipeline.transform(X_test)
```

After the fit, `pipeline.named_transformers_` in combination with `get_feature_names_out()` can retrieve the column names required for coordinates

Any thoughts on this? Using sklearn should provide similar feedback to users about missing columns that is currently part of the mixin logic

wd60622:
This would also touch on #402

PR #482, opened 2024-01-11
Handle new data correctly and extend functionality of MMM posterior predictive methods

### LINKED ISSUE #472: scaling control vars
Hi all - I think it’s easy for an unscaled control to go unnoticed. I think it’d be beneficial to either:
1. provide the option to scale control vars in the same way channels are in the `DelayedSaturatedMMM` class, or
2. raise a warning to the user if controls appear to not be scaled

wd60622:
## Description

Redo of the support for new mmm data

This brings in the feedback about 

- having the `include_last_observations` for all predict methods with `X_pred`
- Support for original scale of target with the `original_scale` keyword that can be passed to all predict methods

## Related Issue

- [ ] Closes #472 #450 #400 #396 
- [ ] Related to #444 #392 

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--482.org.readthedocs.build/en/482/

### DIFF
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -673,10 +673,6 @@ class BaseDelayedSaturatedMMM(MMM):
         with self.model:
             pm.set_data(data, coords=coords)
 
-    def _reset_model_data(self) -> None:
-        """Resets the model data to the original data."""
-        self._data_setter(X=self.X, y=self.y)
-
     @classmethod
     def _model_config_formatting(cls, model_config: Dict) -> Dict:
         """
@@ -903,8 +899,6 @@ class DelayedSaturatedMMM(
             )
             if extend_idata:
                 self.idata.extend(post_pred, join="right")  # type: ignore
-
-        self._reset_model_data()
 
         posterior_predictive_samples = az.extract(
             post_pred, "posterior_predictive", combined=combined
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -9,6 +9,7 @@ import numpy.typing as npt
 import pandas as pd
 import pymc as pm
 import seaborn as sns
+import xarray as xr
 from pytensor.tensor import TensorVariable
 from xarray import DataArray
 
@@ -83,7 +84,7 @@ class BaseDelayedSaturatedMMM(MMM):
     @property
     def output_var(self):
         """Defines target variable for the model"""
-        return "y"
+        return "likelihood"
 
     def _generate_and_preprocess_model_data(  # type: ignore
         self, X: Union[pd.DataFrame, pd.Series], y: Union[pd.Series, np.ndarray]
@@ -100,8 +101,11 @@ class BaseDelayedSaturatedMMM(MMM):
         """
         date_data = X[self.date_column]
         channel_data = X[self.channel_columns]
-        coords: Dict[str, Any] = {
+
+        self.coords_mutable: Dict[str, Any] = {
             "date": date_data,
+        }
+        coords: Dict[str, Any] = {
             "channel": self.channel_columns,
         }
 
@@ -315,7 +319,10 @@ class BaseDelayedSaturatedMMM(MMM):
         )
 
         self._generate_and_preprocess_model_data(X, y)
-        with pm.Model(coords=self.model_coords) as self.model:
+        with pm.Model(
+            coords=self.model_coords,
+            coords_mutable=self.coords_mutable,
+        ) as self.model:
             channel_data_ = pm.MutableData(
                 name="channel_data",
                 value=self.preprocessed_data["X"][self.channel_columns],
@@ -611,19 +618,42 @@ class BaseDelayedSaturatedMMM(MMM):
         -------
         None
         """
+        if not isinstance(X, pd.DataFrame):
+            msg = "X must be a pandas DataFrame in order to access the columns"
+            raise TypeError(msg)
+
         new_channel_data: Optional[np.ndarray] = None
+        coords = {"date": X[self.date_column].to_numpy()}
 
-        if isinstance(X, pd.DataFrame):
-            try:
-                new_channel_data = X[self.channel_columns].to_numpy()
-            except KeyError as e:
-                raise RuntimeError("New data must contain channel_data!", e)
-        elif isinstance(X, np.ndarray):
-            new_channel_data = X
-        else:
-            raise TypeError("X must be either a pandas DataFrame or a numpy array")
+        try:
+            new_channel_data = X[self.channel_columns].to_numpy()
+        except KeyError as e:
+            raise RuntimeError("New data must contain channel_data!", e)
 
-        data: Dict[str, Union[np.ndarray, Any]] = {"channel_data": new_channel_data}
+        def identity(x):
+            return x
+
+        channel_transformation = (
+            identity
+            if not hasattr(self, "channel_transformer")
+            else self.channel_transformer.transform
+        )
+
+        data: Dict[str, Union[np.ndarray, Any]] = {
+            "channel_data": channel_transformation(new_channel_data)
+        }
+
+        if self.control_columns is not None:
+            control_data = X[self.control_columns].to_numpy()
+            control_transformation = (
+                identity
+                if not hasattr(self, "control_transformer")
+                else self.control_transformer.transform
+            )
+            data["control_data"] = control_transformation(control_data)
+
+        if hasattr(self, "fourier_columns"):
+            data["fourier_data"] = self._get_fourier_models_data(X)
 
         if y is not None:
             if isinstance(y, pd.Series):
@@ -634,9 +664,12 @@ class BaseDelayedSaturatedMMM(MMM):
                 data["target"] = y
             else:
                 raise TypeError("y must be either a pandas Series or a numpy array")
+        else:
+            dtype = self.preprocessed_data["y"].dtype  # type: ignore
+            data["target"] = np.zeros(X.shape[0], dtype=dtype)  # type: ignore
 
         with self.model:
-            pm.set_data(data)
+            pm.set_data(data, coords=coords)
 
     @classmethod
     def _model_config_formatting(cls, model_config: Dict) -> Dict:
@@ -814,3 +847,80 @@ class DelayedSaturatedMMM(
             ylabel="contribution",
         )
         return fig
+
+    def _validate_data(self, X, y=None):
+        return X
+
+    def sample_posterior_predictive(
+        self,
+        X_pred,
+        extend_idata,
+        combined,
+        include_last_observations: bool = False,
+        original_scale: bool = True,
+        **kwargs,
+    ):
+        """
+        Sample from the model's posterior predictive distribution.
+
+        Parameters
+        ---------
+        X_pred : array, shape (n_pred, n_features)
+            The input data used for prediction using prior distribution..
+        extend_idata : Boolean determining whether the predictions should be added to inference data object.
+            Defaults to False.
+        combined: Combine chain and draw dims into sample. Won't work if a dim named sample already exists.
+            Defaults to True.
+        include_last_observations: Boolean determining whether to include the last observations of the training data.
+            Defaults to False.
+        original_scale: Boolean determining whether to return the predictions in the original scale of the target variable.
+            Defaults to True.
+        **kwargs: Additional arguments to pass to pymc.sample_posterior_predictive
+
+        Returns
+        -------
+        posterior_predictive_samples : DataArray, shape (n_pred, samples)
+            Posterior predictive samples for each input X_pred
+        """
+        if include_last_observations:
+            X_pred = pd.concat(
+                [self.X.iloc[-self.adstock_max_lag :, :], X_pred], axis=0
+            )
+
+        self._data_setter(X_pred)
+
+        with self.model:  # sample with new input data
+            post_pred = pm.sample_posterior_predictive(self.idata, **kwargs)
+            if extend_idata:
+                self.idata.extend(post_pred)  # type: ignore
+
+        posterior_predictive_samples = az.extract(
+            post_pred, "posterior_predictive", combined=combined
+        )
+
+        if include_last_observations:
+            posterior_predictive_samples = posterior_predictive_samples.isel(
+                date=slice(self.adstock_max_lag, None)
+            )
+
+        if original_scale:
+            # These are lost during the ufunc
+            attrs = posterior_predictive_samples.attrs
+
+            if combined:
+                posterior_predictive_samples = xr.apply_ufunc(
+                    self.get_target_transformer().inverse_transform,
+                    posterior_predictive_samples,
+                )
+            else:
+                posterior_predictive_samples = xr.apply_ufunc(
+                    self.get_target_transformer().inverse_transform,
+                    posterior_predictive_samples.expand_dims(dim={"_": 1}, axis=1),
+                    input_core_dims=[["date", "_"]],
+                    output_core_dims=[["date", "_"]],
+                    vectorize=True,
+                ).squeeze(dim="_")
+
+            posterior_predictive_samples.attrs = attrs
+
+        return posterior_predictive_samples
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -84,7 +84,7 @@ class BaseDelayedSaturatedMMM(MMM):
     @property
     def output_var(self):
         """Defines target variable for the model"""
-        return "likelihood"
+        return "y"
 
     def _generate_and_preprocess_model_data(  # type: ignore
         self, X: Union[pd.DataFrame, pd.Series], y: Union[pd.Series, np.ndarray]
@@ -238,7 +238,7 @@ class BaseDelayedSaturatedMMM(MMM):
         likelihood_dist = self._get_distribution(dist={"dist": likelihood_dist_name})
 
         return likelihood_dist(
-            name="likelihood",
+            name=self.output_var,
             mu=mu,
             observed=observed,
             dims=dims,
@@ -892,7 +892,7 @@ class DelayedSaturatedMMM(
         with self.model:  # sample with new input data
             post_pred = pm.sample_posterior_predictive(self.idata, **kwargs)
             if extend_idata:
-                self.idata.extend(post_pred)  # type: ignore
+                self.idata.extend(post_pred, join="right")  # type: ignore
 
         posterior_predictive_samples = az.extract(
             post_pred, "posterior_predictive", combined=combined
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -9,14 +9,16 @@ import numpy.typing as npt
 import pandas as pd
 import pymc as pm
 import seaborn as sns
-import xarray as xr
 from pytensor.tensor import TensorVariable
 from xarray import DataArray
 
 from pymc_marketing.mmm.base import MMM
 from pymc_marketing.mmm.preprocessing import MaxAbsScaleChannels, MaxAbsScaleTarget
 from pymc_marketing.mmm.transformers import geometric_adstock, logistic_saturation
-from pymc_marketing.mmm.utils import generate_fourier_modes
+from pymc_marketing.mmm.utils import (
+    apply_sklearn_transformer_across_date,
+    generate_fourier_modes,
+)
 from pymc_marketing.mmm.validating import ValidateControlColumns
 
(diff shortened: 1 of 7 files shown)


tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -579,6 +610,63 @@ def test_model_config(
...
+        generate_data,
+        model_name: str,
+        new_dates: pd.DatetimeIndex,
+        original_scale: bool,
+        request,
+    ):
+        mmm = request.getfixturevalue(model_name)
+        X_pred = generate_data(new_dates)
+
+        pp_without = mmm.predict_posterior(
```
wd60622, line 642: @juanitorduz 
I can still break these down a bit more
I'm trying to get a handful of the edge cases via the parametrize tests

Are you thinking one for each predict method?

juanitorduz: Yes, ideally test should be atomic (test one thing) and have all assert statements at the end. Not a strong opinion, more.of a suggestion :)

wd60622: Fair enough. I'm pulling out the date ranges and separating the tests

wd60622:
Introducing he new `predict_posterior` call on mmm model

Screenshot 2024-01-11 at 20 41 28

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -83,7 +84,7 @@ def default_sampler_config(self) -> Dict:
     @property
     def output_var(self):
         """Defines target variable for the model"""
-        return "y"
+        return "likelihood"
```
ricardoV94, line 87: I think y is a better good name. likelihood is about the kind of variable it is. That would be like renaming betas to "prior"

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +847,80 @@ def plot_channel_contributions_grid(
...
+            posterior_predictive_samples = posterior_predictive_samples.isel(
+                date=slice(self.adstock_max_lag, None)
+            )
+
+        if original_scale:
+            # These are lost during the ufunc
+            attrs = posterior_predictive_samples.attrs
+
+            if combined:
+                posterior_predictive_samples = xr.apply_ufunc(
```
ricardoV94, line 911: This kind of complexity/gotchas go away if you make the transforms part of the model itself (but for that you need to write them in PyTensor)

That's why I pushed for ir before. Out of the scope of this PR

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +847,80 @@ def plot_channel_contributions_grid(
...
+            X_pred = pd.concat(
+                [self.X.iloc[-self.adstock_max_lag :, :], X_pred], axis=0
+            )
+
+        self._data_setter(X_pred)
+
+        with self.model:  # sample with new input data
+            post_pred = pm.sample_posterior_predictive(self.idata, **kwargs)
+            if extend_idata:
+                self.idata.extend(post_pred)  # type: ignore
```
ricardoV94, line 895: You need to define join otherwise calling this method twice will ignore the second run

See https://github.com/pymc-devs/pymc-experimental/pull/249

ricardoV94: This applies to prior and fit as well, but maybe it's a good time as any to fix it as it's a pretty serious one.

I had a PR on this but haven't come back to it

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: I change in the model definition then!

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: I know we've had the discussion when talking about #407. Maybe it is easier to start with the target as a pytensor transformation. 
Any change back to original scale (a few spots) would have to be changed

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: Sounds good. This is in the method override. Shall I add to the model_builder method as well?

ricardoV94: Yup. Feel free to reuse as much as you can from the PR of the pymc-experimental ModelBuilder (thinking about tests as well). Just add the author as a co-author if you do that.

wd60622: Sounds good. I will dive into it more later

wd60622: I've just added a check that the idata will be the first sample from two method calls to keep it simple. Hope that's what you expected

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +847,80 @@ def plot_channel_contributions_grid(
...
+        original_scale: bool = True,
+        **kwargs,
+    ):
+        """
+        Sample from the model's posterior predictive distribution.
+
+        Parameters
+        ---------
+        X_pred : array, shape (n_pred, n_features)
+            The input data used for prediction using prior distribution..
```
juanitorduz, line 869: Remove last (repeated point)

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +847,80 @@ def plot_channel_contributions_grid(
...
+        extend_idata : Boolean determining whether the predictions should be added to inference data object.
+            Defaults to False.
+        combined: Combine chain and draw dims into sample. Won't work if a dim named sample already exists.
+            Defaults to True.
+        include_last_observations: Boolean determining whether to include the last observations of the training data.
```
juanitorduz, line 874: Maybe add a reason why this is relevant. I mean, a  comment that this is relevant for the adstock computation.

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +847,80 @@ def plot_channel_contributions_grid(
...
+        **kwargs: Additional arguments to pass to pymc.sample_posterior_predictive
+
+        Returns
+        -------
+        posterior_predictive_samples : DataArray, shape (n_pred, samples)
+            Posterior predictive samples for each input X_pred
+        """
+        if include_last_observations:
+            X_pred = pd.concat(
+                [self.X.iloc[-self.adstock_max_lag :, :], X_pred], axis=0
```
juanitorduz, line 887: Shall we sort by date just in case?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +847,80 @@ def plot_channel_contributions_grid(
...
+        return X
+
+    def sample_posterior_predictive(
+        self,
+        X_pred,
+        extend_idata,
+        combined,
+        include_last_observations: bool = False,
+        original_scale: bool = True,
+        **kwargs,
```
juanitorduz, line 861: Maybe call this `sample_posterior_predictive_kwargs` (or something more informative?) To indicate where these additional parameters will be used.

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +847,80 @@ def plot_channel_contributions_grid(
...
+                    posterior_predictive_samples,
+                )
+            else:
+                posterior_predictive_samples = xr.apply_ufunc(
+                    self.get_target_transformer().inverse_transform,
+                    posterior_predictive_samples.expand_dims(dim={"_": 1}, axis=1),
+                    input_core_dims=[["date", "_"]],
+                    output_core_dims=[["date", "_"]],
+                    vectorize=True,
+                ).squeeze(dim="_")
```
juanitorduz, line 922: I understand this because I wrote it but it might be confusing. So you could write a little helper function to wrap this logic and add a unit-test (just a suggestion)

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: Or as below a better description: "Additional arguments to pass to sample_posterior_predictive method"

juanitorduz (changes requested):
This looks super nice! Thanks! I just added some comments about the docs and other minor things.

juanitorduz:
@wd60622 After we merge this one, we need to add (or create) an example to the docs (let's create an issue).  This one will be a game-changer feature 🚀

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +847,80 @@ def plot_channel_contributions_grid(
...
+        -------
+        posterior_predictive_samples : DataArray, shape (n_pred, samples)
+            Posterior predictive samples for each input X_pred
+        """
+        if include_last_observations:
+            X_pred = pd.concat(
+                [self.X.iloc[-self.adstock_max_lag :, :], X_pred], axis=0
+            )
+
+        self._data_setter(X_pred)
```
ricardoV94, line 890: I wonder if we should reset the data changes after sampling

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -580,6 +609,104 @@ def test_model_config(
...
+    generate_data,
+    model_name: str,
+    new_dates: pd.DatetimeIndex,
+    request,
+) -> None:
+    mmm = request.getfixturevalue(model_name)
+    X_pred = generate_data(new_dates)
+
+    posterior_predictive_mean = mmm.predict(X_pred=X_pred)
+    assert posterior_predictive_mean.shape[0] == new_dates.size
```
ricardoV94, line 707: I would like to test something about the output values, not just their shapes

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -580,6 +609,104 @@ def test_model_config(
...
+    new_dates: pd.DatetimeIndex,
+    combined: bool,
+    original_scale: bool,
+    request,
+) -> None:
+    """This is the method that is used in all the other methods that generate predictions."""
+    mmm = request.getfixturevalue(model_name)
+    X_pred = generate_data(new_dates)
+
+    posterior_predictive = mmm.sample_posterior_predictive(
```
ricardoV94, line 647: What's the difference between the `predict` and `sample_posterior_predictive` MMM methods?

tests/model_builder/test_model_builder.py
```diff
@@ -271,3 +274,27 @@ def test_id():
...
+        "extend_idata": True,
+        "random_seed": rng,
+    }
+    first_sample = method(**kwargs)
+    second_sample = method(**kwargs)
+
+    assert first_sample != second_sample
+
+    sample = getattr(fitted_model_instance.idata, name)
+    assert sample == first_sample
```
ricardoV94, line 300: Shouldn't it override the first with the second when `extend_idata` is True? 

Also we should parametrize this test to test both `extend_idata` cases

ricardoV94: Also I think these changes also apply to `fit`?

tests/mmm/test_delayed_saturated_mmm.py
wd60622: predict is the mean of the sample_posterior_predictive by obs

wd60622:
Thank you guys for the review! I will have some time tomorrow to make some changes

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: done

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: done

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: Will do!

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: I'm hear this as setting the MutableData back to the X and y that it is fit on. i.e. self.X and self.y
Is that correct?
I think that would be good to preserve

tests/model_builder/test_model_builder.py
wd60622: This is the consequent from join="left"
I would think if you call the method again, you'd want to replace or be notified that it is being replaced. 

I think this is out of this PR's scope though\

ricardoV94: We should join right, otherwise extend_inferencedata=True makes little sense?

wd60622: That's what I meant to say. That's how I've change it. join="right"

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: done

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: Added a `_reset_model_data` method to set to the original and it is called after the sample_posterior_predictive call

tests/mmm/test_delayed_saturated_mmm.py
wd60622: added some checks based on the original scale and the original y for context

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: done. This isn't changed for the original model_builder methods however. Would you want that to change in this PR?

juanitorduz: Ok I think we can leave it as it is in the docstrings for now. Let's wait for feedback!

juanitorduz (approved):
This is amazing @wd60622 ! From my side, I just wanna push this out and get feedback. My sleep deprivation (👶 ) might make me miss some details so if @ricardoV94 is happy I  suggest we merge and release 0.4 🚀

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -580,13 +609,126 @@ def test_model_config(
...
+
+
 def test_get_valid_distribution(mmm):
     normal_dist = mmm._get_distribution({"dist": "Normal"})
     assert normal_dist is pm.Normal
 
 
 def test_get_invalid_distribution(mmm):
-    with pytest.raises(ValueError, match="does not exist in PyMC"):
+    with pytest.raises(ValueError, match="does not exist in P"):
```
ricardoV94, line 731: Why this change?

tests/model_builder/test_model_builder.py
ricardoV94: This test is still weird.

I expect the assert to be `assert sample == second_sample`.

The fact it is not failing tells me that assert is too premissive. I guess inferencedata equality is just based on groups or something. We need to check some actual values

ricardoV94 (changes requested):

pymc_marketing/mmm/delayed_saturated_mmm.py
ricardoV94: On a second thought I see that all sampling methods need X as input, so I think we should revert this change. Otherwise we should reset in all the other methods. Just in this one doesn't make sense.

Apologies, I find the ModelBuilder API pretty confusing.

pymc_marketing/model_builder.py
```diff
@@ -582,7 +582,7 @@ def sample_prior_predictive(
                 self.set_idata_attrs(prior_pred)
                 if extend_idata:
                     if self.idata is not None:
-                        self.idata.extend(prior_pred)
+                        self.idata.extend(prior_pred, join="right")
```
wd60622, line 585: @ricardoV94 the join kwarg was added. Maybe you are looking at old commit

pymc_marketing/model_builder.py
```diff
@@ -616,7 +616,7 @@ def sample_posterior_predictive(self, X_pred, extend_idata, combined, **kwargs):
         with self.model:  # sample with new input data
             post_pred = pm.sample_posterior_predictive(self.idata, **kwargs)
             if extend_idata:
-                self.idata.extend(post_pred)
+                self.idata.extend(post_pred, join="right")
```
wd60622, line 619: Same here. Maybe I missed a spot, so let me know

tests/mmm/test_delayed_saturated_mmm.py
wd60622: Not sure. I will change back

tests/model_builder/test_model_builder.py
wd60622: Okay. I will take a look later

pymc_marketing/model_builder.py
ricardoV94: I saw those. What are you referring to? I mentioned `fit` will still override the previous result, which is not something that needs to be fixed in this PR.

The test has the wrong assert because it's checking that the idata matches the first call (not the second). The test is also insufficient, because it doesn't raise (I think you didn't change it when you switched from "left" to "right"

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -814,3 +853,73 @@ def plot_channel_contributions_grid(
...
         )
         return fig
+
+    def _validate_data(self, X, y=None):
+        return X
+
+    def sample_posterior_predictive(
+        self,
+        X_pred,
+        extend_idata,
```
ricardoV94, line 863: Default missing? Docstrings say it defaults to False, All other methods seem to have a default.

I am not sure what the default should be, probably True?

wd60622: I've changed all but the sample_prior_predictive (from model_builder) to default to True. Changed in the docstring as well

pymc_marketing/model_builder.py
wd60622: I modified the tests in order to check for the second sample stays. Please take a look

wd60622:
Hi @ricardoV94. Thanks for the feedback. I just went through all of it and made some changes. 

For the xarray tests, I changed to used check for the second value and am using xr.testing module now for more robust checks.

juanitorduz:
LGTM 🚀 ! Let's wait for @ricardoV94 's feedback :)

ricardoV94:
@wd60622 everything looks kosher, but there is still this open discussion: https://github.com/pymc-labs/pymc-marketing/pull/482#discussion_r1457076712

Given how all sampling/fitting methods require x/y anyway I changed my mind and I don't think the `_reset_data` call is needed after all? The less magic probably the better

wd60622:
> @wd60622 everything looks kosher, but there is still this open discussion: [#482 (comment)](https://github.com/pymc-labs/pymc-marketing/pull/482#discussion_r1457076712)
> 
> Given how all sampling/fitting methods require x/y anyway I changed my mind and I don't think the `_reset_data` call is needed after all? The less magic probably the better|

Perfect. I've changed that back so that the data will not be reset

tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -580,6 +609,122 @@ def test_model_config(
...
+        extend_idata=False,
+        combined=combined,
+        original_scale=original_scale,
+    )
+    pd.testing.assert_index_equal(
+        pd.DatetimeIndex(posterior_predictive.coords["date"]),
+        new_dates,
+    )
+
+    # Data has not been set back
```
ricardoV94, line 659: This check should now also be removed. Having a test suggests it's a feature when it's just what it is. This could be confusing for someone refactoring the code in the future because it suggests it's something that needs to be satisfied.

juanitorduz:
Are we ready to merge? 🤞

juanitorduz:
It's green! 💚

ricardoV94 (approved):

ricardoV94:
Merged, any notebooks that need updating?

juanitorduz:
> Merged, any notebooks that need updating?

I ran the notebook in my first review and it worked perfectly. I can check tomorrow (@wd60622 do we need an update?).

wd60622:
Firstly, thank you both for the reviews! 

> > Merged, any notebooks that need updating?
> 
> I ran the notebook in my first review and it worked perfectly. I can check tomorrow (@wd60622 do we need an update?).

I haven't touched the notebooks. I would like to add an example in docstring / notebook with https://github.com/pymc-labs/pymc-marketing/issues/494
I will check get to that on the weekend

PR #503, opened 2024-01-27
Add scale preserving saturation function

ulfaslak:
## Scale preserving saturation function (with max slope 1)

Saturation function with **single interpretable** parameter and **nice properties**:

$$ f(x) = m \left( \frac{1 - e^{-\frac{2}{m} x}}{1 + e^{-\frac{2}{m} x}} \right) $$

764b9192-bfc6-4c31-91e8-72e8d8a732e4

The function maps $x$ into the range $\[0, m\]$, where $m$ is the asymptote (function value in $x=\inf$). As such it can be interpreted as mapping from *actual spend* to *effective spend*. Since the slope never exceeds one, it is guaranteed that effective is always less than actual spend—which resonates well with our understanding of how spend saturation works in the real world.

### Benefits
* Single easy-to-interpret parameter.
* Since *effective spend* is on the same (or very similar) scale as *spend*, it is **not necessary to transform** coefficient priors. They can simply be the expected channel ROI in units of `[return] / [investment]`.
* Unsaturated case supported, by letting $m \gg \max(\text{spend}_{\text{channel}})$ in which case $f(x) ≈ x$.

### Origins

I don't know if this is a common function used somewhere. Maybe someone reviewing this has seen it before? GPT4 could not suggest a source. I came up with it by trial and error.

## Description
PR adds a new spend saturation function, as well as tests. The saturation function has two desirable properties:
* Slope never exceeds 1
* `m` parameter is the asymptote

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

- [X] Included tests that prove the fix is effective or that the new feature works

## Modules affected

- [X] MMM
- [ ] CLV

## Type of change

- [X] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--503.org.readthedocs.build/en/503/

### DIFF
--- a/tests/mmm/test_transformers.py
+++ b/tests/mmm/test_transformers.py
@@ -7,11 +7,11 @@ from pytensor.tensor.variable import TensorVariable
 from pymc_marketing.mmm.transformers import (
     ConvMode,
     TanhSaturationParameters,
+    asymptotic_logistic_saturation,
     batched_convolution,
     delayed_adstock,
     geometric_adstock,
     logistic_saturation,
-    asymptotic_logistic_saturation,
     tanh_saturation,
     tanh_saturation_baselined,
 )
@@ -220,7 +220,9 @@ class TestSaturationTransformers:
         # When lam == max(x), f(max(x)) is max(x) * 0.76
         x = np.ones(shape=(100))
         y = asymptotic_logistic_saturation(x=x, lam=1)
-        np.testing.assert_array_almost_equal(x=np.ones(shape=(100)) * 0.761594, y=y.eval(), decimal=3)
+        np.testing.assert_array_almost_equal(
+            x=np.ones(shape=(100)) * 0.761594, y=y.eval(), decimal=3
+        )
 
     @pytest.mark.parametrize(
         "x",
--- a/tests/mmm/test_transformers.py
+++ b/tests/mmm/test_transformers.py
@@ -11,6 +11,7 @@ from pymc_marketing.mmm.transformers import (
     delayed_adstock,
     geometric_adstock,
     logistic_saturation,
+    asymptotic_logistic_saturation,
     tanh_saturation,
     tanh_saturation_baselined,
 )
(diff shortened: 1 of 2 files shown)


juanitorduz (changes requested):
This is great! I wanna replace this saturation function in our default MMM model now hehe! 

Still, before we merge I think you need to tweak the docstrings as the documentation is not rendering correctly https://pymc-marketing--503.org.readthedocs.build/en/503/api/generated/pymc_marketing.mmm.transformers.asymptotic_logistic_saturation.html#pymc_marketing.mmm.transformers.asymptotic_logistic_saturation

ulfaslak:
> This is great! I wanna replace this saturation function in our default MMM model now hehe!
> 
> Still, before we merge I think you need to tweak the docstrings as the documentation is not rendering correctly https://pymc-marketing--503.org.readthedocs.build/en/503/api/generated/pymc_marketing.mmm.transformers.asymptotic_logistic_saturation.html#pymc_marketing.mmm.transformers.asymptotic_logistic_saturation

Nice, thanks for catching that!

I'm also gonna change $\lamda$ to $L$, since $L$ is a more common symbol for asymptotes.

ulfaslak:
And also I REALLY want a better name than `asymptotic_logistic_saturation` 😅. Anyone?

juanitorduz:
What about “ell” instead so that we do not use upper case for function arguments?

ulfaslak:
Hmmmm... `ell` seems confusing to me. Maybe lower case `l`? Does the linter not like this, or?

ulfaslak:
Also I don't know what's wrong with the docstring 🥲. Can you spot the error?

ulfaslak:
OK changed it to `m` now, which is seen in literature for logistic functions and sounds a little like "max" (which is what it is) :).

ulfaslak:
How do you rebuild the docs? I think it works now, but I cant tell.

ulfaslak:
OK I think the docs work now...

tests/mmm/test_transformers.py
```diff
@@ -209,6 +210,56 @@ def test_logistic_saturation_min_max_value(self, x, lam):
...
+            (np.linspace(start=0.0, stop=1.0, num=50), 90),
+            (np.linspace(start=200, stop=1000, num=50), 17),
+            (np.zeros(shape=(100)), 200),
+        ],
+    )
+    def test_scale_preserving_logistic_saturation_bounds(self, x, m):
+        # Check that the values are within the range [0, m]
+        y = scale_preserving_logistic_saturation(x=x, m=m)
+        assert y.eval().max() <= m
+        assert y.eval().min() >= 0
```
ulfaslak, line 253: Important test: shows that bounding holds

tests/mmm/test_transformers.py
```diff
@@ -209,6 +210,56 @@ def test_logistic_saturation_min_max_value(self, x, lam):
...
+
+    @pytest.mark.parametrize("m", [10, 100, 1000])
+    def test_scale_preserving_logistic_saturation_slope(self, m):
+        # Check that slope < 1
+        x = np.linspace(0, 10, 100)
+        y = scale_preserving_logistic_saturation(x, m=m).eval()
+        dy_dx = np.diff(y) / np.diff(x)
+        assert np.all(dy_dx <= 1)
```
ulfaslak, line 261: Important test: shows that slope always <= 1

tests/mmm/test_transformers.py
```diff
@@ -209,6 +210,56 @@ def test_logistic_saturation_min_max_value(self, x, lam):
...
         assert y_eval.min() >= 0
 
+    def test_scale_preserving_logistic_saturation_m_zero(self):
+        # All values should be zero
+        x = np.ones(shape=(100))
+        y = scale_preserving_logistic_saturation(x=x, m=0.0)
+        np.testing.assert_array_almost_equal(x=np.zeros(shape=(100)), y=y.eval())
+
+    def test_scale_preserving_logistic_saturation_m_max_x(self):
+        # When m == max(x), f(max(x)) is max(x) * 0.76
```
juanitorduz, line 220: does this 0.76 comes from any known close form or analytical expression?

juanitorduz (approved):
Nice! I just left one comment that is more of a question about a constant. If there is an easy explanation I suggest adding it to the test comments. If not is ok. I also checked the docs and they look great!

Regarding the naming, I think the more explicit the better so I like this one 😅 ! If others have suggestions we can rename it later :)

juanitorduz:
ok! Let's merge this one and iterate 💪 . Thank you @ulfaslak

tests/mmm/test_transformers.py
ulfaslak: It absolutely does, I just didn't do the math yet. I could if it matters :).

juanitorduz: just a comment will be enough ... at least hinting the computation

ulfaslak: added a small comment explaining the math

PR #463, opened 2023-12-13
Add Covariates to `ParetoNBDModel`

ColtAllen:
## Description

This PR provides the option of using time-invariant covariates with `ParetoNBDModel`. This is a draft PR for now because I still need to parametrize some tests and add a few other things, but more importantly get opinions on two points:

Covariates require scaling for model convergence, preferably with `MinMaxScaler` so the coefficients are interpretable. This can easily be added as a utility function, and the covariate Min/Max values saved as `model.idata` attributes for out-of-sample predictions. **Should a scaler be added in this PR, or a future one? Should it be added to `ParetoNBDModel` , or the base `CLVModel`?**

Covariates for existing customers will also require modifications to the "new customer" methods. There are three options for this:

1. Just raise a `ValueError` indicating these methods aren't supported with covariates (easy, but not my preference.)
2. Average all the customer-specific `alpha` and `beta` parameters together for new customer estimates, which is a subset of the next option. 
3. Add arguments to these methods for user-specified covariate values. This will be involved and maybe even clunky to use in practice, but if a user wants to do sensitivity analysis on how covariates can impact new customer estimates, this is the way to go.

**Regardless of the approach, should it added to this PR, or another one? This one is already rather lengthy.**

## Related Issue

- [ ] Closes #
- [x] Related to # https://github.com/pymc-labs/pymc-marketing/issues/183, https://github.com/pymc-labs/pymc-marketing/issues/134

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [x] Added necessary documentation (docstrings and/or example notebooks)
- [x] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Modules affected

- [ ] MMM
- [x] CLV

## Type of change

- [x] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--463.org.readthedocs.build/en/463/

PR #506, opened 2024-01-29
`CLV` Naming Conventions and Test Data

ColtAllen:
## Description

This PR renames two methods in `BetaGeoModel` so that it is consistent with `ParetoNBDModel`. This is required for both models to utilize the same customer lifetime value estimations and plotting functions.

`mean_transaction_value` in `GammaGammaModel` has also been renamed to `monetary_value` so that it is consistent with the rest of the module, with docstrings updated to highlight this is in fact an average value.

While updating the tests for `mean_transaction_value`, I noticed `test_utils.py` uses a test dataset that does not contain a monetary value column 😅 I've deleted this test dataset and pointed all pertinent tests to the CLV Quickstart datafile instead.

## Checklist

## Modules affected

- [ ] MMM
- [x] CLV

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--506.org.readthedocs.build/en/506/


tests/clv/test_utils.py
```diff
@@ -136,79 +122,31 @@ def transaction_data() -> pd.DataFrame:
...
-        self, test_summary_data, fitted_bg
+        self, test_summary_data, fitted_bg, t, discount_rate, expected
     ):
-        # Test borrowed from
-        # https://github.com/CamDavidsonPilon/lifetimes/blob/aae339c5437ec31717309ba0ec394427e19753c4/tests/test_utils.py#L527
-
-        t = test_summary_data.head()
+        data = test_summary_data.head(3)
 
-        expected = np.array([0.016053, 0.021171, 0.030461, 0.031686, 0.001607])
```
ricardoV94, line 147: I insist on keeping the original test with the original values. This is a good independent source that was being used in a third-library.

We can hardcode the old `test_summary_data.head()` if the point was to remove the utility dataset

ricardoV94 (changes requested):
I like the name `transaction_value` better because in theory CLV model don't have to be about money only. I am okay with changing it though if you feel compelled enough.

If we change the name of the methods we should add a deprecation warning for the old names so that code will not mysteriously break when users update pymc_marketing.

```python
def expected_num_purchases(self, *args, **kwargs):
    warnings.warn("Method was renamed to x. This will be removed in a future release)
    self.expected_purchases(*args, **kwargs)

File: tests/clv/test_utils.py (hunk 1 above)
Comment (ColtAllen) [reply tests/clv/test_utils.py:147]: I hardcoded the old test data, but am keeping the pytest parametrization to reduce code repetition.

File: pymc_marketing/clv/models/gamma_gamma.py [hunk 2]
```diff
@@ -174,7 +174,7 @@ class GammaGammaModel(BaseGammaGammaModel):
     data: pd.DataFrame
         DataFrame containing the following columns:
             - customer_id: Customer labels. Must not repeat.
-            - mean_transaction_value: Mean transaction value of each customer.
+            - monetary_value: Average transaction value for each customer.
```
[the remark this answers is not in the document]
ColtAllen: I changed it to `monetary_value` so that results from the `rfm_summary` utility can be used without renaming anything. However, `GammaGammaModelIndividual` requires raw transaction values instead, so I think it's important to keep the distinction.

ColtAllen:
A random `MMM` test is failing in the CI:

`tests/mmm/test_delayed_saturated_mmm.py:716: AssertionError`

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -250,9 +250,9 @@ def __init__(
         try:
             self.mean_transaction_value: Union[
                 np.ndarray, pd.Series, TensorVariable
-            ] = data["mean_transaction_value"]
+            ] = data["monetary_value"]
```
ricardoV94, line 253: Add warning for old name but still accept

pymc_marketing/clv/models/gamma_gamma.py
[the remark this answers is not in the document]
ricardoV94: > However, GammaGammaModelIndividual requires raw transaction values instead, so I think it's important to keep the distinction.

Why call the individual transactions values, but the aggregated monetary value? It's the same quantity, but one of them is the average of the other. I don't see why it should have a different label

juanitorduz:
I re-ran the jobs and it's ✅ now

ricardoV94:
> I re-ran the jobs and it's ✅ now

But we probably have a flaky MMM test, it's good to open an issue linking to the fail CI log when that happens

juanitorduz:
You are right! I'll open an issue (I remember the test name 😅)

pymc_marketing/clv/models/gamma_gamma.py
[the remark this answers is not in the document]
ColtAllen: Monetary Value is the term used in research and every educational article/webinar about CLV, so I'm loathe to change it. `GammaGammaIndividual` is also rather niche; majority of users will be doing population modeling.

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -319,7 +327,7 @@ def expected_probability_alive(
             "chain", "draw", "customer_id", missing_dims="ignore"
         )
 
-    def expected_num_purchases_new_customer(
+    def expected_purchases_new_customer(
```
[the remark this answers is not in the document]
ColtAllen: Whoops, forgot about that one. Warning has been added.

pymc_marketing/clv/models/gamma_gamma.py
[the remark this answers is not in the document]
ricardoV94: Fine but then change the Gamma individual name to sound more like monetary value

ColtAllen:
What I thought was a mere name change has exploded into refactoring half the `CLV` module. 😅 `BetaGeoModel` and `ParetoNBDModel` previously had different APIs, but are now standardized. The BG/NBD tests have also been cleaned up quite a bit. All that's left is to resolve the parametrization for `customer_lifetime_value`.

ColtAllen:
@ricardoV94 I'm getting some unusual errors on three of the tests. Descriptions are in the TODOs I just pushed and am highlighting for review.

ColtAllen:
All tests passing now! Turns out I couldn't force the `customer_lifetime_value` utility into the dataframe API used by the transaction models, so a param per array it is. That utility won't be called directly unless someone wants to use their own value model, so I'm not as concerned with API consistency there.

Also added CLV tests for `ParetoNBDModel`, so this is ready for review (again). FWIW there's a DIV/0 runtime warning with `test_clv_after_thinning`.

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -119,14 +117,13 @@ def expected_new_customer_spend(self) -> xarray.DataArray:
 
         return mean_spend
 
+    # TODO: Why isn't this in the child GammaGammaModel class?
```
ricardoV94, line 120: So it can be used by both child classes?

pymc_marketing/clv/utils.py
```diff
@@ -91,35 +91,40 @@ def _squeeze_dims(x: xarray.DataArray):
...
-        transaction_model.expected_num_purchases(
-            customer_id=customer_id,
-            frequency=frequency,
-            recency=recency,
-            T=T,
-            t=0,
+    # TODO: if data attribute added to transaction model,
+    #       transaction_data can be an optional arg for out-of-sample data
+    # TODO: Set these as transaction_model attributes rather than create a dataframe?
+    transaction_data = pd.DataFrame(
```
ricardoV94, line 102: This should probably be using xarray tools not pandas, since the predictive methods work with xarray as well

ricardoV94 (commented):
This PR seems to he suffering a bit from scope creep. Can we split it into separate independent PRs for faster review and merging?

1. Remove the test dataset
2. Renamimg of methods and data columns, which shouldn't error out in the next release if a user uses the old ones. You didn't add a fallback for old pandas column names.
3. The minimum changes needed to make the clv computation actually work with the ParetoModel. I would like to not touch old tests at all. You can make the new tests more generic so they work with future transaction models but we don't need to touch the old ones. That makes reviewing easier because I'll know nothing that worked before was broken

ColtAllen:
> 3. Expecting DataFrames for methods that used to take separate arrays. This may need some discussion, I am not sure this is an improvement rather than a regression because we loose xarray handy broadcasting behavior.

It's a convenience for end users because passing in individual arrays was one of the most annoying aspects of `lifetimes`. I'll need more clarification around xarray broadcasting, but either way this is more relevant to https://github.com/pymc-labs/pymc-marketing/pull/463/.

> 4. The minimum changes needed to make the clv computation actually work with the ParetoModel.

I think for now we can add an IF block for `hasattr(transaction_model,"expected_puchases")`
 to  the `customer_lifetime_value` utility. This should allow `ParetoNBDModel` support without requiring naming or API changes elsewhere.

pymc_marketing/clv/models/gamma_gamma.py
ColtAllen: Just noticed `GammaGammaIndividual` has an extra `_summarize_mean_data` step that must be called, so it makes sense now.

pymc_marketing/clv/utils.py
ColtAllen: This will need to be included in the IF block mentioned in the comment below, but may require updating the type hints in the transaction models to keep `mypy` happy.

I'll add a TODO for this because I'm not a fan of creating a pandas dataframe here, but don't want to risk a merge conflict with https://github.com/pymc-labs/pymc-marketing/pull/463 either.

ColtAllen:
Closing this because there is half a dozen PRs worth of work here 🤣

PR #515, opened 2024-02-06
Reuse fixtures in slow MMM tests

### LINKED ISSUE #514: `test_new_data_sample_posterior_predictive_method` takes 50 minutes (70%) of CI time
I counted 35 parametrizations, each of which takes roughly 90 seconds long. This accounts for 50 minutes or 70% of the whole CI in this run:

https://github.com/pymc-labs/pymc-marketing/actions/runs/7771863902/job/21193663231?pr=506#step:5:872

wd60622:
## Description

Modify the scope of the fixture from class to module in order to reuse on the out of sample tests

## Related Issue

- [x] Closes #514 
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--515.org.readthedocs.build/en/515/

### DIFF
--- a/tests/mmm/test_delayed_saturated_mmm.py
+++ b/tests/mmm/test_delayed_saturated_mmm.py
@@ -17,7 +17,7 @@ seed: int = sum(map(ord, "pymc_marketing"))
 rng: np.random.Generator = np.random.default_rng(seed=seed)
 
 
-@pytest.fixture(scope="class")
+@pytest.fixture(scope="module")
 def generate_data():
     def _generate_data(date_data: pd.DatetimeIndex) -> pd.DataFrame:
         n: int = date_data.size
@@ -37,7 +37,7 @@ def generate_data():
     return _generate_data
 
 
-@pytest.fixture(scope="class")
+@pytest.fixture(scope="module")
 def toy_X(generate_data) -> pd.DataFrame:
     date_data: pd.DatetimeIndex = pd.date_range(
         start="2019-06-01", end="2021-12-31", freq="W-MON"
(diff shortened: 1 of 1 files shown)


tests/mmm/test_delayed_saturated_mmm.py
```diff
@@ -106,15 +106,15 @@ def mmm_with_fourier_features() -> DelayedSaturatedMMM:
     )
 
 
-@pytest.fixture(scope="class")
+@pytest.fixture(scope="module")
 def mmm_fitted(
     mmm: DelayedSaturatedMMM, toy_X: pd.DataFrame, toy_y: pd.Series
 ) -> DelayedSaturatedMMM:
     mmm.fit(X=toy_X, y=toy_y, target_accept=0.8, draws=3, chains=2)
```
ricardoV94, line 113: Instead of this nearly useless fit, you can take draws from the prior predictive and store those as if they were the posterior. We do this in almost all CLV tests.

There should be one test that actually calls fit and asserts it is converging to something correct, that can be marked with @pytest.mark.slow so it is skipped by default locally.

wd60622:
Tests seem to be a lot faster! So step in the right direction 😄  
Thanks for pointing out @ricardoV94

tests/mmm/test_delayed_saturated_mmm.py
ricardoV94: I dunno. We can simulate data with known parameters and check the posterior converges to those with some relative tolerance.

wd60622: Cool. I will add that in

Some of the tests still call fit with draws=100
Is that an issue?

wd60622:
some other tests that depend on the fit... Will have to change a bit more than expected
Will work through them sometime soon

tests/mmm/test_delayed_saturated_mmm.py
ricardoV94: Depends on whether that is relevant for the test or not

ricardoV94:
So let's merge the easy wins with the scope and optimize in a separate PR?

wd60622:
Going back one commit for the easy win solution. Will create an issue for follow up

ricardoV94 (approved):

juanitorduz:
Thanks @wd60622 !

PR #528, opened 2024-02-12
Replaceable steps and ordering of channel contribution forward pass

wd60622:
## Description

This is an attempt to generalize the forward pass of the mmm in order to make the steps more flexible and open to: 
- different adstocks and saturations functions
- change of order of steps (saw this in a few discussions)

The implementation is done so by adding in:
- `forward_pass` method which is supports general forward pass
- `forward_pass_params` : names of the variable required for the forward pass. All of these have to be keys in the current `model_config` property of the model
- `forward_pass_steps` : the order to the steps defined by callables

Each forward_pass_steps are all callables that take two arguments. **Having the same argument signature is key to generalize the steps!**

I've defined a few based on the transformation functions, but they can be defined by the user as well

```python
def forward_pass_step(x: TensorVariable, params: dict[str, TensorVariable]) -> TensorVariable:
    """Forward pass step
    
    Args: 
        x: data from first or previous step of forward_pass_step
        params: dict[str, TensorVariable]

    Returns: 
        TensorVariable
            Result data to be used in the next step
    """
````

The default behavior is what was available before but can now be extended

# Examples

## Change the order of steps

The default ordering is adstock, saturation, multiply (by beta for each channel). However, that can be changed now.

```python
adstock, saturation, multiply = mmm.forward_pass_steps
mmm.forward_pass_steps = [saturation, adstock, multiply]
```

For a more verbose way of doing this, we can redefine the steps: 

```python
from pymc_marketing.mmm.forward_pass import (
    create_geometric_adstock_handler, 
    logistic_saturation_handler, 
    create_multiply_handler, 
)

# All of these are functions with the signature (x, params)
# adstock and multiply are functions that return functions with that signature
saturation = logistic_saturation_handler
adstock = create_geometric_adstock_handler(l_max=adstock_max_lag, normalize=True, axis=0)
multiply = create_multiply_handler("beta_channel")

mmm = ...
mmm.forward_pass_steps = [
    # This is usually second step!
    saturation, 
    adstock, 
    multiply, 
]
mmm.fit(X, y)
```

## Change the adstock function

1. Create a handler that has the signature defined above (x, params)
2. Make sure that each variable forward_pass_params is in the model_config. If not define
3. Override the forward_pass_steps do the desired

This might look like this

```python
from pymc_marketing.mmm.forward_pass import scale_preserving_saturation_handler

saturation = scale_preserving_saturation_handler
mmm.forward_pass_steps[1] = saturation

mmm.forward_pass_params = [
    # For the default geometric_adstock
    "lam"
    # For the replaced saturation
    "m", 
    # For the default scaling at end
    "beta_channel", 
]
# Same format as before. (Just an example)
mmm.model_config["m"] = {"dist": "HalfNormal", "kwargs": {"sigma": 2.5}}
```
## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc_marketing/mmm/delayed_saturated_mmm.py
+++ b/pymc_marketing/mmm/delayed_saturated_mmm.py
@@ -13,6 +13,13 @@ from pytensor.tensor import TensorVariable
 from xarray import DataArray
 
 from pymc_marketing.mmm.base import MMM
+from pymc_marketing.mmm.forward_pass import (
+    apply_forward_pass,
+    create_geometric_adstock_handler,
+    create_multiply_handler,
+    logistic_saturation_handler,
+    wrap_as_deterministic,
+)
 from pymc_marketing.mmm.preprocessing import MaxAbsScaleChannels, MaxAbsScaleTarget
 from pymc_marketing.mmm.transformers import geometric_adstock, logistic_saturation
 from pymc_marketing.mmm.utils import (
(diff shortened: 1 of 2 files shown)


juanitorduz:
@wd60622 I think this is a great idea, and I like the approach 😎 ! One thing to consider is that we would need to change the plotting functions as well, as I think we are using the parameter names explicitly. 

**Question about generalization:** What about allowing each channel to have its own steps: for example, TV can have saturation -> adstock while paid social adstock -> saturation. I do not think it is super urgent, but maybe it is something we want to allow in the future. This is me just thinking out loud :)

wd60622:
Thanks you for taking a look, @juanitorduz !
Yes, definitely not done yet. So in draft mode 

> One thing to consider is that we would need to change the plotting functions as well, as I think we are using the parameter names explicitly.

Definitely something that I've thought about. I think the good thing is that the plots rely on the final "channel_contributions" rather than the intermediate steps. Please correct me if I am wrong. For instance, the ["channel_adstock" deterministic](https://github.com/search?q=repo%3Apymc-labs%2Fpymc-marketing%20channel_adstock&type=code)

I view this `forward_pass` method to be used for those plotting methods as well. For instance, the [`channel_contributions_forward_pass`](https://github.com/pymc-labs/pymc-marketing/blob/a224cfe7da3fd323579b0d1c16465f305763b454/pymc_marketing/mmm/delayed_saturated_mmm.py#L501-L522) depends on this logic being hard coded. However, this can be reduced to below because it only relies on the final result.

Luckily, having the forward pass defined makes it easy to use for various / new posterior predictive use cases 

```python
with pm.Model(...): 
    channel_contributions = self.forward_pass(channel_data)
    channel_contributions_samples = pm.sample_posterior_predictive(self.fit_results)
    # Scale contributions back to original domain
```

The methods that I know will require attention are: 

- Above mentioned [`channel_contributions_forward_pass`](https://github.com/pymc-labs/pymc-marketing/blob/a224cfe7da3fd323579b0d1c16465f305763b454/pymc_marketing/mmm/delayed_saturated_mmm.py#L483)
- The [`new_spend_contributions` method](https://github.com/wd60622/pymc-marketing/blob/6aa264e9f06338577176828327c7053bdaf4326b/pymc_marketing/mmm/delayed_saturated_mmm.py#L948-L960)


> **Question about generalization:** What about allowing each channel to have its own steps: for example, TV can have saturation -> adstock while paid social adstock -> saturation. I do not think it is super urgent, but maybe it is something we want to allow in the future. This is me just thinking out loud :)

Might be a lot to start with 😉 But I like the idea. I think that this would require changes to the model_config.
There are definitely some people that are looking for different priors on the channels. Maybe make another issue and open to the larger round for discussion 😄 


Definitely a lot of existing to change in order to support the generalization that happens here. But hopefully it'd be worth it and hit some of the flexibility that we've been working towards

wd60622:
Closing in favor of https://github.com/pymc-labs/pymc-marketing/pull/602

PR #545, opened 2024-02-22
Implement ParetoNBD with covariates

ricardoV94:
## Related Issue

- [x] Supersedes and closes #463
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [x] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--545.org.readthedocs.build/en/545/

### DIFF
--- a/pymc_marketing/clv/models/pareto_nbd.py
+++ b/pymc_marketing/clv/models/pareto_nbd.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import Any, Dict, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, Literal, Optional, Sequence, Union, cast
 
 import numpy as np
 import pandas as pd
@@ -51,16 +51,16 @@ pytensor.compile.optdb["specialize"].register(
 
 
 class ParetoNBDModel(CLVModel):
-    """Pareto Negative Binomial Distribution (Pareto/NBD) model for continuous, non-contractual customer populations,
-    first introduced by Schmittlein, et al. [1]_, with additional derivations and predictive methods by
-    Hardie & Fader [2]_ [3]_ [4]_.
+    """Pareto Negative Binomial Model (Pareto/NBD).
 
-    The Pareto/NBD model assumes churn times for the customer population (i.e., amount of time a customer is active)
-    follows a Gamma distribution,
+    Model for continuous, non-contractual customers, first introduced by Schmittlein, et al. [1]_,
+    with additional derivations and predictive methods by Hardie & Fader [2]_ [3]_ [4]_.
+
+    The Pareto/NBD model assumes the time duration a customer is active follows a Gamma distribution,
     and time between purchases is also Gamma-distributed while the customer is still active.
 
     This model requires data to be summarized by recency, frequency, and T for each customer,
-    using `clv.rfm_summary()` or equivalent.
+    using `clv.rfm_summary()` or equivalent. Covariates impacting customer dropouts and transaction rates are optional.
 
     Parameters
     ----------
@@ -70,12 +70,19 @@ class ParetoNBDModel(CLVModel):
             * `recency`: time between the first and the last purchase
             * `T`: time between the first purchase and the end of the observation period; model assumptions require T >= recency
             * `customer_id`: unique customer identifier
+        Along with optional covariate columns.
+
     model_config: dict, optional
-        Dictionary containing model parameters:
-            * `r_prior`: Shape parameter of time between purchases for customer population; defaults to `pymc.Weibull.dist(alpha=2, beta=1)`
-            * `alpha_prior`: Scale parameter of time between purchases for customer population; defaults to `pymc.Weibull.dist(alpha=2, beta=10)`
-            * `s_prior`: Shape parameter of time until churn for customer population; defaults to `pymc.Weibull.dist(alpha=2, beta=1)`
-            * `beta_prior`: Scale parameter of time until churn for customer population; defaults to `pymc.Weibull.dist(alpha=2, beta=10)`
+        Dictionary containing model parameters and covariate column names:
+            * `r_prior`: Shape parameter of time between purchases; defaults to `Weibull(alpha=2, beta=1)`
+            * `alpha_prior`: Scale parameter of time between purchases; defaults to `Weibull(alpha=2, beta=10)`
+            * `s_prior`: Shape parameter of time until dropout; defaults to `Weibull(alpha=2, beta=1)`
+            * `beta_prior`: Scale parameter of time until dropout; defaults to `Weibull(alpha=2, beta=10)`
+            * `purchase_covariates_prior`: Coefficients for purchase rate covariates; defaults to `Normal(0, 3)`
+            * `dropout_covariates_prior`: Coefficients for dropout covariates; defaults to `Normal.dist(0, 3)`
+            * `purchase_covariate_cols`: List containing column names of covariates for customer purchase rates.
+            * `dropout_covariate_cols:`: List containing column names of covariates for customer dropouts.
+
         If not provided, the model will use default priors specified in the `default_model_config` class attribute.
     sampler_config: dict, optional
         Dictionary of sampler parameters. Defaults to None.
@@ -93,10 +100,10 @@ class ParetoNBDModel(CLVModel):
             model = ParetoNBDModel(
                 data=rfm_df,
                 model_config={
-                    "r_prior": pm.Weibull.dist(alpha=2,beta=1),
-                    "alpha_prior": pm.Weibull.dist(alpha=2,beta=10),
-                    "s_prior": pm.Weibull.dist(alpha=2,beta=1),
-                    "beta_prior": pm.Weibull.dist(alpha=2,beta=10),
+                    "r_prior": {"dist": "Weibull", "kwargs": {"alpha": 2, "beta": 1}},
+                    "alpha_prior": {"dist": "Weibull", "kwargs": {"alpha": 2, "beta": 10}},
+                    "s_prior": {"dist": "Weibull", "kwargs": {"alpha": 2, "beta": 1}},
+                    "beta_prior": {"dist": "Weibull", "kwargs": {"alpha": 2, "beta": 10}},
                 },
             )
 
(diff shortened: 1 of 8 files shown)


ColtAllen:
I think an updated UML diagram would be a helpful addition to CONTRIBUTING.MD to better understand all these new class inheritances:

image

 Here are some instructions on how to create one:

https://github.com/pymc-labs/pymc-marketing/issues/178

pymc_marketing/clv/models/pareto_nbd.py
```diff
@@ -750,3 +561,491 @@ def distribution_customer_population(
...
+        List containing column names of covariates for customer purchase rates.
+    dropout_covariate_cols: list, optional
+        List containing column names of covariates for customer dropouts.
+    model_config: dict, optional
+        Dictionary containing model parameters:
+            * `r_prior`: Shape parameter of time between purchases; defaults to `pymc.Weibull.dist(alpha=2, beta=1)`
+            * `alpha_prior`: Scale parameter of time between purchases; defaults to `pymc.Weibull.dist(alpha=2, beta=10)`
+            * `s_prior`: Shape parameter of time until dropout; defaults to `pymc.Weibull.dist(alpha=2, beta=1)`
+            * `beta_prior`: Scale parameter of time until dropout; defaults to `pymc.Weibull.dist(alpha=2, beta=10)`
+            * `purchase_covariates_prior`: Coefficients for purchase rate covariates; defaults to `pymc.Normal.dist(mu=0, sigma=1, n=len(purchase_covariates_cols)`
```
ColtAllen, line 743: Why the change from `StudentT` for default priors?

pymc_marketing/clv/models/pareto_nbd.py
```diff
@@ -211,59 +215,152 @@ def default_model_config(self) -> Dict[str, Dict]:
             "alpha_prior": {"dist": "Weibull", "kwargs": {"alpha": 2, "beta": 10}},
             "s_prior": {"dist": "Weibull", "kwargs": {"alpha": 2, "beta": 1}},
             "beta_prior": {"dist": "Weibull", "kwargs": {"alpha": 2, "beta": 10}},
+            # TODO: Cauchy prior is a bit too extreme? Probably not needed if parameters are standardized
+            "dropout_covariates_prior": {
+                "dist": "StudentT",
+                "kwargs": {"nu": 1},
```
BBDS-Colt, line 221: Did you say `"kwargs": {"nu": 2},` is needed for data to be centered at 0?  For these covariate coefficients to be interpretable, covariates need to be scaled between 0 and 1, with a prior supporting negative values. `StudentT` slows down model fitting considerably, so I'm open to alternatives provided the distribution is centered.

pymc_marketing/clv/models/pareto_nbd.py
```diff
@@ -211,59 +215,152 @@ def default_model_config(self) -> Dict[str, Dict]:
...
+            },
+            "purchase_covariates_prior": {
+                "dist": "StudentT",
+                "kwargs": {"nu": 1},
```
ColtAllen, line 225: Did you say "kwargs": {"nu": 2}, is needed for data to be centered at 0? For these covariate coefficients to be interpretable, covariates need to be scaled between 0 and 1, with a prior supporting negative values. `StudentT slows down model fitting considerably, so I'm open to alternatives.

pymc_marketing/clv/models/pareto_nbd.py
```diff
@@ -389,43 +521,6 @@ def expected_purchases(
             "chain", "draw", "customer_id", missing_dims="ignore"
         )
 
-    def expected_purchases_new_customer(
```
ColtAllen, line 392: It's worth noting `*new_customer*` methods could support user-specified covariates for what-if and sensitivity analysis, but it's a very niche use case.

pymc_marketing/clv/models/pareto_nbd.py
```diff
@@ -637,28 +774,60 @@ def _distribution_new_customers(
...
+                if data:
+                    new_data = {}
+                    if self.purchase_covariate_cols:
+                        new_data["purchase_data"] = data[self.purchase_covariate_cols]
+                    if self.dropout_covariate_cols:
+                        new_data["dropout_data"] = data[self.dropout_covariate_cols]
+                    pm.set_data(new_data, coords={"customer_id": data["customer_id"]})
+                else:
+                    raise ValueError(
+                        "Dataset with new covariates must be provided when sampling new customers"
```
ColtAllen, line 810: With covariates, each customer will have their own `alpha` and `beta` params, so the Xarray dims, will look like `alpha[customer_id0], alpha[customer_id1]`, etc. For a future PR, I wonder if a plotting function like this is doable:

image

pymc_marketing/clv/models/pareto_nbd.py
```diff
@@ -667,86 +836,92 @@ def _distribution_new_customers(
...
+    def distribution_customer_population(
         self,
+        T: Union[None, np.ndarray, pd.Series] = None,
         random_seed: Optional[RandomState] = None,
     ) -> xarray.Dataset:
-        """Sample from the Gamma distribution representing dropout times for new customers.
+        """Pareto/NBD process representing purchases across the customer population.
 
-        This is the duration of time a new customer is active before churning, or dropping out.
+        This is the distribution of purchase frequencies given 'T' observation periods for each customer.
```
ColtAllen, line 856: If this function could support covariates, we can recreate the frequency PPC plot you're familiar with for covariate model evaluation with covariates. It's already conditional on a `T` array of observations.

ColtAllen:
Code is certainly cleaner than the IF statement jungle of the PR I posted haha. Best way I can contribute is to create a child branch and work in parallel in the dev notebook for user-oriented feedback.

ricardoV94:
> Code is certainly cleaner than the IF statement jungle of the PR I posted haha. Best way I can contribute is to create a child branch and work in parallel in the dev notebook for user-oriented feedback.

It's always easier to refactor than write the first pass. The notebook would help a lot! 

I'll be testing the covariate model today, and fix stuff I've certainly broken

tests/clv/models/test_pareto_nbd.py
```diff
@@ -394,15 +372,262 @@ def test_save_load_pareto_nbd(self):
...
+        ] -= 1  # more likely to purchase (negative coefficient)
+        test_data_alt[
+            "dropout_cov"
+        ] -= 1  # less likely to drop out (positive coefficient)
+        res_high_transaction = model.predictive_distribution(test_data_alt).mean(
+            ("chain", "draw")
+        )
+        # FIXME: Test is failing for the first two conditions, signs seem swapped?
+        assert (res_zero["purchase_rate"] < res_high_transaction["purchase_rate"]).all()
+        assert (res_zero["dropout"] > res_high_transaction["dropout"]).all()
```
ricardoV94, line 615: @ColtAllen These two checks fail while the two below pass. Are we perhaps taking draws of the reciprocal of these quantities?

ColtAllen: Yes; these two lines should not be reciprocals:

https://github.com/ricardoV94/pymc-marketing/blob/6ba6ee1e7db2a93467beea974079fb0039cafca8/pymc_marketing/clv/models/pareto_nbd.py#L795

https://github.com/ricardoV94/pymc-marketing/blob/6ba6ee1e7db2a93467beea974079fb0039cafca8/pymc_marketing/clv/models/pareto_nbd.py#L801

ricardoV94:
@ColtAllen finally pushed my last changes. I broke the pre-existing methods API very much, but we can:

1. Go back and make it like it looked before, plus optional argument for covariates in all methods
2. Use a new API but support old API with deprecation warnings
3. Something else. I don't like asking for a whole dataframe in cases where the user is only changing one variable, but for covariates it makes sense to request a dataframe anyway, so it makes sense that other variable also comes in the same dataframe

I'm in favor of 2.

Finally, the slow `TestParetoNBDModelWithCovariates::test_inference` is failing badly. I used synthetic data from fixed true parameters (so I could control the covariate coefficients), but it is not recovering those at all. Should I use different true parameter values?

tests/clv/models/test_pareto_nbd.py
ricardoV94: Fixed

ColtAllen:
> @ColtAllen finally pushed my last changes. I broke the pre-existing methods API very much, but we can:
> 
> 1. Go back and make it like it looked before, plus optional argument for covariates in all methods
> 2. Use a new API but support old API with deprecation warnings
> 3. Something else. I don't like asking for a whole dataframe in cases where the user is only changing one variable, but for covariates it makes sense to request a dataframe anyway, so it makes sense that other variable also comes in the same dataframe
> 
> I'm in favor of 2.

Yes, let's do option 2.

> Finally, the slow `TestParetoNBDModelWithCovariates::test_inference` is failing badly. I used synthetic data from fixed true parameters (so I could control the covariate coefficients), but it is not recovering those at all. Should I use different true parameter values?

Yes - the true parameters you're using are for a model fit without covariates. When I fit a covariate model in the dev notebook of my previous PR, all parameter values changed.

Also, only the covariates were synthesized, and were tacked onto an existing dataset. If we're to test with parameter recovery, we'll need to generate a full dataset from the parameters used for testing.

ColtAllen:
> No, I'm generating new observations (recency and frequency) in that test as well

Buy why is only a single prior predictive sample being taken?

ricardoV94:
> > No, I'm generating new observations (recency and frequency) in that test as well
> 
> Buy why is only a single prior predictive sample being taken?

One draw is enough, it contains an observation for each synthetic customer

ColtAllen:
>  No, I'm generating new observations (recency and frequency) in that test as well
> One draw is enough, it contains an observation for each synthetic customer

I see. The new frequency column name is misspelled though, so it's testing on the original column rather than what was generated for the test.

ColtAllen:
Is the user API still a work in progress? All predictive methods require arguments to be converted into dataframe columns beforehand, which seems rather cumbersome.

ColtAllen (approved):

PR #555, opened 2024-02-29
Run slow tests in CI

### LINKED ISSUE #308: Run `slow` tests on CI
The tests marked as `pytest.mark.slow` are currently not running in the CI. We should consider running them, but have to make sure CI runtime doesn't balloon. If it does, we can perhaps split the tests into more jobs.

ricardoV94:
## Description

Adds a separate CI job that runs only the slow tests. The reason for the split is that these tests won't run locally by default, and on the CI they will run separately so that the faster tests can act as a faster smoke test 

## Related Issue

- [x] Closes #308 
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--555.org.readthedocs.build/en/555/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,12 +38,31 @@ jobs:
           name: ${{ matrix.config.python-version }}
           fail_ci_if_error: false
 
+  test_slow:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
+      - name: Run tests
+        run: |
+          pip install -e .[test]
+          pytest --only-slow --cov-report=xml --no-cov-on-fail --durations=50
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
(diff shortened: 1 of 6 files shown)


ricardoV94:
Only the GammaGamma slow test seems to fail: https://github.com/pymc-labs/pymc-marketing/actions/runs/8110804909/job/22168800146?pr=555#step:4:782

Perhaps @ColtAllen you could take a look sometime?

ricardoV94:
The GammaGamma convergence failure may be due to the strange test config that was being used since the ModelBuilder refactoring

ColtAllen:
> The GammaGamma convergence failure may be due to the strange test config that was being used since the ModelBuilder refactoring

I'm not sure why this config is being specified in the test either. Can it just be ran with the default config?

ricardoV94:
Yeah I changed it but the HalfFlat doesn't converge which may be fair. I'll give slightly better priors although the original test had HalfFlat priors I think

PR #574, opened 2024-03-07
Move watermark in environment.yml

juanitorduz:
Add watermark package for nb versioning. 

Shall we run all notebooks before a release to make sure they all work? 

----
📚 Documentation preview 📚: https://pymc-marketing--574.org.readthedocs.build/en/574/

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -29,6 +29,7 @@ dependencies:
 - sphinxext-opengraph
 - sphinx-notfound-page
 - sphinx-design
+- watermark
 # lint
 - mypy
 - pandas-stubs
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -29,6 +29,7 @@ dependencies:
 - sphinxext-opengraph
 - sphinx-notfound-page
 - sphinx-design
+- watermark
```
wd60622, line 32: This is already above at line 19 from Alex's PR

wd60622 (changes requested):
Seems to already be there. Maybe move to the appropriate section

environment.yml
juanitorduz: haha good catch!

wd60622 (approved):
Looks good

PR #598, opened 2024-03-19
Add time-varying coefficient

ulfaslak:
This PR adds time-varying coefficient options to the `DelayedSaturatedMMM` class.

## Description
Specifically the following model specification is now possible:

```Python
mmm = DelayedSaturatedMMM(
    date_column=my_date_column_name,
    channel_columns=my_media_column_names,
    time_varying_intercept=True,            # <-- These two args
    time_varying_media_effect=True,         # <-- are what's up
)
```

`time_varying_intercept` creates a time-varying prior on the intercept, while `time_varying_media_effect` creates a time-varying prior on the *total media contribution* (i.e. not individual columns, this will be added in a later PR).

🚨 For now let's get some reviews on this and agree what is a good way to add this functionality. Then we can add tests and docs later.

To make this easier to review, maybe start looking inside the `mmm_tvp_example.ipynb`, then check out `tvp.py`, and then go over the modifications in related filed I had to make to get the API changes to work.

## Related Issue

- [x] Related to #231 #415

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--598.org.readthedocs.build/en/598/


pymc_marketing/mmm/tvp.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Returns
+    -------
+    pm.Deterministic
+        Time-varying prior.
+    """  # noqa: W605
+    if cov_func is None:
+        eta = pm.Exponential(f"eta_{name}", lam=eta_lam)
+        ls = pm.InverseGamma(f"ls_{name}", mu=ls_mu, sigma=ls_sigma)
+        cov_func = eta**2 * pm.gp.cov.Matern52(1, ls=ls)
```
wd60622, line 58: These will have to be in the with as well. Maybe this whole function can just be assumed to happen within a with context

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -373,11 +451,21 @@ def build_model(
...
+                    * beta_channel
+                    * tv_multiplier_media[:, None],
+                    dims=("date", "channel"),
+                )
+            else:
+                channel_contributions = pm.Deterministic(
+                    name="channel_contributions",
+                    var=channel_adstock_saturated * beta_channel,
+                    dims=("date", "channel"),
+                )
```
wd60622, line 468: nitpick. Think this makes it a little easier to read
```suggestion
            channel_contributions_var = channel_adstock_saturated * beta_channel
            if self.time_varying_media_effect:
                channel_contributions_var *= tv_multiplier_media[:, None]

            channel_contributions = pm.Deterministic(
                name="channel_contributions", 
                var=channel_contributions_var, 
                dims=("date", "channel"), 
            )
```

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -337,9 +372,52 @@ def build_model(
...
+
+            if self.time_varying_media_effect:
+                tv_multiplier_media = time_varying_prior(
+                    name="tv_multiplier_media",
+                    X=time_index,
+                    X_mid=self._time_index_mid,
+                    positive=True,
+                    m=200,
+                    L=[self._time_index_mid + 365 / self._time_resolution],
+                    ls_mu=365 / self._time_resolution * 2,
```
wd60622, line 416: Maybe pull this out into a constant
DAYS_IN_YEAR = 365

Maybe this can be consolidated with this hard coded value as well
https://github.com/pymc-labs/pymc-marketing/blob/06b111ebe679ad35822e63a0d96b44195eee6cee/pymc_marketing/mmm/delayed_saturated_mmm.py#L478

wd60622 (commented):
Just some initial thoughts. will keep an eye out on changes

pymc_marketing/mmm/delayed_saturated_mmm.py
ulfaslak: Ya this is much nicer thanks man

pymc_marketing/mmm/delayed_saturated_mmm.py
ulfaslak: gotcha

wd60622:
resolved the conflict to see the status of the tests. @ulfaslak

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -34,6 +36,8 @@ def __init__(
         date_column: str,
         channel_columns: List[str],
         adstock_max_lag: int,
+        time_varying_media_effect: bool = False,
+        time_varying_intercept: bool = False,
```
cetagostini-wise, line 40: I think it would be great if we could control the HSGP. Could we add an `hsgp_config` similar to how we defined the priors? So the more "expert" user could play with how much flexibility they want to give to the HSGP?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -338,9 +374,52 @@ def build_model(
...
+                    name="intercept",
+                    var=intercept_base * tv_multiplier_intercept,
+                    dims="date",
+                )
+            else:
+                intercept = self.intercept_dist(
+                    name="intercept", **self.model_config["intercept"]["kwargs"]
+                )
+
+            if self.time_varying_media_effect:
```
cetagostini, line 410: Instead of checking twice, would it be better to check by the end? You load the data and apply the multiplier only if `True` all at once, saving a few lines of code.

pymc_marketing/mmm/tvp.py
```diff
@@ -0,0 +1,73 @@
...
+            hsgp_size = (n_columns, m)
+        else:
+            hsgp_size = m
+        gp = pm.gp.HSGP(m=[m], L=[L], cov_func=cov_func)
+        phi, sqrt_psd = gp.prior_linearized(Xs=X[:, None] - X_mid)
+        hsgp_coefs = pm.Normal(f"_hsgp_coefs_{name}", size=hsgp_size)
+        f = phi @ (hsgp_coefs * sqrt_psd).T
+        if positive:
+            f = softplus(f)
+        return pm.Deterministic(name, f, dims=dims)
```
cetagostini, line 73: I think would be great to create a plot for the varying parameters, showing the recovered latent pattern which is affecting the channels and/or the intercept, what do you think?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -374,10 +453,14 @@ def build_model(
                 var=logistic_saturation(x=channel_adstock, lam=lam),
                 dims=("date", "channel"),
             )
+
+            channel_contributions_var = channel_adstock_saturated * beta_channel
+            if self.time_varying_media_effect:
+                channel_contributions_var *= tv_multiplier_media[:, None]
```
cetagostini, line 459: It is not a strong opinion, but I think that **if we are using the logic of a base contribution, which is then modified**, it would be great to have this join saved in another `Deterministic` variable. So, users can understand their BASE contribution and the latent multiplier effect independently. We would need to have a small extra piece of code, but what do you think?

e.g:
```python
pm.Deterministic(
var = channel_contributions_var * tv_multiplier_media[:, None]
name='varying_contribution' #or something like it
)
```

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -374,10 +453,14 @@ def build_model(
...
             channel_contributions = pm.Deterministic(
-                name="channel_contributions",
-                var=channel_adstock_saturated * beta_channel,
-                dims=("date", "channel"),
+                name="channel_contributions", 
+                var=channel_contributions_var, 
+                dims=("date", "channel"), 
             )
 
             mu_var = intercept + channel_contributions.sum(axis=-1)
```
cetagostini, line 466: Since the HSGP is in dimension "date" from the beginning, I feel that we are first transforming said HSGP to a form that is compatible with `("date", "channel")` to return at the end to `"date"` only. Isn't it better to transform the channels to `"date"` and finally multiply by the HSGP without needing of `[:,None]`?

It's just an opinion, I think it would be more appropriate the other way and we can avoid a somewhat unnecessary transformation, what do you think?

pymc_marketing/mmm/tvp.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Returns
+    -------
+    pm.Deterministic
+        Time-varying prior.
+    """  # noqa: W605
+
+    with pm.modelcontext(model) as model:
+        if cov_func is None:
+            eta = pm.Exponential(f"eta_{name}", lam=eta_lam)
```
cetagostini, line 58: I think using this current implementation, we could move these distributions into the `model_config`?

cetagostini:
Hey hey, amazing work! I made a few comments, but all are initial thoughts. I'll deep dive during the week 💪🏻

pymc_marketing/mmm/tvp.py
nialloulton: oh this would be really clever, it just becomes another supported distribution?

pymc_marketing/mmm/delayed_saturated_mmm.py
ulfaslak: Maybe better to just support that in the model config? The chief thing you'd want to control is the covariance function, m and L. But for MMM this can even be simplified to just lengthscale and m, I believe since optimal L can be calculated form the length of the data and the lengthscale.

pymc_marketing/mmm/tvp.py
ulfaslak: Agree. Will add!

pymc_marketing/mmm/delayed_saturated_mmm.py
ulfaslak: Yeah this is a nice idea. I think I might actually change the time_varying_prior function so it always swings in the positive range. Then when using it, it always works as a multiplier on base contributions. Will also make joining this logic with logic for hierarchical parameters easy.

I'll make this change, it's nice.

pymc_marketing/mmm/delayed_saturated_mmm.py
ulfaslak: Don't fully understand. You can just commit the change if you believe it is best.

In general this time_varying_prior function supports two dimensions but I'm only using the 1D case here, maybe that's confusing.

I should actually add support for individually varying parameters in this PR too... not complicated

pymc_marketing/mmm/tvp.py
ulfaslak: that's better than an hsgp_config I think. I'm trying to work out what's the right level of rigidity to build into this thing. There are some things here you wouldn't want to change as a user I think. And maybe this is just me, but as we put distributions into the config, we pay the price of more obscure code. Opinions?

pymc_marketing/mmm/utils.py
```diff
@@ -329,6 +331,10 @@ def apply_sklearn_transformer_across_dim(
 
     return data
 
+  
+def softplus(x: pt.TensorVariable) -> pt.TensorVariable:
```
twiecki, line 335: I think this exists in `pytensor.tensor.math`.

pymc_marketing/mmm/base.py
```diff
@@ -336,19 +342,26 @@ def plot_posterior_predictive(
             )
 
             ax.fill_between(
-                x=self.X[self.date_column],
+                x=posterior_predictive_data.date,
```
juanitorduz, line 345: What is the reason for this change? The date column can have a different name.

pymc_marketing/mmm/base.py
```diff
@@ -336,19 +342,26 @@ def plot_posterior_predictive(
...
                 label="$50\%$ HDI",  # noqa: W605
             )
 
-            target_to_plot: np.ndarray = np.asarray(
-                self.y if original_scale else self.preprocessed_data["y"]  # type: ignore
+            target_to_plot = np.asarray(
+                self.y if original_scale else self.get_target_transformer().transform(self.y[:, None]).flatten()  # type: ignore
             )
+
+            assert len(target_to_plot) == len(posterior_predictive_data.date), (
```
juanitorduz, line 357: can wee keep `date_col` as generic date name?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -342,9 +378,52 @@ def build_model(
...
+                )
+
+            if self.time_varying_intercept:
+                tv_multiplier_intercept = time_varying_prior(
+                    name="tv_multiplier_intercept",
+                    X=time_index,
+                    X_mid=self._time_index_mid,
+                    positive=True,
+                    m=200,
+                    L=[self._time_index_mid + DAYS_IN_YEAR / self._time_resolution],
```
juanitorduz, line 395: I think it might be cleaner to use the `c` parameeer?
```
c: float
  The proportion extension factor. Used to construct L from X. Defined as S = max|X| such that X is in [-S, S]. 
  L is calculated as c * S. One of c or L must be provided. Further information can be found in Ruitort-Mayol et al.
```

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: I agree :)

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: > I should actually add support for individually varying parameters in this PR too... not complicated
I suggest we keep the scope of this PR small and then iterate

pymc_marketing/mmm/tvp.py
```diff
@@ -0,0 +1,73 @@
...
+            ls = pm.InverseGamma(f"ls_{name}", mu=ls_mu, sigma=ls_sigma)
+            cov_func = eta**2 * pm.gp.cov.Matern52(1, ls=ls)
+
+        if type(dims) is tuple:
+            n_columns = len(model.coords[dims[1]])
+            hsgp_size = (n_columns, m)
+        else:
+            hsgp_size = m
+        gp = pm.gp.HSGP(m=[m], L=[L], cov_func=cov_func)
+        phi, sqrt_psd = gp.prior_linearized(Xs=X[:, None] - X_mid)
```
juanitorduz, line 68: Note for future review: We need to be careful we center the data with respect to the training set (even when we are doing out of sample prediction)

juanitorduz: See doctstrings in https://github.com/pymc-devs/pymc/blob/main/pymc/gp/hsgp_approx.py#L243

juanitorduz (commented):
Added some small comments/ questions fot thie first round :)

juanitorduz:
@ulfaslak could you please rebase from `main` as the tests are failing because https://github.com/pymc-labs/pymc-marketing/pull/615 🙏 ?

cetagostini-wise:
Hey, small comment here:

```
        time_varying_media_effect: bool = False,
        time_varying_intercept: bool = False,
```

This is a matter of taste but I think having two parameters for the same component is a bit strange. What do you think if we evaluate with strings or tuples?

```
time_varying = ('intercept', 'media') #or
```

```
time_varying = 'intercept-media'
```

I think it would be better inside for the API

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -38,6 +40,8 @@ def __init__(
         date_column: str,
         channel_columns: List[str],
         adstock_max_lag: int,
+        time_varying_media_effect: bool = False,
```
cetagostini-wise, line 43: Hey, small comment here... This is a matter of taste but I think having two parameters for the same component is a bit strange. What do you think if we evaluate with strings or tuples?

```
time_varying = ('intercept', 'media') #or
```

```
time_varying = 'intercept-media'
```

I think it would be better inside for the API

nialloulton:
Yep, that approach makes sense!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

cetagostini-wise:
Following @juanitorduz mention, and perhaps to align with https://github.com/transferwise/airflow-marketing/pull/3563, if it's worthy.

We could create a `models.components` folder and a `time.py` file to save the file? This could help to add and save all code related to "time" dependencies.

What about if we treat this implementation as components as well? Similar to the PR attach, all time varying media or intercept happens in a class which then is added to the main model. This and the split between two PRs (media/intercept) could make everything very easy to debug, test and play. What everyone thinks? @wd60622 @ulfaslak

cetagostini-wise:
> The main question from what i see is if we wanna provide all the priors and HSGP parameters configurable via model_config (Personally, i think we should)

@juanitorduz My personal opinion on it, is we should allow it.

juanitorduz:
> Following @juanitorduz mention, and perhaps to align with #602 , if it's worthy.
> 
> We could create a `models.components` folder and a `time.py` file to save the implementation? This could help to add and save all code related to "time" dependencies.
> 
> What about if we treat this implementation as components as well? Similar to the PR attach, all time varying media or intercept happens in a class which then is added to the main model. This and the split between two PRs (media/intercept) could make everything very easy to debug, test and play. What everyone thinks? @wd60622 @ulfaslak

The more modular, the better!  Let's work on smaller PR so that we can iterate faster :)

pymc_marketing/mmm/base.py
ulfaslak: This enables OOS posterior predictive. `self.X` does not get updated upon `self._data_setter(X_new)`. So if the posterior predictive is for new OOS data, then `self.X[self.date_column]` will still be the in-sample dates.

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -28,6 +29,7 @@
 
 __all__ = ["DelayedSaturatedMMM"]
 
+DAYS_IN_YEAR = 365.25
```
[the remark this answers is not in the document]
ulfaslak: Perfect. Will do.

pymc_marketing/mmm/delayed_saturated_mmm.py
ulfaslak: Unfortunately this causes some trouble with predicting out of sample. @bwengals did you find the cause for this?

pymc_marketing/mmm/tvp.py
ulfaslak: Can you elaborate? What is the concern?

pymc_marketing/mmm/delayed_saturated_mmm.py
ulfaslak: I quite like this. Maybe this gets too complicated but:

```Python
time_varying='intercept'
...
time_varying='total_media'
...
time_varying=['intercept', 'total_media']
...
time_varying=['intercept', 'channel1', 'channe2']
```

could work

ulfaslak: Maybe even 

```Python
time_varying=['intercept', ('channel1', 'channel2')] # now channel1 and channel2 are summed and multiplied by a time varying coef
```

ulfaslak:
> @ulfaslak This feature is very exciting! As this PR was about collecting feedback, I have a proposal: Split this PR into two:
> 
> 1. Time-varying intercept
> 2. Time-varying media (as in this PR)
> 
> The reason? to make The review process simpler and faster :)
> 
> For each PR, we should aim for:
> 
> * Implementation with changes just on the scope of the PR.
> * Docstring
> * Tests
> * Example Nb
> 
> This way we can iterate fast, as probably we wanna push this one out soon :) 🚀
> 
> The main question from what i see is if we wanna provide all the priors and HSGP parameters configurable via `model_config` (Personally, i think we should)

I like this.

I will try to bang the first PR together then. Thanks for a lot of high-quality feedback 👌

ulfaslak:
> @ulfaslak could you please rebase from `main` as the tests are failing because #615 🙏 ?

I will close this and make an new PR for intercept-tvp on a branch off of main.

Issue #551, opened 2024-02-27
Is there a way to apply the weibul adstock transformation in model 


shuvayan:
Hello,

I see that there is a weibull adstock transformation function available but it is not currently being called in the delayed_saturated_mmm module (only geometric adstock) is being called.
May I know if there is a way I can use the weibul adstock transformation within the framework?

wd60622:
You're right. The delayed MMM class doesnt currently have support for different saturation functions. 

At the moment, you would have to inherit from the class and override the build_model method to use the adstock, saturation, etc that you would like to use

shuvayan:
Okay. Thank you so much for sharing the methodology , I will see if I can implement it!!

wd60622:
Hi @shuvayan 
Hope you were able to figure out a solution.
FYI: We are working on a #632 where a user can specify the saturation function (as well as adstock)

PR #602, opened 2024-03-23
Allowing different saturation and lagging functions into DelayedSaturatedMMM

cetagostini-wise:
Hello team, as we all know we now have a large number of saturation and lagging functions, however our code was only forcing a specific type.

Create a folder with the lagging and saturation components, this will allow users to define the function they want and have it mapped.

Currently we could choose the function in the following way:
```python
mmm = DelayedSaturatedMMM(
    model_config = my_model_config,
    sampler_config = my_sampler_config,
    lagging_function = "geometric",
    saturation_function = "logistic",
    date_column="date_week",
    channel_columns=["x1", "x2"],
    control_columns=[
        "event_1",
        "event_2",
        "t",
    ],
    adstock_max_lag=8,
    yearly_seasonality=2,
)
```

**Important**: For now the model continues applying lagging and then saturation, this cannot be modified yet but after doing this merge, I will start working on it in another PR.

Right now the model works perfectly with the following functions:
a) Saturation:
- Hill
- Michaelis Menten
- Logistic

b) Lagging:
- Weibull (PDF & CDF)
- Geometric

Any of these combinations train properly. **Important**, the function `plot_channel_contributions_grid` is not available right now because was hard coded to work only with the main principal lagging and saturation functions. I need to think how to make it general, that being said, I do not consider that this should be a blocker to do merge, because it is only to display the saturation functions only. We can work on an improvement in another PR.

## Related Issue

- [ ] Closes #
- [x] Related to #551

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--602.org.readthedocs.build/en/602/


pymc_marketing/mmm/models/components/lagging.py
```diff
@@ -0,0 +1,313 @@
...
+from pymc_marketing.mmm.utils import _get_distribution, _validate_model_config
+
+
+class BaseFunction:
+    def __init__(self, max_lagging, model):
+        self.max_lagging = max_lagging
+        self.model = model
+
+    def apply(self, data):
+        raise NotImplementedError("Subclasses must implement this method.")
```
wd60622, line 15: Think all of the defaults should go here rather than in an ifelse block that is hidden within a method. With defaults here, a user would be able to make their own saturation function that we don't have without having to change the ifelse block

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -462,6 +478,59 @@ def default_model_config(self) -> Dict:
             "gamma_control": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 2}},
             "gamma_fourier": {"dist": "Laplace", "kwargs": {"mu": 0, "b": 1}},
         }
+        if self.lagging_function == "geometric":
```
wd60622, line 481: so in my suggestion, all of these cases would belong to the class of "geometric", etc

pymc_marketing/mmm/models/components/lagging.py
```diff
@@ -0,0 +1,313 @@
...
+import pymc as pm
+
+from pymc_marketing.mmm.transformers import geometric_adstock, weibull_adstock
+from pymc_marketing.mmm.utils import _get_distribution, _validate_model_config
+
+
+class BaseFunction:
+    def __init__(self, max_lagging, model):
+        self.max_lagging = max_lagging
+        self.model = model
```
wd60622, line 12: Think that we can drop the model being passed around overall and just assume that the `apply` method is being called within a model context

i.e 
```python
adstock = MyAdstockClass(max_lagging=12)

with pm.Model(...) as model :
    adstock.apply(channel_data)

File: pymc_marketing/mmm/models/components/saturation.py [hunk 4]
```diff
@@ -0,0 +1,334 @@
...
+    michaelis_menten,
+)
+from pymc_marketing.mmm.utils import _get_distribution, _validate_model_config
+
+
+### SATURATION FUNCTIONS
+class BaseFunction:
+    def __init__(self, max_lagging, model):
+        self.max_lagging = max_lagging
+        self.model = model
```
wd60622, line 17: same recommendation here. 

Also think that these will be very similar so potentially combining in a components.base module

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -39,6 +42,8 @@ def __init__(
         validate_data: bool = True,
         control_columns: Optional[List[str]] = None,
         yearly_seasonality: Optional[int] = None,
+        lagging_function: str = "geometric",
+        saturation_function: str = "logistic",
```
wd60622, line 46: Think that we should allow for class instances as well that fit our component definition and deprecate the `adstock_max_lag` as that is already part of the class. 

For instance, 
```python 
adstock= GeometricAdstock(max_lag=12)
mmm = DelayedSaturatedMMM(
    ...,
    adstock=adstock, 
    ..., 
)

Comment (wd60622) [inline pymc_marketing/mmm/delayed_saturated_mmm.py:46]: I might suggest `adstock` and `saturation` as names. Especially if they are not actually functions

File: pymc_marketing/mmm/models/components/saturation.py (hunk 4 above)
Comment (wd60622) [reply pymc_marketing/mmm/models/components/saturation.py:17]: They might deviate a bit in order to suppor the experimentation integration. 

For all of these, the `add_lift_test_measurements` would just need something like this in order to be supported: https://github.com/wd60622/pymc-marketing/blob/lift-test-calibration/pymc_marketing/mmm/lift_test.py#L372-L378

However we can do that after both are merged

File: pymc_marketing/mmm/models/components/lagging.py [hunk 6]
```diff
@@ -0,0 +1,313 @@
...
+                    lam=adstock_lambda,
+                    k=adstock_shape,
+                    type="PDF",
+                    l_max=self.max_lagging,
+                ),
+                dims=("date", "channel"),
+            )
+
+
+class WeibullCDFAdstockComponent(BaseFunction):
```
wd60622, line 179: Personally think that both of the weibull's can be combined. 

Another reason to just pass the class itself 
For instance, 
```python
adstock_cdf = WeibullAdstock(max_lag=12, kind="CDF")
adstock_pdf = WeibullAdstock(max_lag=12, kind="PDF")

# Choose the adstock out of the class 
adstock = adstock_cdf
mmm = DelayedSaturatedMMM(
    ..., 
    adstock=adstock, 
    ...,

File: pymc_marketing/mmm/models/components/saturation.py [hunk 7]
```diff
@@ -0,0 +1,334 @@
...
+                dims=("channel"),
+            )
+
+            # beta
+            saturation_beta = self.saturation_beta_dist(
+                "saturation_beta",
+                **self.model_config["saturation_beta"]["kwargs"],
+            )
+
+            return pm.Deterministic(
```
wd60622, line 106: General comment for this return: This shouldn't be a Deterministic but the result will be wrapped in a deterministic in the MMM class method

Think this will save a lot of duplication and would be easier to for users to make their own Saturation for our model because this "channel_contributions" variable name is what is used in the plots. It shouldn't be flexible

pymc_marketing/mmm/models/components/saturation.py
```diff
@@ -0,0 +1,334 @@
...
+                "channel_contributions",
+                var=hill_saturation(
+                    x=data,
+                    sigma=saturation_sigma,
+                    beta=saturation_beta,
+                    lam=saturation_lambda,
+                ),
```
wd60622, line 113: If this part is wrapped in a method then this can be used for the lift test integration

for instance: 
```python
class HillSaturation(BaseFunction): 
    def saturation_function(self, spend, sigma, beta, lam): 
        return hill_saturation(spend, sigma=sigma, beta=beta, lam=lam)
    @property 
    def var_mapping(self) -> dict[str, str]: 
        return {
            "sigma": "saturation_sigma", 
            "beta": "saturation_beta", 
            "lam": "saturation_lam", 
        }
    def apply(self, ...): 
        ...
        self.saturation_function(data, saturation_sigma, saturation_beta, saturation_lambda)
    
     # The current methods

**Comment by wd60622** [comment]
Some general comments on how we can better support a more flexible API 

To summarize the comments on the Saturation functions, 
- I really like that the multiplication is brought into this class 
- Think that configuration checkouts and defaults should happen within this class
- People should be able to pass one of these classes and we should work towards deprecation of adstock_max_lag since that would be part of the class instance.
- Just some comments so that this will be working with lift-tests right out of the box 😄 

For a more concrete example, I am seeing that the saturation class would look like this (using LogisticSaturation as example)

```python
class BaseSaturationFunction: 
    def apply(self, data, model_config): 
        """Assumes to happen within a model context."""
        saturation_kwargs = self.initialize_random_variables(model_config)
        return self.saturation_function(**saturation_kwargs)

class LogisticSaturation(BaseSaturationFunction): 
    # Two methods required for experimentation
    def saturation_function(self, spend, lam, beta): 
        return beta * logistic_saturation(spend, lam=lam)

    @property 
    def variable_mapping(self) -> dict[str, str]: 
        """The mapping from saturation function parameters to variables in a model."""
        return {"lam": "saturation_lambda", "beta": "saturation_beta"}

    @property 
    def default_config(self) -> "ModelConfig": 
        return {
            "saturation_beta": {"dist": "HalfNormal", "kwargs": {"sigma": 2}}, 
            "saturation_lambda": {"dist": "Gamma": "kwargs": {"alpha": 3, "beta": 1}}, 
        }

    def initial_random_variables(self, model_config: "ModelConfig") -> dict[str, TensorVariable]: 
        config = {**self.default_config, **model_config}

        lambda_dist = _get_distribution(dist=config["saturation_lambda"])
        beta_dist = _get_distribution(dist=config["saturation_beta"])

        return {
            "lam": lambda_dist("saturation_lambda", **config["saturation_lambda"]["kwargs"], dims="channel"), 
            "beta":  beta_dist("saturation_beta", **config["saturation_beta"]["kwargs"], dims="channel"), 
        }
```

The `BaseDelayedSaturatedMMM` class can rely on the `default_config` so that it wouldn't be hard coded. For instance,

```python
class BaseDelayedSaturationMMM: 
    @property 
    def default_model_config(self) -> dict: 
        base_model_config = {...}
        base_model_config = {**base_model_config, **self.saturation.default_config}
```

Then anything that fits this class can be used in the lift tests: 
```python
saturation= LogisticSaturation()

add_lift_measurements_to_likelihood(
    df_lift_test,
    variable_mapping=saturation.variable_mapping,
    saturation_function=saturation.saturation_function,
    model=model,
    dist=dist,
    name=name,
)
```

wd60622:
Another aspect that becomes difficult here is that saving and loading. That will have to be thought about as well

pymc_marketing/mmm/models/components/lagging.py
cetagostini: I think this option and the other can coexist together, meaning users can define everything as classes or strings.

pymc_marketing/mmm/delayed_saturated_mmm.py
cetagostini: Names changes to `adstock_type` and `saturation_type`

pymc_marketing/mmm/models/components/lagging.py
cetagostini: Agree this can be simplified. Let me think about it a bit more, I'm not fully convinced with the classes because decrease our control and users will need to import the class to use it, none strong opinion here but I recognize doing it has a few things in favor.

pymc_marketing/mmm/models/components/saturation.py
cetagostini: Agree, great catch, already apply it.

cetagostini:
@wd60622 great feedback. I address the majority of your comments at the moment, trying to address the compatibility with calibration, now it should work like:

```python
add_lift_measurements_to_likelihood(
    df_lift_test,
    variable_mapping=dummy_model.saturation.variable_mapping,
    saturation_function=dummy_model.saturation_function,
    model=dummy_model,
    dist=dist,
    name=name,
)

**Comment by cetagostini** [comment]
> Another aspect that becomes difficult here is that saving and loading. That will have to be thought about as well

Why? I was checking, and I don't see the reason. Can you give me more input?

File: pymc_marketing/mmm/delayed_saturated_mmm.py (hunk 5 above)
Comment (cetagostini) [reply pymc_marketing/mmm/delayed_saturated_mmm.py]: Considering considering!

**Comment by cetagostini** [comment]
Small help here, maybe something is missing in the code but all test are getting the following issue:
```
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/602/lib/python3.10/site-packages/sphinx/config.py", line 358, in eval_config_file
    exec(code, namespace)  # NoQA: S102
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/checkouts/602/docs/source/conf.py", line 5, in <module>
    import pymc_marketing  # isort:skip
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/602/lib/python3.10/site-packages/pymc_marketing/__init__.py", line 1, in <module>
    from pymc_marketing import clv, mmm
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/602/lib/python3.10/site-packages/pymc_marketing/mmm/__init__.py", line 1, in <module>
    from pymc_marketing.mmm import base, delayed_saturated_mmm, preprocessing, validating
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/602/lib/python3.10/site-packages/pymc_marketing/mmm/delayed_saturated_mmm.py", line 16, in <module>
    from pymc_marketing.mmm.models.components.lagging import _get_lagging_function
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc-marketing/envs/602/lib/python3.10/site-packages/pymc_marketing/mmm/models/__init__.py", line 1, in <module>
    from pymc_marketing.mmm.models.components.lagging import _get_lagging_function
ModuleNotFoundError: No module named 'pymc_marketing.mmm.models.components'
```

**Comment by cetagostini** [comment]
pre-commit.ci autofix

**Comment by cetagostini** [comment]
"pre-commit.ci autofix"

File: pyproject.toml [hunk 9]
```diff
@@ -48,6 +48,7 @@ test = [
 packages = [
     "pymc_marketing",
     "pymc_marketing.mmm",
+    "pymc_marketing.mmm.models",
```
wd60622, line 51: adding pymc_marketing.mmm.models.components might be required and get you through your issue if I understood the issue correctly. Granted, I don't use setuptools much. 

One suggestion here: Stick with just a models submodule to simplify the imports. Is it required to have such a deep import?

- Before: `from pymc_marketing.mmm.models.components import ...`
- Alternative 1: `from pymc_marketing.mmm.models.lagging import ...`
- Alternative 2: `from pymc_marketing.mmm.lagging import ...`

Adding submodule comes with an extra `__init__.py` file and more folder structure which I find cumbersome

pymc_marketing/mmm/models/components/lagging.py
```diff
@@ -0,0 +1,394 @@
...
+    -------
+    ValueError
+        If the lagging function name is not recognized.
+
+    """
+    lagging_functions = {
+        "geometric": GeometricAdstockComponent,
+        "weibull_pdf": WeibullPDFAdstockComponent,
+        "weibull_cdf": WeibullCDFAdstockComponent,
+        # Add other lagging function classes here as needed
```
wd60622, line 386: Users don't have access to this and is major reason that I am for passing the classes themself. 
Alternatives could 
1. to have this dictionary outside of function allow editing for user to add their own

pymc_marketing/mmm/models/components/lagging.py
wd60622: Our decrease in control = more control for users. 

If the fear is having to import one more thing, then I think their could be a route to use strings or classes. However, having a non-editable mapping from strings to classes shouldn't be hard coded and non-flexible to user. imo

pymc_marketing/mmm/models/components/lagging.py
```diff
@@ -0,0 +1,394 @@
+from typing import Dict, Optional, Union
+
+import pymc as pm
+
+from pymc_marketing.mmm.transformers import geometric_adstock, weibull_adstock
+from pymc_marketing.mmm.utils import _get_distribution
+
+
+class BaseFunction:
```
wd60622, line 9: Think it would be good to setup the requirements of this class here. 
For instance, an ABC can be used and an `abstractmethod` for default_model_config

Also, using this for `apply` method would catch at class initialization instead of method call

pymc_marketing/mmm/models/components/lagging.py
```diff
@@ -0,0 +1,394 @@
...
+    def __init__(self, max_lagging: int = 4, model: Optional[pm.Model] = None):
```
wd60622, line 10: Does `model` have to be passed into init? Or can it just be passed into `apply` instead. 
I think carrying around to just use in one method can become cumbersome

pymc_marketing/mmm/models/components/lagging.py
```diff
@@ -0,0 +1,394 @@
...
+        self.max_lagging = max_lagging
+        self.model = model
+        if model_config is not None and not all(
+            key in model_config for key in self.REQUIRED_KEYS
+        ):
+            self.model_config = {**model_config, **self._default_lagging_config}
+        elif model_config is None:
+            self.model_config = self._default_lagging_config
+        else:
+            self.model_config = model_config
```
wd60622, line 65: Seems to be logic that you apply in every class. Can likely put it in the baseclass

pymc_marketing/mmm/models/components/lagging.py
```diff
@@ -0,0 +1,394 @@
...
+            The adstocked data as a PyMC Deterministic variable.
+
+        """
+        self.adstock_lambda_dist = _get_distribution(
+            dist=self.model_config["adstock_lambda"]
+        )
+
+        self.adstock_shape_dist = _get_distribution(
+            dist=self.model_config["adstock_shape"]
+        )
```
wd60622, line 327: Would these be used anywhere else that is not this method? If so, just using variables that are local to this method makes the most sense to me

pymc_marketing/mmm/models/components/lagging.py
wd60622: This might be repeat comment, but if we assume that the apply happens in a model context, then we don't need to pass the model at all and the class doesn't need it anymore

wd60622 (commented):
Some additional comments. Thanks for making some changes

Some of the failing tests come from the forward pass being redefined (Some from methods that I have added). If you need any help, let me know

juanitorduz:
@cetagostini can you please rebase from `main` to fix the test running (solved in https://github.com/pymc-labs/pymc-marketing/pull/615)

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -346,44 +359,58 @@ def build_model(
...
             )
 
+            # beta_channel = self.beta_channel_dist(
+            #     name="beta_channel",
+            #     **self.model_config["beta_channel"]["kwargs"],
+            #     dims=("channel",),
+            # )
+            # alpha = self.alpha_dist(
+            #     name="alpha",
+            #     dims="channel",
```
juanitorduz, line 383: what up with this commented code XD?

cetagostini-wise: haha, I needed as reference, will be deleted for sure before merge!

PR #606, opened 2024-03-27
Implementing Hill Saturation in PyMC-marketing

JakePiekarski314:
Implements the Hill function, a sigmoidal model often used to describe saturation effects in various contexts,
    such as biochemical reactions, pharmacological dose-response relationships, and media spend efficiency in marketing.
    The function is characterized by a flexible sigmoidal curve that can model a range of response shapes from hyperbolic
    to S-shaped, depending on the parameters.
    The Hill function is given by:
        f(x) = 1 / (1 + (x / ec)^(-slope))
    where:
    - x is the input variable representing the level of an independent factor (e.g., substrate concentration in biochemistry,
      dose in pharmacology, or media spend in marketing).
    - ec is the "half saturation point" or the "effective concentration" at which the function value is half of its maximum,
      indicating a point of inflection where the response starts to exhibit diminishing returns.
    - slope (often denoted as the Hill coefficient in biochemical contexts) is a dimensionless parameter that controls the
      steepness of the curve. It provides insight into the cooperativity or the intensity of the saturation effect:
        - slope < 1 indicates a more gradual increase (concave curve) leading to saturation,
        - slope = 1 represents a hyperbolic response similar to Michaelis-Menten kinetics in enzyme activity,
        - slope > 1 yields an S-shaped curve indicating a sharper transition between the phases of low and high response.
    Parameters:
    - x (float): Input variable (e.g., media spend, dose, substrate concentration) which the response depends on.
    - ec (float): Half saturation point, a critical value of 'x' where the response is half of its maximal value. It must be
                 greater than 0 to avoid division by zero and ensure a meaningful inflection point.
    - slope (float): Shape control parameter that affects the curve's steepness and the responsiveness of the system to changes
                     in 'x'. A positive value is required to maintain the sigmoidal nature of the function.
    Returns:
    - float: The output of the Hill function, representing the saturation effect or response level for the given input 'x'.
             The return value is bounded between 0 (as x approaches 0) and 1 (as x approaches infinity), representing the
             fraction of the maximal possible response.
    Example:
    - To model the effect of increasing media spend on the return on investment with diminishing returns beyond a certain point,
      one might use the Hill function with appropriate 'ec' and 'slope' values to capture the saturation effect.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--606.org.readthedocs.build/en/606/

### DIFF
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -369,6 +369,41 @@ def weibull_adstock(
     else:
         raise ValueError(f"Wrong WeibullType: {type}, expected of WeibullType")
     return batched_convolution(x, w, axis=axis)
+    
+
+def hill_function(x, ec, slope):
+    """
+    Implements the Hill function, a sigmoidal model often used to describe saturation effects in various contexts,
+    such as biochemical reactions, pharmacological dose-response relationships, and media spend efficiency in marketing.
+    The function is characterized by a flexible sigmoidal curve that can model a range of response shapes from hyperbolic
+    to S-shaped, depending on the parameters.
+    The Hill function is given by:
+        f(x) = 1 / (1 + (x / ec)^(-slope))
+    where:
+    - x is the input variable representing the level of an independent factor (e.g., substrate concentration in biochemistry,
+      dose in pharmacology, or media spend in marketing).
+    - ec is the "half saturation point" or the "effective concentration" at which the function value is half of its maximum,
+      indicating a point of inflection where the response starts to exhibit diminishing returns.
+    - slope (often denoted as the Hill coefficient in biochemical contexts) is a dimensionless parameter that controls the
+      steepness of the curve. It provides insight into the cooperativity or the intensity of the saturation effect:
+        - slope < 1 indicates a more gradual increase (concave curve) leading to saturation,
+        - slope = 1 represents a hyperbolic response similar to Michaelis-Menten kinetics in enzyme activity,
+        - slope > 1 yields an S-shaped curve indicating a sharper transition between the phases of low and high response.
+    Parameters:
+    - x (float): Input variable (e.g., media spend, dose, substrate concentration) which the response depends on.
+    - ec (float): Half saturation point, a critical value of 'x' where the response is half of its maximal value. It must be
+                 greater than 0 to avoid division by zero and ensure a meaningful inflection point.
+    - slope (float): Shape control parameter that affects the curve's steepness and the responsiveness of the system to changes
+                     in 'x'. A positive value is required to maintain the sigmoidal nature of the function.
+    Returns:
+    - float: The output of the Hill function, representing the saturation effect or response level for the given input 'x'.
+             The return value is bounded between 0 (as x approaches 0) and 1 (as x approaches infinity), representing the
+             fraction of the maximal possible response.
+    Example:
+    - To model the effect of increasing media spend on the return on investment with diminishing returns beyond a certain point,
+      one might use the Hill function with appropriate 'ec' and 'slope' values to capture the saturation effect.
+    """
+    return 1 / (1 + (x / ec) ** (-slope))
 
 
 def logistic_saturation(x, lam: Union[npt.NDArray[np.float_], float] = 0.5):
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -373,38 +373,49 @@ def weibull_adstock(
 
 def hill_function(x, ec, slope):
     """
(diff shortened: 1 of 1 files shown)


cetagostini:
Hey @JakePiekarski314 great initiative! I think it's a good addition.

I have a few initial comments:
1. Modify the style of the docstring to be [numpy style](https://numpydoc.readthedocs.io/en/latest/format.html).

2. Even if the function came from biochemistry, we must related to marketing. Try to make the docstring more intuitive for marketers. Example, `ec` is located in which axis?

3. Remember the model is scaled, In case the value returned by the model is between 0 and 1, does the saturation occur at the maximum response of the target `y`? Or should this function be multiplied by the `beta` of the channel after? Could you provide an example of how you are using it in your models?

4. Would be nice if you can add plots to the docstring. Check the other functions added already, should be something like:
```python
"""
....
.. plot::
        :context: close-figs

        import numpy as np
        from pymc_marketing.mmm.transformers import hill_saturation

        x = np.linspace(0, 10, 100)

        # Varying sigma
        sigmas = [0.5, 1, 1.5]
        plt.figure(figsize=(12, 4))
        for i, sigma in enumerate(sigmas):
            plt.subplot(1, 3, i+1)
            y = hill_saturation(x, sigma, 2, 5)
            plt.plot(x, y)
            plt.xlabel('x')
            plt.ylabel('Hill Saturation')
            plt.title(f'Sigma = {sigma}')
        plt.tight_layout()
        plt.show()
"""
```

cetagostini:
PS: Make sure the `pre-commit` check are pass.
You just need to run the following command in the terminal `pre-commit run --all-files`

cetagostini (changes requested):
I have been seeing this other parametrisation of `hill`, maybe you want to take a look at.

```
.. math::
        f(x) = \frac{\\sigma}{1 + e^{-\beta(x - \\lambda)}}

    where:
    - :math:`\\sigma` is the maximum contribution from a channel (upper asymptote).
    - :math:`\beta` is the slope parameter which determinate the steepness of the curve.
    - :math:`\\lambda` is the transition point on the X-axis. The midpoint where the curve transitions from growth to saturation.
    - :math:`x` is the independent variable.
```

`y = sigma / (1 + pt.exp(-beta * (x - lam)))`

pymc_marketing/mmm/transformers.py
```diff
@@ -373,39 +373,50 @@ def weibull_adstock(
...
-      one might use the Hill function with appropriate 'ec' and 'slope' values to capture the saturation effect.
+    Estimates the sales effect of marketing spend using the Hill function, designed to model the diminishing returns on sales beyond a certain level of spend.
+
+    The Hill function provides a sigmoidal response curve, which is especially useful for capturing the saturation effect in sales as marketing spend increases. The parameter `ec` represents the spend level at which the sales effect is half of its maximum potential, serving as a pivotal point for optimizing marketing budgets. The `slope` parameter influences the curve's steepness, indicating the rate at which sales approach saturation with increasing spend.
+
+    Parameters
+    ----------
+    x : float
+        The marketing spend for which the sales effect is being calculated.
+    ec : float
```
wd60622, line 384: Is there a more descriptive name for this? `effect_breakpoint` 
Where do these initials come from?

pymc_marketing/mmm/transformers.py
```diff
@@ -373,39 +373,50 @@ def weibull_adstock(
...
+        The spend level at which the sales effect reaches half of its maximal potential, highlighting the efficiency breakpoint of marketing investments.
+    slope : float
+        Dictates the curve's steepness, with higher values depicting a more abrupt transition to the saturation phase of sales.
+
+    Returns
+    -------
+    float
```
wd60622, line 391: Might be better to put as tensor

wd60622 (commented):
A few comments. 
Also, some tests to check values would be appreciated!

wd60622:
Hey @JakePiekarski314 
Carlos and I have been working on a change to allow users to specify adstock and saturation in this PR: #632
Carlos also took a stab at the hill saturation function as well in the PR as well.
It might be good to close this PR and consolidate there with the larger changes.
Would you be able to give your feedback on the Hill saturation implementation?

cetagostini:
Hey team, given the new PR #632 I'll close this one!

wd60622:
Thanks for starting this PR @JakePiekarski314 
Give the new version a try and let us know what you think.
If you have a take a prior defaults, please let us know as well.

PR #624, opened 2024-04-11
Add run notebook script

juanitorduz:
~~Attempt to run notebooks on the CI.~~

It was not so easy to add to the CI so we can manually run this script before the releases.

----
📚 Documentation preview 📚: https://pymc-marketing--624.org.readthedocs.build/en/624/


juanitorduz:
https://github.com/pymc-labs/pymc-marketing/pull/624/commits/890ac2f129acab4879c4b47b492cbb7db2e6b111 Fails because of the nb -> html step, which we do not need. I will continue in a different path: do it via a custom simple script

juanitorduz:
I'll close this PR as I wanna see how it is done for the core pymc notebooks as pointed out by @OriolAbril 🙏

PR #636, opened 2024-04-16
Add ruff `RUF` rules

juanitorduz:
Add `RUF` rules as the PyMC configuration.

----
📚 Documentation preview 📚: https://pymc-marketing--636.org.readthedocs.build/en/636/


pymc_marketing/mmm/base.py
```diff
@@ -665,7 +665,7 @@ def plot_budget_scenearios(
         standardize_scenarios_dict_keys(base_data, ["contribution", "budget"])
 
         fig, axes = plt.subplots(nrows=1, ncols=2, figsize=(15, 6))
-        scenarios = [base_data] + list(scenarios_data)
+        scenarios = [base_data, *list(scenarios_data)]
```
wd60622, line 668: Was this whole method not tested to begin with?

cetagostini (approved):

pyproject.toml
```diff
@@ -59,7 +59,12 @@ repository = "https://github.com/pymc-labs/pymc-marketing"
 #changelog = ""
 
 [tool.ruff.lint]
-select = ["E", "F", "I", "W", "UP"]
+select = ["E", "F", "I", "W", "UP", "RUF"]
+ignore = [
+    "RUF001", # String contains ambiguous character (such as Greek letters)
+    "RUF002", # Docstring contains ambiguous character (such as Greek letters)
+    "RUF012", # Mutable class attributes should be annotated with `typing.ClassVar`
+]
```
wd60622, line 67: For the test suite, you might want you use something file specific ruff rules. For instance, the assert statements

```toml
[tool.ruff.lint.per-file-ignores]
"__init__.py" = ["F401"]
"**/tests/*" = ["S101"]

pymc_marketing/mmm/base.py
ricardoV94: Why? that's equivalent code

juanitorduz: This was an automatic ruff refactor 😄

pyproject.toml
juanitorduz: Ok! Actually these ignores I took from the PyMC config as they appear on the code base. I'll create a tickett to improve the rules in the test suite!

pymc_marketing/mmm/base.py
wd60622: Says there is a code coverage difference

wd60622 (approved):
Looks good!

pymc_marketing/mmm/base.py
ricardoV94: Ah... maybe it's only used in NB plots?

juanitorduz: I think so :)

PR #637, opened 2024-04-16
Add Flake8-BugBear and Bandit (Security!)

juanitorduz:
- This is a suggestion to use https://github.com/PyCQA/flake8-bugbear . In many cases, it has helped me find bugs!
-  I also suggest using https://github.com/tylerwince/flake8-bandit as this has security best practices.

Let me know what you think about it :)

----
📚 Documentation preview 📚: https://pymc-marketing--637.org.readthedocs.build/en/637/

### DIFF
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -57,7 +57,8 @@ class CLVModel(ModelBuilder):
                 category=UserWarning,
                 message="The group fit_data is not defined in the InferenceData scheme",
             )
-            assert self.idata is not None
+            if self.idata is None:
+                raise ValueError("idata is not defined")
             self.idata.add_groups(fit_data=data.to_xarray())
 
     def fit(  # type: ignore
@@ -208,7 +209,8 @@ class CLVModel(ModelBuilder):
         """
         if not hasattr(self, "fit_result"):
             raise RuntimeError("The model hasn't been fit yet, call .fit() first")
-        assert self.idata is not None
+        if self.idata is None:
+            raise RuntimeError("The model hasn't been fit yet, call .fit() first")
         new_idata = self.idata.isel(draw=slice(None, None, keep_every)).copy()
         return type(self)._build_with_idata(new_idata)
 
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -218,10 +218,8 @@ class CLVModel(ModelBuilder):
     def _create_distribution(dist: dict, shape=()):
         try:
             return getattr(pm, dist["dist"]).dist(**dist.get("kwargs", {}), shape=shape)
-        except AttributeError as err:
-            raise ValueError(
-                f"Distribution {dist['dist']} does not exist in PyMC"
-            ) from err
+        except AttributeError:
+            raise ValueError(f"Distribution {dist['dist']} does not exist in PyMC")
 
     @property
     def default_sampler_config(self) -> dict:
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -242,7 +242,7 @@ class CLVModel(ModelBuilder):
         if self.idata is None:
             self.idata = res
         elif "posterior" in self.idata:
-            warnings.warn("Overriding pre-existing fit_result", stacklevel=2)
+            warnings.warn("Overriding pre-existing fit_result", stacklevel=1)
             self.idata.posterior = res
         else:
             self.idata.posterior = res
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -240,7 +240,7 @@ class CLVModel(ModelBuilder):
         if self.idata is None:
             self.idata = res
         elif "posterior" in self.idata:
-            warnings.warn("Overriding pre-existing fit_result")  # noqa: B028
+            warnings.warn("Overriding pre-existing fit_result", stacklevel=1)
             self.idata.posterior = res
         else:
             self.idata.posterior = res
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -206,7 +206,8 @@ class CLVModel(ModelBuilder):
             )
 
         """
-        self.fit_result  # Raise Error if fit didn't happen yet
+        if not hasattr(self, "fit_result"):
+            raise RuntimeError("The model hasn't been fit yet, call .fit() first")
         assert self.idata is not None
         new_idata = self.idata.isel(draw=slice(None, None, keep_every)).copy()
         return type(self)._build_with_idata(new_idata)
@@ -215,8 +216,10 @@ class CLVModel(ModelBuilder):
     def _create_distribution(dist: dict, shape=()):
         try:
             return getattr(pm, dist["dist"]).dist(**dist.get("kwargs", {}), shape=shape)
-        except AttributeError:
-            raise ValueError(f"Distribution {dist['dist']} does not exist in PyMC")
+        except AttributeError as err:
+            raise ValueError(
+                f"Distribution {dist['dist']} does not exist in PyMC"
+            ) from err
 
     @property
     def default_sampler_config(self) -> dict:
@@ -237,7 +240,7 @@ class CLVModel(ModelBuilder):
         if self.idata is None:
             self.idata = res
         elif "posterior" in self.idata:
-            warnings.warn("Overriding pre-existing fit_result")
+            warnings.warn("Overriding pre-existing fit_result")  # noqa: B028
             self.idata.posterior = res
         else:
             self.idata.posterior = res
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -57,8 +57,7 @@ class CLVModel(ModelBuilder):
                 category=UserWarning,
                 message="The group fit_data is not defined in the InferenceData scheme",
             )
-            if self.idata is None:
-                raise ValueError("idata is not defined")
+            assert self.idata is not None  # noqa: S101
             self.idata.add_groups(fit_data=data.to_xarray())
 
     def fit(  # type: ignore
@@ -207,10 +206,8 @@ class CLVModel(ModelBuilder):
             )
 
         """
-        if not hasattr(self, "fit_result"):
-            raise RuntimeError("The model hasn't been fit yet, call .fit() first")
-        if self.idata is None:
-            raise RuntimeError("The model hasn't been fit yet, call .fit() first")
+        self.fit_result  # noqa: B018 (Raise Error if fit didn't happen yet)
+        assert self.idata is not None  # noqa: S101
         new_idata = self.idata.isel(draw=slice(None, None, keep_every)).copy()
         return type(self)._build_with_idata(new_idata)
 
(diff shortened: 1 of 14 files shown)


ricardoV94:
The missing asserts are embarrassing.
The zip is fine

Not convinced by the raises and the stscklevels

juanitorduz:
> The missing asserts are embarrassing.
This is why I like this ! haha

> Not convinced by the raises and the stacklevels
I can #noqa these changes :)

ricardoV94:
> I can #noqa these changes :)

The raises are very opinionated, it looked like the code was specifically looking for the exception and converting it to something else for the user. I don't know of noqa makes more sense than disabling altogether 

The stscklevel I simply don't get. Isn't 1 the default?

ricardoV94:
Also for the zip, is it forcing a default or asking you to be explicit?

juanitorduz:
> Also for the zip, is it forcing a default or asking you to be explicit?

Asking to be explicit

> The raises are very opinionated, it looked like the code was specifically looking for the exception and converting it to something else for the user. I don't know of noqa makes more sense than disabling altogether

I do not have a strong opinion here. Maybe we can simply ignore that rule.

> The stacklevel I simply don't get. Isn't 1 the default?

The stacklevel needs to be explicit to I am putting the default. Qw can #noqa these ones.

WDYT?

wd60622:
> > I can #noqa these changes :)
> 
> The raises are very opinionated, it looked like the code was specifically looking for the exception and converting it to something else for the user. I don't know of noqa makes more sense than disabling altogether
> 
> The stacklevel I simply don't get. Isn't 1 the default?

It might be the default, but isn't it recommended to have at least 2 or more? else nothing shows up?

wd60622:
> The missing asserts are embarrassing.

Think I am the culprit for a few a these 🙈

juanitorduz:
> > > I can #noqa these changes :)
> > 
> > 
> > The raises are very opinionated, it looked like the code was specifically looking for the exception and converting it to something else for the user. I don't know of noqa makes more sense than disabling altogether
> > The stacklevel I simply don't get. Isn't 1 the default?
> 
> It might be the default, but isn't it recommended to have at least 2 or more? else nothing shows up?

Copilot suggests 2 XD so we can use that and not the default hehe

juanitorduz:
> > The missing asserts are embarrassing.
> 
> Think I am the culprit for a few a these 🙈

This is why I like this plugin, I did it many times and probably will. So I need a system to prevent this haha

ricardoV94:
the level depends on what you want the user to see as the source of the warning. It's usually one, but sometimes it's called by a helper or a very internal function and 2 (or 3 or whatever) is more appropriate. I don't think 2 is recommended blindly?

juanitorduz:
So then we can take it case by case. This is good imo to have to think about this :)

juanitorduz:
> the level depends on what you want the user to see as the source of the warning. It's usually one, but sometimes it's called by a helper or a very internal function and 2 (or 3 or whatever) is more appropriate. I don't think 2 is recommended blindly?

I suggest we leave `stacklevel=1` as the warnings are raised in public methods.

juanitorduz:
Personally, I like to know where the errors are coming from so I am in for keeping the `from err` addition to the warnings. @ricardoV94 , you have a better experience with user-facing errors, so if you think these are generally more confusing for the user, then I will ignore the result and revert the changes 🙏

ricardoV94:
> Personally, I like to know where the errors are coming from so I am in for keeping the `from err` addition to the warnings. @ricardoV94 , you have a better experience with user-facing errors, so if you think these are generally more confusing for the user, then I will ignore the result and revert the changes 🙏

Some errors are being expected/ handled by us, so there's not much to gain from passing the original error. No strong opinion though.

In this example, I don't see what's the point of raising the original exception. The user may have no idea about how PyMC model contexts work, and I would guess the whole trace-back would be more confusing than our curated error.
```python
def user_facing_foo():
  # This function must be called in a model context.
  try:
    pm.model_context(None)
  except ModelContextError:   # or whatever
    raise RuntimeError("You must call this function from within a PyMC model")
```

juanitorduz:
Yep! That makes sense! Then let's reverb and ignore this rule and make sure we think about the user when reporting errors (as we have been doing) 🙌

juanitorduz:
@ricardoV94  done ✅

pymc_marketing/clv/models/basic.py
```diff
@@ -57,7 +57,8 @@ def _add_fit_data_group(self, data: pd.DataFrame) -> None:
                 category=UserWarning,
                 message="The group fit_data is not defined in the InferenceData scheme",
             )
-            assert self.idata is not None
+            if self.idata is None:
+                raise ValueError("idata is not defined")
```
ricardoV94, line 61: I don't like these, because they are just a work-around stupid mypy. We know that this method can only be called after we already checked idata exists

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -56,7 +56,8 @@ def expected_customer_spend(
         mean_transaction_value, frequency = to_xarray(
             customer_id, mean_transaction_value, frequency
         )
-        assert self.idata is not None, "Model must be fitted first"
+        if self.idata is None:
+            raise ValueError("Model must be fitted first")
```
ricardoV94, line 60: This one is fine, but could be replaced by `posterior = self.fit_result`, then used below. Again that will take care of checking idata exists and raising a useful error.

The point of all this is to avoid code duplication of `model must be fit first`

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -89,7 +90,8 @@ def distribution_new_customer_spend(
     def expected_new_customer_spend(self) -> xarray.DataArray:
         """Expected transaction value for a new customer"""
 
-        assert self.idata is not None, "Model must be fitted first"
+        if self.idata is None:
+            raise ValueError("Model must be fitted first")
```
ricardoV94, line 94: same as above

tests/model_builder/test_model_builder.py
```diff
@@ -31,20 +31,19 @@
...
-    X = pd.DataFrame({"input": x})
-    return X
+    return pd.DataFrame({"input": x})
 
 
 @pytest.fixture(scope="module")
 def toy_y(toy_X):
     y = 5 * toy_X["input"] + 3
-    y = y + np.random.normal(0, 1, size=len(toy_X))
+    y = y + rng.normal(0, 1, size=len(toy_X))
```
ricardoV94, line 40: How do these changes work. Is there a global `rng` object?

tests/model_builder/test_model_builder.py
```diff
@@ -182,7 +180,7 @@ def test_save_load(fitted_model_instance):
     test_builder2 = ModelBuilderTest.load(temp.name)
     assert fitted_model_instance.idata.groups() == test_builder2.idata.groups()
     assert fitted_model_instance.id == test_builder2.id
-    x_pred = np.random.uniform(low=0, high=1, size=100)
+    x_pred = rng.uniform(low=0, high=1, size=100)
```
ricardoV94, line 183: In general seeds should be local to the tests, so that they don't depend on the order or subset of tests that are run. The old tests were not seeded in the first place, which was even worse

juanitorduz: Ups! You are right! 🙈 let me fix this 🙏

juanitorduz: fixed in https://github.com/pymc-labs/pymc-marketing/pull/637/commits/f9dc0dc2371a3375ddc094d36d26ede25a665cf0

tests/model_builder/test_model_builder.py
juanitorduz: yes, but now I change it so that they are independent, see https://github.com/pymc-labs/pymc-marketing/pull/637/commits/f9dc0dc2371a3375ddc094d36d26ede25a665cf0

pymc_marketing/clv/models/gamma_gamma.py
juanitorduz: solved in https://github.com/pymc-labs/pymc-marketing/pull/637/commits/f9dc0dc2371a3375ddc094d36d26ede25a665cf0, however I could not use self.idata.posterior because mypy 🙈

pymc_marketing/clv/models/basic.py
```diff
@@ -206,8 +207,10 @@ def thin_fit_result(self, keep_every: int):
             )
 
         """
-        self.fit_result  # Raise Error if fit didn't happen yet
-        assert self.idata is not None
+        if not hasattr(self, "fit_result"):
+            raise RuntimeError("The model hasn't been fit yet, call .fit() first")
```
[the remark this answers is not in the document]
juanitorduz: Reverted  https://github.com/pymc-labs/pymc-marketing/commit/f9dc0dc2371a3375ddc094d36d26ede25a665cf0 :)

pymc_marketing/clv/models/basic.py
juanitorduz: Reverted  https://github.com/pymc-labs/pymc-marketing/commit/f9dc0dc2371a3375ddc094d36d26ede25a665cf0 :)

pymc_marketing/clv/models/gamma_gamma.py
juanitorduz: Solved in https://github.com/pymc-labs/pymc-marketing/pull/637/commits/89ca03516b41f0610e56f141fb9ce70f065b40a4

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -56,10 +56,12 @@ def expected_customer_spend(
         mean_transaction_value, frequency = to_xarray(
             customer_id, mean_transaction_value, frequency
         )
-        assert self.idata is not None, "Model must be fitted first"
-        p = self.idata.posterior["p"]
-        q = self.idata.posterior["q"]
-        v = self.idata.posterior["v"]
+        if (posterior := self.fit_result) is None:
+            raise ValueError("Model must be fitted first")
```
wd60622: Won't this never reach since fit_result either returns InferenceData or raises?

wd60622:
```python
# Still would have the RuntimeError if not fit, right?
posterior = self.fit_results
```

ricardoV94, line 60: This doesn't work? self.fit_result will already raise if needed

```suggestion
        posterior = self.fit_result
```

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -89,10 +91,12 @@ def distribution_new_customer_spend(
     def expected_new_customer_spend(self) -> xarray.DataArray:
         """Expected transaction value for a new customer"""
 
-        assert self.idata is not None, "Model must be fitted first"
-        p_mean = self.idata.posterior["p"]
-        q_mean = self.idata.posterior["q"]
-        v_mean = self.idata.posterior["v"]
+        if (posterior := self.fit_result) is None:
```
ricardoV94, line 94: Same here

juanitorduz: Ahh got it! Fixed in https://github.com/pymc-labs/pymc-marketing/pull/637/commits/7e0689358e5f268dd6a8db50d1fad0ba025a9954

ricardoV94:
> @ricardoV94  It was my first time using the `:=` operator 😆 

Me too haven't found a good use for it yet

ricardoV94 (approved):

wd60622 (commented):
Comment on the fit_result attribute already causing an error to raise

wd60622 (approved):
Was commenting same as ricardo 
Looks good to me!

juanitorduz:
Thank you @wd60622 and @ricardoV94  🎸 !

PR #662, opened 2024-05-01
Remove unnecessary NonImplemented errors from abstract methods

juanitorduz:
We want to have a more precise metric of code coverage.

----
📚 Documentation preview 📚: https://pymc-marketing--662.org.readthedocs.build/en/662/


wd60622:
Which lines do you have in mind? Looking at the cov report, NotImplementedError seems to be used in various ways. Maybe alternative errors can be raised

juanitorduz:
It is indeed for the abstract methods 😄. Some while ago I wrote trivial tests for them so that the coverage did not go down because if this. Ricardo pointed me out this was an overkill. So shall we maybe do not raise any error (I did not in now it was raised automatically).

wd60622:
Well can the function body be removed for abstract methods?

wd60622 (approved):
LGTM!

Issue #666, opened 2024-05-02
Example NB plots are compressed


twiecki:
<img width="1235" alt="image" src="https://github.com/pymc-labs/pymc-marketing/assets/674200/a130a99f-e3e6-4ea2-82bf-4e6d0dc6708d">
 
https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_lift_test.html

wd60622:
noticed that too ☹️

ColtAllen:
This is also impacting the CLV notebooks. I tried changing the fig size in the plotting code, but it made no difference. I'm thinking we need to specify something in `conf.py` related to `pydata-sphinx-theme`.

twiecki:
Maybe @OriolAbril has an idea.

juanitorduz:
can it be related to https://github.com/executablebooks/MyST-NB/pull/588?

juanitorduz:
mmmm when I build the docs locally `make html` it works fine :/

juanitorduz:
Here is a quick fix https://github.com/pymc-labs/pymc-marketing/pull/667 . We might need a better solution. I think one has to tweak the `nb_cell_metadata_key`, but I did not manage to do it :(

juanitorduz:
I will re-open this issue as the fix in https://github.com/pymc-labs/pymc-marketing/pull/667 should not be final as we do not want to restrict the versions :)

OriolAbril:
I added a CSS patch for this to be able to remove the pin on myst-nb in https://github.com/pymc-labs/pymc-marketing/pull/1984/commits/4881e95e51bce0d4dc4f880c1df4c5374ffec32a. I will try looking into the root issue and contribute a fix upstream. Support for translatable notebooks requires a version above 1.1.2 or something like that

PR #667, opened 2024-05-03
[Try] Fix compressed images in docs.

juanitorduz:
Trying to fix https://github.com/pymc-labs/pymc-marketing/issues/666

----
📚 Documentation preview 📚: https://pymc-marketing--667.org.readthedocs.build/en/667/


juanitorduz:
ok It seems restricting the `myst_nb` version fixes the problem. @OriolAbril is there a better way? 

I suggest we merge this one and then come back with a better fix. WDYT?

twiecki:
That does fix it.

twiecki (approved):

Issue #523, opened 2024-02-09
Add `rfm_segmentation` Utility


ColtAllen:
[RFM Segmentation](https://www.optimove.com/resources/learning-center/rfm-segmentation) is a way to segment customers on their past purchasing behavior. This is a complementary technique to CLV/BTYD modeling as customers in different segments benefit from different marketing strategies, and would be a good bridge between [combining MMM and CLV](https://github.com/pymc-labs/pymc-marketing/issues/64).

A traditional rules-based RFM segmentation would be a straightforward addition, and I also have some ideas on a Bayesian clustering variant.

sarim-zafar:
`I also have some ideas on a Bayesian clustering variant.`
My recent experience with an RFM segmentation project led me to discover the effectiveness of utilizing [jenks breaks](https://en.wikipedia.org/wiki/Jenks_natural_breaks_optimization) methodology over simple percentile-based segmentation. I found Jenks breaks logic to offer more meaningful segmentations. I suggest considering Jenks breaks as it could potentially enhance the segmentation process. There is a C-based implementation that can be found [here](https://pypi.org/project/jenkspy/).

ColtAllen:
Hey @sarim-zafar,

Thanks for sharing. The Jenks Breaks methodology has some parallels with [Dirichlet processes](https://en.wikipedia.org/wiki/Dirichlet_process), which can be used to automatically infer the optimal number of clusters.

PR #680, opened 2024-05-11
Rfm segment

ColtAllen:
## Description

This PR adds the RFM segmentation functionality described in the [recent CLV webinar](https://www.youtube.com/watch?v=u3oMWgStIZY). This [article](https://www.barilliance.com/rfm-analysis/#) is a good deep dive into how the segmentation is performed.

There are two TODOs left in `clv.utils.rfm_segments` I'm interested in getting some feedback on, but they're quite minor. The dev notebook used for testing could also be expanded into a tutorial in the future, but we'll leave that for another PR as there are some plotting methods I wish to add first.

## Related Issue

- [x] Closes # https://github.com/pymc-labs/pymc-marketing/issues/523
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [x] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):


ColtAllen:
I'm confused about this Codecov report, because `pytest-cov` reports 100% code coverage in my IDE and many of the code lines in that report aren't even relevant to this PR.

pymc_marketing/clv/utils.py
```diff
@@ -597,3 +598,185 @@ def rfm_train_test_split(
...
     return train_test_rfm_data
+
+
+def rfm_segments(
+    transactions: pd.DataFrame,
+    customer_id_col: str,
+    datetime_col: str,
+    monetary_value_col: str,
+    segment_config: dict | None = None,
+    observation_period_end: str | pd.Period | datetime | None | None = None,
```
juanitorduz, line 609: There are two `None`s :)

juanitorduz: Also below in all the remaining function arguments ;)

pymc_marketing/clv/utils.py
```diff
@@ -597,3 +598,185 @@ def rfm_train_test_split(
...
+    datetime_format: str | None | None = None,
+    time_unit: str = "D",
+    time_scaler: float | None | None = 1,
+    sort_transactions: bool | None | None = True,
+) -> pd.DataFrame:
+    """
```
juanitorduz, line 615: I think this function will be very useful! Marketing practitioners love RFM segments. Hence, I suggest we add more context about the method (high level) in the docstring as in https://www.pymc-marketing.io/en/stable/api/generated/pymc_marketing.mmm.delayed_saturated_mmm.DelayedSaturatedMMM.html#pymc_marketing.mmm.delayed_saturated_mmm.DelayedSaturatedMMM (for example)

pymc_marketing/clv/utils.py
```diff
@@ -597,3 +598,185 @@ def rfm_train_test_split(
...
+        Dictionary containing segment names and list of RFM score assignments;
+        key/value pairs should be formatted as `{"segment": ['111', '123', '321'], ...}`.
+        If not provided, the following segment names and definitions are used:
+            "Premium Customer": Customers in top 2 quartiles for all variables.
+            "Repeat Customer": Customers in top 2 quartiles for frequency, and either recency or monetary value.
+            "Top Spender": Customers in top 2 quartiles for monetary value, and either frequency or recency.
+            "At-Risk Customer": Customers in bottom 2 quartiles for two or more variables.
+            "Inactive Customer": Customers in bottom quartile for two or more variables.
+        Customers with unspecified RFM scores will be assigned to a segment named "Other".
+    observation_period_end: Union[str, pd.Period, datetime], optional
```
juanitorduz, line 647: These can be None right? So `Union[str, pd.Period, datetime, None]` ?

juanitorduz: Same for the variables below

pymc_marketing/clv/utils.py
```diff
@@ -597,3 +598,185 @@ def rfm_train_test_split(
...
+    for column_name in zip(
+        ["r_quartile", "f_quartile", "m_quartile"],
+        ["recency", "frequency", "monetary_value"],
+        strict=False,
+    ):
+
+        def _quartile_labels(column_name, max_label_range):
+            # recency labels must be reversed because lower values are more desirable
+            if column_name == "r_quartile":
+                return list(reversed(list(range(1, max_label_range))))
```
juanitorduz, line 698: What if we do `list(range(max_label_range - 1, 0, -1))` to simplify the code? 

Also, shall we extract this function and add a simple unitest? Even `list(range(m - 1, 0, -1))` might  not be straightforward to verify on the fly ;)

pymc_marketing/clv/utils.py
```diff
@@ -597,3 +598,185 @@ def rfm_train_test_split(
...
+            else:
+                return range(1, max_label_range)
+
+        try:
+            labels = _quartile_labels(column_name[0], 4)
```
juanitorduz, line 703: Will there be any case where we would like to have `max_label_range != 4`?

juanitorduz: (and `max_label_range != 3`, `max_label_range != 5` below?

pymc_marketing/clv/utils.py
```diff
@@ -597,3 +598,185 @@ def rfm_train_test_split(
...
+                    rfm_data[column_name[1]], q=4, labels=labels, duplicates="drop"
+                ).astype(str)
+            except ValueError:
+                labels = _quartile_labels(column_name[0], 5)
+                rfm_data[column_name[0]] = pd.qcut(
+                    rfm_data[column_name[1]], q=4, labels=labels, duplicates="drop"
+                ).astype(str)
+
+    # TODO: Is `finally` block also needed?
+    rfm_data["rfm_score"] = (
```
juanitorduz, line 720: Just as a personal preference (no objective reason), I like https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.eval.html for simple arithmetic operations between columns :D

pymc_marketing/clv/utils.py
```diff
@@ -597,3 +598,185 @@ def rfm_train_test_split(
...
+        rfm_data["r_quartile"] + rfm_data["f_quartile"] + rfm_data["m_quartile"]
+    )
+
+    if segment_config is None:
+        segment_config = {
```
juanitorduz, line 725: Shall we put this default config in a `constants.py` module with an explanation of the meaning in the docstrings . For example

```python
# constants.py

def _get_default_rfm_config() -> dict[str, list[str]]:
    " This default configuration for the RFM model consists of the keys ... "
    return { ... }
```

juanitorduz: Especially, explaining what does `[334]` mean (for example).

pymc_marketing/clv/utils.py
```diff
@@ -597,3 +598,185 @@ def rfm_train_test_split(
...
+    segment_names = list(segment_config.keys())
+
+    # create catch-all "Other" segment and assign defined segments from config
+    rfm_data["segment"] = "Other"
+
+    for key in segment_names:
+        rfm_data.loc[rfm_data["rfm_score"].isin(segment_config[key]), "segment"] = key
+
+    # drop unnecessary columns
+    # TODO: Is dropping the rfm_score column prudent?
```
juanitorduz, line 777: What about if we add this as a boolean option so that the user can choose?

tests/clv/test_utils.py
```diff
@@ -849,4 +850,34 @@ def test_rfm_train_test_split_with_monetary_value(self, transaction_data):
...
+
+        actual = rfm_segments(
+            transaction_data,
+            "id",
+            "date",
+            "monetary_value",
+            segment_config=config,
+        )
+
+        assert (actual["segment"] == expected).all()
```
juanitorduz, line 883: Can we add tests on the logic and expected outcome of the segmentation itself?

juanitorduz (changes requested):
RFM segmentation is always in marketeers tool box 🚀 . 

I left some comments :) Great work!

juanitorduz:
In a different PR we can tackle the example notebook and maybe add som plots on how to evaluate and understand these segments

juanitorduz:
> I'm confused about this Codecov report, because `pytest-cov` reports 100% code coverage in my IDE and many of the code lines in that report aren't even relevant to this PR.

Unfortunately, I do not understand why this is happening either :( . Maybe once so add the missing parts of this PR it will come back ¯\\_(ツ)_/¯.

pymc_marketing/clv/utils.py
juanitorduz: Thank you for the explanation! It might be worth adding this comments in the doc-strings (or code comment)

pymc_marketing/clv/utils.py
ColtAllen: Anything else you can think of (including from the MMM side) that can be moved into this `constants.py` module? I support relocating this config into an external variable to clean up the function code, but I'm loathe to create a entirely new module for a single dictionary.

pymc_marketing/clv/utils.py
ColtAllen: I'll get it added. Can also add columns for the bin edges.

pymc_marketing/clv/utils.py
ColtAllen: Whoops; updating the minimum python version to 3.10 seems to have changed the type hinting syntax. Will get this sorted out.

pymc_marketing/clv/utils.py
ColtAllen: Probably more memory safe, so this has been modified.

ColtAllen:
Still need to add a test or two and address https://github.com/pymc-labs/pymc-marketing/issues/697, which may necessitate some refactoring to allow users to use their own segmentation methods. Might even roll https://github.com/pymc-labs/pymc-marketing/pull/698 into this too depending on timing.

juanitorduz:
Maybe we can leave out https://github.com/pymc-labs/pymc-marketing/pull/698 as the PR is open and we are just missing some tests. It should not be a blocker for this one ;)

ColtAllen:
> Maybe we can leave out #698 as the PR is open and we are just missing some tests. It should not be a blocker for this one ;)

Extra tests are unnecessary; just needed to rename a column in the small dataset used for testing. The entire fix took five minutes.

ColtAllen:
Also added a test and warning if only 2 bins are created. This is unlikely unless dealing with very small and/or invalid datasets that a model probably wouldn't converge on anyway, but nonetheless a custom segment config would be required.

juanitorduz (approved):

juanitorduz:
@ColtAllen Would you mind if we merge https://github.com/pymc-labs/pymc-marketing/pull/698 and then you do a small rebase? We wanna make sure we welcome contributions from everyone :)

Issue #703, opened 2024-05-26
Refactor `logp` in BG/BB to remove Scan


ColtAllen:
`logp` in the `BetaGeoBetaBinom` distribution block contains an iterable currently serviced by a `Scan` from `pytensor`.  It's possible to refactor this so that `Scan` is no longer needed:

```python
i = pt.scalar("i", dtype=int)
died = pt.lt(t_x + i, T)

unnorm_logp_died_at_tx_plus_i = pt.where(
    pt.ge(t_x, i),
    (
        betaln(alpha + x, beta + t_x - x + i)
        + betaln(gamma + died, delta + t_x + i)
    ),
    -np.inf
)

#Maximum prevents invalid T - t_x values from crashing logp
max_range = pt.maximum(pt.max(T - t_x), 0)
i_vec = pt.arange(max_range + 1)
unnorm_logp_died_at_tx_plus_i_vec = vectorize_graph(
    unnorm_logp_died_at_tx_plus_i,
    replace={i: i_vec},
)

unnorm_logp = pt.logsumexp(unnorm_logp_died_at_tx_plus_i_vec, axis=0)
```

I compared both approaches in a dev notebook, and sans Scan is about 3x faster: 
```
# w/ Scan
267 ms ± 6.69 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# w/o Scan
85.2 ms ± 339 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

However, the above code requires modification because tests are failing with the returned `logp` values.

ricardoV94:
Scan may be plenty fast in other backends: numba and jax, the first will be the default sometime in the future. I would benchmark that before trying to get rid of it. Also for ragged datasets the non scan will probably be slower as it does a lot of useless computations

juanitorduz:
ok! thanks for the input! I took the PR because I always wanna play with `scan`, but we can close it and have other benchmarks. We can always come back and change it, as we have the code in a branch already.

ColtAllen:
@ricardoV94 do you have a time estimate on when `numba` will became the new default backend? I'm working on the BG/BB model right now, and currently NUTS is taking over an hour on my Macbook M2 Pro with a dataset of 11.2k rows.

ricardoV94:
You can select other backend manually, don't need to wait for the default to change

juanitorduz:
[...]

PR #707, opened 2024-05-30
Replace scan in BetaGeoBetaBinom

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/703

----
📚 Documentation preview 📚: https://pymc-marketing--707.org.readthedocs.build/en/707/


juanitorduz:
@ColtAllen Thanks for the nice issue description. I removed the `pt.ge(t_x, i)` from the `unnorm_logp_died_at_tx_plus_i` expression and this solved the scalar case, but something is missing in the vectorized one:

```python
tests/clv/test_distributions.py::TestBetaGeoBetaBinom::test_logp_matches_lifetimes[scalar] PASSED                   [ 50%]
tests/clv/test_distributions.py::TestBetaGeoBetaBinom::test_logp_matches_lifetimes[vector] FAILED                   [100%]
```

with

```python
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           
E           Mismatched elements: 4 / 5 (80%)
E           Max absolute difference: 0.0505898
E           Max relative difference: 0.18415497
E            x: array([ -0.360687, -21.684232,  -6.955498,  -0.224123,  -0.360694])
E            y: array([ -0.36069 , -21.696809,  -6.9558  ,  -0.274713,  -0.360694])
```

any ideas?

pymc_marketing/clv/distributions.py
```diff
@@ -609,28 +608,23 @@ def logp(value, alpha, beta, gamma, delta, T):
...
+        unnorm_logp_died_at_tx_plus_i = betaln(alpha + x, beta + t_x - x + i) + betaln(
+            gamma + died, delta + t_x + i
+        )
 
-        unnorm_logp, _ = scan(
-            fn=logp_customer_died,
-            outputs_info=[None],
-            sequences=[t_x, x, alpha, beta, gamma, delta, T],
+        # Maximum prevents invalid T - t_x values from crashing logp
+        max_range = pt.maximum(pt.max(T - t_x), 0)
```
juanitorduz, line 619: Actually I think here might be the issue as this maximum varies over “i” in the scan implementation and therefore this global maximum will not match 🤔

juanitorduz: And not I think I got the reason of pt.ge(t_x, i) from the unnorm_logp_died_at_tx_plus_i on the first place 🙈 🤦

ColtAllen:
CC'ing @ricardoV94 (he wrote this code).

pymc_marketing/clv/distributions.py
ColtAllen: `pt.maximum` seems redundant to me because `T` is the same value across all customers for this model; it'll never be less than 0. Maybe add/subtract 1 from `max_range` so that it indexes properly?

tests/clv/test_distributions.py
```diff
@@ -278,7 +278,9 @@ def test_pareto_nbd_sample_prior(
 
 
 class TestBetaGeoBetaBinom:
-    @pytest.mark.parametrize("batch_shape", [(), (5,)])
+    @pytest.mark.parametrize(
+        argnames="batch_shape", argvalues=[(), (5,)], ids=["scalar", "vector"]
+    )
     def test_logp_matches_lifetimes(self, batch_shape):
```
ColtAllen, line 284: Lines 298:
```
T = pm.draw(pm.DiscreteUniform.dist(1, 10, shape=batch_shape), random_seed=rng)
```
Setting `T` to a constant of 10 would be more realistic, but I'm not sure this will fix things.

This test was added out of a desire to compare with `lifetimes._likelihood`, but the ground truth values for when `lifetimes` was originally built & tested are in `test_logp_matches_excel`. If a vectorized likelihood matches those values, then this is an unnecessary test.

pymc_marketing/clv/distributions.py
ricardoV94: The maximum is just a sanity check. For constant data it will be constant folded (evaled once and remove from the graph)

ricardoV94: > And not I think I got the reason of pt.ge(t_x, i) from the unnorm_logp_died_at_tx_plus_i on the first place 🙈 🤦

I don't understand this comment

ricardoV94: The formula is slightly different after the subject "died", in that case it needs a +1 in one of the terms, that what it was doing

juanitorduz: So about the comment above: "I was just saying I understood the reason of the `np.where` and I think I fixed in https://github.com/pymc-labs/pymc-marketing/pull/707/commits/da6568c846858723cc31ea931e51f3a6c0d7c375

tests/clv/test_distributions.py
juanitorduz: I think it was worth to have as it helped me to get https://github.com/pymc-labs/pymc-marketing/pull/707/commits/da6568c846858723cc31ea931e51f3a6c0d7c375

juanitorduz:
@ricardoV94 I think I got it and the tests are passing :) . Anything else I am missing?

ricardoV94:
Changes look fine, although I would suggest a more careful benchmark before deciding on it.

Specifically other backends and cases that should be more favourable to Scan (large variation between T and tx across rows).

I gave some more details on the original issue

ricardoV94:
And more generally when we do performance driven changes I would add a benchmark as part of the CI (pytest-benchmark is pretty nice). That's the conceptual equivalent to a regression test for bugfixes.

juanitorduz:
Despite the changes, do what we want. We wanna do more benchmarking to decide on the implementation to choose. See https://github.com/pymc-labs/pymc-marketing/issues/703#issuecomment-2141495580

Issue #527, opened 2024-02-09
CLV API Standardization


ColtAllen:
## TODO List (best to give each one its own PR)
- [ ] Naming consistencies between models
- [ ] Individual array vs Dataframe args. **What `xarray` broadcasting functionality is lost with the latter?**
- [ ] Make args optional for in-sample data
- [ ] Fix `BetaGeoModel` documentation errors
- [ ] Add CLV support for `ShiftedBetaGeoModelIndividual` 
- [ ] Add `UserWarnings` for API breaking changes

There are API inconsistencies in the CLV module. Here's a breakdown by model:

## Beta-Geo/NBD Transactions Model
[I found some errors in the BG/NBD docstring examples while writing this](https://www.pymc-marketing.io/en/stable/api/generated/pymc_marketing.clv.models.beta_geo.BetaGeoModel.html#pymc_marketing.clv.models.beta_geo.BetaGeoModel)
```
rfm_data = pd.DataFrame(
            {
                "customer_id": customer_id,
                "frequency": frequency,
                "recency": recency,
                "T": T,
            }
        )

model = BetaGeoModel(data=rfm_df)
model.build_model()
model.fit()

model.expected_num_purchases(
    customer_id=rfm_data["customer_id"],
    t=10,
    frequency=rfm_data["frequency"],
    recency=rfm_data["recency"],
    T=rfm_data["T"],
)
```

Note how fit data is provided as a dataframe, but in the predictive methods individual arrays must be provided. Specifying one array at a time was one of the most annoying aspects about using the legacy `lifetimes` library, and sometimes even created indexing issues that caused the underlying `scipy` functions to crash.

For `ParetoNBDModel`, I streamlined this nonsense with a dataframe argument, and made it optional if running predictions on the fit dataset:

## Pareto/NBD Transactions Model
```
rfm_data = pd.DataFrame(
            {
                "customer_id": customer_id,
                "frequency": frequency,
                "recency": recency,
                "T": T,
            }
        )

model = ParetoNBDModel(data=rfm_data)
model.build_model()
model.fit()

# Data param is optional and only required for out-of-sample data
model.expected_purchases(future_t=10)

model.expected_purchases(
    data=future_rfm_df,
    future_t=10,
)
```

(We will also need to resolve the naming inconsistencies between these models.)

[...]

PR #709, opened 2024-06-01
Update `BetaGeoModel` API

ColtAllen:
## Description

This PR upgrades the user API for `BetaGeoModel` and standardizes it with `ParetoNBDModel`. This should also speed up the CI runs a bit because there was a lot of redundancy in the old BG/NBD tests.

The `expected_num_purchases` method needed to be retained with a deprecation warning to prevent breaking changes elsewhere. These changes will be implemented in follow-up PRs for https://github.com/pymc-labs/pymc-marketing/issues/527.

## Related Issue

- [x] Related to # https://github.com/pymc-labs/pymc-marketing/issues/527

## Checklist

## Modules affected

- [ ] MMM
- [x] CLV

## Type of change

- [ ] Bug fix
- [x] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--709.org.readthedocs.build/en/709/


pymc_marketing/clv/models/beta_geo.py
```diff
@@ -92,33 +84,22 @@ class BetaGeoModel(CLVModel):
                 "cores": 2,
             },
         )
-        model.build_model()
-        model.fit()
+        # Fit model quickly to large datasets via the default Maximum a Posteriori method
+        model.fit(fit_method='map')
```
juanitorduz, line 88: can you use " instead of ' for the string?

pymc_marketing/clv/plotting.py
```diff
@@ -332,33 +332,26 @@ def plot_probability_alive_matrix(
...
-    #  We should harmonize them!
+    transaction_data = pd.DataFrame(
+        {
+            "customer_id": np.arange(mesh_recency.size),  # placeholder
+            "frequency": mesh_frequency.ravel(),
+            "recency": mesh_recency.ravel(),
+            "T": max_recency,
+        }
+    )
     if isinstance(model, ParetoNBDModel):
```
juanitorduz, line 343: Can we remove these if statements with this channel? If not, what is missing?

Also, can we do something similar in the function above https://github.com/ColtAllen/pymc-marketing/blob/bgnbd_api/pymc_marketing/clv/plotting.py#L227 ?

pymc_marketing/clv/plotting.py
```diff
@@ -332,33 +332,26 @@ def plot_probability_alive_matrix(
...
-                "recency": mesh_recency.ravel(),
-                "T": max_recency,
-            }
-        )
-
         Z = (
             model.expected_probability_alive(
                 data=transaction_data,
-                future_t=0,  # TODO: This can be a function parameter in the case of ParetoNBDModel
+                future_t=0,  # TODO: This if block can be removed if
```
juanitorduz, line 347: Can be removed if what XD ?

juanitorduz (commented):
Hey! Thanks for this PR! I left some comments, which are more questions :)

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -11,13 +11,14 @@
...
 #   limitations under the License.
+import warnings
 from collections.abc import Sequence
 
 import numpy as np
 import pandas as pd
 import pymc as pm
 import pytensor.tensor as pt
-import xarray as xr
+import xarray
```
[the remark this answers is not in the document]
ColtAllen: Consistency with `ParetoNBDModel`. I can change it back, but it's splitting hairs.

pymc_marketing/clv/plotting.py
ColtAllen: I suppose there are worse things to rush through than TODO comments 😆

pymc_marketing/clv/plotting.py
ColtAllen: I don't want to introduce too much scope creep in this PR, and this would be the very next thing I work on after the `BetaGeoModel` changes get merged.

pymc_marketing/clv/models/beta_geo.py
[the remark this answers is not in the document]
ColtAllen: There are way more issues in that docstring than just this haha. The docs prebuild preview was very helpful in finding them.

My general writing approach is to be concise, and limit technical jargon as much as possible. Let me know if the updated docstrings pass judgment.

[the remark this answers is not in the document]
wd60622: Doesn't the ruff formatter change ' to "?

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -92,44 +101,35 @@ class BetaGeoModel(CLVModel):
...
-            t=future_t,
-        )
+        # Predict number of purchases for a new customer over 't' time periods
+        expected_purchases_new_customer = model.expected_purchases_new_customer(t=10)
 
     References
     ----------
-    .. [1] Fader, P. S., Hardie, B. G., & Lee, K. L. (2005). “Counting your customers”
-           the easy way: An alternative to the Pareto/NBD model. Marketing science,
+    .. [1] Fader, P. S., Hardie, B. G., & Lee, K. L. (2005). “Counting your customers
```
wd60622, line 125:
```suggestion
    .. [1] Fader, P. S., Hardie, B. G., & Lee, K. L. (2005). "Counting your customers
```
Seems like different kind of quote

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -277,9 +316,15 @@ def expected_num_purchases(
...
-        xr.DataArray
+        xarray.DataArray
             Expected number of purchases in the next time interval of length :math:`t`.
         """
+        warnings.warn(
+            "Deprecated method. Use 'expected_purchases' instead.",
+            FutureWarning,
+            stacklevel=1,
+        )
+
```
wd60622, line 327: If this method is meant to be the same / replacement of `expected_num_purchases`, why not just use that method's logic now and not have it in two different places

pymc_marketing/clv/models/beta_geo.py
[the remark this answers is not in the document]
juanitorduz: No, all good. Thanks!

pymc_marketing/clv/plotting.py
juanitorduz: ok! makes sense!

pymc_marketing/clv/models/beta_geo.py
juanitorduz: I think it is because of the data input format (one gets arrays, and another gets a data frame). Still, in the next iteration, we will harmonize this. See https://github.com/pymc-labs/pymc-marketing/pull/709/files/b692cf3e123acd5ee2e36ac4020430e71ce414d7#r1625726624

juanitorduz (approved):

ColtAllen:
Thanks for approving! Give me a day or two to double-check the docstrings for formatting, typos, etc. and I'll get it merged.

pymc_marketing/clv/models/beta_geo.py
ColtAllen: Yes; the new API uses a single dataframe parameter, but the old array logic is still expected in the plotting and `customer_lifetime_value` methods. Those will be given separate , follow-up PRs to keep changes focused and easier to review.

pymc_marketing/clv/models/beta_geo.py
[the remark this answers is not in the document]
ColtAllen: @wd60622 I'm not seeing this single-double quote change in the docs preview. Maybe because it's inside of a reStructuredText python block inside of a docstring?

PR #725, opened 2024-06-07
Set upper bound pymc 5.16

juanitorduz:
Expected breaking changes for RandonVariables in the next release, see https://github.com/pymc-devs/pytensor/releases/tag/rel-2.23.0

----
📚 Documentation preview 📚: https://pymc-marketing--725.org.readthedocs.build/en/725/


juanitorduz:
Thank you for https://github.com/pymc-labs/pymc-marketing/pull/725/commits/9870be7afc8237ae0173ea441fbb01939df5415c copilot and chatgpt!

.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 env:
   # The lower bound from the pyproject.toml file
-  OLDEST_PYMC_VERSION: "$(grep -E 'pymc *>' pyproject.toml | sed 's/[^0-9\\.]*//g')"
+  OLDEST_PYMC_VERSION: "$(grep -E 'pymc *>' pyproject.toml | sed -n 's/.*>=\\([0-9]*\\.[0-9]*\\.[0-9]*\\).*/\\1/p')"
```
wd60622, line 11: Does this still work when we switch back to strictly greater than?
i.e. >=5.16.0

wd60622 (commented):
Can you make some issues on where we will have to change to migrate to >=5.16.0.

My understanding is that it would only be where custom ops are define. i.e. [(only) here?](https://github.com/pymc-labs/pymc-marketing/blob/6edc4ca4e9353d10804c8817252a7e45879c0827/pymc_marketing/clv/distributions.py)

.github/workflows/ci.yml
juanitorduz: Not immediately, but I think it is a good practice to bound the versions anyway. WDYT?

juanitorduz: Wait! It actually works! hehe

ricardoV94: Some people are against it, as it can make certain environments impossilble (you need libraries A and B, but A pins to a different version of C than B does, and you simply can't have both)

juanitorduz: I don't have a strong opinion here as I see both sides. @ricardoV94 ehat is your suggestion to handle the pymc 5.16 release? Shall we just add this upper bound now and removed after we have done the migration?

ricardoV94: Yes, although as @maresb mentioned this will only prevent the last version of pymc-marketing from being installed with pymc5.16, older versions, because they had no upper pin, will gladly try to install it

juanitorduz: ok! We will do so!

maresb: I have high hopes that we can test lower versions more systematically with https://github.com/mamba-org/rattler/pull/660

> ...it can make certain environments impossible...

If we had some probably-not-yet-existent infrastructure that automatically updates the dependencies to the latest and ensures that we're actually testing with those latest versions, given the release velocity of PyMC stuff I don't really see how this can lead to a significant issue. (In my experience this is only problematic when the software with the pins lags behind.)

If we had such a solution, potentially utilizing pixi/rattler, would your concerns be mitigated @ricardoV94?

wd60622: > I'm not too concerned as I have not seen it fail yet. If we can automate upper pins that would be super

What would this look like?

maresb: I'm not sure exactly. But probably something involving dependabot or similar, together with some extra infrastructure in the tests to ensure that the upper bounds are simultaneously installed.

More challenging is the lower bounds, where hopefully we could leverage the pixi/rattler stuff to constantly test against the lower bounds, and then we'll know when we need to bump those.

wd60622 (approved):
Think this looks good

PR #743, opened 2024-06-13
Hierarchical Model Configuration

wd60622:
## Related Issue

- [ ] Closes #708 
- [ ] Closes #739 
- [ ] Related to #599 #497 #334 #648 #460

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [x] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--743.org.readthedocs.build/en/743/

PR #747, opened 2024-06-13
Allowing Hierarchical Non Centered Parametrization

cetagostini:
## Related Issue

- [x] Closes #743 
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--747.org.readthedocs.build/en/747/


pymc_marketing/model_config.py
```diff
@@ -388,13 +466,12 @@ def create_distribution(
     **kwargs
         Additional keyword arguments for the distribution.
 
-    Returns
-    -------
-    TensorVariable
-        A PyMC random variable.
-
```
juanitorduz, line 395: why did you erase this?

juanitorduz (commented):
Just some little questions

pymc_marketing/model_config.py
```diff
@@ -369,6 +388,65 @@ def handle_parameter_distributions(
...
+def create_hierarchical_non_center(
+    name: str,
+    distribution_kwargs: dict[str, Any],
+    **kwargs,
+) -> pt.TensorVariable:
+    """
+    Create a hierarchical non-centered distribution.
+
+    This function constructs a hierarchical non-centered distribution using the provided
+    distribution parameters for offset, mu, and sigma. It returns a deterministic variable
```
juanitorduz, line 400: Why do we need this offset random variable? Usually the non-center parametrization in the Normal distribution case means using a standard normal  multiply it with the variance and add the mean as in https://www.pymc.io/projects/examples/en/latest/generalized_linear_models/multilevel_modeling.html#non-centered-parameterization Can you provide some explanation in the doctrines?

juanitorduz:
@cetagostini, I do not see any real change in the mmm notebook, right (or did I miss something)? If so, can you remove it from this PR?

cetagostini:
> @cetagostini, I do not see any real change in the mmm notebook, right (or did I miss something)? If so, can you remove it from this PR?

Yes, previous notebook has "likelihood sigma", now we replace to be call "y". 

I update the notebook, but we should update all. I was playing with TVP, and same thing!

pymc_marketing/model_config.py
wd60622: +1 and I have a counter example for why this is wrong. It's pretty clear with 2D case

pymc_marketing/model_config.py
cetagostini: Mistake, adding this back!

pymc_marketing/model_config.py
cetagostini: @juanitorduz How I see it, when we use Normal(0,1) then the global mean is the baseline (mu), making the model simpler to interpret and ensuring that the deviations are centered around the global mean. So, global_mu + offset + global_sigma will be the posterior parameter for the each specific hierarchy.
```
#non-centered
offset = pm.Normal(
  'offset', 
  mu=0, sigma=1
  dims=("geo")
 )
  mu_global = pm.HalfNormal(name="mu_global",sigma=2)
  sigma_global = pm.HalfNormal(name="sigma_global",sigma=1)

  vmax = pm.Deterministic(
    "vmax", 
    var=mu_global + offset * sigma_global, 
    dims=("geo")
)
```

This as well mean, offset controls how much each geo deviates from the global mean. @twiecki mentioned in their article:

> Instead of saying that our individual slopes b are normally distributed around a group mean (i.e. modeling their absolute values directly), we can say that they are offset from a group mean by a certain value (b_offset; i.e. modeling their values relative to that mean).

So, by having the prior distribution where the offset equal 0, we basically say that it does not deviate from the global mean mean. Example picture:
image

In the opposite case, offset becomes how much we deviate. And depending on how much we deviate, the final parameter will have a value, Example, the following image where offset is different from zero, but is still a normal distribution:
image

If our offset is set for a non-normal distribution with a non-zero mean (e.g., gamma distribution) shifts the global mean of the parameters but does not inherently lead to a wrong model. It changes the interpretation.

Example: Using `Gamma` because it's cap to zero, transform the global mean into a global minimum. So, our posterior distribution for the parameter could not have a value lower than that.

image

This is not directly a wrong model, it's just a model with a different interpretation (Which I accept could be not common). But it's a choice and depending the parameter to inform, and your problem, you could use it or not.

Happy to be challenged :) 

cc: @wd60622

wd60622: My biggest issue is that this still requirements on the user to know how to correctly provide the dims of offset.

## 1D Case
Even in the 1D, the user can deviate from the actual

```python
distribution = {
    "dist": "Normal", 
    "kwargs": {
        "mu": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}}, 
        "sigma": {"dist": "HalfNormal", "kwargs": {"sigma": 1}}, 
    }, 
    "dims": ("channel"), 
}
```

Since there are "channel" draws from the parent normal distributions, this in my mind is only one option for non-centered. That is: 

```python
distribution = {
    "dist": "Normal", 
    "kwargs": {
        "offset": {
            "dist": "Normal", 
            "kwargs": {"mu": 0, "sigma": 1}, 
            # This has to be center dims in order to get the correct number of draws
            "dims": "channel",
        }, 
        "mu": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}}, 
        "sigma": {"dist": "HalfNormal", "kwargs": {"sigma": 1}}, 
    }, 
    "dims": ("channel"), 
}
```

But user could have offset.dims = () or use mu.dims = channel instead? There is a lot of room for error

## 2D Case

This centered parameterization assumes "channel" x "geo" draws from the Normal

```python
distribution = {
    "dist": "Normal", 
    "kwargs": {
        "mu": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}, "dims": channel"}, 
        "sigma": {"dist": "HalfNormal", "kwargs": {"sigma": 1}, "dims": "channel"}, 
    }, 
    "dims": ("channel", "geo"),
}
```

However, I need to be explicit about offset if I want that! But that is not required to be so which puts the burden onto the user 😢 

For instance, the user could do: 

- offset.dims = "geo"
- offset.dims = "channel"
- offset.dims = ()
- offset.dims = ("channel", "geo") # Correct number of draws in my mind

The offset.dims needs to be the correct number of draws or you end up just broadcasting vectors which doesn't give the write answer

wd60622 (changes requested):
Left comment about issue with dims in the thread

pymc_marketing/model_config.py
cetagostini: Sure, this make sense but we are talking about an user usability issue. 

I feel not allowed is to radical, we are basically saying, "better not to give driver license to anyone because a few drivers could be involved in accidents".  

Nevertheless, I have a few things which could work to solve this, based on what you said, we just need to highlight a warning or error when final parameter and offset has different dimensions.

We can explain as well in the notebook, I really don't see this as a dramatic issue!

juanitorduz: Great discussion 💪 ! What if we, for this first iteration, simply use the flag `"centered": False / True` and get feedback and iterate ?

cetagostini: I just did an small changes to raise warning and overwrite the dims in case don't match. This should solve the issue, as well should we get better feedback no?

If we just move to `"centered": False / True` will always work and user will never play with anything else, so I don't feel any feedback will come. Adding the warning, they could share if they are playing with the distributions, and getting warnings or not. Which, I believe is more constructive. no?

cc: @juanitorduz @wd60622

cetagostini: Final point about not allowing moving the offset. If the user wants to make `beta` a non-centered hierarchical parameter, and we have offset always `Normal`. It could have negative values ​​even if global `mu` and `sigma` are positive only.

I think that would force us to then have to make an adjustment when they want to force positive values. But if we can play with all the distributions they can make all the distributions Gamma, HalfNormal, TruncatedNormal and that's it. :)

juanitorduz: >  if we just move to "centered": False / True will always work and user will never play with anything else,

For this PR, and given the user base we have now, I think is a wise bet. We can do more experiments and test them in a different PR. I would like, for example, to test edge cases. But having this boolean flag can help our users with divergences which can be avoidable by this easy parametrization change. WDYT?

cetagostini: > I would like, for example, to test edge cases.

Sure, are you talking about model flexibility? or mathematical properties? Based on those cases, we can usually cover those on the unit tests.

> But having this boolean flag can help our users with divergences which can be avoidable by this easy parametrization change.

Totally agree, but then are we discussing a line of code? 😅 because the same can be achieve if:

```python
"kwargs": {
        "mu": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}}, 
        "sigma": {"dist": "HalfNormal", "kwargs": {"sigma": 1}}, 
    }
   . . .
"centered": False,
```
or

```python
"kwargs": {
        "offset": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}},
        "mu": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}}, 
        "sigma": {"dist": "HalfNormal", "kwargs": {"sigma": 1}}, 
    }
   . . .
```

Both cases is a single extra line of code, and using the `default` users get the same output.

For me the conversation should be around:
a) What is more flexible?
b) What is more sustainable?
c) What aligns with our final goal?

As I see it, the idea is our users have an API that is flexible enough to build different models, where **imagination is the limit**. Limiting it to a single type of parameterization would lead us to find particular cases that are not addressed and create solutions to handle those (Example, non-negative parameters from the top of my head, more will appear for sure).

On the other hand we can leave the configuration flexible, by `default` the user has the same experience and output (one line of code will not make their life miserable) but we have many more cases covered due to these same fact, which means, less work in the future for us. More important, we align with different tastes and ideas.

For me that resonates more with us but I'm a democrat and I stick to what the majority decides! 😂

cc: @juanitorduz @wd60622

wd60622: Think having the centered=False key in the dictionary is an easy user experience, makes most sense to me, and is my preferred API. I am in favor of keeping the keys of kwargs close to actually distribution parameters because that is predictable in my eyes which makes me not like the "offset" option.

I've never heard of a non-centered hierarchical Beta distribution. Please dig one up if I you have a reference! @carlosagostini 

Personally think that the hierarchical support already is quite flexible! We do support 0-1 distributions and positive distributions with the current implementation so people do have options! For example: 
```python
# hierarchical beta example
positive_distribution = {"dist": "HalfNormal, "kwargs": {"alpha": 1, "beta": 1}
hierarchical_beta_distribution = {
    "dist": "Beta", 
    "kwargs": {"alpha": positive_distribution, "beta": positive_distribution},
    "dims": "channel", 
}
# hierarchical gamma example
hierarchical_positive_distribution = {
    "name": "Gamma", 
    "kwargs": {"alpha", positive_distribution, "beta": positive_distribution},
    "dims": "channel",
}
```
Another alternative to fix the domain could be that we add is a "transform" key. For instance, transform="exp" would be performed to after the variable is created. 
For example, 
```python
mu = {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}}
non_centered_hierarchical_positive_distribution = {
    "name": "Normal", 
    "kwargs": {"mu", mu, "sigma": positive_distribution},
    "dims": "channel",
    "centered": False,
    "transform": "exp",
}
```
And the `transform` key would have to come from `pt` or `pm.math` or similar namespace. This might be inlined with non-centered but support different domains. [Ref](https://discourse.mc-stan.org/t/non-centered-parameterization-for-likelihood/31023/2)

juanitorduz: I think this is the key aspect: I also did not find any reference about non-centered parameterization for general distributions and also ended up in the exact Stan blog 😄

Note that Bob Carpenter (and he is really an expert) says:

> The beta distribution isn’t a location/scale family, so the notion of centering/non-centering doesn’t apply directly.

Hence, in view of this, I also like the centered bean and transform API suggested by @wd60622 .

My biggest concern with the offset option is that this could potentially mess up the geometry and hurt the sampler. Plus making the parameters unidentifiable because of too many degrees of freedom.

cetagostini: I was talking about "beta" like coefficient, "beta_channel" or a parameter in specific, definitely not sure about beta non center parametrization.

Nevertheless, I hear you loud and clear :) I'll make the change, and push soon.

cetagostini:
Hey requested changes are done, and tests are working properly!
cc: @wd60622 @juanitorduz

pymc_marketing/model_config.py
```diff
@@ -369,10 +383,66 @@ def handle_parameter_distributions(
...
+
+    sigma_global = create_distribution(
+        f"{name}_sigma_global",
+        sigma_dist["dist"],
+        sigma_dist["kwargs"],
+        dims=sigma_dims,
+    )
+
+    return pm.Deterministic(
+        name=name, var=mu_global + offset * sigma_global, dims=desired_dims
```
wd60622, line 437: Gotta use the dim_hander or broadcasting can break

pymc_marketing/model_config.py
```diff
@@ -369,10 +383,66 @@ def handle_parameter_distributions(
...
+
+    """
+    desired_dims = kwargs.get("dims", ())
+
+    mu_dist = distribution_kwargs["mu"]
+    mu_dims = mu_dist.get("dims", ())
+    sigma_dist = distribution_kwargs["sigma"]
+    sigma_dims = sigma_dist.get("dims", ())
+
+    offset = pm.Normal(name="offset", mu=0, sigma=2, dims=desired_dims)
```
wd60622, line 420: sigma=1

wd60622, line 420: sigma=1

wd60622, line 420: name=f"{name}_offset"

pymc_marketing/model_config.py
```diff
@@ -369,10 +383,66 @@ def handle_parameter_distributions(
...
+
+    mu_global = create_distribution(
+        f"{name}_mu_global",
```
wd60622, line 423: {name}_mu is how centered is so use that instead

pymc_marketing/model_config.py
```diff
@@ -369,10 +383,66 @@ def handle_parameter_distributions(
...
+        mu_dist["dist"],
+        mu_dist["kwargs"],
+        dims=mu_dims,
+    )
+
+    sigma_global = create_distribution(
+        f"{name}_sigma_global",
```
wd60622, line 430: Same as above

pymc_marketing/model_config.py
```diff
@@ -369,10 +383,66 @@ def handle_parameter_distributions(
...
+    desired_dims = kwargs.get("dims", ())
+
+    mu_dist = distribution_kwargs["mu"]
+    mu_dims = mu_dist.get("dims", ())
+    sigma_dist = distribution_kwargs["sigma"]
+    sigma_dims = sigma_dist.get("dims", ())
+
+    offset = pm.Normal(name="offset", mu=0, sigma=2, dims=desired_dims)
+
+    mu_global = create_distribution(
```
wd60622, line 422: What happens if mu includes a nested distribution? Doesnt seem to be handled

pymc_marketing/model_config.py
```diff
@@ -392,9 +462,11 @@ def create_distribution(
     -------
     TensorVariable
         A PyMC random variable.
-
     """
-    dim_handler = create_dim_handler(kwargs.get("dims"))
+    if centered_hierarchy is False:
```
wd60622, line 466: if not centered_hierarchy:

pymc_marketing/model_config.py
```diff
@@ -369,10 +383,66 @@ def handle_parameter_distributions(
...
+    return pm.Deterministic(
+        name=name, var=mu_global + offset * sigma_global, dims=desired_dims
+    )
+
+
 def create_distribution(
     name: str,
     distribution_name: str,
     distribution_kwargs: dict[str, Any],
+    centered_hierarchy: bool | None = None,
```
wd60622, line 445: I prefer name centered because it is shorter. 

Just make bool = True as default

tests/test_model_config.py
```diff
@@ -224,6 +257,25 @@ def coords() -> dict[str, list[str]]:
...
+            ],
+            [(3, 2), (3,), (2,)],
+        ),
+        (
+            "hierarchical_non_centered",
+            [
+                "hierarchical_non_centered",
+                "hierarchical_non_centered_mu_global",
+                "hierarchical_non_centered_sigma_global",
+                "offset",
```
wd60622, line 275: Add the prefix

tests/test_model_config.py
```diff
@@ -224,6 +257,25 @@ def coords() -> dict[str, list[str]]:
...
+        (
+            "hierarchical_centered",
+            [
+                "hierarchical_centered",
+                "hierarchical_centered_mu",
+                "hierarchical_centered_sigma",
+            ],
+            [(3, 2), (3,), (2,)],
+        ),
+        (
```
wd60622, line 269: Can you test a 2D version with centered=False

pymc_marketing/model_config.py
wd60622, line 466: Need to check that dist=Normal and because that is assumed

wd60622 (changes requested):

pymc_marketing/model_config.py
cetagostini: Solve!

pymc_marketing/model_config.py
cetagostini: solve!

pymc_marketing/model_config.py
cetagostini: solve!

pymc_marketing/model_config.py
cetagostini: solve!

pymc_marketing/model_config.py
cetagostini: I manage cases by default is `True` if exist. We must have default `None` otherwise when loop is `if not None` the values pass and mess up everything.

```python
centered_flag = parameter_config.get("centered", True)
```

pymc_marketing/model_config.py
cetagostini: That generates problem when is `None` and we need to handle the case when is none in case don't exist in the dict because they are not using hierarchies. Related to next comments!

tests/test_model_config.py
cetagostini: Done!

pymc_marketing/model_config.py
cetagostini: Done!

pymc_marketing/model_config.py
cetagostini: Done!

wd60622:
Should we continue with this or go with #759

pymc_marketing/model_config.py
cetagostini: Done!

cetagostini:
I address all your comments @wd60622. My suggestion would be merge to have the feature, which probably will improve with #759 but users can start from now to play with it and give some feedback.

tests/test_model_config.py
wd60622: @cetagostini

tests/test_model_config.py
```diff
@@ -224,6 +257,25 @@ def coords() -> dict[str, list[str]]:
...
         ("delta", ["delta"], [(3, 1)]),
+        (
+            "hierarchical_centered",
+            [
+                "hierarchical_centered",
+                "hierarchical_centered_mu",
+                "hierarchical_centered_sigma",
+            ],
+            [(3, 2), (3,), (2,)],
+        ),
```
wd60622, line 268: Similar to this one but with non centered

wd60622 (changes requested):
Just one more check then looks good

pymc_marketing/model_config.py
```diff
@@ -392,9 +506,13 @@ def create_distribution(
     -------
     TensorVariable
         A PyMC random variable.
-
     """
-    dim_handler = create_dim_handler(kwargs.get("dims"))
+    if centered is False:
+        if distribution_name != "Normal":
+            NonCenterInvalidDistributionError(distribution_name)
```
wd60622, line 512: Need to raise this

wd60622 (changes requested):

pymc_marketing/model_config.py
cetagostini: My bad done!

PR #741, opened 2024-06-12
`model.fit` doesn't remove prior samples

wd60622:
## Description

The following workflow is now supported

```python
model.sample_prior_predictive(X)
model.fit(X, y)
assert "prior" in model.idata
```

Doesn't touch on the "fit_data_group" introduced with #414 

## Related Issue

- [x] Closes #410
- [x] Related to #414 

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [x] Included tests that prove the fix is effective or that the new feature works
- [x] Added necessary documentation (docstrings and/or example notebooks)
- [x] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Modules affected

- [x] MMM
- [x] CLV

## Type of change

- [ ] New feature / enhancement
- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--741.org.readthedocs.build/en/741/

PR #759, opened 2024-06-19
WIP: Prior class

wd60622:
## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc_marketing/prior.py
+++ b/pymc_marketing/prior.py
@@ -1,0 +1,451 @@
+#   Copyright 2024 The PyMC Labs Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+from __future__ import annotations
+
+import copy
+from inspect import signature
+from typing import Any
+
+import numpy as np
+import pymc as pm
+import pytensor.tensor as pt
+import xarray as xr
+from graphviz import Digraph
+from pymc.distributions.shape_utils import Dims
+
+
+class UnsupportedShapeError(Exception):
+    """Error for when the shape of the hierarchical variable is not supported."""
+
+
+class UnsupportedDistributionError(Exception):
+    """Error for when an unsupported distribution is used."""
+
+
+class MuAlreadyExistsError(Exception):
+    """Error for when 'mu' is present in Prior."""
+
+    def __init__(self, distribution: Prior) -> None:
+        self.distribution = distribution
+        self.message = f"The mu parameter is already defined in {distribution}"
+        super().__init__(self.message)
+
+
+def handle_dims(x: pt.TensorLike, dims: Dims, desired_dims: Dims) -> pt.TensorVariable:
+    """Take a tensor of dims `dims` and align it to `desired_dims`.
+
+    Doesn't check for validity of the dims
+
+    Examples
+    --------
+    1D to 2D with new dim
+
+    .. code-block:: python
+
+        x = np.array([1, 2, 3])
+        dims = "channel"
+
+        desired_dims = ("channel", "group")
+
+        handle_dims(x, dims, desired_dims)
+
+
+    """
+    x = pt.as_tensor_variable(x)
+
+    if np.ndim(x) == 0:
+        return x
+
+    dims = dims if isinstance(dims, tuple) else (dims,)
+    desired_dims = desired_dims if isinstance(desired_dims, tuple) else (desired_dims,)
+
+    aligned_dims = np.array(dims)[:, None] == np.array(desired_dims)
+
+    missing_dims = aligned_dims.sum(axis=0) == 0
+    new_idx = aligned_dims.argmax(axis=0)
+
+    args = [
+        "x" if missing else idx
+        for (idx, missing) in zip(new_idx, missing_dims, strict=False)
+    ]
+    return x.dimshuffle(*args)
+
+
+def create_dim_handler(desired_dims: Dims):
+    """Wrapper to act like the previous `create_dim_handler` function."""
+
+    def func(x: pt.TensorLike, dims: Dims) -> pt.TensorVariable:
+        return handle_dims(x, dims, desired_dims)
+
+    return func
+
+
+def dims_to_str(obj: tuple[str, ...]) -> str:
+    if len(obj) == 1:
+        return f'"{obj[0]}"'
+
+    return (
+        "(" + ", ".join(f'"{i}"' if isinstance(i, str) else str(i) for i in obj) + ")"
+    )
+
+
+def get_pymc_distribution(name: str) -> type[pm.Distribution]:
+    if not hasattr(pm, name):
+        raise Exception(f"pymc doesn't have a distribution of name {name!r}")
+
+    return getattr(pm, name)
+
+
+def get_transform(name: str):
+    for module in (pt, pm.math):
+        if hasattr(module, name):
+            break
+    else:
+        module = None
+
+    if not module:
+        raise Exception(f"Neither pytensor or pm.math have the function {name!r}")
+
+    return getattr(module, name)
+
+
+def pymc_parameters(distribution: pm.Distribution) -> set[str]:
+    return set(signature(distribution.dist).parameters.keys()) - {"kwargs", "args"}
+
+
+class Prior:
+    """A class to represent a prior distribution."""
+
+    def __init__(
+        self,
+        distribution: str,
+        dims: Dims | None = None,
+        centered: bool = True,
+        transform: str | None = None,
+        **parameters,
+    ) -> None:
+        self.distribution = distribution
+        self.parameters = parameters
+        self.dims = dims
+        self.centered = centered
+        self.transform = transform
+
+        self._checks()
+
+    @property
+    def distribution(self) -> str:
+        return self._distribution
+
+    @distribution.setter
+    def distribution(self, distribution: str) -> None:
+        if not isinstance(distribution, str):
+            raise ValueError("Distribution must be a string")
+
+        self._distribution = distribution
+        self.pymc_distribution = get_pymc_distribution(distribution)
+
+    @property
+    def transform(self) -> str | None:
+        return self._transform
+
+    @transform.setter
+    def transform(self, transform: str | None) -> None:
+        if not isinstance(transform, str) and transform is not None:
+            raise ValueError("Transform must be a string or None")
+
+        self._transform = transform
+        self.pytensor_transform = not transform or get_transform(transform)
+
+    @property
+    def dims(self) -> Dims:
+        return self._dims
+
+    @dims.setter
+    def dims(self, dims) -> None:
+        if isinstance(dims, str):
+            dims = (dims,)
+
+        self._dims = dims or ()
+
+        self._param_dims_work()
+        self._unique_dims()
+
+    def __getitem__(self, key: str) -> Prior | Any:
+        return self.parameters[key]
+
+    def _checks(self) -> None:
+        if not self.centered:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -16,20 +16,20 @@ repos:
...
-        files: ^pymc_marketing/
-        additional_dependencies: [numpy>=1.20, pandas-stubs]
   - repo: https://github.com/nbQA-dev/nbQA
     rev: 1.8.5
     hooks:
       - id: nbqa-ruff
         args: ["--fix", "--output-format=full"]
         files: ^docs/source/notebooks/
         exclude: ^docs/source/notebooks/clv/dev/
+  - repo: https://github.com/pre-commit/mirrors-mypy
```
wd60622, line 26: Changing the order so that ruff formatting all happens before mypy (something slow)

juanitorduz: Can we leave this out of this PR? We want to do this before 0.8

wd60622: Yes. I will change back before this is merged

wd60622:
Hey @ColtAllen, will pick up this PR after 0.7.0 is released. I am willing to do the migration of the CLV module. 

I see in many of the cases that the `default_model_config` will have the suffix `_prior` to it. Would you like to keep that convention? 

Also, this is backwards compatible with the previous dictionaries that are used via the `Prior.from_json` method. My plan would be to warn users 
if their configuration has the old format, convert, and suggest using new format to avoid the warning. Any thoughts on this? 

  If you like an overview of how these will work, check out the top comment or the notebook that is in this PR. 

  Interested to hear your thoughts on this.

wd60622:
Think the migration for CLV would be pretty small too.
1. Remove the [`_create_distribution` method](https://github.com/pymc-labs/pymc-marketing/blob/897cf3d9a9d589e6ab103a071d48a000e9a4aa68/pymc_marketing/clv/models/basic.py#L227-L232)
2. Change `default_model_config` to use `Prior` class. All seems in the same format
3. Change `_create_distribution`, `model.registered_rv` workflow to `self.model_config["var_prior"].create_variable("var")` which can happen in one line rather than 2
    - Will have to be explicit about the shapes [here](https://github.com/pymc-labs/pymc-marketing/blob/897cf3d9a9d589e6ab103a071d48a000e9a4aa68/pymc_marketing/clv/models/pareto_nbd.py#L227-L234) using dims
4. Make sure that tests work

Use a warning from `parse_model_config` function to signify to users that model config values will change to `Prior`

juanitorduz:
@wd60622 You pushed this one quite fast! As in 0.7.0 we will have an MMM new class, do you think this should be part of 0.7.0 as well? I do not think there is any rush and we can think about releasing 0.7.0 in 2-3 weeks as many people are taking summer holidays anyway. WDYT?

I hope to review this one in the next days. 💪

ColtAllen:
> I see in many of the cases that the `default_model_config` will have the suffix `_prior` to it. Would you like to keep that convention?

Yes, I think it's good to differentiate between priors and posteriors for the model parameters.

> Also, this is backwards compatible with the previous dictionaries that are used via the `Prior.from_json` method. My plan would be to warn users if their configuration has the old format, convert, and suggest using new format to avoid the warning. Any thoughts on this?

I agree it's good to add a warning and support the old format until a 1.0 release.

wd60622:
> Yes, I think it's good to differentiate between priors and posteriors for the model parameters.

Yeah, I agree with that which is one of the reasons I was going with "Prior" instead of something like "Distribution". Though, I like "Distribution" as well since in MMM, there is a target "distribution"

Regardless, the MMM model config now looks like this: 

```python
mmm_model_config = {
    "beta": Prior("Normal", mu=0, sigma=1), 
    ...
}
```
So the word "Prior" hopefully makes it clear to the user.

Because of the suffix, the CLV config would have "prior" twice. Not sure if that is redundant or not an issue. Let me know your thoughts @ColtAllen 
```python
    @property
    def default_model_config(self) -> dict:
        return {
            "alpha_prior": Prior("HalfFlat"),
            "beta_prior": Prior("HalfFlat"),
        }

```
Came from this:
https://github.com/pymc-labs/pymc-marketing/blob/e81b4031e7e26369a5d55f08d1b1ce47fed0ad94/pymc_marketing/clv/models/shifted_beta_geo.py#L128-L133

Changing the key names will be a bit hard to deprecate than just switching the values though

Just some things to be aware of!

> I agree it's good to add a warning and support the old format until a 1.0 release.

Yup, there is a warning which will print out what the person would have to replace it with. Nice to have the __repr__ of the class also evaluate to the same class 😄

wd60622:
> @wd60622 You pushed this one quite fast! As in 0.7.0 we will have an MMM new class, do you think this should be part of 0.7.0 as well? I do not think there is any rush and we can think about releasing 0.7.0 in 2-3 weeks as many people are taking summer holidays anyway. WDYT?

Think it would be good to not hold up a 0.7.0 release as this functionality still exists in the dictionaries due to #743 . This is just a prettier version 😄 with a bunch of new functionality. 

> I hope to review this one in the next days. 💪

Sounds good. Thanks.

The items that I have left are: 

- [ ] Migrate the CLV priors
- [ ] CLV tests
- [ ] Run all the notebooks
- [ ] Make sure all the documentation is migrated over
- [ ] Any additional tests & documentation that would be helpful

ColtAllen:
Will removing the "_prior" suffix from the `model_config` parameter keys create issues for prior and posterior predictive checks?

wd60622:
> Will removing the "_prior" suffix from the `model_config` parameter keys create issues for prior and posterior predictive checks?

Which workflow do you have in mind?

ColtAllen:
> > Will removing the "_prior" suffix from the `model_config` parameter keys create issues for prior and posterior predictive checks?
> 
> Which workflow do you have in mind?

https://github.com/pymc-labs/pymc-marketing/pull/741

Will these variables have the same names, but reside in different groups? Just don't want users confusing prior & posterior prediction checks.

wd60622:
> > > Will removing the "_prior" suffix from the `model_config` parameter keys create issues for prior and posterior predictive checks?
> > 
> > 
> > Which workflow do you have in mind?
> 
> #741
> 
> Will these variables have the same names, but reside in different groups? Just don't want users confusing prior & posterior prediction checks.

This is separate from the Dataset that are populated in the InferenceData. 
People will still have sample_prior_predictive and fit for prior / prior_predictive and posterior data group in InferenceData. 

This is just replacing the parameter configurations that are currently stored in dictionaries with a more defined class. 

There are some additional methods that come with this class. For instance, sampling of the distributions defined specifically. However, that is separate from the model and any InferenceData that would be populated from the `ModelBuilder`-like methods

wd60622:
> > > Will removing the "_prior" suffix from the `model_config` parameter keys create issues for prior and posterior predictive checks?
> 
> > 
> 
> > Which workflow do you have in mind?
> 
> 
> 
> https://github.com/pymc-labs/pymc-marketing/pull/741
> 
> 
> 
> Will these variables have the same names, but reside in different groups? Just don't want users confusing prior & posterior prediction checks.

Also to clarify, the name of the key of the config is not the name of the variable for clv module like it is in mmm. The name of the variable is defined by what is passed to create_variable method of Prior

pymc_marketing/mmm/components/base.py
```diff
@@ -177,18 +186,19 @@ def update_priors(self, priors: dict[str, Any]) -> None:
         --------
         Update the priors for a transformation after initialization.
 
-        from pymc_marketing.mmm.components.base import Transformation
-
         .. code-block:: python
 
+            from pymc_marketing.prior import Prior
+            from pymc_marketing.mmm.components.base import Transformation
```
juanitorduz, line 192: small nit: are these isorted :D)

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -877,15 +857,14 @@ class MMM(
 
     .. code-block:: python
 
+        import numpy as np
+
+        from pymc_marketing.prior import Prior
+        from pymc_marketing.mmm import MMM
```
juanitorduz, line 863: are these isorted?

pymc_marketing/prior.py
```diff
@@ -0,0 +1,893 @@
...
+        )
+
+    return getattr(module, name)
+
+
+def _get_pymc_parameters(distribution: pm.Distribution) -> set[str]:
+    return set(signature(distribution.dist).parameters.keys()) - {"kwargs", "args"}
+
+
+class Prior:
```
juanitorduz, line 219: Here is an opinionated suggestion: Inherint from PyDAntic BAse model and use model validators https://docs.pydantic.dev/latest/concepts/validators/#dataclass-validators to do the checks. See also https://docs.pydantic.dev/latest/concepts/validation_decorator/

Given its popularity, I do not think Pydantic is a strong dependency.  WDYT?

pymc_marketing/prior.py
```diff
@@ -0,0 +1,893 @@
...
+        if self.transform:
+            raise ValueError("Can't constrain a transformed variable")
+
+        new_parameters = pm.find_constrained_prior(
+            self.pymc_distribution,
+            lower=lower,
+            upper=upper,
+            init_guess=self.parameters,
+            **kwargs,
+        )
```
juanitorduz, line 711: We need to handle the exception when this `find_constrained_prior` fails. Maybe add a reasonable fallback mechanism? Or just handle the exception and inform the user with an informative error message?

pymc_marketing/prior.py
```diff
@@ -0,0 +1,893 @@
...
+            self.distribution,
+            dims=copy.copy(self.dims),
+            centered=self.centered,
+            transform=self.transform,
+            **copy.deepcopy(self.parameters),
+        )
+        memo[id(self)] = copy_obj
+        return copy_obj
+
+    def copy(self) -> Prior:
```
juanitorduz, line 798: shall we simply call this `decopy`?

pymc_marketing/prior.py
```diff
@@ -0,0 +1,893 @@
...
+        """Return a deep copy of the prior."""
+        return copy.deepcopy(self)
+
+    def graph(self):
+        """Generate a graph of the variables.
+
+        Examples
+        --------
+        Create the graph for a complicated variable.
```
juanitorduz, line 807: Does it need to be "complicated"? hehe

pymc_marketing/prior.py
```diff
@@ -0,0 +1,893 @@
...
+            **copy.deepcopy(self.parameters),
+        )
+        memo[id(self)] = copy_obj
+        return copy_obj
+
+    def copy(self) -> Prior:
+        """Return a deep copy of the prior."""
+        return copy.deepcopy(self)
+
+    def graph(self):
```
juanitorduz, line 802: what about `to_graph`?

juanitorduz (commented):
@wd60622 This looks vey nice! I just left some small suggestions and comments. I think is in a great state and we can merge relatively soon :)

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: changed

pymc_marketing/mmm/components/base.py
wd60622: changed

pymc_marketing/prior.py
wd60622: Just changed to "hierarchical" instead

pymc_marketing/prior.py
wd60622: The `MMM` object does have a `graphviz` object, should that be used?
I like `to_graph` better though

pymc_marketing/prior.py
wd60622: How about `deepcopy` specifically then? 
I read that as de-copy

pymc_marketing/prior.py
wd60622: Think it is a great idea. Let's save off for another PR

wd60622:
I touched just reran the clv notebooks to check. I dont expect major differences @ColtAllen 

Did you see much variation between runs of the time varying media notebooks? @cetagostini 
I see some different parameters but the curves looks practically the same

juanitorduz (approved):

PR #763, opened 2024-06-20
fix np typing

juanitorduz:
We have a new error with MyPy

```python
 error: Name "np.float_" is not defined  [name-defined]
```

----
📚 Documentation preview 📚: https://pymc-marketing--763.org.readthedocs.build/en/763/


wd60622 (approved):

juanitorduz:
Also the docs are failing! I will try tomorrow ....

Issue #774, opened 2024-06-23
Different defaults for HSGP TVPs


bwengals:
I think some of the TVP fits in the TVP notebook can be improved by using different priors on the GP.

Basically, I'd propose to:
1. Use the PC prior as the default prior for a GP TVP.
2. Allow the user to overwrite the particular settings of that PC prior.
3. Allow the user to override the PC prior with a more informative Inverse Gamma prior where they choose `mu` and `sigma` or the upper and lower tail probabilities and mass, a la `pm.find_constrained_prior`.  
4. Use [this function](https://github.com/pymc-devs/pymc/blob/ebea3b8435cdbc7b9ba1254c2f43bc6634e53693/pymc/gp/hsgp_approx.py#L94) to set `m` and `c` to match the prior given to the lengthscale.  This will require assuming how far into the future the user wants to make predictions.  If the user wants reliable predictions further into the future, they will need to specify the number of days they need.

The main idea is to use the PC prior for the GP scale (`eta`) and lengthscale (`ls`) as the default prior.  I think this could potentially be done using the data and without any user input.  The PC prior is derived as joint prior.  It works out to be an Exponential on eta and a Frechet or inverse Weibull on the lengthscale. 

PC priors are not uninformative, so the user potentially needs to set two tail probabilities:
1. For the scale eta: $p(\eta > U) = \alpha$.  User must choose $U$ and $\alpha$.` eta is just like a standard deviation, so some fraction of the overall standard deviation of the data could be used here.  
2. For the lengthscale ls: $p(\ell < L) = \alpha$. User must choose $L$ and $\alpha$.  If the data is daily, is it reasonable to not measure variation on a scale of less than one week?  If so, we could automatically choose L = 7 and $\alpha=0.01$.  For daily data though, there must be a hard cutoff at L = 1, the resolution of the data.  

What do people think about setting sensible defaults for the scale and the lengthscale using just the data, or should the user be forced to make a choice?  Any guidance on the implementation would be greatly appreciated.

juanitorduz:
This is awesome! Our users are definitively not experts in GPs; having suitable default priors is key 🙏! I'll take a look into the PR this week! 

Thanks for bringing your expertise to the package 🙌

bwengals:
[...]

PR #775, opened 2024-06-23
Change priors for TVPs

bwengals:
## Description

This PR changes the TVP HSGP prior parameters for the scale and lengthscale to default to the PC prior, and be allowed to change to the Inverse Gamma.  

Posting this PR as draft since I got pretty hung up on how best to integrate this into PyMC-Marketing, so hopefully I could get some guidance on how to do that. 

I also reran the results in the TVP example nb.  The first sine wave example looks OK to me, but the GP just reverts to it's mean (which looks to be estimated a bit off).  The second example with the linear trend looks much much better.  The PC prior allows the data to choose a very long lengthscale now.  The third example shouldn't change much, a short lengthscale will work the best here.  I do think it'd be nice to include an example where the GP can really shine, where the TVP varies irregularly with a moderate frequency.  

## Related Issue
- [ ] Closes #
- [x] Related to #774

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--775.org.readthedocs.build/en/775/


wd60622:
Carlos made some recent changes so their will have to be a rebase. Not only is there the intercept tvp kwargs (now config) but media tvp config

pymc_marketing/mmm/tvp.py
```diff
@@ -17,20 +17,94 @@
 import pymc as pm
 import pytensor.tensor as pt
 
-from pymc_marketing.constants import DAYS_IN_YEAR
+
+def pc_prior_1d(name, alpha=0.1, lower=1.0):
+    """
+    One dimensional PC prior for GP lengthscales, parameterized by tail probability:
+    p(lengthscale < lower) = alpha.
```
juanitorduz, line 24: This looks very nice! Would it be possible to get a bit more info about the nature of this prior and the benefits ? This can help us (and our users) whoa re not familiar with this :)

pymc_marketing/mmm/tvp.py
```diff
@@ -78,29 +164,54 @@ def time_varying_prior(
...
-        ls = pm.InverseGamma(f"{name}_ls", mu=ls_mu, sigma=ls_sigma)
-        cov_func = eta**2 * pm.gp.cov.Matern52(1, ls=ls)
+    eta = pm.Exponential(f"{name}_eta", lam=-np.log(eta_mass) / eta_upper)
+
+    if ls_upper is None:
+        ls = pc_prior_1d(f"{name}_ls", alpha=1.0 - ls_mass, lower=ls_lower)
+        ls_upper = 2 * X_mid
+    else:
+        # Note: this function sometimes fails, how to handle?
+        # If not using PC prior, should user set ls_mu and ls_sigma instead?
```
juanitorduz, line 176: Do we have any simple heuristics for a reasonable fallback mechanism?

pymc_marketing/mmm/tvp.py
```diff
@@ -78,29 +164,54 @@ def time_varying_prior(
...
 
-    gp = pm.gp.HSGP(m=[m], L=[L], cov_func=cov_func)
+    cov_funcs = {
+        "expquad": pm.gp.cov.ExpQuad,
+        "matern52": pm.gp.cov.Matern52,
+        "matern32": pm.gp.cov.Matern32,
+    }
+    cov_func = eta**2 * cov_funcs[cov_func.lower()](input_dim=1, ls=ls)
+
+    gp = pm.gp.HSGP(m=[m], L=[L], cov_func=cov_func, drop_first=True)
```
juanitorduz, line 204: shall we pass the `drop_true` from the function signature here?

pymc_marketing/mmm/tvp.py
```diff
@@ -78,29 +164,54 @@ def time_varying_prior(
...
     phi, sqrt_psd = gp.prior_linearized(Xs=X[:, None] - X_mid)
-    hsgp_coefs = pm.Normal(f"{name}_hsgp_coefs", dims=hsgp_dims)
-    f = phi @ (hsgp_coefs * sqrt_psd).T
-    centered_f = f - f.mean(axis=0) + 1
```
juanitorduz, line 102: Do we want this to oscillate around `1.0` as before?

pymc_marketing/mmm/tvp.py
```diff
@@ -146,7 +257,7 @@ def create_time_varying_intercept(
         )
         return pm.Deterministic(
             name="intercept",
-            var=intercept_base * multiplier,
+            var=pt.exp(intercept_base + multiplier),
```
juanitorduz, line 260: I like this parametrization much better!

pymc_marketing/mmm/tvp.py
```diff
@@ -125,14 +236,14 @@ def create_time_varying_intercept(
...
-                DAYS_IN_YEAR / time_resolution * 2
-            )
+        # if model_config["intercept_tvp_kwargs"]["L"] is None:
+        #    model_config["intercept_tvp_kwargs"]["L"] = (
+        #        time_index_mid + DAYS_IN_YEAR / time_resolution
+        #    )
+        # if model_config["intercept_tvp_kwargs"]["ls_mu"] is None:
+        #    model_config["intercept_tvp_kwargs"]["ls_mu"] = (
+        #        DAYS_IN_YEAR / time_resolution * 2
+        #    )
```
juanitorduz, line 246: Is this to be deleted? 

Are we using the `DAYS_IN_YEAR` at all? 🤔

juanitorduz (commented):
Thanks @bwengals! This is super nice and definitively appreciate your input on better GP priors. I left some comments, that are more questions :)

bwengals:
> Carlos made some recent changes so their will have to be a rebase. Not only is there the intercept tvp kwargs (now config) but media tvp config

Gotcha thanks @wd60622.  I'll rebase and check out those updates.  This should benefit that as well.

pymc_marketing/mmm/tvp.py
bwengals: I took out `f - f.mean(axis=0) + 1` since I think that will make predicting into the future difficult.  And a better way to zero mean this is to de-mean the basis `phi` directly.  I think if z ~ Normal(0, 1), then exp(z) will have a median at 1, so still oscillates around 1.0.

pymc_marketing/mmm/tvp.py
bwengals: I got hung up on the best default.  There can be identifiability issues between the intercept and a GP, particularly for smaller intercepts (relative to the GP) and at longer lengthscales.  This is evident in the HSGP when c gets a bit larger, the first basis vector becomes more and more constant.  For a default, I think the best option is to either, keep the intercept parameter and set drop_first = True, or have drop_first=False and not have an intercept parameter.  

I'm not sure what would be better though, it depends on what the data "usually" looks like.  Like if the intercept for the TVP is like 1000 and the GP part oscillates between -1 and 1, then there wont be identifiability issues, regardless of lengthscale.  But if the intercept is like 0.2 and the GP oscillates between -1 and 1, then it'll be a problem.

pymc_marketing/mmm/tvp.py
bwengals: None that I know of, I think @ulfaslak told me he wrote something to fix this though?

pymc_marketing/mmm/tvp.py
cetagostini: Hey team, can we deep-dive a bit more about this implementation?

In the original implementation, The mean of 𝛽𝑡 would be the 
`**_beta` (since the mean of multiplier is 1). Then variance of 𝛽𝑡 will depend on the variance of the `multiplier`.

Because of it, the initial implementation is more straightforward and interpretable (Personal opinion). Mostly around the response curves.

1. What would be the benefit of this parameterization, if it is not more interpretable? 
2. Being exponential, it forces the final parameter to be positive. This depending on the parameter we want to be time dependent may or may not be needed.
3. I think that if the result of `**_base + latent` is a normal distribution, there should be no problem, the result is a LogNormal distribution. But if it is `Gamma` or `HalfNormal`, the result would be an even more skewed distribution -This might be feasible because the config controls the base distributions- I'm not sure if it's a problem or not, but have we thought about how this transformation would apply to different arbitrary distributions?

A couple of questions that may be very obvious and I am a little slow to see the answers! 😅

juanitorduz:
@bwengals now that we have released 0.7.0 after some refactor I think this is a great time to work on this improvements :) Would you mind rebasing or (maybe even better to avoid fixing conflicts) creating a new PR and work on the integration? We can also work on smaller PRs to work in shorter iterations :)

wd60622:
I will take this over using `main` branch. I will get to it when I can

PR #784, opened 2024-06-27
Fix typo in Control Section of MMM Example notebook

wd60622:
## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [ ] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--784.org.readthedocs.build/en/784/

PR #787, opened 2024-06-27
Add prior predictive example notebook

juanitorduz:
- Add prior predictive plot.
- Fix the double scaling bug.
- Typo from https://github.com/pymc-labs/pymc-marketing/pull/784

----
📚 Documentation preview 📚: https://pymc-marketing--787.org.readthedocs.build/en/787/


juanitorduz:
When adding the prior-predictive step, the processing is wrong:

image

wd60622:
Would the self.y be getting scaled twice? 
If you remove the sample_prior_predictive lines, this doesn't happen, right?

juanitorduz:
The bug was fixed in https://github.com/pymc-labs/pymc-marketing/pull/787/commits/b765b38c89fb2606df90c5aada6c84519e990733

wd60622:
Is it then the combination of `build_model` and `_data_setter`? As the `sample_posterior_predictive` method will call `_data_setter` but not the `build_model`...

wd60622 (approved):
Notebook looks good! Just question but not a blocker

juanitorduz:
> Is it then the combination of `build_model` and `_data_setter`? As the `sample_posterior_predictive` method will call `_data_setter` but not the `build_model`...

I think we need to take a detailed look at how this is happening and that is why I created the issue above 😬. Let me merge this one and then work on the other issue.

PR #791, opened 2024-06-28
Run example notebooks CI

juanitorduz:
I am revisiring https://github.com/pymc-labs/pymc-marketing/pull/624 . We can do it via `readthedocs` but I want to get feedback on this simpler solution.

----
📚 Documentation preview 📚: https://pymc-marketing--791.org.readthedocs.build/en/791/


scripts/run_notebooks/runner.py
wd60622, line 1: I'd recommend using `pathlib.Path` and the glob method. 
For instance
```python
NOTEBOOKS_PATH = Path("docs/source/notebooks")

notebooks: list[Path] = list(NOTEBOOKS_PATH.glob("*/*.ipynb"))
```

This would:
- assure that they all exist already
- be only one level deep which would avoid all the dev notebooks as well

Could also keep as a generator as well

scripts/run_notebooks/runner.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def verify_notebooks_exist() -> None:
+    for notebook_path in get_notebooks_paths():
+        if not os.path.exists(notebook_path):
+            raise FileNotFoundError(f"Notebook not found: {notebook_path}")
+
+
+def run_notebook(notebook_path: str) -> None:
+    pm.execute_notebook(
```
wd60622, line 46: pymc? 😆

scripts/run_notebooks/runner.py
juanitorduz: Great tip thanks!

juanitorduz: I wanna leave out the clv/dev notebooks though

juanitorduz:
@wd60622 @carlosagostini We have had many API changes, and the notebooks are hard to keep up 🙈 (see, for example, https://github.com/pymc-labs/pymc-marketing/pull/793, which I discovered when working on this).  I suggest this solution that is flexible and easy to manage. It runs locally very fast and in the CI takes 10 min, which is comparable to the tests we have in place. WDYT?

scripts/run_notebooks/runner.py
```diff
@@ -0,0 +1,50 @@
...
+from pathlib import Path
+
+import papermill
+from joblib import Parallel, delayed
+from tqdm import tqdm
+
+KERNEL_NAME: str = "python3"
+NOTEBOOKS_PATH = Path("docs/source/notebooks")
+NOTEBOOKS_SKIP: list[str] = [
+    "mmm_budget_allocation_example.ipynb",  # This works locally but not on GitHub Actions
```
juanitorduz, line 13: It seems a problem when reading the model file. I want to investigate in a future PR as this way we are already cheking most of the notebooks

wd60622 (approved):
Looks great and will be a lifesafer. 

How about a command in the Makefile as well? `make notebooks` or something like that to do the runner

juanitorduz:
> Looks great and will be a lifesafer.
> 
> How about a command in the Makefile as well? `make notebooks` or something like that to do the runner

Good idea! Added in https://github.com/pymc-labs/pymc-marketing/pull/791/commits/fe3cdf7457c9ccc283603be404bdcb000efc89bc

wd60622:
> > Looks great and will be a lifesafer.
> > How about a command in the Makefile as well? `make notebooks` or something like that to do the runner
> 
> Good idea! Added in [fe3cdf7](https://github.com/pymc-labs/pymc-marketing/commit/fe3cdf7457c9ccc283603be404bdcb000efc89bc)

Can you run that twice in a row? i.e. does the runner still work if the files already exist?

juanitorduz:
@wd60622 im sorry but I don't understand the question 🙈😄

wd60622:
> @wd60622 im sorry but I don't understand the question 🙈😄

does 
```bash
make run_notebooks
make run_notebooks
```
work? 
Sometimes things don't work if the files already exist

juanitorduz:
> > @wd60622 im sorry but I don't understand the question 🙈😄
> 
> does
> 
> ```shell
> make run_notebooks
> make run_notebooks
> ```
> 
> work? Sometimes things don't work if the files already exist

Woks locally ✅

Issue #220, opened 2023-03-28
Proposal: Use a more intuitive parametrization of `logistic_saturation`


tomicapretto:
The `logistic_saturation` function maps the reals to (-1, 1). If we use values in $[0, \infty)$, the output is in $[0, 1)$. Currently, the parametrization of `logistic_saturation` is:

$$
f(x, \lambda) = \frac{1 - e^{-x\lambda}}{1 + e^{-x\lambda}}
$$ 

$\lambda$ is called the "saturation parameter". In [Bayesian Media Mix Modeling using PyMC3, for Fun and Profit](https://engineering.hellofresh.com/bayesian-media-mix-modeling-using-pymc3-for-fun-and-profit-2bd4667504e6) it says the parameter represents the "half-saturation point" but I think it's not completely right.

If we evaluate $f$ at $1 / \lambda$  the result is $\approx 0.462$, meaning that at $1 / \lambda$ you reach almost half saturation.  So $\lambda$ would be something close to the inverse of the half saturation point.

Thus, I propose to use 

$$
f(x, \lambda) = \frac{1 - e^{-\frac{x}{\lambda}}}{1 + e^{-\frac{x}{\lambda}}}
$$ 

and $\lambda$ could be interpreted as "almost half-saturation" point.

arthurmello:
hi,
I want to contribute to pymc-marketing and this seemed like a good first issue :)
quick question: by changing the formula for `logistic_saturation`, I assume this also means changing all of the occurrences of lambda whenever `logistic_saturation` is called (ex.: in `saturation_function`), including the examples notebooks, in addition to the tests for `logistic_saturation`.
is that correct?

wd60622:
Also had the thought...
What if we scale by constant ln(3) in numerator and denominator such that f(lam) = 0.5 

I.e.
$$ f(x, \lambda) = \frac{1 - e^{-\frac{x\ln(3)}{\lambda}}}{1 + e^{-\frac{x\ln(3)}{\lambda}}} $$

wd60622:
Thoughts on this @juanitorduz

arthurmello:
So we could have `logistic_saturation` and `inverse_scaled_logistic_saturation` (and thus the corresponding `LogisticSaturation` and `InverseScaledLogisticSaturation`).
In `logistic_saturation` nothing changes, while in `InverseScaledLogisticSaturation`, we inverse lambda and multiply it by a scaling parameter (ln(3) by default).
What do you think?

[...]

PR #827, opened 2024-07-10
Create inverse_scaled_logistic_saturation and the corresponding class

arthurmello:
## Description
Offer a more intuitive alternative to `logistic_saturation`, where we use
$$f(x,λ)=\frac{1−e^{−xln⁡(3)/λ}}{1+e^{−xln⁡(3)/λ}}$$

Instead of the original:
$$f(x,λ)=\frac{1−e^{−xλ}}{1+e^{−xλ}}$$

Allowing for lambda to be interpreted as the half saturation point.

## Related Issue

- [ ] Closes #
- [x] Related to #220 

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--827.org.readthedocs.build/en/827/

### DIFF
--- a/pymc_marketing/mmm/components/saturation.py
+++ b/pymc_marketing/mmm/components/saturation.py
@@ -76,6 +76,7 @@ from pydantic import Field, InstanceOf, validate_call
 from pymc_marketing.mmm.components.base import Transformation
 from pymc_marketing.mmm.transformers import (
     hill_saturation,
+    inverse_scaled_logistic_saturation,
     logistic_saturation,
     michaelis_menten,
     tanh_saturation,
@@ -194,6 +195,39 @@ class LogisticSaturation(SaturationTransformation):
 
     def function(self, x, lam, beta):
         return beta * logistic_saturation(x, lam)
+
+    default_priors = {
+        "lam": Prior("Gamma", alpha=3, beta=1),
+        "beta": Prior("HalfNormal", sigma=2),
+    }
+
+
+class InverseScaledLogisticSaturation(SaturationTransformation):
+    """Wrapper around inverse scaled logistic saturation function.
+
+    For more information, see :func:`pymc_marketing.mmm.transformers.inverse_scaled_logistic_saturation`.
+
+    .. plot::
+        :context: close-figs
+
+        import matplotlib.pyplot as plt
+        import numpy as np
+        from pymc_marketing.mmm import InverseScaledLogisticSaturation
+
+        rng = np.random.default_rng(0)
+
+        adstock = InverseScaledLogisticSaturation()
+        prior = adstock.sample_prior(random_seed=rng)
+        curve = adstock.sample_curve(prior)
+        adstock.plot_curve(curve, sample_kwargs={"rng": rng})
+        plt.show()
+
+    """
+
+    lookup_name = "inverse_scaled_logistic"
+
+    def function(self, x, lam, beta):
+        return beta * inverse_scaled_logistic_saturation(x, lam)
 
     default_priors = {
         "lam": Prior("Gamma", alpha=3, beta=1),
@@ -339,6 +373,7 @@ SATURATION_TRANSFORMATIONS: dict[str, type[SaturationTransformation]] = {
     cls.lookup_name: cls
     for cls in [
         LogisticSaturation,
+        InverseScaledLogisticSaturation,
         TanhSaturation,
         TanhSaturationBaselined,
         MichaelisMentenSaturation,
--- a/pymc_marketing/mmm/components/saturation.py
+++ b/pymc_marketing/mmm/components/saturation.py
@@ -230,7 +230,7 @@ class InverseScaledLogisticSaturation(SaturationTransformation):
         return beta * inverse_scaled_logistic_saturation(x, lam)
 
     default_priors = {
-        "lam": Prior("Gamma", alpha=3, beta=1),
+        "lam": Prior("Gamma", alpha=0.3, beta=0.6),
         "beta": Prior("HalfNormal", sigma=2),
     }
 
--- a/pymc_marketing/mmm/components/saturation.py
+++ b/pymc_marketing/mmm/components/saturation.py
@@ -230,7 +230,7 @@ class InverseScaledLogisticSaturation(SaturationTransformation):
         return beta * inverse_scaled_logistic_saturation(x, lam)
 
     default_priors = {
-        "lam": Prior("Gamma", alpha=0.3, beta=0.6),
+        "lam": Prior("Gamma", alpha=0.5, beta=1),
         "beta": Prior("HalfNormal", sigma=2),
     }
 
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -482,6 +482,9 @@ def inverse_scaled_logistic_saturation(
     x, lam: npt.NDArray[np.float64] | float = 0.5, eps: float = np.log(3)
 ):
     """Inverse scaled logistic saturation transformation.
+        It offers a more intuitive alternative to logistic_saturation,
+        allowing for lambda to be interpreted as the half saturation point,
+        when using default values for lam and eps.
 
     .. math::
         f(x) = \\frac{1 - e^{-x*\epsilon/\lambda}}{1 + e^{-x*\epsilon/\lambda}}
@@ -514,7 +517,7 @@ def inverse_scaled_logistic_saturation(
     lam : float or array-like, optional, by default 0.5
         Saturation parameter.
     eps : float or array-like, optional, by default ln(3)
-        Scaling parameter.
+        Scaling parameter. ln(3) results in halfway saturation for lam = 0.5
 
     Returns
     -------
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -517,7 +517,7 @@ def inverse_scaled_logistic_saturation(
     lam : float or array-like, optional, by default 0.5
         Saturation parameter.
     eps : float or array-like, optional, by default ln(3)
-        Scaling parameter. ln(3) results in halfway saturation
+        Scaling parameter. ln(3) results in halfway saturation at lam
 
     Returns
     -------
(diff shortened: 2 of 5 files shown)


pymc_marketing/mmm/components/saturation.py
```diff
@@ -197,6 +198,39 @@ def function(self, x, lam, beta):
...
+
+    """
+
+    lookup_name = "inverse_scaled_logistic"
+
+    def function(self, x, lam, beta):
+        return beta * inverse_scaled_logistic_saturation(x, lam)
+
+    default_priors = {
+        "lam": Prior("Gamma", alpha=3, beta=1),
```
wd60622, line 229: Since we scale the x values between zero and 1, does it make sense to have the prior with much of the density between 0 and 1? 
What does this distribution looks like?

wd60622 (changes requested):
Amazing stuff. Thanks for this contribution!

Can you test that the tranformation at lam is about 0.5? It should be test in transformations.py

pymc_marketing/mmm/transformers.py
```diff
@@ -478,6 +478,52 @@ def logistic_saturation(x, lam: npt.NDArray[np.float64] | float = 0.5):
...
+        plt.show()
+
+    Parameters
+    ----------
+    x : tensor
+        Input tensor.
+    lam : float or array-like, optional, by default 0.5
+        Saturation parameter.
+    eps : float or array-like, optional, by default ln(3)
+        Scaling parameter.
```
wd60622, line 517: add that this results in halfway saturation at lam

pymc_marketing/mmm/transformers.py
```diff
@@ -478,6 +478,52 @@ def logistic_saturation(x, lam: npt.NDArray[np.float64] | float = 0.5):
     return (1 - pt.exp(-lam * x)) / (1 + pt.exp(-lam * x))
 
 
+def inverse_scaled_logistic_saturation(
+    x, lam: npt.NDArray[np.float64] | float = 0.5, eps: float = np.log(3)
+):
+    """Inverse scaled logistic saturation transformation.
+
```
wd60622, line 485: Think a callout for this parameterization property is all good here too! It is a neat property and very explainable

pymc_marketing/mmm/components/saturation.py
wdtccc: Maybe something like: 
Gamma(alpha=alpha beta=2 * alpha)

which would put average at 0.5 but choose alpha such that there is most of probability between 0 and 1 (because of the x domain)

wd60622: Maybe something like: 
Gamma(alpha=alpha beta=2 * alpha)

which would put average at 0.5 but choose alpha such that there is most of probability between 0 and 1 (because of the x domain)

This comes with the assumption that the half-way saturation is (likely) below the max spend of the media in training.  

Would be interested in hearing thoughts here!

arthurmello: thinking of Gamma(alpha=0.3, beta=0.6), so that:
P(lam>1) = 0.0338 (low enough)
P(lam>0.5) = 0.10811 (not too skewed towards 0)

tests/mmm/test_transformers.py
```diff
@@ -343,6 +344,32 @@ def test_logistic_saturation_min_max_value(self, x, lam):
         assert y_eval.max() <= 1
         assert y_eval.min() >= 0
 
+    def test_inverse_scaled_logistic_saturation_lam_half(self):
+        x = np.array([0.5] * 100)
+        y = inverse_scaled_logistic_saturation(x=x, lam=0.5, eps=np.ln(3))
```
wd60622, line 349: Can we test different values of lam then pass x=lam, lam=lam

tests/mmm/test_transformers.py
```diff
@@ -343,6 +344,32 @@ def test_logistic_saturation_min_max_value(self, x, lam):
...
+        "x, lam",
+        [
+            (np.ones(shape=(100)), 0.5),
+            (np.linspace(start=0.0, stop=1.0, num=50), 10),
+            (np.linspace(start=200, stop=1000, num=50), 0.001),
+            (np.zeros(shape=(100)), 1),
+        ],
+    )
+    def test_inverse_scaled_logistic_saturation_min_max_value(self, x, lam):
+        y = inverse_scaled_logistic_saturation(x=x, lam=lam)
```
wd60622, line 368:
```suggestion
x = np.array([0, 1, 100, 500, 5000])
lam = np.array([...])[:, None]
y = inversed_scaled_logistic_saturation(x=x, lam=lam)
```

Lets remove the parametrize and make used of vectorized inputs. Think it is a bit more clear and consise

wd60622 (changes requested):

pymc_marketing/mmm/transformers.py
```diff
@@ -478,6 +478,55 @@ def logistic_saturation(x, lam: npt.NDArray[np.float64] | float = 0.5):
...
+        plt.show()
+
+    Parameters
+    ----------
+    x : tensor
+        Input tensor.
+    lam : float or array-like, optional, by default 0.5
+        Saturation parameter.
+    eps : float or array-like, optional, by default ln(3)
+        Scaling parameter. ln(3) results in halfway saturation for lam = 0.5
```
wd60622, line 520: This sounds incorrect. Ln3 results in halfway at lam. Not just lam = 0.5

tests/mmm/test_transformers.py
arthurmello: done!

pymc_marketing/mmm/transformers.py
arthurmello: @wd60622 you're right, my bad. just fixed it.

tests/mmm/test_transformers.py
wd60622: Lam needs actual values though. 
The ... was place holder. 
Maybe np.array([0.01, 0.25, 0.75, 1.5, 5.0, 10.0, 15.0])

pymc_marketing/mmm/components/saturation.py
```diff
@@ -201,6 +202,39 @@ def function(self, x, lam, beta):
...
+
+    """
+
+    lookup_name = "inverse_scaled_logistic"
+
+    def function(self, x, lam, beta):
+        return beta * inverse_scaled_logistic_saturation(x, lam)
+
+    default_priors = {
+        "lam": Prior("Gamma", alpha=0.3, beta=0.6),
```
wd60622, line 233: This seems to have a lot of density close to zero. Ie immediate saturation. Maybe increase both of these a bit to shift the density while keeping reasonable mean

pymc_marketing/mmm/transformers.py
```diff
@@ -478,6 +478,55 @@ def logistic_saturation(x, lam: npt.NDArray[np.float64] | float = 0.5):
     return (1 - pt.exp(-lam * x)) / (1 + pt.exp(-lam * x))
 
 
+def inverse_scaled_logistic_saturation(
+    x, lam: npt.NDArray[np.float64] | float = 0.5, eps: float = np.log(3)
+):
+    """Inverse scaled logistic saturation transformation.
+        It offers a more intuitive alternative to logistic_saturation,
+        allowing for lambda to be interpreted as the half saturation point,
+        when using default values for lam and eps.
```
wd60622, line 487: This is not right. It is halfway for default eps and for all lam

wd60622 (changes requested):

tests/mmm/test_transformers.py
arthurmello: sorry, read too fast and assumed that it would generate an array full of "None".
in that case, wouldn't it be better to stick to parametrize, for consistency with `test_logistic_saturation_min_max_value`?
in other words, why should they be different?

pymc_marketing/mmm/transformers.py
arthurmello: fixed it!

pymc_marketing/mmm/components/saturation.py
arthurmello: just updated it, to `alpha=0.5, beta=1`

tests/mmm/test_transformers.py
wd60622: I was interpreting this test as checking result domain being between 0 and 1 for various x and lam. We can just make use of the vectorized inputs which, imo, is a bit more clear and concise than the parameterized.

arthurmello: ok, I have just updated it for the vectorized version

tests/mmm/test_transformers.py
```diff
@@ -343,6 +344,26 @@ def test_logistic_saturation_min_max_value(self, x, lam):
         assert y_eval.max() <= 1
         assert y_eval.min() >= 0
 
+    def test_inverse_scaled_logistic_saturation_lam_half(self):
+        x = np.array([0.5] * 100)
```
wd60622, line 348: Can these be different values then pass as both x and lam in function

wd60622 (changes requested):
One final request ☺️

tests/mmm/test_transformers.py
arthurmello: not sure I understood, do you mean something like this?:

```py
def test_inverse_scaled_logistic_saturation_lam_half(self):
        x = np.array([0.01, 0.1, 0.5, 1, 100])
        y = inverse_scaled_logistic_saturation(x=x, lam=x, eps=np.log(3))
        expected = np.array([0.5] * len(x))
        np.testing.assert_almost_equal(
            y.eval(),
            expected,
            decimal=5,
            err_msg="The function does not behave as expected at eps = ln(3).",
        )
```

wd60622: Yup seems right. Want to check f(x=lam, lam=lam) = 0.5 for various values. Also remove the eps in order to test the default value had this property

arthurmello: got it, thanks! just updated the PR

wd60622 (approved):
Looks good! Just running tests one last time

wd60622:
Thanks @arthurmello !

PR #829, opened 2024-07-11
Better default priors for TVP & TVP component

### LINKED ISSUE #751: Pull out TVP as component
Having an API like: 

```python
from pymc_marketing.mmm import TVP

tvp = TVP(...)

coords = {
    "date": dates_in_model, 
}
with pm.Model(coords=coords): 
    # Ideally one thing that the user need to know about. Example 
    days_since_reference = pm.Data("days_since_reference", ..., dims="date")
    intercept= pm.Determistic(
        "tvp_intercept", 
        tvp.apply(days_since_reference), 
        dims="date",  
    )

    mu = intercept # + media_contributions + controls + various
    # Rest of model


```
Related to #711 

CC: @ulfaslak

wd60622:
## Description

Combine Bill's work and different API for TVP components. WIP 

Looks like this: 

```python
import pandas as pd

import pymc as pm
import numpy as np

from pymc_marketing.hsgp_kwargs import HSGP
from pymc_marketing.mmm.plot import plot_curve

hsgp = HSGP(
    ls_lower=1,
    ls_upper=15,
    drop_first=True,
    cov_func="matern52",
   dims="time",
)

n = 52
X = np.arange(n)
hsgp.register_data(X)

dates = pd.date_range("2022-01-01", periods=n, freq="W-MON")
coords = {
    "time": dates,
}
with pm.Model(coords=coords) as model:
    f = hsgp.create_variable("f")

    prior = pm.sample_prior_predictive().prior

plot_curve(prior["f"], {"time"})
plt.show()
```

And out of sample prediction just requires `register_data` on `Data` container: 

```python
dates = pd.date_range("2022-01-01", periods=n, freq="W-MON")

hsgp = HSGP(
    ls_lower=1,
    ls_upper=15,
    drop_first=True,
    cov_func="matern52",
   dims=("time", "channel"),
)
coords = {
    "time": dates,
    "channel": ["A", "B"]
}
with pm.Model(coords=coords) as model:
    data = pm.Data("data", X, dims="time")
    f = (
        hsgp
        .register_data(data)
        .create_variable("f")
    )
    prior = pm.sample_prior_predictive().prior

with model:
    pm.set_data(
        {
            "data": np.arange(n, n + 10),
        },
        coords={"time": pd.date_range("2023-01-01", periods=10, freq="W-MON")},
    )
    post = pm.sample_posterior_predictive(prior, var_names=["f"])
```

## Related Issue

- [ ] Closes #751
- [ ] Related to #775 

## Checklist

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--829.org.readthedocs.build/en/829/


pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -433,17 +433,17 @@ def build_model(
...
+                    .register_data(time_index)
+                    .create_variable(
+                        name="intercept_temporal_latent_multiplier",
+                        dims="date",
+                    )
                 )
                 intercept = pm.Deterministic(
                     name="intercept",
-                    var=baseline_intercept * intercept_latent_process,
+                    var=pt.exp(baseline_intercept + intercept_latent_process),
```
cetagostini-wise, line 446: I'll add the same comment than before here:

Can we deep-dive a bit more about this implementation?

In the original implementation, The mean of 𝛽𝑡 would be the
**_beta (since the mean of multiplier is 1). Then variance of 𝛽𝑡 will depend on the variance of the multiplier.

Because of it, the initial implementation is more straightforward and interpretable (Personal opinion). Mostly around the response curves.

What would be the benefit of this parameterization, if it is not more interpretable?
Being exponential, it forces the final parameter to be positive. This depending on the parameter we want to be time dependent may or may not be needed.

I think that if the result of **_base + latent is a normal distribution, there should be no problem, the result is a LogNormal distribution. But if it is Gamma or HalfNormal, the result would be an even more skewed distribution -This might be feasible because the config controls the base distributions- I'm not sure if it's a problem or not, but have we thought about how this transformation would apply to different arbitrary distributions?

A couple of questions that may be very obvious and I am a little slow to see the answers! 😅

juanitorduz:
@wd60622 I think we should aim to get this one for 0.10.0 :D . But I think we need a clean branch ;)

PR #840, opened 2024-07-18
unpinned pymc version

### LINKED ISSUE #936: Run tests with dependency changes
No description provided.

wd60622:
## Description

Checking to see if the tests pass

## Related Issue

- [ ] Closes #936
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--840.org.readthedocs.build/en/840/


wd60622:
Ref to what fails when no pymc cap @juanitorduz @ColtAllen 

`make_node` method seems to be main issue. The tests with oldest supported pymc werent run due to auto stopping but should pass to my understanding. 

As we eventually work through this change, would it make sense to support current oldest and latest? Or do we expect to see a jump in the pymc min version? Wdyt?

juanitorduz:
Thanks @wd60622 ! I think we could tackle this issue once 5.17 is released and support from 5.16 onwards (so that we support two versions) WDYT?

wd60622:
> Thanks @wd60622 ! I think we could tackle this issue once 5.17 is released and support from 5.16 onwards (so that we support two versions) WDYT? 

Probably the easiest. Alternative would be to do conditional classes based on the version of pymc?

juanitorduz:
> > Thanks @wd60622 ! I think we could tackle this issue once 5.17 is released and support from 5.16 onwards (so that we support two versions) WDYT?
> 
> Probably the easiest. Alternative would be to do conditional classes based on the version of pymc?

No, I think this will be hard to maintain IMO

pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
     "pandas",
     "pydantic>=2.1.0",
     # NOTE: Used as minimum pymc version with ci.yml `OLDEST_PYMC_VERSION`
-    "pymc>=5.13.0,<5.16.0",
+    "pymc>=5.13.0",
```
juanitorduz, line 35: I think we should always maintain an upper bound. WDYT?

wd60622:
> > > Thanks @wd60622 ! I think we could tackle this issue once 5.17 is released and support from 5.16 onwards (so that we support two versions) WDYT?
> > 
> > 
> > Probably the easiest. Alternative would be to do conditional classes based on the version of pymc?
> 
> No, I think this will be hard to maintain IMO

Let's wait until 5.17 is released then and do a new lower bound of >=5.16.0

PR #867, opened 2024-07-24
Check for missing attrs after `sample_prior_predictive` and `fit`

wd60622:
## Description

Trying this solution out of separating out 
- the attrs creation from setting used after `fit` and `sample_prior_predictive`
  - check that all of the `__init__` parameters exist
- the init kwargs creations from attrs from actual loading

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--867.org.readthedocs.build/en/867/


pymc_marketing/model_builder.py
```diff
@@ -277,42 +299,59 @@ def set_idata_attrs(
 
         Raises
         ------
+        ValueError
+            If the attrs are missing.
```
wd60622, line 303:
```suggestion
            If the attrs are missing for a property initialization of the class
```

pymc_marketing/model_builder.py
```diff
@@ -277,42 +299,59 @@ def set_idata_attrs(
...
+        init_parameters: set[str] = set(signature(self.__init__).parameters.keys())  # type: ignore
+        # Remove since this will be stored in the fit_data group of InferenceData
+        init_parameters -= {"data"}
+
+        if missing_keys := init_parameters - attrs_keys:
+            msg = (
+                f"__init__ has parameters that are not in the attrs: {missing_keys}. "
+                "The save and load functionality will not work correctly."
+            )
+            raise ValueError(msg)
```
wd60622, line 352: Think we should do this step at initialization. WDYT? 
Then we'd know immediately if there is an issue with saving the model. Don't think this is too heavy of a calculation

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -298,19 +296,21 @@ def _generate_and_preprocess_model_data(  # type: ignore
...
+        attrs["adstock"] = json.dumps(self.adstock.lookup_name)
+        attrs["saturation"] = json.dumps(self.saturation.lookup_name)
+        attrs["adstock_first"] = json.dumps(self.adstock_first)
+        attrs["control_columns"] = json.dumps(self.control_columns)
+        attrs["channel_columns"] = json.dumps(self.channel_columns)
+        attrs["adstock_max_lag"] = json.dumps(self.adstock.l_max)
+        attrs["validate_data"] = json.dumps(self.validate_data)
+        attrs["yearly_seasonality"] = json.dumps(self.yearly_seasonality)
+        attrs["time_varying_intercept"] = json.dumps(self.time_varying_intercept)
+        attrs["time_varying_media"] = json.dumps(self.time_varying_media)
```
juanitorduz, line 311: I wonder if we could store these attributes in a configuration yml file so that we can add or remove one we just do it in one place and no need to manually check

juanitorduz (commented):
Overall looks good! just left a comment / question

pymc_marketing/mmm/delayed_saturated_mmm.py
wd60622: Are you saying yml file *instead* of the InferenceData attrs?

juanitorduz: No, in addition. I'm open to ideas ;) Just a centralized object that carries all these names. Imagine that we need to add another field. Ideally we just add it in one place and it just works.

wd60622: Wouldnt we have to always write parsing to and parsing from the current attrs / dict, json? Have a small example in mind?

juanitorduz: yes! you are probably right! Then we need to make sure our unit-tests are good to help us with additional updates and changes :)

juanitorduz (approved):
I think this is a step in the right direction! Thanks @wd60622 !

wd60622:
> I think this is a step in the right direction! Thanks @wd60622 !

Will give the check at init instead of after sample a try first. 
Will make issue for yml

pymc_marketing/model_builder.py
wd60622: This actually has a bit of difficulty which the creation of the `id` which depends on the model_config attribute

PR #870, opened 2024-07-24
Allow plot MMM components in the original scale

juanitorduz:
We can not get the plot in the example notabook 

image

----
📚 Documentation preview 📚: https://pymc-marketing--870.org.readthedocs.build/en/870/


pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -1072,6 +1072,166 @@ def plot_channel_parameter(self, param_name: str, **plt_kwargs: Any) -> plt.Figu
...
+        if original_scale:
+            return apply_sklearn_transformer_across_dim(
+                data=contributions,
+                func=self.get_target_transformer().inverse_transform,
+                dim_name="date",
+            )
+
+        return contributions
+
+    def plot_components_contributions(
```
juanitorduz, line 1104: This overwrites the method from the base class as in the MMM class we are assuming we have scalers.

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -1072,6 +1072,166 @@ def plot_channel_parameter(self, param_name: str, **plt_kwargs: Any) -> plt.Figu
...
 
+    def get_ts_contribution_posterior(
+        self, var_contribution: str, original_scale: bool = False
+    ) -> DataArray:
+        """Get the posterior distribution of the time series contributions of a given variable.
+
+        Parameters
+        ----------
+        var_contribution : str
+            The variable for which to get the contributions.
```
wd60622, line 1083: Does this work for all variables? Or is there only a few that work? Maybe list out

wd60622 (approved):
Looks good. Just comment on doc

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: Added better description in the doctring 👍

PR #881, opened 2024-07-28
`json.loads` with python types bug

wd60622:
## Description

`json.loads` doesn't support boolean types. The values would "true" and "false" if they existed

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--881.org.readthedocs.build/en/881/


juanitorduz (approved):

juanitorduz:
Do you want to add a unit-test? (why was it not failing before?)

wd60622:
> Do you want to add a unit-test? (why was it not failing before?)

Yes! Wasn't failing because we don't have any tests for backwards compat of the attrs in the MMM load

wd60622:
This should be good to go. Also found bug on the adstock and saturation and made the fix. However, those will change with #882

PR #882, opened 2024-07-28
Save off media transformations

### LINKED ISSUE #826: saving with media transformation config
MMM with `adstock` and `saturation` configurations can potentially not be saved correctly since the additional kwargs are not saved off. 

 Having these classes serializable and taking advantage of them in `save` and `load` methods would be a good idea.

 For instance, a model that would not be able to be loaded correctly would be:

 ```python
 from pymc_marketing.mmm import MMM, GeometricAdstock, LogisticSaturation

 mmm = MMM(
     ..., 
     adstock=GeometricAdstock(l_max=4, normalize=False, mode="Before"), 
     ..., 
)
```

This is because both the `normalize` and `mode` are not saved off.

This would involve reconstructing the `adstock` and `saturation` upon loading here: 

https://github.com/pymc-labs/pymc-marketing/blob/2be2664557fa84471e448d9ba12fe78ccacd69f3/pymc_marketing/mmm/delayed_saturated_mmm.py#L637-L659

wd60622:
## Description

Add `to_dict` method for media transformation and leverage in the `MMM` in order to save off configurations completely

## Related Issue

- [ ] Closes #826 
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--882.org.readthedocs.build/en/882/

### DIFF
--- a/tests/mmm/test_delayed_saturated_mmm.py
+++ b/tests/mmm/test_delayed_saturated_mmm.py
@@ -1190,8 +1190,6 @@ def test_missing_attrs_to_defaults(toy_X, toy_y) -> None:
         time_varying_media=False,
     )
     mmm = mock_fit(mmm, toy_X, toy_y)
-    mmm.idata.attrs.pop("adstock")
-    mmm.idata.attrs.pop("saturation")
     mmm.idata.attrs.pop("adstock_first")
     mmm.idata.attrs.pop("time_varying_intercept")
     mmm.idata.attrs.pop("time_varying_media")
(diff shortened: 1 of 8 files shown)


juanitorduz (approved):
Thanks @wd60622

juanitorduz:
@wd60622 we can merge this one after we you solve the merge conflicts :)

tests/mmm/test_delayed_saturated_mmm.py
wd60622, line 1: Any thoughts on this change? @juanitorduz 

If the adstock and saturation keys are not present than we *could* inject a default. However, I am not sure that is good idea

wd60622:
Any idea if the model `version` attribute needs to be updated?

tests/mmm/test_delayed_saturated_mmm.py
juanitorduz: Well, I think we need this to recover the model after saving it, right? What could be the drawbacks of injecting the default?

juanitorduz:
> Any idea if the model `version` attribute needs to be updated?

We actually do not have any explicit convention. I usually do it when there is a big refactor, but we can change that.

tests/mmm/test_delayed_saturated_mmm.py
wd60622: Injecting a default would likely not recreate the same model at all

PR #886, opened 2024-07-30
Automate UML creation

wd60622:
## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--886.org.readthedocs.build/en/886/


.github/workflows/uml.yml
```diff
@@ -0,0 +1,37 @@
+name: Update the UML Diagrams
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
```
wd60622, line 6: This would only be on merge to main branch. With that, we won't have to have it run all the time. i.e. all the pngs updating every PR

WDYT?

.github/workflows/uml.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Update the UML Diagrams
+        run: |
+          sudo apt-get update && sudo apt-get install -y graphviz
+          make uml
+
+      - name: Configure Git
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email "github-actions@github.com"
```
wd60622, line 31: This doesn't seem to render well. I am not sure of an alternative. Maybe it can be left blank

juanitorduz (approved):

.github/workflows/uml.yml
juanitorduz: we can leave it blank

.github/workflows/uml.yml
wd60622: Set it up that it doesn't push if there wasn't a change. So "every PR" is likely not true. 

Does knowing this change when this action should be run @juanitorduz ?

Issue #845, opened 2024-07-20
Example notebooks to not specify what version of pymc-marketing was used when last executed


cluhmann:
The MMM example notebook, the Budget Allocation notebook, the Lift Test Calibration notebook, and likely others specify the versions of python, matplotlib, arviz, numpy, pandas,  pymc, and pytensor. Strangely, the version of pymc-marketing is not included. This makes it very difficult to know what features described in the notebook might require in terms of package versions.  The version of pymc-marketing seems much more important than the version of something like matplotlib.

Furthermore, the CLV quickstart notebook has no package versions listed at all.

The only notebooks I could find that include a pymc-marketing version are the Pareto/NBD Model notebook, the "Shifted Beta Geometric model with individual customer data" notebook, the "Custom Models with MMM components", and the "other samplers" notebook.

wd60622:
Good thinking. Weird because the watermark command seems to be the same and version of that package is the same. Likely best to just be explicit with pymc marketing. 

How are these as steps?
- [ ] ensure all notebooks have watermark
- [ ] be explicit in watermark command at end of notebook

anevolbap:
This looks resolved on main: every notebook the docs build includes now prints the `pymc_marketing` version in its watermark output, including `clv/clv_quickstart.ipynb` (`%watermark -n -u -v -iv -w -p pymc_marketing,pymc,pytensor`). The only 4 notebooks without a watermark are under `clv/dev/`, which `docs/source/conf.py` excludes. OK to close?

PR #904, opened 2024-08-02
Add link model deployment to example notebook

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/897

Issue #860, opened 2024-07-23
Dynamic version based on git commit


wd60622:
No description provided.

wd60622:
PyMC uses [versioneer](https://pypi.org/project/versioneer/). The installation instructions are in the versioneer link

williambdean:
CC: @cetagostini

williambdean:
Maybe this
https://github.com/ofek/hatch-vcs

LeonxLJX:
Hi! I'd like to implement dynamic versioning based on git commit. Thanks!

LeonxLJX:
Hi! I'd like to make the dynamic version work when installing from git. Plan:

1. Switch the version in `pyproject.toml` to a `setuptools-scm`-style dynamic version so a git checkout always differs from the last PyPI release.
2. Verify `pip install` from the git URL replaces the old version.
3. Add docs note.

May I be assigned?

PR #907, opened 2024-08-02
Add pymc-marketing version to some MMM notebooks

juanitorduz:
Related to https://github.com/pymc-labs/pymc-marketing/issues/845

----
📚 Documentation preview 📚: https://pymc-marketing--907.org.readthedocs.build/en/907/


wd60622:
The issue with this is that it actually isn't 0.8.0 but what will come out in the next release. I.e. it is actually always ahead of the current release, right? What are your thoughts?

juanitorduz:
> The issue with this is that it actually isn't 0.8.0 but what will come out in the next release. I.e. it is actually always ahead of the current release, right? What are your thoughts?

I think for now we can add at least the lastest version. 

Another thing we can do is to update the version to 0.9.0dev while development

wd60622:
> I think for now we can add at least the lastest version.

Yeah, I am fine with that

> Another thing we can do is to update the version to 0.9.0dev while development

Have something like that for this #860 
Have you done this before?

wd60622 (approved):
Good stuff

juanitorduz:
Let me remove the mmm example nb from this PR as I added it in https://github.com/pymc-labs/pymc-marketing/pull/904

juanitorduz:
> Have you done this before?

No 🙈  hehe

PR #919, opened 2024-08-09
test running budget allocator nb

juanitorduz:
~~Just a test, do not merge!~~ It actually worked in https://github.com/pymc-labs/pymc-marketing/pull/919/commits/72100773b57095cb7a5e68ff9533d57ea90f37cc

**We need to comment on file saving, which is something we can do, as it is much more beneficial when trying to find bugs!**

----
📚 Documentation preview 📚: https://pymc-marketing--919.org.readthedocs.build/en/919/


juanitorduz:
When I run all the notebooks it fails ... maybe a memory issue?

juanitorduz:
> When I run all the notebooks it fails ... maybe a memory issue?

**We need to comment on file saving, which is something we can do, as it is much more beneficial when trying to find bugs!**

twiecki (approved):

Issue #917, opened 2024-08-07
does hill_sat go through (0, 0)


dirknbr:
Is it possible that your hill_saturation function doesn't pass through (0, 0), which would be necessary since we assume 0 sales if a single channel is 0

![Screenshot from 2024-08-07 16-36-12](https://github.com/user-attachments/assets/4198688d-0cf9-44d3-b90d-b085002761dc)

juanitorduz:
Thanks for reporting this @dirknbr ! We have a fix in https://github.com/pymc-labs/pymc-marketing/pull/920

dirknbr:
but does this fix mean that sigma is still the aymptote for all positive values of lambda and beta?

juanitorduz:
> but does this fix mean that sigma is still the aymptote for all positive values of lambda and beta?

Actually, is going to be sigma minus the offset. Let me reopen this issue and see is we bring the original parametrization.

dirknbr:
why not use equation 4 of https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46001.pdf

juanitorduz:
Ok! Seems the way to go. I am unsure about the current parameterization via exponentials.

juanitorduz:
> why not use equation 4 of https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46001.pdf

Ok! Sterted a PR https://github.com/pymc-labs/pymc-marketing/pull/925

juanitorduz:
We added the original implementation of the paper (with the corresponding reference)

PR #925, opened 2024-08-12
Original hill function definition

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/917

In https://github.com/pymc-labs/pymc-marketing/issues/917#issuecomment-2284338338 it is suggested to use Equation 5 from https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46001.pdf to have the original definition. I renamed the previous one as people might find it useful.

----
📚 Documentation preview 📚: https://pymc-marketing--925.org.readthedocs.build/en/925/


juanitorduz:
@wd60622 What do you think about this proposal? I think it is very reasonable.

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -63,11 +63,11 @@
 
 class BaseMMM(BaseValidateMMM):
     """
-    Base class for a media mix model using Delayed Adstock and Logistic Saturation (see [1]_).
+     Base class for a media mix model using Delayed Adstock and Logistic Saturation (see [1]_).
 
     References
-    ----------
-    .. [1] Jin, Yuxue, et al. “Bayesian methods for media mix modeling with carryover and shape effects.” (2017).
+     ----------
```
wd60622, line 69: what's up this this new space?

wd60622 (approved):
Looks good. Just question about the spaces. Don't think that causes any issues with docs though

pymc_marketing/mmm/delayed_saturated_mmm.py
juanitorduz: super weird! fixed!

juanitorduz:
> Looks good. Just question about the spaces. Don't think that causes any issues with docs though

Thanks! Good catch! Fixed!

PR #933, opened 2024-08-14
Solving optimizer issues & typos

cetagostini:
## Description

We got a few concerns regarding the optimizer in the [forum](https://discourse.pymc.io/t/mmm-allocate-budget-to-maximize-response-generating-weird-decimal-places-results/14828/10?u=cluhmann) and the [discussions](https://github.com/pymc-labs/pymc-marketing/discussions/843) regarding the response the optimizer. Deeper checks reveal an scale problem once we transform back the response from the optimizer.

On the other hand we got a [typo](https://discourse.pymc.io/t/pymc-marketing-question-on-total-budget-parameter-for-budget-allocation-model/14259/2) in the function discussed here.

Finally, we where adding implicit noise to the final optimization. We added the noise level parameter to let the user to control if they want or not to add noise to their estimations.

## Related Issue

- [x] Closes #843 
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--933.org.readthedocs.build/en/933/


pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -2153,9 +2159,11 @@ def allocate_budget_to_maximize_response(
             custom_constraints=custom_constraints,
         )
 
-        inverse_scaled_channel_spend = self.channel_transformer.inverse_transform(
+        inverse_scaled_channel_spend = (
             np.array([list(self.optimal_allocation_dict.values())])
+            * self.channel_transformer["scaler"].scale_.max()
```
juanitorduz, line 2164: Is there a way to add a test to ensure this change is working as expected?

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -2102,8 +2108,8 @@ def allocate_budget_to_maximize_response(
             The total budget to be allocated.
         time_granularity : str
             The granularity of the time periods (e.g., 'daily', 'weekly', 'monthly').
-        num_days : int
-            The number of days over which the budget is to be allocated.
+        horizon : int
+            The number of time units over which the budget is to be allocated.
```
cluhmann, line 2112: Maybe refer to these as "periods" because that's what the description of `time_granularity` refers to (assuming that `horizon` is measured in units of `time_granularity`)?  So:

```
horizon : int
    The number of time periods over which the budget is to be allocated.
```

cetagostini: Correct, I'll replace ✌🏻

cluhmann: Also, I am not sure what standard procedure for the project is given that we're at v0.8.0 at this point, but because this is a breaking change, I would typically recommend throwing a warning if the old argument name is used (I mean, usually you would want through a whole deprecation cycle with  future warnings, etc. but probably not necessary for versions < 1.0).

pymc_marketing/mmm/delayed_saturated_mmm.py
cetagostini: Totally agree, just added three test to verify if the input budget and estimated budget are the same. Same thing with bounds, and periods.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -50,7 +50,7 @@ class BudgetOptimizer(BaseModel):
         The adstock class.
     saturation : SaturationTransformation
         The saturation class.
-    num_days : int
+    periods : int
         The number of days.
```
cluhmann, line 54: Number of days or the number of periods?  Also, why `periods` and not `num_periods`?  The latter seems less ambiguous.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -65,7 +65,11 @@ class BudgetOptimizer(BaseModel):
     saturation: SaturationTransformation = Field(
         ..., description="The saturation transformation class."
     )
-    num_days: int = Field(..., gt=0, description="The number of days.")
+    periods: int = Field(
+        ...,
+        gt=0,
+        description="The number of time units over which the budget is to be allocated.",
```
cluhmann, line 71: This seems to suggest that `periods` is in units of periods, not days.

pymc_marketing/mmm/delayed_saturated_mmm.py
```diff
@@ -2102,8 +2108,8 @@ def allocate_budget_to_maximize_response(
             The total budget to be allocated.
         time_granularity : str
             The granularity of the time periods (e.g., 'daily', 'weekly', 'monthly').
-        num_days : int
-            The number of days over which the budget is to be allocated.
+        periods : float
```
cluhmann, line 2111: Same comment as above. Why `periods` instead of `num_periods`?

cluhmann (commented):
Focused on the code rather than the notebooks (which seem like they were simply re-executed). The test also seems ok, but I am not super confident in my evaluation of it.  Left a couple of comments about argument names and docstrings.

juanitorduz:
We have the following test failing because the change of parameter name for the budget optimizer 

```python
FAILED tests/mmm/test_budget_optimizer.py::test_allocate_budget[default_minimizer_kwargs]
FAILED tests/mmm/test_budget_optimizer.py::test_allocate_budget[custom_minimizer_kwargs]
FAILED tests/mmm/test_budget_optimizer.py::test_allocate_budget_zero_total[0-budget_bounds0-parameters0-expected_optimal0-2.38e-10]
FAILED tests/mmm/test_budget_optimizer.py::test_allocate_budget_custom_minimize_args
FAILED tests/mmm/test_budget_optimizer.py::test_allocate_budget_infeasible_constraints[100-budget_bounds0-parameters0-custom_constraints0]
```

We need to update the tests so that the optimizer uses  `periods` instead of `num_days`.

pymc_marketing/mmm/budget_optimizer.py
cetagostini: Okay trying to be more explicit ✌🏻

pymc_marketing/mmm/delayed_saturated_mmm.py
cetagostini: Made the changes!

cetagostini:
@juanitorduz the test should be done now!

@cluhmann the new description should be more understandable?

juanitorduz (approved):
Thank you @cetagostini !

cetagostini:
@juanitorduz Thanks to you, will wait for Chris opinion before merge!

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -50,8 +50,8 @@ class BudgetOptimizer(BaseModel):
         The adstock class.
     saturation : SaturationTransformation
         The saturation class.
-    num_days : int
-        The number of days.
+    num_periods : int
+        The number of time units.
```
cluhmann, line 54: In general, I think this doc string (not just this line) could use with a bit more elaboration. For example, the number of time units...what? But given that this is largely a bug fix, I say merge it and we can improve docstrings later.

cluhmann (approved):

Issue #918, opened 2024-08-09
Geometric ad stock accepts alpha values outside its domain


daniel-saunders-phil:
The docstring says alpha must be between 0 and 1. But you can pass negative and larger alpha values and it will produce strange results, silently. Negative alpha subtracts impressions from each day. Adstock greater than 1 causes exponential accumulation of impressions over time. I can't see users ever finding this behaviour helpful.

```
import numpy as np
from pymc_marketing.mmm.transformers import geometric_adstock

geometric_adstock(np.array([100,200,50]),alpha = 1.1).eval()
geometric_adstock(np.array([100,200,50]),alpha = -0.1).eval()
```

**Proposed solution**: Adding a check would make it easier to spot mistakes, especially in large hierarchical models with complex parameterizations.

```
    """
    if np.any(alpha >= 1) or np.any(alpha < 0):
            raise ValueError("alpha must be between greater than or equal to 0 and less than 1.")

    w = pt.power(pt.as_tensor(alpha)[..., None], pt.arange(l_max, dtype=x.dtype))
    w = w / pt.sum(w, axis=-1, keepdims=True) if normalize else w
    return batched_convolution(x, w, axis=axis, mode=mode)
```

wd60622:
Think this is a good idea. Probably want to use pymc / pytensor tooling though for the checks instead of numpy. The `pymc.logprob.utils.CheckParameterValue` can be used.

Example: 
https://github.com/pymc-devs/pymc-experimental/blob/af91b423095eb37017eb84ded91c14f0eae08494/pymc_experimental/distributions/continuous.py#L331

wd60622:
Thoughts @juanitorduz ?

juanitorduz:
Yes! We should use pytensor.

wd60622:
Checklist:

- [x] geometric adstock #960
- [ ] ...

Any other parameters that would have a restricted domain?

ricardoV94:
Copying my reply from #2874 

> As someone who doesn't care much about tools like mypy and pydantic... beware these sorts of checks can have big impacts in model speed (sampling, optimization, etc...). I wouldn't push too hard to enforce them.

> People usually love JAX for speed, and one of the first compromises JAX does is ignore any forms of checks whatsoever for the sake of speed.

juanitorduz:
Thank you @ricardoV94 ! Great input!

[...]

PR #960, opened 2024-08-22
Enforce `check_parameters` for `alpha` in `geometric_adstock`

PabloRoque:
## Description

* Introduces `pymc.distributions.dist_math.check_parameters` to validate `geometric_adstock`'s `alpha`.
  - I took the idea to use that function from pymc's [implementing_distribution.md](implementing_distribution.md)
  - Basically a wrapper around @wd60622 idea of using `pymc.logprob.utils.CheckParameterValue`
* Add a test to assert failure with bad `alpha` values

## Related Issue

- [ ] Closes #
- [x] Related to #918 

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--960.org.readthedocs.build/en/960/


pymc_marketing/mmm/transformers.py
```diff
@@ -235,6 +236,11 @@ def geometric_adstock(
        with carryover and shape effects." (2017).
 
     """
+    alpha = check_parameters(
+        alpha, [pt.gt(alpha, 0), pt.lt(alpha, 1)], msg="0 <= alpha <= 1"
+    ).eval()
+    l_max = check_parameters(l_max, [pt.gt(l_max, 0)], msg="l_max >= 0").eval()
```
wd60622, line 242: We won't want to use eval here

PabloRoque: Yep, I am seeing issues. 

Removing and having a look.

wd60622: It should only raise when the eval is called on the result of function

```python
# Fine
result = geometric_adstock(np.linspace(0, 1), alpha=-5)
# Causes error
result.eval()

Comment (wd60622) [reply pymc_marketing/mmm/transformers.py:242]: Just like the PyMC distributions: 

```python
# Fine
dist = pm.Normal.dist(mu=0, sigma=-1)
# ValueError
dist.eval()
```

tests/mmm/test_transformers.py
```diff
@@ -131,22 +132,38 @@ def test_geometric_adstock_x_zero(self, mode):
...
+    def test_geometric_adstock_check_parameters(self, x, alpha, l_max, suceeds):
+        if suceeds:
+            y = geometric_adstock(x=x, alpha=alpha, l_max=l_max)
+            y_np = y.eval()
+            assert y_np[0] == x[0]
+            assert y_np[1] == x[1] + alpha * x[0]
+            assert y_np[2] == x[2] + alpha * x[1] + (alpha**2) * x[0]
+        else:
+            with pytest.raises(ParameterValueError):
+                geometric_adstock(x=x, alpha=alpha, l_max=l_max)
```
wd60622, line 166: There should already be some tests for the sucesses. So we can just focus on the behavior of the bad params here

pymc_marketing/mmm/transformers.py
wd60622: Can this check not all be done in the same line? like in the example I linked?

PabloRoque: I removed the check on l_max > 0:
* If passed, an exception is already raised: `ValueError: length should not be negative`
* `batched_convolution` is called ahead of `.eval()`. So the previous exception is already raised and we never reach `ParameterValueError`

PabloRoque: `.eval()` now part of the test as: 

```
y = geometric_adstock(x=x, alpha=alpha, l_max=l_max)
y.eval()
```

tests/mmm/test_transformers.py
PabloRoque: Left the previous `test_geometric_adstock_good_alpha` as it was. Introduced `test_geometric_adstock_bad_alpha`

tests/mmm/test_transformers.py
```diff
@@ -148,6 +149,21 @@ def test_geometric_adstock_good_alpha(self, x, alpha, l_max):
...
+            (np.ones(shape=(100)), -1.7, 100),
+            (np.ones(shape=(100)), 22.5, 7),
+            (np.linspace(start=0.0, stop=1.0, num=50), -0.8, 3),
+            (np.linspace(start=0.0, stop=1.0, num=50), 8, 50),
+        ],
+    )
+    def test_geometric_adstock_bad_alpha(self, x, alpha, l_max):
+        with pytest.raises(ParameterValueError):
+            y = geometric_adstock(x=x, alpha=alpha, l_max=l_max)
+            y.eval()
```
wd60622, line 165:
```suggestion
        y = geometric_adstock(x=x, alpha=alpha, l_max=l_max)
        with pytest.raises(ParameterValueError):
            y.eval()
```

tests/mmm/test_transformers.py
```diff
@@ -148,6 +149,21 @@ def test_geometric_adstock_good_alpha(self, x, alpha, l_max):
         assert y_np[1] == x[1] + alpha * x[0]
         assert y_np[2] == x[2] + alpha * x[1] + (alpha**2) * x[0]
 
+    @pytest.mark.parametrize(
+        "x, alpha, l_max",
```
wd60622, line 153: Is the `x` important here? can it be the same for all the tests? Would be less moving parts 😄

wd60622 (commented):
Think this looks good. Just 2 suggestions

pymc_marketing/mmm/transformers.py
wd60622: Perf. Don't think we are passing l_max as TensorVariable anyway

tests/mmm/test_transformers.py
PabloRoque: I basically copied the values from `test_geometric_adstock_good_alpha`. 
Let me just test an int and a float smaller and greater than the bounds. `x = np.ones(shape=(100)` for all of them

tests/mmm/test_transformers.py
```diff
@@ -148,6 +149,20 @@ def test_geometric_adstock_good_alpha(self, x, alpha, l_max):
...
+        [
+            (np.ones(shape=(100)), -0.3, 10),
+            (np.ones(shape=(100)), -2, 10),
+            (np.ones(shape=(100)), 22.5, 10),
+            (np.ones(shape=(100)), 2, 10),
+        ],
+    )
+    def test_geometric_adstock_bad_alpha(self, x, alpha, l_max):
+        y = geometric_adstock(x=x, alpha=alpha, l_max=l_max)
```
juanitorduz, line 162: As x is the same for all cases I think it is better to put `x=np.ones(shape=(100)` inside the `test_geometric_adstock_bad_alpha` and just have 

```python
    @pytest.mark.parametrize(
        "alpha, l_max",
    . . . 
```

juanitorduz (changes requested):
Thanks @PabloRoque ! Just a small request and this one is good to go!

tests/mmm/test_transformers.py
wd60622, line 162:
```suggestion
    @pytest.mark.parametrize(
        "alpha",
        [-0.3, -2, 22.5, 2],
        ids=["less_than_zero_0", "less_than_zero_1", "greater_than_one_0", "greater_than_one_1"],
    )
    def test_geometric_adstock_bad_alpha(self, alpha):
        l_max = 10
        x = np.ones(shape=100)
        y = geometric_adstock(x=x, alpha=alpha, l_max=l_max)
```

juanitorduz: haha or this XD

wd60622: Easier to see what is happening imo

PabloRoque:
Thanks for the suggestions, and having a look at the PR. 
Indeed cleaner this way

wd60622:
pre-commit.ci autofix

wd60622 (approved):
Looks good! thanks for the edits

PR #970, opened 2024-08-23
scaling should be done before hdi computation posterior predictive

juanitorduz:
The scaling step  should be done before  the hdi computation in the posterior predictive plot. Tested locally in mmm notebook ✅ 

----
📚 Documentation preview 📚: https://pymc-marketing--970.org.readthedocs.build/en/970/


wd60622 (approved):
Good stuff

Issue #819, opened 2024-07-09
Time Varying Media not considered in `channel_contributions_forward_pass`


wd60622:
The time varying contributions are not multiplied out in the method.

Having the time_varying forward pass method might be the way to go

Reference: 
https://github.com/pymc-labs/pymc-marketing/blob/a97e272de174aa32859c858fe62bb192b8262b45/pymc_marketing/mmm/delayed_saturated_mmm.py#L544-L548

juanitorduz:
Shall we add this to the 0.8.0 milestone?

juanitorduz:
@wd60622 Do you want to take a stab at this one, as https://github.com/pymc-labs/pymc-marketing/pull/998 seems forgotten?

wd60622:
Yes, I will try to complete tomorrow

PR #998, opened 2024-09-04
Handle TVP media in `channel_contributions_forward_pass`

radiokosmos:
## Related Issue

- [ ] Closes #
- [x] Related to #819 

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--998.org.readthedocs.build/en/998/


radiokosmos:
I am not sure if it's right logic to do time_varying_channel_contributions_forward_pass and need some help with relevant test

pymc_marketing/mmm/mmm.py
```diff
@@ -615,6 +615,68 @@ def channel_contributions_forward_pass(
 
         return idata.posterior_predictive.channel_contributions.to_numpy()
 
+    def time_varying_channel_contributions_forward_pass(
```
wd60622, line 618: Can we just use the channel_contributions_forward_pass method? What is the benefit of having two separate ones?

pymc_marketing/mmm/mmm.py
```diff
@@ -982,8 +1050,13 @@ def get_channel_contributions_forward_pass_grid(
             channel_data = (
                 delta * self.preprocessed_data["X"][self.channel_columns].to_numpy()
             )
+            time_index = (
+                np.arange(0, self.preprocessed_data["X"].shape[0])
+                if self.time_varying_media
+                else None
+            )
```
wd60622, line 1057: Will we know if the dates are correct? Is this only used for the insampel data? What if the `channel_contributions` input are from a subset of the input. We will likely need to use the `infer_time_index` if this is used for new data. This a complaint of mine from the original implementation. If it is always used in the context of the insample, can we not just use the self._time_index variables create in `fit` call.

wd60622 (commented):
Just some thoughts to deduplicate logic

pymc_marketing/mmm/mmm.py
radiokosmos: Yeap, makes sense. I'll fix it

pymc_marketing/mmm/mmm.py
radiokosmos: This is a bit trickier. Should we

wd60622:
Hi @radiokosmos 
Are you still interested in this issue?

PR #1000, opened 2024-09-04
Update to lift test calibration notebook

### LINKED ISSUE #999: Suggested edits to Lift Test Calibration notebook
I have some recommended edits for the [Lift Test Calibration](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_lift_test.html) notebook.

## 1. Under the 'Requirements' section we have

> Today, we won’t be discussing how to conduct various tests, but instead, we will focus on their utilization. If you wish to acquire knowledge on how to generate results that are compatible with your MMM models, you can check out [CausalPy](https://causalpy.readthedocs.io/en/latest/) for conducting experiments.

I think it makes sense for us to build an example like that into CausalPy. There is already an issue (in pymc-marketing) to add an example using CausalPy: https://github.com/pymc-labs/pymc-marketing/issues/255. I don't really mind, but I think now that it could be a good idea to base this example in CausalPy.

So I'm just focussing the spotlight on https://github.com/pymc-labs/pymc-marketing/issues/255 again I guess. Happy to talk about how to do that. I'm happy to implement it.

## 2. Increase clarity around saturation curves

Under the "Indistinguishable Parameter Estimates" heading we have this:

Screenshot 2024-09-04 at 12 49 00

And just below in the "Unobserved Saturation Curves" section we have more saturation plots

Screenshot 2024-09-04 at 12 49 38

It was hard to tell what the differences were between these, so I think we need more explanation. I think the first sentence of this section does attempt to do that, but I think we need more elaboration on these being 'theoretical', hence we can see the whole continuous curve as a function of channel spend. Are they mean estimates based on a model fit, or are they the 'true' curves from our generated data?

And we can perhaps elaborate back up in the first plot that these are 'empirical'. I'm not 100% sure about this, but as far as I understand the terminology of 'direct response curves' means that these are plotted from empirically observed data? Maybe draw from some the docstring explanation of [`plot_direct_contribution_curves`](https://www.pymc-marketing.io/en/stable/api/generated/pymc_marketing.mmm.delayed_saturated_mmm.DelayedSaturatedMMM.plot_direct_contribution_curves.html#pymc_marketing.mmm.delayed_saturated_mmm.DelayedSaturatedMMM.plot_direct_contribution_curves)

It might even be pretty cool to directly plot the theoretical and the empirical over the top of each other. It might be easier to visually parse that the curves are pretty different, and to explain why.

In fact, we do this later on in the "Add lift tests to model" section... 

Screenshot 2024-09-04 at 13 16 39

Which is a great plot. But can we get some description of what the points (rug) are about? Do they represent `x + delta_x`? Needs to be made explicit and clear.

And I'm actually wondering: Would it be better to replace the points (direct response curves) with model estimated response curves with HDI's. 

## 3. Increase clarity about lift tests

I know this notebook focusses on adding in data from lift tests to calibrate/refine the MMM parameter estimates, but I think a _bit_ more explanation of the following could be a good idea:

> An experiment characterized in this way can be viewed as two points on the saturation curve for the channel.

I also think the functionally important parts of a lift test could be made more clear by explaining (either in text or a docstring of the custom `create_lift_test_from_actual_curve` function). From what I can tell it's something like:

> `delta_y` is the change in contribution going from a spend of `x` to `x + delta_x`. 

So, functionally a lift test (or the mysterious output of CausalPy applied to lift test results) is an estimate of the causal impact of the contribution of a channel when it's spend `x` was changed to `x + delta_x`. And we can link to a CausalPy example that we put together on this to talk about how we practically get that estimate.

I think the plot of lift test triangles is almost genius level, but needs a bit more explanation. Maybe this could be a 2 panel plot. The upper panel could show the triangles on top of the saturation function, and the lower panel (i.e. the current plot) could remain the same.

But I think it could be a good idea to demonstrate an example of a lift test where the spend (`delta_x`) goes _down_. So we'd have a triangle pointing downwards as the contribution will decrease. Or does this mess with the Gamma prior over `lift_measurements`? If there is a limitation that lift test data has to involve _increases_ in media spend, then that needs to be clearly mentioned.

We could maybe plot `df_lift_test` within an admonition box and add some text to expand out on "This is what you need from your lift tests in order to use it with pymc-marketing. Here is the definition of the columns. Does the `delta_y` value _have_ to be estimated channel contribution change, or can it be estimated _total change in outcome variable_? If it's the former then I think estimating this would require an MMM and be outside the scope of CausalPy? But if it's the latter, then I guess there are a number of ways to do that in CausalPy, including synthetic control.

## 4. Increased clarity about influence of lift tests

The forest plots showing the influence of the lift tests are great. But can we add some more description to pre-digest this for the reader. For example, the first set of revised parameter estimates are "only a bit better". So we should flag this up explicitly, saying that more lift tests will help you more, and here's an example of doing that.

The conclusion does an excellent job of doing that:

> ... can be used in order to incorporate experiments into our model likelihood and nudge the model parameters closer to the actuals in this example.

but I think it is beneficial to 'hold-hold' the reader a bit more by giving interpretations of the forest plots as they appear.

drbenvincent:
This issue closes #999 in that it provides various updates and clarifications to the [lift test calibration](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_lift_test.html) notebook.

## Checklist

## Modules affected
- [X] MMM
- [ ] CLV

## Type of change
- [ ] Bug fix
- [X] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1000.org.readthedocs.build/en/1000/


cetagostini:
I like the improvement in the doc, I leave some comments with ideas and opinions!

juanitorduz:
<p>+1</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1000/discussion/#comment-2332402720'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
@drbenvincent I love the changes 💪 ! Think about Carlo's suggestions and when you are done ping me for a quick last change ant merge !

drbenvincent:
<p>Good point - I now explicitly mention the adstock function. See what you think.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1000/discussion/#comment-2332402720'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Good idea, have done something along these lines, but retaining the max abs transformation</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1000/discussion/#comment-2332402740'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>I've given this a go. Can you double check the line <code>y=true_params["saturation_beta"][i] * df[f"x{i+1}_adstock_saturated"], # ??????</code> in the utility function <code>plot_channel_contributions</code> please? I think it's ok, but would appreciate a sanity check.  </p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1000/discussion/#comment-2332402767'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>done</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1000/discussion/#comment-2332402799'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
Thanks for comments @carlosagostini. I _think_ I've addressed your comments. I did drop in a question for you to double check I'm calculating the true channel contributions correctly.

juanitorduz:
<p>nice!</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1000/discussion/#comment-2332402740'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz (approved):

juanitorduz:
@carlosagostini if you have time can you please double check @drbenvincent 's las questions :)?

cetagostini:
Hey @drbenvincent and @juanitorduz the implementation is correct but I'll recommend do the following:

```python
# Save the true contribution for each channel
df[["true_contribution_x1", "true_contribution_x2"]] = pm.draw(fixed_model["channel_contributions"], random_seed=rng)
```

Then replace in the function by:
```python
def plot_channel_contributions(mmm, df):
    channels_contribution_original_scale = (
        mmm.compute_channel_contribution_original_scale()
    )
    channels_contribution_original_scale_hdi = az.hdi(
        ary=channels_contribution_original_scale
    )

    get_mean_contributions_over_time_df = mmm.compute_mean_contributions_over_time(
        original_scale=True
    )

    fig, ax = plt.subplots(
        nrows=2,
        figsize=(15, 8),
        ncols=1,
        sharex=True,
        sharey=False,
        layout="constrained",
    )

    for i, x in enumerate(["channel 1", "channel 2"]):
        # Estimate true contribution in the original scale from the data generating process
        sns.lineplot(
            x=df["date"],
            y=df[f"true_contribution_x{i+1}"], ## Change everything here.
            color="black",
            label=f"{x} true contribution",
            ax=ax[i],
        )
        # HDI estimated contribution in the original scale
        ax[i].fill_between(
            x=df["date"],
            y1=channels_contribution_original_scale_hdi.sel(channel=x)["x"][:, 0],
            y2=channels_contribution_original_scale_hdi.sel(channel=x)["x"][:, 1],
            color=f"C{i}",
            label=rf"{x} $94\%$ HDI contribution",
            alpha=0.4,
        )
        # Mean estimated contribution in the original scale
        sns.lineplot(
            x=df["date"],
            y=get_mean_contributions_over_time_df[x].to_numpy(),
            color=f"C{i}",
            label=f"{x} posterior mean contribution",
            alpha=0.8,
            ax=ax[i],
        )
        ax[i].legend(loc="center left", bbox_to_anchor=(1, 0.5))
        ax[i].set(title=x, ylim=(0, 1))
```

If we are using `pm.draw` already, then we can keep it simple estimating the true contribution from the start. What do you think?

cetagostini:
@juanitorduz @drbenvincent 

Another comment, we are adding numpyro to all our notebooks but it is not a dependency, neither is Jax. So a new user will have problems running our notebooks, sometimes I forget, for example now, trying this I got errors.

I would recommend not using numpyro in the examples, otherwise we should add it to the dependencies. What do you think? I had removed numpyro in a PR but apparently it came back 😅

cetagostini (approved):

juanitorduz:
Good point @cetagostini !

juanitorduz:
Thanks @drbenvincent !

drbenvincent:
Thanks. I'll get on these changes today/tomorrow.

PR #1011, opened 2024-09-09
use maxent from preliz instead of find_constrained_prior

aloctavodia:
PyMC's `find_constrained_prior` has been deprecated. This PR changes `constrain` function to use PreliZ's maxent.


juanitorduz (approved):
Thank you @aloctavodia !

PR #1026, opened 2024-09-12
Model Evaluation, Diagnostics and MLFlow Registry

### LINKED ISSUE #911: Evaluation and Model Diagnostic Functions
Hello!

Related to
- #901 
- #891 

It could be nice to have some standardised model evaluation and diagnostic functions added to pymc-marketing. Ideally they'd be formulated in a way that makes them easy to log in MLFLow later on.

Standard model metrics could include:
- Bayesian R2
- MAPE, RMSE, MAE
- Normalised RMSE and MAE (to allow comparisons across datasets and methodologies - in particular with Robyn models, who use NRMSE as one of their 2 key metrics)
- etc.

Diagnostic metrics could include:
- Step size, divergences
- LOOCV metrics e.g.
```    # Calculate elemwise log_likelihood of model given posteriors
    pm.compute_log_likelihood(
        mmm.idata, model=mmm.model, progressbar=False
    )  # Used for LOOCV
    model_loo = az.loo(mmm.idata)

    # Log LOOCV metrics
    loocv_metrics = {
        "loocv_elpd_loo": model_loo.elpd_loo,  # expected log pointwise predictive density
        "loocv_se": model_loo.se,  # standard error of elpd
        "loocv_p_loo": model_loo.p_loo,  # effective number of parameters
    }```
    
    
    I will open a draft PR on this ☺️

### LINKED ISSUE #973: Add Support for Model Registering to MLFlow Module
Hello! 

I was looking through the MLFlow module and I thought it might be nice to add a wrapper to allow people to actually register their MMM in MLFlow. Having access to the Registry would unlock a lot of functionality for MMM x MLFlow users. It could be good for faster deployment / serving of MMMs, better model lifecycle management and could help with refreshes.

[MLFlow Registry](https://mlflow.org/docs/latest/model-registry.html)

I already have some code written (and tested locally) for this, which I will add in [this PR](https://github.com/pymc-labs/pymc-marketing/pull/912)

### PR DESCRIPTION
**Pull request by louismagowan**
## Description

This PR is a recreation of [this one](https://github.com/pymc-labs/pymc-marketing/pull/912) which I was told to close and restart due to git issues on the pymc-marketing repo.

It could be nice to have some standardised model evaluation and diagnostic functions added to pymc-marketing. Ideally they'd be formulated in a way that makes them easy to log in MLFLow later on.
It would also be cool to build on top of the MLFlow module, to create a custom mlflow.pyfunc.PythonModel class to allow users to be able to register their models in the MLFlow registry. This would allow people to serve and maintain their MMMs more easily, and could help with MMM refreshes too.

Standard model metrics could include:
- Bayesian R2
- MAPE, RMSE, MAE
- Normalised RMSE and MAE (to allow comparisons across datasets and methodologies - in particular with Robyn models, who use NRMSE as one of their 2 key metrics)
etc.
- Wrapper functions to calculate those across the entire distributions of posteriors, not just against the means (so we can have HDI lower, upper for each metric too)

Diagnostic metrics could include:
- Step size, divergences
- LOOCV metrics e.g.

Some additional plots (also useful for diagnosing models):
- Plot prior vs posterior distributions, see how data is shifting things
- Plot HDI forests for a given var, along with r-hat value

Model Registry / Additional Logging Code:

A wrapper for an MMM model to make it conform to the MLFlow api, enabling registering and easier deployment
Also an option to load models from the registry as well / or download the idata from MLflow too.
I'll open this as a Draft for now - since I'll need advice on how where best to put this code, as well as overall design etc.

## Related Issue

- [x] Closes #911 
- [x] Closes #973 
- [x] Related to #891 #901 

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1026.org.readthedocs.build/en/1026/

### DIFF
--- a/pymc_marketing/model_builder.py
+++ b/pymc_marketing/model_builder.py
@@ -29,7 +29,6 @@ import xarray as xr
 from pymc.util import RandomState
 
 from pymc_marketing.hsgp_kwargs import HSGPKwargs
-from pymc_marketing.prior import Prior
 from pymc_marketing.utils import from_netcdf
 
 # If scikit-learn is available, use its data validator
@@ -353,8 +352,8 @@ class ModelBuilder(ABC):
         """
 
         def default(x):
-            if isinstance(x, Prior):
-                return x.to_json()
+            if hasattr(x, "to_dict"):
+                return x.to_dict()
             elif isinstance(x, HSGPKwargs):
                 return x.model_dump(mode="json")
             return x.__dict__
--- a/docs/source/notebooks/general/model_configuration.ipynb
+++ b/docs/source/notebooks/general/model_configuration.ipynb
@@ -2112,7 +2112,7 @@
    "source": [
     "### Serialization\n",
     "\n",
-    "The `to_json` and `from_json` methods can be helpful for storage of the distributions."
+    "The `to_dict` and `from_dict` methods can be helpful for storage of the distributions."
    ]
   },
   {
@@ -2142,7 +2142,7 @@
     }
    ],
    "source": [
-    "variable_2d_dict = variable_2d.to_json()\n",
+    "variable_2d_dict = variable_2d.to_dict()\n",
     "variable_2d_dict"
    ]
   },
@@ -2315,7 +2315,7 @@
     }
    ],
    "source": [
-    "Prior.from_json(variable_2d_dict).to_graph()"
+    "Prior.from_dict(variable_2d_dict).to_graph()"
    ]
   },
   {
@@ -2332,7 +2332,7 @@
     " \n",
     "### Backwards compatibility\n",
     "\n",
-    "The `from_json` method will create a `Prior` instance from the previous format.\n",
+    "The `from_dict` method will create a `Prior` instance from the previous format.\n",
     "\n",
     "For instance, take this previous configuration: "
    ]
@@ -2367,7 +2367,7 @@
    "id": "d63987c0-2525-4b12-8415-cd62dd66265f",
    "metadata": {},
    "source": [
-    "This can be parsed with the `Prior.from_json` constructor for each key. Much more consise too!"
+    "This can be parsed with the `Prior.from_dict` constructor for each key. Much more consise too!"
    ]
   },
   {
@@ -2390,7 +2390,7 @@
    ],
    "source": [
     "new_model_config = {\n",
-    "    name: Prior.from_json(key) for name, key in old_model_config.items()\n",
+    "    name: Prior.from_dict(key) for name, key in old_model_config.items()\n",
     "}\n",
     "\n",
     "new_model_config"
@@ -2414,7 +2414,7 @@
     }
    ],
    "source": [
-    "new_model_config[\"alpha\"].to_json()"
+    "new_model_config[\"alpha\"].to_dict()"
    ]
   },
   {
--- a/pymc_marketing/customer_choice/mv_its.py
+++ b/pymc_marketing/customer_choice/mv_its.py
@@ -219,9 +219,9 @@ class MVITS(ModelBuilder):
     @property
     def _serializable_model_config(self) -> dict[str, int | float | dict]:  # type: ignore
         result: dict[str, int | float | dict] = {
-            "intercept": self.model_config["intercept"].to_json(),
-            "likelihood": self.model_config["likelihood"].to_json(),
-            "market_distribution": self.model_config["market_distribution"].to_json(),
+            "intercept": self.model_config["intercept"].to_dict(),
+            "likelihood": self.model_config["likelihood"].to_dict(),
+            "market_distribution": self.model_config["market_distribution"].to_dict(),
         }
 
         return result
--- a/pymc_marketing/mmm/components/base.py
+++ b/pymc_marketing/mmm/components/base.py
@@ -579,9 +579,6 @@ class Transformation:
 
 
 def _serialize_value(value: Any) -> Any:
-    if isinstance(value, Prior):
-        return value.to_json()
-
     if hasattr(value, "to_dict"):
         return value.to_dict()
 
--- a/pymc_marketing/mmm/fourier.py
+++ b/pymc_marketing/mmm/fourier.py
@@ -321,7 +321,7 @@ class FourierBase(BaseModel):
 
         Parameters
         ----------
-        prior : Prior
+        prior : VariableFactory | Prior
             The prior distribution to serialize.
 
         Returns
@@ -330,10 +330,7 @@ class FourierBase(BaseModel):
             The serialized prior distribution.
 
         """
-        if hasattr(prior, "to_dict"):
-            return prior.to_dict()
-
-        return prior.to_json()
+        return prior.to_dict()
 
     @property
     def nodes(self) -> list[str]:
--- a/pymc_marketing/mmm/hsgp.py
+++ b/pymc_marketing/mmm/hsgp.py
@@ -319,7 +319,7 @@ class HSGPBase(BaseModel):
         data = self.model_dump()
 
         def handle_prior(value):
-            return value if not hasattr(value, "to_json") else value.to_json()
+            return value if not hasattr(value, "to_dict") else value.to_dict()
 
         return {key: handle_prior(value) for key, value in data.items()}
 
@@ -770,7 +770,7 @@ class HSGP(HSGPBase):
         """
         for key in ["eta", "ls"]:
             if isinstance(data[key], dict):
-                data[key] = Prior.from_json(data[key])
+                data[key] = Prior.from_dict(data[key])
 
         return cls(**data)
 
@@ -1002,6 +1002,6 @@ class HSGPPeriodic(HSGPBase):
         """
         for key in ["scale", "ls"]:
             if isinstance(data[key], dict):
-                data[key] = Prior.from_json(data[key])
+                data[key] = Prior.from_dict(data[key])
 
         return cls(**data)
--- a/pymc_marketing/prior.py
+++ b/pymc_marketing/prior.py
@@ -97,6 +97,7 @@ Create a prior with a custom transform function by registering it with
 from __future__ import annotations
 
 import copy
+import warnings
 from collections.abc import Callable
 from inspect import signature
 from typing import Any, Protocol, runtime_checkable
@@ -641,8 +642,10 @@ class Prior:
 
         return getattr(pz, self.distribution)(**self.parameters)
 
-    def to_json(self) -> dict[str, Any]:
-        """Convert the prior to the previous dictionary format.
+    def to_dict(self) -> dict[str, Any]:
+        """Convert the prior to dictionary format.
+
+        This is equivalent to the older PyMC-Marketing dictionary format.
 
         Returns
         -------
@@ -659,7 +662,7 @@ class Prior:
 
             dist = Prior("Normal", mu=0, sigma=1)
 
-            dist.to_json()
+            dist.to_dict()
             # {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}}
 
         Convert a hierarchical prior to the dictionary format.
@@ -673,7 +676,7 @@ class Prior:
                 dims="channel",
             )
 
-            dist.to_json()
+            dist.to_dict()
             # {
             #     "dist": "Normal",
             #     "kwargs": {
@@ -684,14 +687,14 @@ class Prior:
             # }
 
         """
-        json: dict[str, Any] = {
+        data: dict[str, Any] = {
             "dist": self.distribution,
         }
         if self.parameters:
 
             def handle_value(value):
                 if isinstance(value, Prior):
-                    return value.to_json()
+                    return value.to_dict()
 
                 if isinstance(value, pt.TensorVariable):
                     value = value.eval()
@@ -704,27 +707,45 @@ class Prior:
 
                 return value
 
-            json["kwargs"] = {
+            data["kwargs"] = {
                 param: handle_value(value) for param, value in self.parameters.items()
             }
         if not self.centered:
-            json["centered"] = False
+            data["centered"] = False
 
         if self.dims:
-            json["dims"] = self.dims
+            data["dims"] = self.dims
 
         if self.transform:
-            json["transform"] = self.transform
+            data["transform"] = self.transform
 
-        return json
+        return data
+
+    def to_json(self) -> dict[str, Any]:
+        """Convert the prior to dictionary format.
+
+        Deprecated in favor of :function:`pymc_marketing.prior.Prior.to_dict`.
+
+        Returns
+        -------
+        dict[str, Any]
+            The dictionary format of the prior.
+
+        """
+        warnings.warn(
+            "The `to_json` method is deprecated in favor of `to_dict`",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.to_dict()
 
     @classmethod
-    def from_json(cls, json) -> Prior:
+    def from_dict(cls, data) -> Prior:
         """Create a Prior from the dictionary format.
 
         Parameters
         ----------
-        json : dict[str, Any]
+        data : dict[str, Any]
             The dictionary format of the prior.
 
         Returns
@@ -740,25 +761,25 @@ class Prior:
 
             from pymc_marketing.prior import Prior
 
-            json = {
+            data = {
                 "dist": "Normal",
                 "kwargs": {"mu": 0, "sigma": 1},
             }
 
-            dist = Prior.from_json(json)
+            dist = Prior.from_dict(data)
             dist
             # Prior("Normal", mu=0, sigma=1)
 
         """
-        if not isinstance(json, dict):
+        if not isinstance(data, dict):
             msg = (
                 "Must be a dictionary representation of a prior distribution. "
-                f"Not of type: {type(json)}"
+                f"Not of type: {type(data)}"
             )
             raise ValueError(msg)
 
-        dist = json["dist"]
-        kwargs = json.get("kwargs", {})
+        dist = data["dist"]
+        kwargs = data.get("kwargs", {})
 
         def handle_value(value):
             if isinstance(value, dict):
@@ -770,13 +791,37 @@ class Prior:
             return value
 
         kwargs = {param: handle_value(value) for param, value in kwargs.items()}
-        centered = json.get("centered", True)
-        dims = json.get("dims")
+        centered = data.get("centered", True)
+        dims = data.get("dims")
         if isinstance(dims, list):
             dims = tuple(dims)
-        transform = json.get("transform")
+        transform = data.get("transform")
 
         return cls(dist, dims=dims, centered=centered, transform=transform, **kwargs)
+
+    @classmethod
+    def from_json(cls, json) -> Prior:
+        """Create a Prior from the dictionary format.
+
+        Deprecated in favor of :function:`pymc_marketing.prior.Prior.from_dict`.
+
+        Parameters
+        ----------
+        json : dict[str, Any]
+            The dictionary format of the prior.
+
+        Returns
+        -------
+        Prior
+            The prior distribution.
+
+        """
+        warnings.warn(
+            "The `from_json` method is deprecated in favor of `from_dict`",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return cls.from_dict(json)
 
     def constrain(
         self, lower: float, upper: float, mass: float = 0.95, kwargs=None
@@ -1133,7 +1178,7 @@ class Censored:
         return {
             "class": "Censored",
             "data": {
-                "dist": self.distribution.to_json(),
+                "dist": self.distribution.to_dict(),
                 "lower": handle_value(self.lower),
                 "upper": handle_value(self.upper),
             },
@@ -1144,7 +1189,7 @@ class Censored:
         """Create a censored distribution from a dictionary."""
         data = data["data"]
         return cls(  # type: ignore
-            distribution=Prior.from_json(data["dist"]),
+            distribution=Prior.from_dict(data["dist"]),
             lower=data["lower"],
             upper=data["upper"],
         )
@@ -1295,5 +1340,5 @@ def _is_censored_type(data: dict) -> bool:
     return data.keys() == {"class", "data"} and data["class"] == "Censored"
 
 
-register_deserialization(is_type=_is_prior_type, deserialize=Prior.from_json)
+register_deserialization(is_type=_is_prior_type, deserialize=Prior.from_dict)
 register_deserialization(is_type=_is_censored_type, deserialize=Censored.from_dict)
--- a/tests/mmm/components/test_adstock.py
+++ b/tests/mmm/components/test_adstock.py
@@ -153,7 +153,7 @@ def test_adstock_from_dict_without_priors(adstock, deserialize_func) -> None:
 
     adstock = deserialize_func(data)
     assert adstock.default_priors == {
-        k: Prior.from_json(v) for k, v in adstock.to_dict()["priors"].items()
+        k: Prior.from_dict(v) for k, v in adstock.to_dict()["priors"].items()
     }
 
 
--- a/tests/mmm/components/test_saturation.py
+++ b/tests/mmm/components/test_saturation.py
@@ -243,7 +243,7 @@ def test_saturation_from_dict_without_priors(saturation) -> None:
 
     saturation = saturation_from_dict(data)
     assert saturation.default_priors == {
-        k: Prior.from_json(v) for k, v in saturation.to_dict()["priors"].items()
+        k: Prior.from_dict(v) for k, v in saturation.to_dict()["priors"].items()
     }
 
 
--- a/tests/test_prior.py
+++ b/tests/test_prior.py
@@ -311,8 +311,8 @@ def test_transform() -> None:
         assert fast_eval(model[var_name]).shape == dim
 
 
-def test_to_json(large_var) -> None:
-    data = large_var.to_json()
+def test_to_dict(large_var) -> None:
+    data = large_var.to_dict()
 
     assert data == {
         "dist": "Normal",
@@ -347,9 +347,9 @@ def test_to_json(large_var) -> None:
     }
 
 
-def test_to_json_numpy() -> None:
+def test_to_dict_numpy() -> None:
     var = Prior("Normal", mu=np.array([0, 10, 20]), dims="channel")
-    assert var.to_json() == {
+    assert var.to_dict() == {
         "dist": "Normal",
         "kwargs": {
             "mu": [0, 10, 20],
@@ -358,8 +358,8 @@ def test_to_json_numpy() -> None:
     }
 
 
-def test_json_round_trip(large_var) -> None:
-    assert Prior.from_json(large_var.to_json()) == large_var
+def test_dict_round_trip(large_var) -> None:
+    assert Prior.from_dict(large_var.to_dict()) == large_var
 
 
 def test_constrain_with_transform_error() -> None:
@@ -433,7 +433,7 @@ def mmm_default_model_config():
 
 def test_backwards_compat(mmm_default_model_config) -> None:
     result = {
-        param: Prior.from_json(value)
+        param: Prior.from_dict(value)
         for param, value in mmm_default_model_config.items()
     }
     assert result == {
@@ -479,7 +479,7 @@ def test_to_graph() -> None:
     assert isinstance(G, Digraph)
 
 
-def test_from_json_list() -> None:
+def test_from_dict_list() -> None:
     data = {
         "dist": "Normal",
         "kwargs": {
@@ -489,13 +489,13 @@ def test_from_json_list() -> None:
         "dims": "channel",
     }
 
-    var = Prior.from_json(data)
+    var = Prior.from_dict(data)
     assert var.dims == ("channel",)
     assert isinstance(var["mu"], np.ndarray)
     np.testing.assert_array_equal(var["mu"], [0, 1, 2])
 
 
-def test_from_json_list_dims() -> None:
+def test_from_dict_list_dims() -> None:
     data = {
         "dist": "Normal",
         "kwargs": {
@@ -505,14 +505,14 @@ def test_from_json_list_dims() -> None:
         "dims": ["channel", "geo"],
     }
 
-    var = Prior.from_json(data)
+    var = Prior.from_dict(data)
     assert var.dims == ("channel", "geo")
 
 
-def test_to_json_transform() -> None:
+def test_to_dict_transform() -> None:
     dist = Prior("Normal", transform="sigmoid")
 
-    data = dist.to_json()
+    data = dist.to_dict()
     assert data == {
         "dist": "Normal",
         "transform": "sigmoid",
@@ -655,7 +655,7 @@ def test_serialize_with_pytensor() -> None:
     sigma = pt.arange(1, 4)
     dist = Prior("Normal", mu=0, sigma=sigma)
 
-    assert dist.to_json() == {
+    assert dist.to_dict() == {
         "dist": "Normal",
         "kwargs": {
             "mu": 0,
@@ -809,7 +809,7 @@ def test_censored_to_dict() -> None:
     data = censored_normal.to_dict()
     assert data == {
         "class": "Censored",
-        "data": {"dist": normal.to_json(), "lower": 0, "upper": float("inf")},
+        "data": {"dist": normal.to_dict(), "lower": 0, "upper": float("inf")},
     }
 
 
@@ -850,7 +850,7 @@ def test_create_prior_with_arbitrary_serializable(arbitrary_serialized_data) -> 
         dims=("channel", "geo"),
     )
 
-    assert dist.to_json() == {
+    assert dist.to_dict() == {
         "dist": "Normal",
         "kwargs": {
             "mu": arbitrary_serialized_data,
@@ -898,7 +898,7 @@ def test_censored_with_tensor_variable() -> None:
     assert censored_normal.to_dict() == {
         "class": "Censored",
         "data": {
-            "dist": normal.to_json(),
+            "dist": normal.to_dict(),
             "lower": [0, 1, 2],
             "upper": float("inf"),
         },
@@ -1009,3 +1009,18 @@ def test_censored_logp(mu) -> None:
 
     point = {}
     np.testing.assert_allclose(logp(point), expected_logp(point))
+
+
+def test_to_json_deprecation() -> None:
+    match = "The `to_json` method is deprecated"
+    with pytest.warns(DeprecationWarning, match=match):
+        Prior("Normal").to_json()
+
+
+def test_from_json_deprecation() -> None:
+    data = {
+        "dist": "Normal",
+    }
+    match = "The `from_json` method is deprecated"
+    with pytest.warns(DeprecationWarning, match=match):
+        Prior.from_json(data)

### CODE REVIEW

**Comment by juanitorduz** [comment]
@louismagowan, thanks for opening this one up again! Some tests (and the pre-commit) are failing. Do you need some help?

**Comment by louismagowan** [comment]
@juanitorduz - thanks but I think I'm alright for now (or at least I think I can try and fix it myself later). I think I probably need to recreate my conda env (as I think it's a bit old now).
Apologies for being slow on this, I have a deadline at work that I've been rushing for - so really not had much time to work on this recently.

@wd60622 - no need to review again yet, I'm still working on addressing your earlier feedback (I'll ping myself when I think I'm ready again)

Thanks!

**Comment by louismagowan** [comment]
Hi @wd60622 👋  - what do you think of this now? ☺️ 

I still need to add some tests, but I thought I'd ask for some more feedback before I start doing that 

Thanks!

**Comment by juanitorduz** [comment]
Thanks @louismagowan ! We are takibg aonw days off to recharge 😄! We will come ti this one in a couple of days 🙏

File: pymc_marketing/mlflow.py [hunk 1]
```diff
@@ -393,6 +396,69 @@ def log_sample_diagnostics(
...
+    """
+    if "posterior" not in idata or "sample_stats" not in idata:
+        logging.warning(
+            "Skipping LOOCV metrics: InferenceData object does not contain\
+                         the group posterior and sample stats."
+        )
+        return
+
+    # try-excepts necessary in case of empty models, like `test_log_data_no_data`
+    if "log_likelihood" not in idata:
```
wd60622, line 429: I would suggest the computation is done separately from the logging. The if block makes it hard to extend for new models.

pymc_marketing/mlflow.py
```diff
@@ -176,7 +179,7 @@ def _backwards_compatiable_data_vars(model: Model) -> list[TensorVariable]:
 
 
 def log_data(model: Model, idata: az.InferenceData) -> None:
```
wd60622, line 181: `log_metadata` a better name then?

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,535 @@
...
+        Raises
+        ------
+        ValueError
+            If the required attributes (idata) are not found.
+        """
+        if not hasattr(self.model, "idata") or self.model.idata is None:
+            raise ValueError(
+                "The model object must have a non-None 'idata' attribute. "
+                "Ensure you've called model.fit() before running diagnostics."
+            )
```
wd60622, line 137: Is this different than the check at initialization?

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,535 @@
...
+            hdi_prob=hdi_prob,
+            r_hat=True,
+            **plot_kwargs,
+        )
+
+        fig.suptitle(f"Posterior Distributions: {hdi_prob:.1%} HDI")
+
+        return fig
+
+    def plot_prior_vs_posterior(
```
wd60622, line 154: Might be better suited in the MMM class itself. Heavy coupling using the channel_columns attributes, etc

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,535 @@
...
+
+    def __init__(self, model: MMM):
+        if not hasattr(model, "idata") or model.idata is None:
+            raise ValueError("The model object must have a non-None 'idata' attribute.")
+        if not hasattr(model, "model") or model.model is None:
+            raise ValueError("The model object must have a non-None 'model' attribute.")
+
+        self.model = model
+        self.metric_distributions: dict[str, np.ndarray] = {}
+        self.metric_summaries: dict[str, dict[str, float]] = {}
```
wd60622, line 56: I find this a bit confusing. Also with the required order of the method calls

pymc_marketing/mlflow.py
juanitorduz: Agree! Let's make functions do one thing only ;)

pymc_marketing/mlflow.py
```diff
@@ -412,14 +478,324 @@ def log_inference_data(
...
+                var_names=self.var_names,
+                **self.sample_kwargs,
+            )
+        elif predict_method == "sample_prior_predictive":
+            return self.model.sample_prior_predictive(
+                model_input,
+                extend_idata=self.extend_idata,
+                combined=self.combined,
+                var_names=self.var_names,
+                **self.sample_kwargs,  # type: ignore[arg-type]
```
juanitorduz, line 631: Why don't you have this `# type: ignore[arg-type]` in the functions above?

pymc_marketing/mmm/evaluation.py
juanitorduz: Can we also add some information about these attributes to the class doctrings?

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,535 @@
...
+
+    # Same error metric as Robyn
+    @staticmethod
+    def nrmse(y_true: np.ndarray, y_pred: np.ndarray) -> float:
```
juanitorduz, line 60: Shall we put these metrics in https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/metrics.py?

pymc_marketing/mmm/evaluation.py
juanitorduz: yes! I agree! this does not need to be part of an "evaluator"

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,535 @@
...
+        # Hide any unused subplots
+        for j in range(i + 1, len(axs)):
+            fig.delaxes(axs[j])
+
+        # Adjust layout
+        plt.tight_layout(rect=[0, 0.03, 1, 0.97])  # Adjust layout to fit the title
+
+        return fig
+
+    def calculate_metric_distributions(
```
juanitorduz, line 306: I do not see where we are using the models in these methods. What about using just functions instead of the `MMMEvaluator` class?

pymc_marketing/model_builder.py
```diff
@@ -777,6 +777,43 @@ def sample_posterior_predictive(
 
         return az.extract(post_pred, variable_name, combined=combined)
 
+    def compute_log_likelihood(
```
juanitorduz, line 780: It seems this is a wrapper of `pm.compute_log_likelihood` and, therefore, redundant. Shall we remove this and simply use `pm.compute_log_likelihood` when needed

juanitorduz (commented):
Hey @louismagowan This is looking great! I left some questions and comments :)

pymc_marketing/mmm/evaluation.py
louismagowan: Deleted the class and moved to functions, so no longer relevant -thanks! ☺️

pymc_marketing/mlflow.py
louismagowan: Yep good spot, added! 🙇

louismagowan:
@wd60622 @juanitorduz  - thanks both for your comments! I made some more changes, hopefully it's looking a bit better now 🙇 

I'll still hold off on tests until we're happy with what's already been submitted ☺️

wd60622:
Thanks for making the edit! I will take at the end of the weekend

pymc_marketing/metrics.py
```diff
@@ -127,3 +131,45 @@ def crps(
     - For an introduction to CRPS, see https://towardsdatascience.com/crps-a-scoring-function-for-bayesian-machine-learning-models-dd55a7a337a8
     """
     return np.average(per_observation_crps(y_true, y_pred), weights=sample_weight)
+
+
+# Same error metric as Robyn
```
juanitorduz, line 136: I would remove this comment (are hard to maintain)

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,238 @@
...
+            "min": np.min(distribution),
+            "max": np.max(distribution),
+            f"{hdi_prob:.0%}_hdi_lower": hdi[0],
+            f"{hdi_prob:.0%}_hdi_upper": hdi[1],
+        }
+
+    return metric_summaries
+
+
+def evaluate_model(
```
juanitorduz, line 141: I would rename this method to something more explicitly like: `compute_summary_metrics` (or similar) as it is closer to what is doing, what do you think?

juanitorduz (commented):
Looking very close to the end! I left two small comments, but I will let @wd60622 look deeper into the mlflow part as he is more experienced with this module  :) (for me the changes look reasonable)

pymc_marketing/mmm/evaluation.py
louismagowan: Yep, good idea - thanks!

pymc_marketing/mlflow.py
```diff
@@ -412,14 +460,319 @@ def log_inference_data(
...
+
+    Parameters
+    ----------
+    model : MMMWrapper
+        The MMM to be logged.
+    artifact_path : str, optional
+        The path to the artifact to be logged. Defaults to "mmm_model".
+    conda_env : dict, optional
+        A dictionary representation of a Conda environment. Defaults to the default conda environment.
+    registered_model_name : str, optional
```
wd60622, line 632: doesn't match up with the parameter name. I prefer this name over one with a prefix of "mmm"

pymc_marketing/mlflow.py
```diff
@@ -412,14 +460,319 @@ def log_inference_data(
...
+            os.rmdir("idata")
+
+    return model
+
+
 @autologging_integration(FLAVOR_NAME)
 def autolog(
     log_sampler_info: bool = True,
-    log_datasets: bool = True,
+    log_metadata_info: bool = True,
```
wd60622, line 769: I find "_info" redundant with name of metadata

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,238 @@
...
+
+    if metrics_to_calculate is None:
+        metrics_to_calculate = list(metric_functions.keys())
+    else:
+        invalid_metrics = set(metrics_to_calculate) - set(metric_functions.keys())
+        if invalid_metrics:
+            raise ValueError(
+                f"Invalid metrics: {invalid_metrics}. "
+                f"Valid options are: {list(metric_functions.keys())}"
+            )
```
wd60622, line 72: No harm in checking again and reduce some indentation
```suggestion
      invalid_metrics = set(metrics_to_calculate) - set(metric_functions.keys())
      if invalid_metrics:
          raise ValueError(
              f"Invalid metrics: {invalid_metrics}. "
              f"Valid options are: {list(metric_functions.keys())}"
          )
```

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,238 @@
...
+    Sets
+    ----
+    self.metric_summaries : dict of str to dict
+        Stores the calculated metric summaries as an instance attribute.
+    """
+    if metric_distributions is None:
+        raise ValueError(
+            "Metric distributions have not been calculated.\
+            Call `calculate_metric_distributions` first."
+        )
```
wd60622, line 123: We shouldn't support the None case any more
```suggestion
```

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,238 @@
...
+                for i in range(y_pred.shape[1])
+            ]
+        )
+        results[metric] = metric_values
+
+    return results
+
+
+def summarize_metric_distributions(
+    metric_distributions: dict[str, np.ndarray] | None = None,
```
wd60622, line 90: remove the None case
```suggestion
    metric_distributions: dict[str, np.ndarray],

File: pymc_marketing/mlflow.py [hunk 17]
```diff
@@ -579,6 +939,22 @@ def new_fit(*args, **kwargs):
...
             log_inference_data(idata, save_file="idata.nc")
 
+            if log_loocv:
+                log_loocv_metrics(model=self, idata=idata)
+
+            posterior_preds = self.sample_posterior_predictive(self.X)
+            log_summary_metrics(
+                y_true=self.y,
+                y_pred=posterior_preds[
+                    self.output_var[0]
```
wd60622, line 949: I thought MMM.output_var was a string

pymc_marketing/mlflow.py
```diff
@@ -412,14 +460,319 @@ def log_inference_data(
...
+    metric_summaries = compute_summary_metrics(
+        y_true=y_true_np,
+        y_pred=y_pred_np,
+        metrics_to_calculate=metrics_to_calculate,
+        hdi_prob=hdi_prob,
+    )
+
+    for metric, stats in metric_summaries.items():
+        for stat, value in stats.items():
+            mlflow.log_metric(f"{metric}_{stat.replace('%', '')}", value)
```
wd60622, line 497: Same comment as above `log_metrics`

wd60622 (commented):
Abstraction looks really good. Let's build out the tests!

pymc_marketing/mlflow.py
louismagowan: But isn't it necessary otherwise you're passing an arg which is also the name of a function so it throws an error? Is there a name for the arg you like better?

pymc_marketing/mlflow.py
```diff
@@ -531,29 +887,33 @@ def autolog(
...
+                    )
+
+                if log_model_info:
+                    log_model_derived_info(model)
+
+                if log_metadata_info:
+                    log_metadata(model=model, idata=idata)
+
+                if log_loocv:
+                    log_loocv_metrics(model=model, idata=idata)
```
wd60622, line 916:
```suggestion
                    log_loocv_metrics(idata=idata)
```

pymc_marketing/mlflow.py
wd60622: We'd have to separate out into module. However, we can wait on that

wd60622:
Hi @louismagowan 
Let me know if you need any support with writing tests. That's something I gladly do 😄

pymc_marketing/mlflow.py
```diff
@@ -393,6 +396,51 @@ def log_sample_diagnostics(
...
+    model_loo = az.loo(idata)
+
+    loocv_metrics = {
+        "loocv_elpd_loo": model_loo.elpd_loo,
+        "loocv_se": model_loo.se,
+        "loocv_p_loo": model_loo.p_loo,
+    }
+
+    for metric_name, metric_value in loocv_metrics.items():
+        mlflow.log_metric(metric_name, metric_value)
```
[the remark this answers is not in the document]
louismagowan: Yes haha, good point!

pymc_marketing/mmm/evaluation.py
```diff
@@ -63,13 +63,12 @@ def calculate_metric_distributions(
...
-            raise ValueError(
-                f"Invalid metrics: {invalid_metrics}. "
-                f"Valid options are: {list(metric_functions.keys())}"
-            )
+      invalid_metrics = set(metrics_to_calculate) - set(metric_functions.keys())
+      if invalid_metrics:
+          raise ValueError(
+              f"Invalid metrics: {invalid_metrics}. "
+              f"Valid options are: {list(metric_functions.keys())}"
+          )
```
wd60622, line 71: Dedent this again

pymc_marketing/mlflow.py
louismagowan: Ahh this one is just because MLFlow doesn't support logging with "%" in the name - so it was either this or a dict comprehension followed by `log_metrics()` - which pretty much are same-same in my mind.

I added a comment though!

pymc_marketing/mlflow.py
```diff
@@ -531,29 +887,33 @@ def autolog(
...
-                )
-
             model = pm.modelcontext(kwargs.get("model"))
-            if log_model_info:
-                log_model_derived_info(model)
+            with model:
+                idata = sample(*args, **kwargs)
+                mlflow.log_param("pymc_marketing_version", __version__)
+                mlflow.log_param("pymc_version", pm.__version__)
+                mlflow.log_param("nuts_sampler", kwargs.get("nuts_sampler", "pymc"))
```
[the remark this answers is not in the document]
louismagowan: Sorry not sure I understand this one - do you mean something like this:

```
def log_sample_diagnostics(
    idata: az.InferenceData,
    tune: int | None = None,
) -> None:
    # ... (existing code) ...

    if inference_library := posterior.attrs.get("inference_library"):
        mlflow.log_param("inference_library", inference_library)
        mlflow.log_param(
            "inference_library_version",
            posterior.attrs["inference_library_version"],
        )
    mlflow.log_param("arviz_version", posterior.attrs["arviz_version"])
    # move this here
    mlflow.log_param("pymc_marketing_version", __version__)

# ... (rest of the file)

def patch_sample(sample):
    @wraps(sample)
    def new_sample(*args, **kwargs):
        model = pm.modelcontext(kwargs.get("model"))
        with model:
            idata = sample(*args, **kwargs)
            
            # Remove these lines as they're now handled in log_sample_diagnostics
            # mlflow.log_param("pymc_marketing_version", __version__)
            # mlflow.log_param("pymc_version", pm.__version__)
            
            mlflow.log_param("nuts_sampler", kwargs.get("nuts_sampler", "pymc"))

            # ... (rest of the function)

    return new_sample
```

pymc_marketing/mlflow.py
louismagowan: Yep good catch, changed ✅

louismagowan:
> Hi @louismagowan Let me know if you need any support with writing tests. That's something I gladly do 😄

Thanks @wd60622 ! I might have a first stab at some tests this weekend if that's alright, as it's something I'd like to get a bit better at. I'll see how I go this weekend and then reach out if I need some support 🙏

pymc_marketing/mlflow.py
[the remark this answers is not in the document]
wd60622: Those are the areas I had in mind. Are some of them referring to the same values and be duplicated? I just want to consolidate if we can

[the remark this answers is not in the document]
louismagowan: Hmmm there seems to be something I don't understand / have context for in the tests - seems like we need to log the pymc version differently depending on what sampler is used? Would have to change the tests quite a bit if we wanted to only log one of `pymc_version` and `inference_library_version`.

I did add the sampler logging to the MMM `fit` patch though, this was missing haha

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,320 @@ def log_inference_data(
...
+    )
+    run_id = mlflow.active_run().info.run_id
+    model_uri = f"runs:/{run_id}/{artifact_path}"
+
+    if registered_model_name:
+        mlflow.register_model(model_uri, registered_model_name)
+
+
+def load_model(
+    run_id: str | None = None,
```
wd60622, line 721: This shouldn't allow None unless we get the active run from the context. Do other mmlflow load functions get the run_id as well?

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,320 @@ def log_inference_data(
...
+    else:
+        idata_path = mlflow.artifacts.download_artifacts(
+            run_id=run_id, artifact_path="idata.nc", dst_path="idata"
+        )
+        model = MMM.load(idata_path)
+        if not keep_idata:
+            os.remove(idata_path)
+            os.rmdir("idata")
+
+    return model
```
wd60622, line 767: Nit:

```suggestion
    if not full_model:
        model = mlflow.pyfunc.load_model(model_uri)
        return model

    idata_path = mlflow.artifacts.download_artifacts(
        run_id=run_id, artifact_path="idata.nc", dst_path="idata"
    )
    model = MMM.load(idata_path)
    if not keep_idata:
        os.remove(idata_path)
        os.rmdir("idata")

    return model
```

pymc_marketing/mlflow.py
```diff
@@ -577,8 +941,35 @@ def new_fit(*args, **kwargs):
...
             log_inference_data(idata, save_file="idata.nc")
 
+            if log_loocv:
+                log_loocv_metrics(idata=idata)
+
+            posterior_preds = self.sample_posterior_predictive(self.X)
+            log_summary_metrics(
+                y_true=self.y,
+                y_pred=posterior_preds[self.output_var],
+            )
```
wd60622, line 966: Don't know how I feel about the `fit` method also calling the `sample_posterior_predictive`. Seems a bit surprising to me. Does this restrict calling the `sample_posteior_predictive` a second time? Maybe patch the `sample_posterior_predictive` method with this logic. That puts a wrench in the `log_model` though. (Typing out load)

Thoughts here? CC @juanitorduz?

wd60622:
I am having some issue with the `load_model` with `keep_idata` set to `False`.

```python
run_id = ...
loaded_mmm = load_model(run_id, full_model=True, keep_idata=False)

parameter_in_model = \"saturation_beta\"
param = az.extract(loaded_mmm.idata.posterior, var_names=[parameter_in_model])
```

wd60622:
Can you add an example of loading the model here: https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_budget_allocation_example.html#loading-a-pre-trained-model

pymc_marketing/mlflow.py
juanitorduz: I agree with your suggestion Will :)

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,320 @@ def log_inference_data(
...
+    model = load_model(run_id="your_run_id", full_model=True, keep_idata=True)
+    """
+    model_uri = f"runs:/{run_id}/{artifact_path}"
+
+    if not full_model:
+        model = mlflow.pyfunc.load_model(model_uri)
+
+    else:
+        idata_path = mlflow.artifacts.download_artifacts(
+            run_id=run_id, artifact_path="idata.nc", dst_path="idata"
```
wd60622, line 760: Comment in the thread about this. It might be good to expose the `dst_path` in order to load multiple registered models at the same time

pymc_marketing/mlflow.py
```diff
@@ -446,6 +803,9 @@ def autolog(
         Additional keyword arguments to pass to `az.summary`.
     log_mmm : bool, optional
         Whether to log PyMC-Marketing MMM models. Default is True.
+    log_loocv : bool, optional
+        Whether to log LOOCV metrics for model. Default is False.
+        Requires the user to have already called `pm.compute_log_likelihood()` on the model.
```
wd60622, line 808: Think this logic is a bit tangled. We need to call the compute_log_likelihood before calling `pm.sample` / `MMM.fit` which cannot be done

It might be best to patch the `pm.compute_log_likelihood` directly then it can be used in arbitrary PyMC models as well 😄

pymc_marketing/mlflow.py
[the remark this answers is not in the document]
wd60622: The inference library will not always be pymc. For instance it can be nutpie and the `inference_library_version` would be the version of nutpie instead of pymc

pymc_marketing/mlflow.py
```diff
@@ -531,29 +891,31 @@ def autolog(
     def patch_sample(sample):
         @wraps(sample)
         def new_sample(*args, **kwargs):
-            idata = sample(*args, **kwargs)
-            mlflow.log_param("pymc_marketing_version", __version__)
-            mlflow.log_param("pymc_version", pm.__version__)
-            mlflow.log_param("nuts_sampler", kwargs.get("nuts_sampler", "pymc"))
+            model = pm.modelcontext(kwargs.get("model"))
+            with model:
+                idata = sample(*args, **kwargs)
```
wd60622, line 896: whats happening with these new indents? Are they needed?

pymc_marketing/mlflow.py
```diff
@@ -577,8 +941,35 @@ def new_fit(*args, **kwargs):
...
+            posterior_preds = self.sample_posterior_predictive(self.X)
+            log_summary_metrics(
+                y_true=self.y,
+                y_pred=posterior_preds[self.output_var],
+            )
+
+            log_model(
+                self,
+                registered_model_name=registered_model_name,
+            )
```
wd60622, line 971: I conceptually like the idea of this being included in the autologging but the PyMC models are a bit tough. I lean a bit more to keeping this separate because someone's model flow might be: 

```python
with mlflow.start_run() as run: 
    mmm.fit(X, y)
    mmm.sample_posterior_predictive(X)
    mmm.sample_posterior_predictive(X_new, predictions=True)
    load_model(mmm, registered_model_name="my-registered-model")
```

So having this on `fit` is a bit restrictive. But I think the signature is easy enough to throw in where you want to!

tests/mmm/test_evaluation.py
```diff
@@ -0,0 +1,222 @@
...
+import pytest
+
+from pymc_marketing.mmm.evaluation import (
+    calculate_metric_distributions,
+    compute_summary_metrics,
+    summarize_metric_distributions,
+)
+
+
+@pytest.fixture(scope="module")
```
wd60622, line 25: Might be good to have scope = "function" instead

louismagowan:
Thanks again for your feedback and help guys! Really great learning experience - much appreciated! 🫶 

I'm travelling for the next few days but I'll get around to addressing this ASAP ☺️

pymc_marketing/mlflow.py
louismagowan: Ah yep, good point - I've added it as an arg 👍

pymc_marketing/mlflow.py
louismagowan: I've pushed a change for this - but I'm not fully happy with it. I didn't think about `mmm.fit()` calls `pm.sample()` internally, so you run the risk of creating a log_likelihood_group twice when you do mmm autologging.

My solution just checks that there isn't already a log-likelihood group present, but this seems a bit of a hacky fix

pymc_marketing/mlflow.py
louismagowan: Ah yeah, I see what you mean - I've removed it from the fit patch ☺️

louismagowan: All good, just removed it ☺️

pymc_marketing/mlflow.py
louismagowan: Ok yep, makes sense - I just separated it into its own patched function, hopefully that's better now 😁

pymc_marketing/mlflow.py
louismagowan: Feel free to lemme know if you can think of a better way on this @wd60622! Otherwise, that should hopefully be the comments addressed now ☺️

wd60622:
> Any idea why those tests are failing @wd60622 ? They aren't in the scope of my PR, I don't think 🤔 

The tests are with fresh installs of libraries. Pydantic's latest version has changed error messages. 

I pushed a changed to Carlo's posterior pr. Feel free to ignore. Out of scope

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,344 @@ def log_inference_data(
...
+        The MMM to be logged.
+    artifact_path : str, optional
+        The path to the artifact to be logged. Defaults to "mmm_model".
+    conda_env : dict, optional
+        A dictionary representation of a Conda environment. Defaults to the default conda environment.
+    registered_model_name : str, optional
+        The name of the registered model to be logged. Defaults to None.
+        If specified, the model will be registered under this name, otherwise it will not be registered.
+    wrapper_kwargs : dict, optional
+        Additional keyword arguments to pass to the MMMWrapper, controlling inference behaviour of logged model.
```
wd60622, line 640: Just be explicit here. It'll be easier to follow in the documentation

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,344 @@ def log_inference_data(
...
+    Returns
+    -------
+    model : mlflow.pyfunc.PyFuncModel | MMM
+        The loaded MLflow PyFuncModel or MMM model.
+
+
+    Examples
+    --------
+    # Load model using run_id
+    model = load_model(run_id="your_run_id", full_model=True, keep_idata=True)
```
wd60622, line 755: Wrap this in the code-block like in the other examples. Required to render in the documentation

pymc_marketing/mlflow.py
wd60622: Looks fine to me. Thanks for the changes

pymc_marketing/mlflow.py
```diff
@@ -548,12 +950,15 @@ def new_sample(*args, **kwargs):
...
                 log_model_derived_info(model)
 
-            if log_datasets:
-                log_data(model=model, idata=idata)
+            if log_metadata_info:
+                log_metadata(model=model, idata=idata)
+
+            # mmm.fit() calls pm.sample() internally, so we need to make sure log-likelihood is only added once
+            if log_loocv and "log_likelihood" not in idata.groups():
+                pm.compute_log_likelihood(idata=idata, model=model)
```
wd60622, line 961: Think this is fine. I think we should ideally have it separate though and expose an MMM method. Let's hold off on that in this PR

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,344 @@ def log_inference_data(
...
+    mmm: MMM,
+    artifact_path: str = "model",
+    registered_model_name: str | None = None,
+    **wrapper_kwargs,
+) -> None:
+    """Log a PyMC-Marketing MMM as an MLflow artifact for the current run.
+
+    Parameters
+    ----------
+    model : MMMWrapper
```
wd60622, line 630: This doesn't match with the signature

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,344 @@ def log_inference_data(
...
+                var_names=self.var_names,
+                **self.sample_kwargs,  # type: ignore[arg-type]
+            )
+        else:
+            raise ValueError(
+                f"The prediction method '{predict_method}' is not supported."
+            )
+
+
+def log_model(
```
wd60622, line 620: Think `log_mmm` might be better name as we have items for general PyMC models and this wouldn't work for our CLV models

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,344 @@ def log_inference_data(
...
+
+            # Additional specific logging
+            fig = mmm.plot_components_contributions()
+            mlflow.log_figure(fig, "components.png")
+
+            model_info = log_model(mmm,
+                    registered_model_name="my_amazing_mmm",
+                    include_last_observations=True,
+                    original_scale=False,
+                    )
```
wd60622, line 696:
```suggestion
            model_info = log_model(
                mmm=mmm,
                registered_model_name="my_amazing_mmm",
                include_last_observations=True,
                original_scale=False,
            )
```

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,344 @@ def log_inference_data(
...
+        python_model=mlflow_mmm,
+    )
+    run_id = mlflow.active_run().info.run_id
+    model_uri = f"runs:/{run_id}/{artifact_path}"
+
+    if registered_model_name:
+        mlflow.register_model(model_uri, registered_model_name)
+
+
+def load_model(
```
wd60622, line 720: `load_mmm` here as well

wd60622 (changes requested):
Looking very close to the finish line. Thanks for making the changes!

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,359 @@ def log_inference_data(
...
+        python_model=mlflow_mmm,
+    )
+    run_id = mlflow.active_run().info.run_id
+    model_uri = f"runs:/{run_id}/{artifact_path}"
+
+    if registered_model_name:
+        mlflow.register_model(model_uri, registered_model_name)
+
+
+def load_mmmm(
```
wd60622, line 733: Too many Ms

pymc_marketing/mlflow.py
```diff
@@ -528,12 +928,29 @@ def autolog(
     """
     arviz_summary_kwargs = arviz_summary_kwargs or {}
 
+    def patch_compute_log_likelihood(compute_log_likelihood):
+        """Patch the compute_log_likelihood function to log LOOCV metrics.
+
+        We don't need to call this directly again in MMM.fit patch, since that function calls
+        pm.sample() internally anyway.
```
wd60622, line 935: sample never calls compute_log_likelihood to my understand. CC @juanitorduz 
Nor would we want it to

Do we need to adjust any logic to fit accordingly?

pymc_marketing/mlflow.py
```diff
@@ -548,12 +965,15 @@ def new_sample(*args, **kwargs):
...
                 log_model_derived_info(model)
 
-            if log_datasets:
-                log_data(model=model, idata=idata)
+            if log_metadata_info:
+                log_metadata(model=model, idata=idata)
+
+            # mmm.fit() calls pm.sample() internally, so we need to make sure log-likelihood is only added once
+            if log_loocv and "log_likelihood" not in idata.groups():
+                pm.compute_log_likelihood(idata=idata, model=model)
```
wd60622, line 976: Dont automatically call with sample. It can be high computation

pymc_marketing/mlflow.py
```diff
@@ -563,12 +983,14 @@ def new_sample(*args, **kwargs):
...
         @wraps(fit)
-        def new_fit(*args, **kwargs):
-            idata = fit(*args, **kwargs)
+        def new_fit(self, *args, **kwargs):
+            idata = fit(self, *args, **kwargs)
 
             mlflow.log_params(
                 idata.attrs,
             )
+            mlflow.log_param("nuts_sampler", kwargs.get("nuts_sampler", "pymc"))
```
wd60622, line 992: Why did this move. This is helpful for *all* pymc models

pymc_marketing/mlflow.py
```diff
@@ -528,12 +928,29 @@ def autolog(
...
+
+    pm.compute_log_likelihood = patch_compute_log_likelihood(pm.compute_log_likelihood)
+
     def patch_sample(sample):
         @wraps(sample)
         def new_sample(*args, **kwargs):
+            model = pm.modelcontext(kwargs.get("model"))
             idata = sample(*args, **kwargs)
-            mlflow.log_param("pymc_marketing_version", __version__)
-            mlflow.log_param("pymc_version", pm.__version__)
```
wd60622, line 536: Can we keep for *all* pymc models

pymc_marketing/mlflow.py
```diff
@@ -577,11 +999,41 @@ def new_fit(*args, **kwargs):
                 "saturation_name",
                 json.loads(idata.attrs["saturation"])["lookup_name"],
             )
+
+            # Align with the default values in pymc.sample
+            tune = kwargs.get("tune", 1000)
```
wd60622, line 1004: Shoud be logged because of patched sample

pymc_marketing/mlflow.py
```diff
@@ -548,12 +965,15 @@ def new_sample(*args, **kwargs):
...
-            model = pm.modelcontext(kwargs.get("model"))
             if log_model_info:
                 log_model_derived_info(model)
 
-            if log_datasets:
-                log_data(model=model, idata=idata)
+            if log_metadata_info:
+                log_metadata(model=model, idata=idata)
+
+            # mmm.fit() calls pm.sample() internally, so we need to make sure log-likelihood is only added once
```
wd60622, line 974: Wrong to my understanding

pymc_marketing/mlflow.py
```diff
@@ -412,14 +463,359 @@ def log_inference_data(
     os.remove(save_file)
 
 
+def log_summary_metrics(
```
wd60622, line 466: Can we rename to include mmm and something with evaluation

pymc_marketing/mlflow.py
```diff
@@ -391,6 +394,54 @@ def log_sample_diagnostics(
...
+
+    Parameters
+    ----------
+    idata : az.InferenceData
+        The InferenceData object returned by the sampling method.
+
+    Sets
+    ----
+    idata.log_likelihood : az.InferenceData
+        The InferenceData object with the log_likelihood group, unless it already exists.
```
wd60622, line 421: Outdated

pymc_marketing/mmm/evaluation.py
```diff
@@ -0,0 +1,228 @@
...
+            * `max`: Maximum value of the metric distribution.
+            * `hdi_lower`: Lower bound of the Highest Density Interval.
+            * `hdi_upper`: Upper bound of the Highest Density Interval.
+
+    Examples
+    --------
+    Evaluation (error and model metrics) for a PyMC-Marketing MMM.
+
+    .. code-block:: python
+        import pandas as pd
```
wd60622, line 178: Think a space is needed here

wd60622 (changes requested):
Would like to keep support for all pymc models when not only for mmm

cetagostini:
Hey @louismagowan, great PR. I really like your work, but I have something to ask haha, could we have a notebook that is something like "Evaluating model" where we show all the new metrics and how we can import and use them. It would be great if each addition had its own notebook!

pymc_marketing/mlflow.py
cetagostini, line 1: I see the repeated patterns occur in logging functions, Could we encapsulate and re-use?

```python
def log_and_remove_artifact(path: str | Path):
    mlflow.log_artifact(str(path))
    os.remove(path)
```

Then we replace on places like `log_arviz_summary`, `log_model_graph`, and `log_inference_data`.

pymc_marketing/mlflow.py
```diff
@@ -528,12 +928,29 @@ def autolog(
     """
     arviz_summary_kwargs = arviz_summary_kwargs or {}
 
+    def patch_compute_log_likelihood(compute_log_likelihood):
```
cetagostini, line 931: Type hints.

pymc_marketing/mlflow.py
```diff
@@ -577,11 +999,41 @@ def new_fit(*args, **kwargs):
...
+                    **arviz_summary_kwargs,
+                )
+
             log_inference_data(idata, save_file="idata.nc")
 
             return idata
 
         return new_fit
 
+    def patch_mmm_sample_posterior_predictive(sample_posterior_predictive):
```
cetagostini, line 1021: Type hints

louismagowan:
Thanks again guys! Will address comments this weekend 😁

louismagowan:
Apologies - I had some work stuff come up, been pretty slammed recently (before my holidays), but will get to this ASAP. I'm travelling this weekend but should have some long flights at least to work on this

louismagowan:
Ok @wd60622 that's fine with me too! Thanks 😁

pymc_marketing/mlflow.py
louismagowan: Ok I've updated my comments accordingly. I think with the `if log_loocv and "log_likelihood" not in idata.groups():` line it should be fine, but I'm not totally clear on what that's necessary tbh. Where else is log_likelihood being created? 🤔

pymc_marketing/mlflow.py
louismagowan: But isn't this taken care of by `log_loocv=False` as default for the arg? Like the computation only happens if the user specifies they want loocv metrics?

pymc_marketing/mlflow.py
louismagowan: Ah yep my mistake - I was getting confused with how the scope of the pm.sample works after being patched and then used in mmm.fit().

Should be fixed now I hope!

pymc_marketing/mlflow.py
louismagowan: Just added, thanks! 😁

pymc_marketing/mlflow.py
wd60622: It is only created when pm.compute_log_likelihood is called and not pm.sample. Ie idata wont have it unless someone calls it

pymc_marketing/mlflow.py
wd60622: In my mind, the automatic logging should be triggered with pm.compute_log_likelihood call or equivalent method wrapper

juanitorduz:
pre-commit.ci autofix

pymc_marketing/mlflow.py
louismagowan: Hmm, isn't this a bit confusing though? Like the creation of LOOCV metrics is handled by `az.loo` - I would find it surprising that `pm.compute_log_likelihood` does this automatically, since usually people would have to do this manually.

I suppose I could wrap `az.loo` instead, but I'm not sure that's simpler since people would still need to know to do `pm.compute_log_likelihood` first. 

What do you think about just updating the doc-string to warn people that if `log_loocv=True` then  `pm.compute_log_likelihood` is called, which can be high computation? @wd60622

wd60622:
Hi @louismagowan 
Can you resolve the git conflict

pymc_marketing/mlflow.py
wd60622: I think wrapping az.loo makes more sense  then. Since we don't have that functionality in ModelBuilder yet, let's hold off on this implementation so this PR doesn't become more than it already is

wd60622:
Hey @louismagowan, if you need any support, let us know. 

BTW, I will take a look at #1259 soon and will be working with the same files. Not the end the world, but there might be more conflicts for one of us to resolve.

pymc_marketing/mlflow.py
louismagowan: Ok I'll create an issue and move the existing code over there ☺️

louismagowan: Issue is [here](https://github.com/pymc-labs/pymc-marketing/issues/1347) - I'll delete the code now.

docs/source/notebooks/mmm/mmm_budget_allocation_example.ipynb
```diff
@@ -103,6 +103,25 @@
     "mmm = MMM.load(\"model.nc\")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Alternatively, load a model that has been autologged to MLflow via `pymc_marketing.mlflow.autolog(log_mmm=True)`, from the [PyMC-Marketing MLflow module](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mlflow.py)."
```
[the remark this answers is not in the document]
louismagowan: Hopefully fixed this now - the important part is that the user `full_model=True`, so they get the actual idata and not the registry-compatible version (based on my wrapper).

louismagowan:
Oops, closing this was an accident - I'll undo

louismagowan:
I pushed new commits which are showing in [my branch](https://github.com/louismagowan/pymc-marketing/tree/model_evaluation2), but aren't showing here. Anyone know why?

wd60622:
I am not too sure. Let me look into it

louismagowan:
Ah yeah, there's now quite a few merge conflicts to resolve - but I'm on it ⏲️ 😁

Issue #1022, opened 2024-09-11
Expand causal/counterfactual functionality and add the do-operator


drbenvincent:
This is perhaps a meta-proposal which may be fully addressed by multiple separate PR's. But the core idea is to build out the functionality (and visibility of that) on the causal inference side. So that will involve code and documentation changes.

And right now I am focussing on MMM's. CLV components could be dealt with later.

## Add the `do` operator

We should add a method to allow a user to intervene upon the DAG. I believe it is currently possible for a user to use `set_data`. Though I believe the user does not do this directly - see the [example of evaluating out of sample predictions](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_example.html#out-of-sample-predictions).

The addition of the do operator would enable greater flexibility in the interventions that a user can apply to the causal DAG. Currently, doing this with `set_data` alone restricts users to updating the values of data nodes. But the do operator would additionally allow interventions on latent / random variables.

## Showcase counterfactual inference with MMM's

We should also showcase the causal inference capabilities of `pymc-marketing`. Though these capabilities require a relatively high degree of familiarity and custom coding. Below I outline an example use case. This could be used as the basis of new docs. 

**GOAL:** We have conducted a media campaign where we changed media spend in some channel (or region, or relating to a product) and we want to evaluate the causal impact of that intervention upon our outcome variable.

The algorithmic approach would be:

1. Collect our dataset of predictors and our outcome variable. We will record the date at which an intervention (change in media spend) was made. The dataset will ideally cover dates before the intervention period.
2. Build an MMM and fit the model to the data.
3. Calculate the posterior predictions of the model.
4. With this fitted model, we want to evaluate what the model believed would have happened (to the outcome variable) in a counterfactual situation where the change in media spend did not happen. This will involve specifying (or estimating - perhaps through synthetic control methods) what the media spend in that channel would have been in the absence of the intervention. We will do this by intervening on the model graph by (in this case) intervening upon the spend data. We then run posterior predictive sampling in order to obtain what we believe the outcome would have been in the counterfactual scenario.
5. We calculate the causal impact of the intervention by looking at the difference between the posterior predictions (or actual empirical outcome observations). We should only start to see deviation from zero causal impact (if at all) from the point of the intervention.
6. Using the cumulative sum operation, we can also evaluate the cumulative causal impact of the intervention.

[...]

PR #1032, opened 2024-09-13
New notebook: causal inference with MMM's

drbenvincent:
## Description

This PR adds a new docs page which demonstrates ability to do causal inference with MMM's. In particular we showcase an example where we run a campaign with elevated media spend for a period of time. Our goal is to estimate the causal impact of the campaign. This is illustrated with simulated data in order to run a parameter recovery. So we know the true causal impact of the campaign and can therefore evaluate if the estimated causal impact is close to the true causal impact.

Overall, this notebooks showcases the ability to conduct causal inference in MMM's in `pymc-marketing`.

## Related Issue
- [ ] Partially addresses #1022

## Checklist

## Modules affected
- [X] MMM
- [ ] CLV

## Type of change
- [ ] Bug fix
- [X] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1032.org.readthedocs.build/en/1032/


juanitorduz:
@drbenvincent What if we do a train test split (before and after intervention) and generate predictions and counterfactuals as https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_time_slice_cross_validation.html ? This is how we use it in practice. WDYT?

drbenvincent:
I will think about it. There are at least 2 scenarios:
- Using prior to an experiment to help in planning an experiment. For example, guiding you on level of spend increase for different minimum detectable effects. 
- Using after the experiment - to analyse the causal impact of a campaign. In this case I don't think there's a data leakage issue. Or are you thinking about a case where there's an unobserved confounding variable which changed at the same time as the intervention?

I'm also planning on taking about (but maybe not including a worked example) what you'd do in real work situations when you don't have counterfactual spends. That could be approached by using CausalPy & synthetic control to predict what the media spend would have been in the absence of the campaign. Though it might be the case that a company knows what they were going to spend in the absence of a campaign.

juanitorduz:
>  I'm not sure I understand the issue with data leakage in this particular scenario. But it's been a long week.

For example, if you train in the whole data set, then you are already using information on the post-intervention period to infer parameters to generate counterfactuals.

Now that I think about it, according to https://smithamilli.com/blog/causal-ladder/ (and similar references). Your approach is indeed at the last level: counterfactuals (and fitting on the whole data is ok I think). Whereas how I would use it (for scenario planing) is on level 2: interventions. This can actually be a nice discussion to have in the notebook (if we want to)

drbenvincent:
I understand the point now - you could argue that training on the full data may overestimate the trend. But I think that any increase in the outcome should be accounted for by the media + transformations. 

Good point about counterfactual versus interventions. I think it is worth taking about that in the notebook.

drbenvincent:
Just to say... this notebook is going to get substantial modification / TLC. But the input already is great - looks like this topic is interesting to people 🙂

drbenvincent:
Very excited about plans I have for this post. Though changes will have to wait until Monday. Watch this space...

wd60622 (commented):
Very cool. 

Do you think we can do away with the large functions and duplication in the notebook? It seems the API should expose: 

- parameter freezing via `pm.do` instead redefining the `forward_pass`
  - Someone can initialize model without data (that defines the structure) and call something like: `params = {"intercept: 1, ...}; y = mmm.generate_target(X, params=params, target_scaler=...)`. This opens door to `X_counterfactual = some_modification(X); y_counterfactual = mmm.generate_target(X_counterfactual, params=params, ...)`
- The counterfactual comparision plot. We just need two targets, right? which we can store in xarray to have the date dim
  - `plot_counterfactual(y, y_counterfactual, along="date", intervention_date="...")`

WDYT?

drbenvincent:
Yep - so this is the first very crude implementation. I'm trying to move away from generating synthetic data from separate numpy code and doing it with the DAG, sampling for the prior. I didn't know about `mmm.generate_target` so this is very useful. I'll see how far I get with that approach and maybe ask for some guidance. I know how to do it with base pymc, but I'm still unfamiliar with much of the pymc-marketing functionality/methods.

And I'll move towards getting counterfactual samples into the `idata.predictions` field. Is that what you're suggesting?

This notebook is going to change quite a lot before I flag as ready for review 👍🏻

wd60622:
> Yep - so this is the first very crude implementation. I'm trying to move away from generating synthetic data from separate numpy code and doing it with the DAG, sampling for the prior. I didn't know about `mmm.generate_target` so this is very useful. I'll see how far I get with that approach and maybe ask for some guidance. I know how to do it with base pymc, but I'm still unfamiliar with much of the pymc-marketing functionality/methods.
> 
> And I'll move towards getting counterfactual samples into the `idata.predictions` field. Is that what you're suggesting?
> 
> This notebook is going to change quite a lot before I flag as ready for review 👍🏻

That doesn't exist but just some ideas are API. I think we should make whatever is useful to us

drbenvincent:
<p>Done in an upcoming commit</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1032/discussion/#comment-2350727880'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Have slightly modified this text. Let's see how it all reads once the larger scale structure of the notebook has settled down a bit.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1032/discussion/#comment-2350727876'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Yep - so this initial commit was very preliminary - I've expanded on the text and narrative of the notebook much more now</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1032/discussion/#comment-2350727906'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Good point. I've removed these and we now use default priors</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1032/discussion/#comment-2350727910'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
Sorry, took me a moment to understand that this was a proposal @wd60622. 

> * parameter freezing via `pm.do` instead redefining the `forward_pass`
>   
>   * Someone can initialize model without data (that defines the structure) and call something like: `params = {"intercept: 1, ...}; y = mmm.generate_target(X, params=params, target_scaler=...)`. This opens door to `X_counterfactual = some_modification(X); y_counterfactual = mmm.generate_target(X_counterfactual, params=params, ...)`

Yep - I will have a go at defending these functions (quickly and crudely) in the notebook. Then they can be built out either in this PR or a different one.

> * The counterfactual comparision plot. We just need two targets, right? which we can store in xarray to have the date dim
>   
>   * `plot_counterfactual(y, y_counterfactual, along="date", intervention_date="...")`

Yes, but I think we would need a way to set the `predictions=True` flag in the kwargs of `pm.sample_posterior_predictive`. Is that possible now, or would that be new functionality.

I have to shift focus for a day or so, but I'll see what I can do code-wise in this notebook to inspire some new functionality.

cetagostini:
<p>🚀</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1032/discussion/#comment-2350727876'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
<p>🚀</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1032/discussion/#comment-2350727906'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

wd60622:
Feel free to make some issues if you don't touch in this PR

> Yes, but I think we would need a way to set the `predictions=True` flag in the kwargs of `pm.sample_posterior_predictive`. Is that possible now, or would that be new functionality.
> 
> I have to shift focus for a day or so, but I'll see what I can do code-wise in this notebook to inspire some new functionality.

The `sample_posterior_predictive` method allows for passing sample_posterior_predictive kwargs. 
My idea was to just pass the xarray.DataArray objects for a function

drbenvincent:
Boom! Ready for review.

For iteration 1 I think it's best to leave the data simulation as it is now. But I'm more than happy to come back at a later time to generate data from an empty MMM model by sampling from the prior.

Any comments on the causal DAG? I'm not sure if I should include the time node, or whether time should also causally affect $x1$ and $x2$. The direct effect of time on sales is modelled by the trend and seasonality components - I'm thinking that can be said in text rather than be specific nodes. But happy to get thoughts on this.
Screenshot 2024-10-10 at 11 00 13

drbenvincent:
<p>I wasn't going to add the intervention lines to the legend - it might get a bit cluttered. And I was hoping to rely on the text to make it clear that those represent the intervention period.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1032/discussion/#comment-2408474157'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
Thanks for the review @juanitorduz. I think I've addressed most of your comments. I was going to avoid legends for the campaign interval - but if that was important I could either add them, or have it as a shaded region perhaps.

Happy to get more feedback based on this iteration.

juanitorduz:
<p>ok! I just thought because the <code>label="intervention date"</code> which id not doing anything then </p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1032/discussion/#comment-2408474157'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz (approved):
Let's push this one out!

cetagostini:
Hey @drbenvincent 

I'm very agree here:

>The direct effect of time on sales is modelled by the trend and seasonality components. **I'm thinking that can be said in text** 

Maybe it's a bit picky but putting time in a DAG is something I've never seen before, I've been looking a bit on the internet and I think one of the main reasons is that, time is often implicit in the temporal ordering of variables. The causal relationships are understood to occur over time without explicitly including “time” as a variable.

I think it is better to say exogenous variables and that's it, but maybe I'm incorrect 😅

drbenvincent:
Thanks @cetagostini. What do you think @juanitorduz  about time in the DAG? I agree that you don't see it very much, but we should have _something_. I'll have a think too and edit it in a follow on PR.

juanitorduz:
I do not have a preference. In the example notebook we use "seasonality". I think time can combine both trend and seasonality.

PR #1053, opened 2024-09-19
comment save notebook in mmm example

juanitorduz:
This saving cell is problematic in the CI. We can done this in the past but the change was undone XD

----
📚 Documentation preview 📚: https://pymc-marketing--1053.org.readthedocs.build/en/1053/


wd60622:
Maybe there is a magic cell or tag that can be used. I will check the papermill docs

wd60622 (approved):
Comments arent blocker. Just some thoughts

PR #1054, opened 2024-09-20
`michaelis_menten` transformation as `pt.TensorVariable`

### LINKED ISSUE #1050: `MMM.plot_direct_contribution_curves` errors when using `MichaelisMentenSaturation`
Model initialized as `model = MMM( ..., saturation=MichaelisMentenSaturation(), ...)`

Then, after calling `model.fit(X=X, y=y, ...)`, `model.sample_posterior_predictive(X)`, we try to plot the direct contributions by channel with  `model.plot_direct_contribution_curves(show_fit=True, ...)`:

`AttributeError: 'numpy.ndarray' object has no attribute 'eval'` 

The issue arises from the call to `_plot_response_curve_fit` trying to evaluate a `pytensor.tensor.TensorVariable` in [L1671](https://github.com/pymc-labs/pymc-marketing/blob/bc4b248b8da508c1e5aa03bbcfdf487de0aaeabb/pymc_marketing/mmm/mmm.py#L1671), [L1675](https://github.com/pymc-labs/pymc-marketing/blob/bc4b248b8da508c1e5aa03bbcfdf487de0aaeabb/pymc_marketing/mmm/mmm.py#L1675), and [L1678](https://github.com/pymc-labs/pymc-marketing/blob/bc4b248b8da508c1e5aa03bbcfdf487de0aaeabb/pymc_marketing/mmm/mmm.py#L1678).

Note that `MichaelisMentenSaturation.function` currently returns `float | Any`, contrary to other transformations

1. Hacky fix: comment `.eval()`. 
2. Better fix: modify `michaelis_menten` to return `pytensor.tensor.TensorVariable`

I'll try to open a PR with solution (2) soon

PabloRoque:
Make `michaelis_meten` consistent with the rest of `SaturationTransformation` by wrapping it as `pt.TensorVariable` 

## Description

* ~~Rename old `michaelis_menten` as `michaelis_menten_function`. Made to avoid major changes in `mmm.utils.estimate_menten_parameters`'s  [L70](https://github.com/pymc-labs/pymc-marketing/blob/f70db336ebf8766e7792279ca4148261f6b18edf/pymc_marketing/mmm/utils.py#L70).~~
* ~~Wrap previous `michaelis_menten_function` as `pt.TensorVariable` inside `michaelis_menten`~~
* ~~Minor change in  `test_michaelis_menten`. The TensorVariable now needs to be evaluated.~~
* ~~Minor changes to type hints~~
* Wrap `MichaelisMentenSaturation.function`with `pt.as_tensor_variable`
* Expand `test_plotting.mock_mmm` fixture with `MichaelisMentenSaturation`
* `.eval()` in example notebook

## Related Issue

- [x] Closes #1050 
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1054.org.readthedocs.build/en/1054/


pymc_marketing/mmm/transformers.py
```diff
@@ -914,6 +914,31 @@ def michaelis_menten(
     return alpha * x / (lam + x)
 
 
+def michaelis_menten(
```
wd60622, line 917: Do we need two functions?

wd60622: Can all of the `SaturationTransformations` be wrapped in as_tensor_variable instead of making these changes?

PabloRoque: > Do we need two functions?

We don't need them, but made my life easier to avoid the pesky [L70](https://github.com/pymc-labs/pymc-marketing/blob/f70db336ebf8766e7792279ca4148261f6b18edf/pymc_marketing/mmm/utils.py#L70)

> Can all of the `SaturationTransformations` be wrapped in as_tensor_variable instead of making these changes?
Do you have anything in mind? I can think of 2 ways:
* Adding boilerplate to all the classes. Perhaps defining a decorator to reduce boilerplate. 
* Rely on `Transformation._checks()` but here we would need to trust the type hints to use something like `signature(class_function).return_annotation` and wrap `pt.as_tensor_variable` if needed.

Perhaps you have something less convoluted in mind?

wd60622: Can you not add it in the one location of the `apply` method. That is a common method that is not overwritten by the subclasses

pymc_marketing/mmm/transformers.py
```diff
@@ -826,10 +826,10 @@ def tanh_saturation_baselined(
     return gain * x0 * pt.tanh(x * pt.arctanh(r) / x0) / r
 
 
-def michaelis_menten(
+def michaelis_menten_function(
     x: float | np.ndarray | npt.NDArray[np.float64],
-    alpha: float | np.ndarray | npt.NDArray[np.float64],
-    lam: float | np.ndarray | npt.NDArray[np.float64],
+    alpha: float,
+    lam: float,
```
wd60622, line 832: I still don't know the best type hints here. But TensorVariable is also accepted here

pymc_marketing/mmm/utils.py
```diff
@@ -22,7 +22,7 @@
 import xarray as xr
 from scipy.optimize import curve_fit, minimize_scalar
 
-from pymc_marketing.mmm.transformers import michaelis_menten
+from pymc_marketing.mmm.transformers import michaelis_menten_function
 
 
 def estimate_menten_parameters(
```
wd60622, line 28: Where is this still used? I thought there was a generalization of this? @cetagostini

pymc_marketing/mmm/transformers.py
wd60622: I think having it for apply would make sure that other custom saturation transformations will not encounter this found bug as well

wd60622: It shouldn't hurt to has pt.as_tensor_variable wrapping an already TensorVariable. pytensor will figure out an optimization

pymc_marketing/mmm/utils.py
wd60622: Created an issue here: #1055

wd60622:
I am looking over the `_plot_response_curve_fit` internals. I think it would be better to rely on the `SaturationTransformation.sample_curve` method so that the solution doesn't directly call the `function` method. 

There needs to be a solution that works for others custom saturation transformations

I will create an issue for this. We can address in the future

PabloRoque:
> Can you write a test for the case that discovered this bug

Added `saturation` fixture. Now `test_mmm_plots` includes `MichaelisMentenSaturation`. 
Would fail without TensorVariable in  `plot_direct_contribution_curves` fixture, but does not with the fix in this PR (pending suggested changes)

pymc_marketing/mmm/transformers.py
PabloRoque: > Can you not add it in the one location of the `apply` method? (where I linked) That is a common method that is not overwritten by the subclasses

The linked `apply` method needs to be called within a model context. Not the case in the plotting function.

PabloRoque: The proposed `sample_curve` might work though, since it open a model context. Having a look

PabloRoque: Do we want to add another pm.Deterministic to the model? 
It would be the case if we use `_sample_curve`

PabloRoque:
> SaturationTransformation.sample_curve

This will add 3 extra `pm.Deterministic` to the model. 
Are we happy we that behavior?

wd60622:
> This will add 3 extra `pm.Deterministic` to the model. 
> 
> Are we happy we that behavior?

Lets skip if that is required. Dont think that would be good

pymc_marketing/mmm/transformers.py
wd60622: We wouldnt want to use the private methods of the saturation transformation. Does sample_curve with fit_result not get us close to the data that is plotted? (Just not scaled on x and y)

pymc_marketing/mmm/transformers.py
```diff
@@ -826,10 +826,10 @@ def tanh_saturation_baselined(
     return gain * x0 * pt.tanh(x * pt.arctanh(r) / x0) / r
 
 
-def michaelis_menten(
+def michaelis_menten_function(
```
wd60622, line 829: This is only used in the functions we will deprecate, right?

PabloRoque: > This is only used in the functions we will deprecate, right?

Yes, they are only used in the methods linked in #1055. 
If we remove those, there is no need to define the extra function.

wd60622: Mind closing that now?

pymc_marketing/mmm/transformers.py
PabloRoque: > I still don't know the best type hints here. But TensorVariable is also accepted here

Reverted changes to type hints

pymc_marketing/mmm/transformers.py
PabloRoque: Hi @wd60622,

having doubts about completely removing functions in #1055. 

As an example, take `estimate_menten_parameters`. It was introduced in #329, and never used outside of the current test. I am guessing the original purpose was to double check that the introduced function gives correct results. If we remove the tests, we would never be checking that again.

As alternative I've simplified the PR to wrap only `MichaelisMentenSaturation.function` with `pt.as_tensor_variable`.

Note that the newly introduced fixture now passes the tests

wd60622: If we need to remove tests, that is fine. 

This is only saturation transformation which needs to be defined with two functions (which I find silly). Just cosolidate

wd60622:
Thanks for all the edits @PabloRoque 

I am out atm but @juanitorduz will take a look

juanitorduz:
I'll review on the weekend 🙏💪 (busy week ahead 😄)

PabloRoque:
No rush! I'll be taking some time off myself until the 28th.

juanitorduz (approved):
This is looking good! Shall we merge this one and continue with other smaller PRs?

wd60622 (approved):
Looks good. Let's merge and continue elsewhere if needed

PR #1068, opened 2024-09-25
 #990 - Plot FourierBase along date rather than index

### LINKED ISSUE #990: Plot `FourierBase` along date rather than index
The plotting of the fourier is done along the index rather than a date which can be confusing to the meaning of the x-axis. The can be changed by allowing the `sample_curve` function to set the coords to datetimes instead of integers. 

What date would be the start of the fourier curve though? Might want to derive from the current year or current month. 

Overall, this would be optional and be an alternative to the current functionality.

Ishaanjolly:
Enhancement: Plot FourierBase along date rather than index

## Description

I added the following:  

```
def sample_curve(
        self,
        parameters: az.InferenceData | xr.Dataset,
        use_dates: bool = False,
        start_date: datetime.datetime | None = None,
    ) -> xr.DataArray:
        """Create full period of the Fourier seasonality.

        Parameters
        ----------
        parameters : az.InferenceData | xr.Dataset
            Inference data or dataset containing the Fourier parameters.
            Can be posterior or prior.
        use_dates : bool, optional
            If True, use datetime coordinates for the x-axis. Defaults to False.
        start_date : datetime.datetime, optional
            Starting date for the Fourier curve. If not provided and use_dates is True,
            it will be derived from the current year or month. Defaults to None.

        Returns
        -------
        xr.DataArray
            Full period of the Fourier seasonality.

        """
        # Determine the full period
        full_period = np.arange(int(self.days_in_period) + 1)

        coords = {}
        if use_dates:
            if start_date is None:
                # Derive start_date based on the type of Fourier seasonality
                today = datetime.datetime.now()
                if isinstance(self, YearlyFourier):
                    start_date = datetime.datetime(year=today.year, month=1, day=1)
                elif isinstance(self, MonthlyFourier):
                    start_date = datetime.datetime(
                        year=today.year, month=today.month, day=1
                    )
                else:
                    raise ValueError("Unknown Fourier type for deriving start_date")

            # Create a date range
            date_range = pd.date_range(
                start=start_date,
                periods=int(self.days_in_period) + 1,
                freq="D",
            )
            coords["date"] = date_range.to_numpy()
            dayofyear = date_range.dayofyear.to_numpy()

        else:
            coords["day"] = full_period
            dayofyear = full_period

        # Include other coordinates from the parameters
        for key, values in parameters[self.variable_name].coords.items():
            if key in {"chain", "draw", self.prefix}:
                continue
            coords[key] = values.to_numpy()

        with pm.Model(coords=coords):
            name = f"{self.prefix}_trend"
            pm.Deterministic(
                name,
                self.apply(dayofyear=dayofyear),
                dims=tuple(coords.keys()),
            )

            return pm.sample_posterior_predictive(
                parameters,
                var_names=[name],
            ).posterior_predictive[name]

```

and following within each plot_* function as I was not able to define new attributes: 

```
if "date" in curve.coords:
      x_coord_name = "date"
elif "day" in curve.coords:
    x_coord_name = "day"
else:
    raise ValueError("Curve must have either 'day' or 'date' as a coordinate")

```

## Related Issue

- [x] Closes #990 
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1068.org.readthedocs.build/en/1068/


pymc_marketing/mmm/fourier.py
```diff
@@ -422,25 +424,62 @@ def sample_prior(self, coords: dict | None = None, **kwargs) -> xr.Dataset:
...
+        if use_dates:
+            if start_date is None:
+                # Derive start_date based on the type of Fourier seasonality
+                today = datetime.datetime.now()
+                if isinstance(self, YearlyFourier):
+                    start_date = datetime.datetime(year=today.year, month=1, day=1)
+                elif isinstance(self, MonthlyFourier):
+                    start_date = datetime.datetime(
+                        year=today.year, month=today.month, day=1
+                    )
```
wd60622, line 465: Is it possible to not rely on the child classes? This makes an inheritance from the `FourierBase` not possible.

This can usually be solved by a method in each child class. For instance, `get_default_start_date(today: datetime)`

pymc_marketing/mmm/fourier.py
```diff
@@ -422,25 +424,62 @@ def sample_prior(self, coords: dict | None = None, **kwargs) -> xr.Dataset:
         coords[self.prefix] = self.nodes
         return self.prior.sample_prior(coords=coords, name=self.variable_name, **kwargs)
 
-    def sample_curve(self, parameters: az.InferenceData | xr.Dataset) -> xr.DataArray:
-        """Create full period of the fourier seasonality.
+    def sample_curve(
+        self,
+        parameters: az.InferenceData | xr.Dataset,
+        use_dates: bool = False,
+        start_date: datetime.datetime | None = None,
```
wd60622, line 431: How about allowing for string dates as well?

wd60622 (commented):
Thanks for taking this @Ishaanjolly !

Some initial comments and suggestions

Could you write a test for the sample plot workflow using dates

pymc_marketing/mmm/fourier.py
```diff
@@ -333,6 +333,72 @@ def nodes(self) -> list[str]:
...
+                try:
+                    return pd.to_datetime(start_date)
+                except ValueError as e:
+                    raise ValueError(f"Unable to parse start_date: {e}") from e
+            elif isinstance(start_date, datetime.datetime):
+                return start_date
+            else:
+                raise TypeError(
+                    "start_date must be a datetime.datetime object, a string, or None"
+                )
```
wd60622, line 380: nit: turn into elif or remove else entirely

pymc_marketing/mmm/fourier.py
```diff
@@ -422,25 +490,58 @@ def sample_prior(self, coords: dict | None = None, **kwargs) -> xr.Dataset:
...
+                    start_date = datetime.datetime(year=today.year, month=1, day=1)
+                elif isinstance(self, MonthlyFourier):
+                    start_date = datetime.datetime(
+                        year=today.year, month=today.month, day=1
+                    )
+                else:
+                    raise ValueError("Unknown Fourier type for deriving start_date")
+
+            date_range = pd.date_range(
+                start=start_date,
```
wd60622, line 534: pd.date_range should handle the conversions above, does it not?

Ishaanjolly: elaborate?

wd60622: no need to convert to datetime to put into this function as it is pretty flexible itself: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.date_range.html

But getting start date might require that conversion. I need to look at it in its entirety

Ishaanjolly: will have a look today and some tests too! low brain cells yesterday!

Ishaanjolly: Please have a look now @wd60622?

- I have removed all the pd.DateTime conversions where it was not needed as mentioned by you

> pd.date_range should handle the conversions above.


- and I added more tests in `test_fourier.py`

> Could you write a test for the sample plot workflow using dates

pymc_marketing/mmm/fourier.py
```diff
@@ -331,6 +333,63 @@ def nodes(self) -> list[str]:
...
+        """
+        if start_date is None:
+            return self._get_default_start_date(today)
+        else:
+            if isinstance(start_date, str) or isinstance(start_date, datetime.datetime):
+                return start_date
+            else:
+                raise TypeError(
+                    "start_date must be a datetime.datetime object, a string, or None"
+                )
```
wd60622, line 371: nit: This is what I meant. Bit easier to read imo
```suggestion
        elif isinstance(start_date, (str, datetime.datetime)):
            return start_date
        else:
            raise TypeError(
                "start_date must be a datetime.datetime object, a string, or None"
            )
```

pymc_marketing/mmm/fourier.py
wd60622, line 1: Can you remove the "NON_GRID_NAMES" at the top if they are no longer used

pymc_marketing/mmm/fourier.py
```diff
@@ -331,6 +333,63 @@ def nodes(self) -> list[str]:
             f"{func}_{i}" for func in ["sin", "cos"] for i in range(1, self.n_order + 1)
         ]
 
+    def get_default_start_date(
+        self,
+        today: datetime.datetime,
+        start_date: str | datetime.datetime | None = None,
```
wd60622, line 339: The two parameters confuse me a bit. Is there any way of doing it with only one? 

Maybe it is the function and parameter name that is confusing

The test using `fourier.get_default_start_date(today, None)` is what was throwing me off

pymc_marketing/mmm/fourier.py
Ishaanjolly: I had first done this only but the pre-commit hooks was failing on exactly this line for some reason `isinstance(start_date, (str, datetime.datetime))` and suggested the usage of `X|Y` instead of `X,Y`

wd60622: If it does that then use: "# type: ignore"

pymc_marketing/mmm/fourier.py
Ishaanjolly: Latest push has : 

- `today` removed from params 
- `isinstance(start_date, (str, datetime.datetime))` added with comment for ruff ignore

Hopefully, the above two fixes are OK for you.

tests/mmm/test_fourier.py
```diff
@@ -272,3 +279,83 @@ def test_change_name() -> None:
...
+@pytest.fixture
+def yearly_fourier() -> YearlyFourier:
+    prior = Prior("Laplace", mu=0, b=1, dims="fourier")
+    return YearlyFourier(n_order=2, days_in_period=365, prior=prior)
+
+
+@pytest.fixture
+def monthly_fourier() -> MonthlyFourier:
+    prior = Prior("Laplace", mu=0, b=1, dims="fourier")
+    return MonthlyFourier(n_order=2, days_in_period=30, prior=prior)
```
wd60622, line 293: Why is `days_in_periods` being redefined here?

juanitorduz: +1

pymc_marketing/mmm/fourier.py
```diff
@@ -331,6 +334,51 @@ def nodes(self) -> list[str]:
...
+
+        Raises
+        ------
+        TypeError
+            If `start_date` is neither a string nor a datetime object.
+        """
+        if start_date is None:
+            return self._get_default_start_date()
+        else:
+            if isinstance(start_date, (str, datetime.datetime)):  # type: ignore  # noqa: UP038  # Ignored to maintain compatibility with Python <3.10
```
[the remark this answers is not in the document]
Ishaanjolly: sadly, if this not added then ruff checks fail. It suggests to use `X|Y` instead of `X,Y`. Maybe you and @wd60622  can agree on what's the best approach here, as was suggested by Will to use this.

tests/mmm/test_fourier.py
Ishaanjolly: sure, no worries. I noticed you have constants for this, but it should work without it too.

Ishaanjolly: I'll push something tonight!

Ishaanjolly: Only `# noqa: UP038` remains in the latest commit; have removed` <3.10` comment. Hope that works, if anything else let me know please. 

and yeah `days_in_period` from `test_fourier.py`  has been removed

pymc_marketing/mmm/fourier.py
```diff
@@ -331,6 +334,51 @@ def nodes(self) -> list[str]:
...
+
+        Raises
+        ------
+        TypeError
+            If `start_date` is neither a string nor a datetime object.
+        """
+        if start_date is None:
+            return self._get_default_start_date()
+        else:
+            if isinstance(start_date, (str, datetime.datetime)):  # noqa: UP038
```
wd60622, line 364: If | also works with isistance without the noqa, lets go with that

wd60622, line 364: Elif 🤓

Ishaanjolly: have a look now!

pymc_marketing/mmm/fourier.py
```diff
@@ -331,6 +334,50 @@ def nodes(self) -> list[str]:
...
+
+        Raises
+        ------
+        TypeError
+            If `start_date` is neither a string nor a datetime object.
+        """
+        if start_date is None:
+            return self._get_default_start_date()
+        elif isinstance(start_date, str) | isinstance(start_date, datetime.datetime):
+            return start_date
```
wd60622, line 364: Subtlety here is that user can pass date that doesn't start at beginning of month or year and the plot starts.

I think that's fine if someone adjusts the zero point in x / `dayofyear`. Though conventionally that'd be beginning of month or year.

Just thinking out loud

PR #1044, opened 2024-09-17
MMM Case Study from PyData Global

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/1014

----
📚 Documentation preview 📚: https://pymc-marketing--1044.org.readthedocs.build/en/1044/

PR #1070, opened 2024-09-27
remove spend scaling step from budget optimizer

juanitorduz:
When looking into https://github.com/pymc-labs/pymc-marketing/pull/1044 I encountered the following issue: The budget to allocate in the example is of the order of ~ 1e6 and I was getting an uniform split 

```python
pd.Series(mmm.optimal_allocation_dict)

Direct Mail       1.972773e+06
Insert            1.972773e+06
Newspaper         1.972773e+06
Online Display    1.972773e+06
Radio             1.972773e+06
Social Media      1.972773e+06
TV                1.972773e+06
dtype: float64
```

Nevertheless the plot 

```python
fig, ax = mmm.plot_budget_allocation(samples=response, figsize=(12, 8))
fig.suptitle("Response vs spent per channel");
```

showed

image

which was values of the order ~ 1e12. 

I talked with @carlosagostini and notice that with the latest changes we do not need to scale the spends.

----
📚 Documentation preview 📚: https://pymc-marketing--1070.org.readthedocs.build/en/1070/


pymc_marketing/mmm/mmm.py
```diff
@@ -2239,28 +2239,14 @@ def plot_budget_allocation(
...
+        channel_contributions = (
+            samples["channel_contributions"].mean(dim=["date", "sample"]).to_numpy()
+        )
 
-            allocate_spend = (
-                np.array(list(self.optimal_allocation_dict.values()))
-                * self.channel_transformer["scaler"].scale_
-            )
+        if original_scale:
+            channel_contributions *= self.get_target_transformer()["scaler"].scale_
```
juanitorduz, line 2247: @carlosagostini, Is this still valid, or should we remove the scaling everywhere?

juanitorduz:
In https://github.com/pymc-labs/pymc-marketing/pull/1070/commits/bc78e643b442f10650323877ffd6a002641874e1 I re-ran the notebook. The results look as expected right?

juanitorduz:
BTW: I changed the colours so that they can be synced with the user palettes (that is why we are using the "C0" and "C1") syntax

pymc_marketing/mmm/mmm.py
cetagostini: Contributions still need to be scaled, but not the budget allocation, because the optimizer now returns it in original scale. So post processing is not necessary 🙌🏻

juanitorduz: Ok! Thanks! Si is this one good to go ?

cetagostini (approved):
Everything looks good here!

pymc_marketing/mmm/mmm.py
cetagostini: Indeed!

PR #1086, opened 2024-10-16
Automatic labeling of untouched issues

wd60622:
Liked this idea from scikit-learn. Idea is automatically label issues "Needs Triage" if they are not labeled. Once they are touched by someone, the label can be removed.

Any thoughts on this? There are a few more automation ideas I think that would be good.

----
📚 Documentation preview 📚: https://pymc-marketing--1086.org.readthedocs.build/en/1086/


juanitorduz (approved):
Great to learn from other projects! 🚀

Issue #413, opened 2023-10-27
Negative values not allowed for MMM channel variables


Brites101:
A consequence of the usage of HalfNormal distributions to estimate the beta parameter of the channel variable is that it **cannot account for channels with negative impacts**, such as competitor activity.

An expected solution would be to make the channel with competitor impact negative, therefore keeping the beta parameter positive. Currently, this results in an ValueError.

Attempting to fit the MMM with a negative channel returns the following error:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[34], line 4
      1 X = df_train.drop(kpi,axis=1)
      2 y = df_train[kpi]
----> 4 mmm.fit(X=X, y=y, target_accept=0.95, chains=4, random_seed=42)

File /opt/conda/lib/python3.11/site-packages/pymc_marketing/model_builder.py:511, in ModelBuilder.fit(self, X, y, progressbar, predictor_names, random_seed, **kwargs)
    509     y = np.zeros(X.shape[0])
    510 y_df = pd.DataFrame({self.output_var: y})
--> 511 self.generate_and_preprocess_model_data(X, y_df.values.flatten())
    512 if self.X is None or self.y is None:
    513     raise ValueError("X and y must be set before calling build_model!")

File /opt/conda/lib/python3.11/site-packages/pymc_marketing/mmm/delayed_saturated_mmm.py:126, in BaseDelayedSaturatedMMM.generate_and_preprocess_model_data(self, X, y)
    124 self.model_coords = coords
    125 if self.validate_data:
--> 126     self.validate("X", X_data)
    127     self.validate("y", y)
    128 self.preprocessed_data: Dict[str, Union[pd.DataFrame, pd.Series]] = {
    129     "X": self.preprocess("X", X_data),
    130     "y": self.preprocess("y", y),
    131 }

File /opt/conda/lib/python3.11/site-packages/pymc_marketing/mmm/base.py:129, in BaseMMM.validate(self, target, data)
    126     validation_methods = self.validation_methods[1]
    128 for method in validation_methods:
--> 129     method(self, data)

File /opt/conda/lib/python3.11/site-packages/pymc_marketing/mmm/validating.py:63, in ValidateChannelColumns.validate_channel_columns(self, data)
     59     raise ValueError(
     60         f"channel_columns {self.channel_columns} contains duplicates"
     61     )
     62 if (data.filter(list(self.channel_columns)) < 0).any().any():
[...]

PR #1092, opened 2024-10-19
Allow Negative Values in Media Channels

louismagowan:
## Description
Currently the MMM by default informs our channels to be >=0 via the HalfNormal and inclusion of only non-negative values.

It would be nice to have an option to pass in a column of channels that contain negative values, such that the effect of that channel is allowed to be negative on model target - but while not having to blanket-disable the data validation `validate_data=False`

## Related Issue

- [915] Closes #
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -92,6 +92,14 @@ class BaseMMM(BaseValidateMMM):
         ),
         model_config: dict | None = Field(None, description="Model configuration."),
         sampler_config: dict | None = Field(None, description="Sampler configuration."),
+        allow_negative_channels: Annotated[
+            list[str],
+            Field(
+                min_length=1,
+                description="Column names of media channel variables whose effect should be <=0",
+            ),
+        ]
+        | None = None,
         validate_data: bool = Field(
             True, description="Whether to validate the data before fitting to model"
         ),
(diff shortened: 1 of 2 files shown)


louismagowan:
Still need to do tests and docs ⏳

pymc_marketing/mmm/mmm.py
```diff
@@ -92,6 +92,14 @@ def __init__(
         ),
         model_config: dict | None = Field(None, description="Model configuration."),
         sampler_config: dict | None = Field(None, description="Sampler configuration."),
+        allow_negative_channels: Annotated[
```
juanitorduz, line 95: What would be the effect of this new flag on the model? Change the prior? What about simply passing the data multiplied by -1 (for the desired channels) to the original model?

louismagowan: No change really! No change to prior etc- it's just that currently to be able to pass in data that's negative / you've multiplied by -1 you have to set `validate_data=False` in `MMM()` and I'd prefer to be able to retain the validation while still passing in the negative values.

Or perhaps I'm misunderstanding something haha. A very lightweight PR so far so happy to close if I'm misunderstanding something ☺️

juanitorduz: ok! I get the idea :) 

I generally try to avoid flags as this hints at design flaws (from our side)

An alternative would be to remove:

https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/validating.py#L118

Or make it optional in a better way .... Ideally with Pydantic ... 

Let me think about it (we can keep this PR open) 🙏 

@wd60622 any thoughts or ideas?

juanitorduz: Ok! @louismagowan, would you mind removing the ValueError and adding a Warning instead?

louismagowan: @juanitorduz should be good now I think! ☺️

wd60622:
Closes #413

PR #1096, opened 2024-10-21
add `hdi_list` kwarg to `plot_posterior_predictive`

### LINKED ISSUE #1041: Allow user to specify list of HDI's in `plot_posterior_predictive`
At the moment `plot_posterior_predictive` has hard-wired HDI's. Plots look like this (taken from MMM docs page):

Screenshot 2024-09-16 at 16 17 10

This is the relevant line:

https://github.com/pymc-labs/pymc-marketing/blob/aa5fa1d280eae270a7a992dbc82f77586a02f237/pymc_marketing/mmm/base.py#L414

Can we provide a kwarg, called something like `hdi` where the user provides a scalar (or list or scalars). That way, we can decide if we just want one or multiple HDI's and what values they are.

Maybe also expose the `alpha`s, or maybe automatically compute them from the provided `hdi` value(s).

drbenvincent:
## Description
Adds an optional kwarg to the `plot_posterior_predictive` function. Allows user to provide a list of hdi levels, e.g. `[0.94]`. Default is the same as before, `[0.94, 0.5]`. 

## Related Issue
- [ ] Closes #1041

## Checklist
- [X] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)

## Modules affected
- [X] MMM
- [ ] CLV

## Type of change
- [X] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1096.org.readthedocs.build/en/1096/


pymc_marketing/mmm/base.py
```diff
@@ -363,6 +363,7 @@ def plot_posterior_predictive(
         self,
         original_scale: bool = False,
         add_hdi: bool = True,
+        hdi_list: list[float] | None = None,
```
juanitorduz, line 366: Shall we add a warning when we have `add_hdi=False` and we have a non None `hdi_list`? (In this case we ignore the hdi params?)

Can we please add one simple test case in https://github.com/pymc-labs/pymc-marketing/blob/main/tests/mmm/test_plotting.py#L195 where we pass the hdi list?

wd60622: Is there a way to do this with one parameter? That would be ideal

drbenvincent: Just addressed @juanitorduz 's requests, but agree that having both kwargs is excessive. Maybe we can just remove `add_hdi` and if the user doesn't want one then they provide a `None` or `[]` to `hdi_list`?

drbenvincent: Personally I think we definitely want HDI's by default. 

Though I'd make the default `[0.94]` to align with `arviz`. A user could then change to a list or `None` or `[]`?

drbenvincent: Ok. That last commit (https://github.com/pymc-labs/pymc-marketing/pull/1096/commits/38415454e70394dd712121c67f7382cc08bb82f8) now leaves us with:

* default is to plot HDI with 0.94
* `hdi_list=None` ends up with the default. Because the pre-commit checks won't let me set the default value to be `[0.94]` then it has to be `None` and then set that to the default value in the function body.
* So if you want to omit plotting the HDI, then you have to do that by `hdi_list=[]`.

drbenvincent: If this is acceptable, I should maybe rebuild the docs so that the posterior prediction plots show the update... only the 94% HDI.

juanitorduz:
There is one notebook that needs to be updated

```python
 mmm.plot_posterior_predictive(add_hdi=False, add_mean=False, add_gradient=True);   # <-- Here
```

drbenvincent:
Should probably do the equivalent changes for `sample_prior_predictive` and `plot_errors`? What do you think @juanitorduz / @wd60622 ?

wd60622:
Yeah, that is a good idea. Feel free to do here or in a separate PR.

pymc_marketing/mmm/base.py
```diff
@@ -436,8 +437,13 @@ def plot_posterior_predictive(
...
 
-        if add_hdi:
-            for hdi_prob, alpha in zip((0.94, 0.50), (0.2, 0.4), strict=True):
+        if hdi_list is None:
+            hdi_list = [0.94]
+
+        if hdi_list:
+            # skipped if hdi_list is empty
+            alpha_list = np.linspace(0.2, 0.4, len(hdi_list))
+            for hdi_prob, alpha in zip(hdi_list, alpha_list, strict=True):
```
wd60622, line 446: Doesn't the `strict` cause an issue here? Should it be set to `False`?

juanitorduz: +1

pymc_marketing/mmm/base.py
```diff
@@ -436,8 +437,13 @@ def plot_posterior_predictive(
         else:
             fig = ax.figure
 
-        if add_hdi:
-            for hdi_prob, alpha in zip((0.94, 0.50), (0.2, 0.4), strict=True):
+        if hdi_list is None:
+            hdi_list = [0.94]
+
+        if hdi_list:
+            # skipped if hdi_list is empty
```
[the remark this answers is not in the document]
wd60622:
```suggestion
```

pymc_marketing/mmm/base.py
```diff
@@ -436,8 +437,13 @@ def plot_posterior_predictive(
         else:
             fig = ax.figure
 
-        if add_hdi:
-            for hdi_prob, alpha in zip((0.94, 0.50), (0.2, 0.4), strict=True):
+        if hdi_list is None:
+            hdi_list = [0.94]
```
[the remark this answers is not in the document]
drbenvincent: Oh, I maybe misinterpreted a previous reply as being in favour of copying arviz defaults of just the 0.94. But fair enough, I'll change the defaults back to `[0.94, 0.5]`

pymc_marketing/mmm/base.py
drbenvincent: So the `strict=True` was there when I got here :)

It doesn't cause any issues because the output of the `np.linspace` for the alphas is equal to the number of hdi's.

Screenshot 2024-10-24 at 11 35 37


There's no combination which causes it to fail.
Screenshot 2024-10-24 at 11 36 16

So probably makes sense to remove the `strict` kwarg as there's no way for the list of hdi's and alpha's to be of different lengths. Though you could argue that leaving it in there would be good for defensive programming when someone makes more changes to the function in the future. happy to do whatever -

wd60622: Ah didnt see using len of list

juanitorduz (approved):

drbenvincent:
"Pull Request Labeler" not passing. No idea what that is.

juanitorduz:
> "Pull Request Labeler" not passing. No idea what that is.

Yeah, this is a new thing we are working on (not a blocker, once is green we can merge)

juanitorduz:
@wd60622 the labeling is 🟢 🚀 !

drbenvincent:
I'll let you merge this when you are happy with everything @juanitorduz

wd60622 (approved):
LGTM

PR #803, opened 2024-07-03
Add new `customer_choice` module

drbenvincent:
This PR will add new functionality to pymc-marketing. Namely it will allow causal analysis of product incrementality/cannibalization.

In this initial PR we will focus on:
a) adding docs to introduce the idea behind incrementality/cannibalization analysis
b) in terms of modeling we will start with a multivariate interrupted time series model. This will include a notebook to illustrate its use, as well as docs in the guide section to convey the basic workings.

Later PR's will add additional modeling options, such as nested discrete choice models.

## TODO
- [X] Finish "Introduction to product incrementality" docs page
- [x] Finish MV-ITS model explanation
- [x] Add tests
- [x] Finish how-to notebook demonstrating use
  - [X] Add module code
  - [x] Add Scenario 3 where we operate on product level data
  - [x] Add scenario where market is not saturated + update relevant docs + tests
- [x] Add a QuickStart on the Getting Started docs page
  - [x] add python code block for QuickStart 
- [x] Rename module to `consumer-choice`. Update docs and imports and re-run notebooks etc.
- [x] Fix typo "isoluate" -> "isolate"
- [ ] Update docs homepage and `README.md`

## Known limitations
At the moment the MVITS model is very simple. It only has an intercept term, embodying the assumption that sales are relatively static over time. Clearly there are many ways to expand this model, so the proposal is that we get this in as a first attempt and stepwise add features over time. We can be lead by feature requests (if there are any), but this establishes an MVP which can be built upon.

## Checklist
- [ ] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Modules affected
This PR will create a new module and will not affect the MMM or CLV modules.

## Type of change
- [X] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--803.org.readthedocs.build/en/803/

PR #1101, opened 2024-10-22
Use ModelBuilder for incrementality model

wd60622:
Use the ModelBuilder for #803

Also uses the Prior class to generalize the logic a bit. The previous priors can be created with:

```python
def set_priors_from_existing_sales(model, existing_sales):
    model.model_config["intercept"].parameters["mu"] = existing_sales.mean().values
    model.model_config["intercept"].parameters["sigma"] = existing_sales.std().values
    model.model_config["likelihood"]["sigma"].parameters["sigma"] = (
        existing_sales.std().mean()
    )
```

Using the ModelBuilder allows user to save and load the model.

```python
test_file = "test-mvits.nc"
model.save(test_file)

loaded = MVITS.load(test_file)
```

CC: @juanitorduz @drbenvincent

----
📚 Documentation preview 📚: https://pymc-marketing--1101.org.readthedocs.build/en/1101/


pymc_marketing/model_builder.py
```diff
@@ -580,11 +580,17 @@ def fit(
         Initializing NUTS using jitter+adapt_diag...
 
         """
+        if isinstance(y, pd.Series) and not X.index.equals(y.index):
```
juanitorduz, line 583: can we add a test? :)

pymc_marketing/product_incrementality/mv_its.py
```diff
@@ -13,176 +13,217 @@
...
+        """Default model configuration."""
+        return {
+            "intercept": Prior("Normal", dims="background_product"),
+            "likelihood": Prior(
+                "TruncatedNormal",
+                lower=0,
+                sigma=Prior("HalfNormal", dims="background_product"),
+                dims=("time", "background_product"),
+            ),
+            "alpha_background": 0.5,
```
juanitorduz, line 80: what is the meaning of this `alpha_background` :D ?

pymc_marketing/product_incrementality/mv_its.py
```diff
@@ -13,176 +13,217 @@
...
-                    var_names=["mu", "y"],
-                    random_seed=self.rng,
+                    var_names=["mu", self.output_var],
+                    random_seed=random_seed,
                     predictions=True,
                 )
             )
 
-    def causal_impact(self, variable="mu"):
+    def sample(self, X, y, random_seed: int | None = None):
```
juanitorduz, line 213: can we add a self-like return type?

juanitorduz (commented):
Thanks @wd60622 ! I think this is a great addition! It makes the code more cohesive to what we already have. I have added a couple of small comments and asked for a small test ;)

pymc_marketing/product_incrementality/mv_its.py
wd60622: @drbenvincent

wd60622: I have moved this entirely to the `model_config` which will have a default like before but it also exposes it to the user to create (with some limitations)

pymc_marketing/product_incrementality/mv_its.py
```diff
@@ -13,176 +13,255 @@
...
+
+        return result
+
+    def _generate_and_preprocess_model_data(
+        self,
+        X: pd.DataFrame | pd.Series,
+        y: np.ndarray,
+    ) -> None:
+        if isinstance(X, pd.Series):
+            raise ValueError("X must be a DataFrame, not a Series")  # pragma: no cover
```
wd60622, line 132: Message doesnt make sense

wd60622 (commented):
Thoughts on this API? Does using "y" for treatment sales make sense or should that be part of X as well? 

Would this treatment sales ever be a DataFrame as well? Is that an extension we foresee?

drbenvincent (changes requested):
Thanks for the updates.

I can get the tests in `test_mv_its.py` to pass.

However, the notebooks (`mv_its_saturated.ipynb` and `mv_its_unsaturated.ipynb`) seem to not work:
* `treatment_time` seems to be absent as a kwarg from `MVITS.__init__`
* `MVITS.plot_data` is called in the notebook like a static method, but that isn't working / won't work

Would you be able to update these docs notebooks with the new API?

wd60622:
> Thanks for the updates 🥳
> 
> 
> 
> I can get the tests in `test_mv_its.py` to pass.
> 
> 
> 
> However, the notebooks (`mv_its_saturated.ipynb` and `mv_its_unsaturated.ipynb`) seem to not work:
> 
> * `treatment_time` seems to be absent as a kwarg from `MVITS.__init__`
> 
> * `MVITS.plot_data` is called in the notebook like a static method, but that isn't working / won't work
> 
> 
> 
> Would you be able to update these docs notebooks with the new API?

Yes, I can make those changes. The full CI suite isnt running as far as I can tell. Shall I merge into your previous branch and continue there? @drbenvincent

drbenvincent:
> Yes, I can make those changes. The full CI suite isnt running as far as I can tell. Shall I merge into your previous branch and continue there? @drbenvincent

Yes, makes sense. It was good to branch off of the branch, but I think it's clear that we do want this inheritance from `ModelBuilder`. As you say, more refinement can be done in the original `incrementality` branch.

Issue #1112, opened 2024-10-23
Iterate on PR auto labeling


wd60622:
No description provided.

PR #1113, opened 2024-10-23
Sync PR labels with closing issues

wd60622:
Related to #1087
Follow up to #1112

----
📚 Documentation preview 📚: https://pymc-marketing--1113.org.readthedocs.build/en/1113/

### DIFF
--- a/.github/workflows/pr-auto-label.yml
+++ b/.github/workflows/pr-auto-label.yml
@@ -9,4 +9,30 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


wd60622:
Might need to figure out the permissions on this but I think this might be good. Could be a trial and error

juanitorduz (approved):
Let's try this out! (I am not a JavaScript expert)

juanitorduz:
@wd60622 I think we are missing some permissions, see https://github.com/pymc-labs/pymc-marketing/actions/runs/11498688990/job/32005011527?pr=1096

PR #1115, opened 2024-10-24
Add `plot_expected_purchases` function

ColtAllen:
## Description

This PR consolidates the `plot_cumulative_transactions` and `plot_incremental_transactions` functions from `lifetimes` into a single plotting function, which is very useful for model evaluation:

image

So useful in fact that when I added it to the Quickstart, it revealed a data discrepancy causing model bias! I left it in the notebook because I felt like it was a teachable moment, but at the same time we're showcasing this library with a skewed model. I'm curious to hear your thoughts on this.

I added `plt.clf()` to the plotting function to reset the `pyplot` object between plots, which is useful when logging multiple figures to MLflow. I also took the opportunity to make a few other revisions to the Quickstart, such as use of the `Prior` class in model configs and a watermark. 

## Related Issue

- [ ] Closes #
- [x] Related to  https://github.com/pymc-labs/pymc-marketing/issues/326

## Checklist

## Modules affected

- [ ] MMM
- [x] CLV

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1115.org.readthedocs.build/en/1115/


pymc_marketing/clv/plotting.py
```diff
@@ -344,15 +346,138 @@ def plot_probability_alive_matrix(
...
+    time_unit: str = "D",
+    time_scaler: float | None = 1,
+    sort_transactions: bool | None = True,
+    set_index_date: bool | None = False,
+    title: str | None = None,
+    xlabel: str = "Time Periods",
+    ylabel: str = "Purchases",
+    ax: plt.Axes | None = None,
+    **kwargs,
+):
```
juanitorduz, line 375: Can we add the return type hint `plt.Axes` (or shall we return the Figure object instead?, we do this for the MMM plots)

pymc_marketing/clv/plotting.py
```diff
@@ -344,15 +346,138 @@ def plot_probability_alive_matrix(
...
+        purchase_history,
+        customer_id_col,
+        datetime_col,
+        t,
+        datetime_format,
+        time_unit,
+        time_scaler,
+        sort_transactions,
+        set_index_date,
+    )
```
juanitorduz, line 457: This function has a lot of arguments. Shall we use keyword arguments to make sure we do not pass the wrong order

```python
    df_cum_transactions = _expected_cumulative_transactions(
        model=model,
        ...
    )
```

juanitorduz (approved):
I really like this type of diagnostic plots!

I left some suggestions :D

pymc_marketing/clv/plotting.py
wd60622: +1

pymc_marketing/clv/plotting.py
```diff
@@ -344,15 +346,138 @@ def plot_probability_alive_matrix(
...
+    datetime_col :  string
+        Column in the *transactions* DataFrame denoting datetimes purchase were made.
+    t : int
+        Number of time units since earliest purchase to include in plot.
+    plot_cumulative : bool
+        Default: *True*
+        Plot cumulative purchases over time. Set to *False* to plot incremental purchases.
+    t_cal : int, optional
+        If testing model on unobserved data, specify number of time units in training data to add an indicator for
+        the start of the testing period.
```
wd60622, line 403: I am having a bit of a hard time understand the name `t_cal`. Could be the suffix `cal` throwing me off.

wd60622 (commented):
Overall looks good. Just some comments on the verbiage

pymc_marketing/clv/plotting.py
ColtAllen: Ah, a `lifetimes` leftover I overlooked. This marks the start of unobserved data if evaluating the model with `rfm_train_test_split`. I've renamed it to `t_unobserved`.

pymc_marketing/clv/plotting.py
```diff
@@ -344,15 +346,138 @@ def plot_probability_alive_matrix(
...
+    axes: matplotlib.AxesSubplot
+
+    References
+    ----------
+    .. [1] Fader, Peter S., Bruce G.S. Hardie, and Ka Lok Lee (2005),
+    A Note on Implementing the Pareto/NBD Model in MATLAB.
+    http://brucehardie.com/notes/008/
+    """
+    # clear any existing pyplot figures
+    plt.clf()
```
wd60622, line 441: I understand where you are coming from here but shouldn't it be on the user to do this? Also, does this obstruct the passing of the ax?
```suggestion
```

ColtAllen: You may have a point about the `ax` arg, but the unit tests will error out if this isn't provided, so I've moved it out of the function and into the test.

juanitorduz:
Thanks @ColtAllen !

PR #1121, opened 2024-10-27
Different Attempt on UML Automation

wd60622:
Trying a different trigger which has different permissions 

Reference: 
- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request_target-workflow-when-a-pull-request-merges


juanitorduz (approved):
Let's try it out!

PR #1147, opened 2024-11-01
Speed up notebook run action

### LINKED ISSUE #1141: Speed up `make notebook`
The `test_notebook` action is slow and takes about 30 minutes to run. Largely due to actual sampling of models taking place.

I few the purpose of this action is to ensure they run without errors and not that they are "correct"

The `scripts/run_notebooks/runner.py` script already uses `Parallel` with `n_jobs=-1` to run the notebooks in parallel.

Some ideas to speed up this action:

- [ ] Use the larger runners offered by GitHub: https://github.blog/changelog/2022-10-20-github-actions-larger-hosted-runners-are-now-automatically-created-for-customers/
- [ ] Patch `pm.sample` and other slow functions to return dummy values (pm.sample_prior_predictive)

I lean toward looking into mocking the `pm.sample` call in the notebook. We should be able to monkey patch this.

Any one else have any ideas?

wd60622:
Closes #1141

This runs into some issues with sampling Flat distribution, loading the model, and some sample_stats

This potentially could reduce the run time of the notebooks to 1/3 of the original time. One local notebook run when from 2.5 minutes to 19 seconds.

CC: @juanitorduz


scripts/run_notebooks/runner.py
```diff
@@ -25,27 +30,102 @@ def setup_logging() -> None:
     )
 
 
-def get_cwd_from_notebook_path(notebook_path: Path) -> str:
-    return str(notebook_path).rsplit("/", 1)[0]
+INJECTED_CODE = """
+import pymc as pm
+import arviz as az
+import xarray as xr
+import numpy as np
```
juanitorduz, line 37: can we isort these :)

scripts/run_notebooks/runner.py
```diff
@@ -25,27 +30,102 @@ def setup_logging() -> None:
...
+    run = mock_run if mock else actual_run
+    try:
+        run(notebook_path)
+    except Exception as e:
+        logging.error(f"Error running notebook: {notebook_path.name}")
+        raise e
+
+
 if __name__ == "__main__":
+    SLICE = slice(-2, None)
```
juanitorduz, line 118: why -2?

scripts/run_notebooks/runner.py
```diff
@@ -25,27 +30,102 @@ def setup_logging() -> None:
...
+
+    NOTEBOOKS = NOTEBOOKS[SLICE]
+
     setup_logging()
     logging.info("Starting notebook runner")
     logging.info(f"Notebooks to run: {NOTEBOOKS}")
+    run = partial(run_notebook, mock=MOCK)
     Parallel(n_jobs=-1)(
-        delayed(run_notebook)(notebook_path) for notebook_path in tqdm(NOTEBOOKS)
+        delayed(run)(notebook_path) for notebook_path in tqdm(NOTEBOOKS)
```
juanitorduz, line 128: Are we using the run_notebook function anywhere?

scripts/run_notebooks/runner.py
```diff
@@ -25,27 +30,102 @@ def setup_logging() -> None:
...
+    try:
+        run(notebook_path)
+    except Exception as e:
+        logging.error(f"Error running notebook: {notebook_path.name}")
+        raise e
+
+
 if __name__ == "__main__":
+    SLICE = slice(-2, None)
+    MOCK = True
```
juanitorduz, line 119: Can you add a comment on the Mocking idea please?

juanitorduz (commented):
Thanks @wd60622 ! I left some small questions and comments! Looks cool!

scripts/run_notebooks/runner.py
wd60622: It was used in the partial, but I switched it to run some fully and some mocked. It seems like the main issue in the HalfFlat distribution now

scripts/run_notebooks/runner.py
wd60622: Was testing locally. I've removed it now

scripts/run_notebooks/runner.py
wd60622: I've put in a file itself so it will go through the linter and formatting

wd60622:
I am looking at some of the logs. Seems like it was taken down from 40 minutes to 10 minutes 😆

scripts/run_notebooks/runner.py
wd60622: Added to the docstring of the file. Let me know your thoughts

scripts/run_notebooks/runner.py
```diff
@@ -2,21 +2,41 @@
...
+    # "sBG.ipynb",
+    # Uses the MAP estimate
+    # "bg_nbd.ipynb",
+    # "gamma_gamma.ipynb",
+    # "pareto_nbd.ipynb",
+]
 NOTEBOOKS_SKIP: list[str] = [
-    "mmm_tvp_example.ipynb",  # This notebook takes too long to run
+    # This notebook takes too long to run
+    # "other_nuts_samplers.ipynb",
```
juanitorduz, line 30: @wd60622 So if I understand this correctly, you will uncomment these notebook lists before merging as these lists decide full runs and mock runs?

juanitorduz: Why does "other_nuts_samplers.ipynb" appear in FULL RUNS and NOTEBOOKS_SKIP?

juanitorduz:
@wd60622 This is a new trick for me, hehe. How does the injected code "replace" the sampling step in the notebook? Can you maybe push an error in a notebook and see if it get caught (we can then revert the commit). This seems like magic XD

scripts/run_notebooks/runner.py
wd60622: I can actually remove all of the comments now. We can leave empty lists or populate if we want to either skip or run without a mock. 

Should I leave empty without comments or remove entirely?

wd60622:
> @wd60622 This is a new trick for me, hehe. How does the injected code "replace" the sampling step in the notebook? Can you maybe push an error in a notebook and see if it get caught (we can then revert the commit). This seems like magic XD

The injected first cell will run the "injected.py". This is before any imports which makes python think that pm.sample is mock_sample and pm.Flat = pm.Normal and pm.HalfFlat = pm.HalfFlat because they are cached even when they are imported from our modules

scripts/run_notebooks/runner.py
juanitorduz: I think we can remove entirely if we are not using it

juanitorduz (approved):
ok! I thnik I get it now :D Very nice! After removing the commented lines (see comment above), feel free to merge 🚀

PR #1153, opened 2024-11-03
Conditional run of tests

wd60622:
This will have the tests run only if any of the files in the list have changed:

- pyproject.toml
- anything in pymc_marketing folder
- anything in the tests

This would remove running tests if there are changes to other python files that might exist like in the scripts folder like in
#1147

----
📚 Documentation preview 📚: https://pymc-marketing--1153.org.readthedocs.build/en/1153/


juanitorduz (approved):
yes! good call!

PR #1166, opened 2024-11-04
Creating Causal Identification module

cetagostini:
## Description

Short description: Integration of CausalGraphModel in BaseMMM Class

This update integrates a CausalGraphModel into the BaseMMM class, allowing for automated causal identification based on backdoor criteria, assuming a given Directed Acyclic Graph (DAG).

#### Summary of Changes
1. **Added Causal Graph Option**: 
   - The `BaseMMM` class now accepts an optional `dag` parameter, which can be provided either as a string (DOT format) or a `networkx.DiGraph`. 
   - If `dag` is provided, a `CausalGraphModel` is instantiated to analyze causal relationships and determine necessary adjustment sets.

2. **Automatic Minimal Adjustment Set Handling**:
   - The `BaseMMM` initialization now includes logic to calculate the minimal adjustment set required to estimate the causal effect of the treatment variables (assume to be media channels) on the outcome.
   - `control_columns` are automatically updated to include variables from the minimal adjustment set only.
   - If the variable `yearly_seasonality` is not in the minimal adjustment set, the `yearly_seasonality` parameter is set to `None`, effectively disabling it in the model.

3. **Warnings for Missing Adjustment Sets**:
   - If a minimal adjustment set cannot be identified, a warning is issued, and not modifications are made during the initialization.

#### Code Example

Here's how to initialize `BaseMMM` with a DAG for causal inference:

```python
dag_str = """
digraph {
    x1 -> y;
    x2 -> y;
    yearly_seasonality -> y;
    event_1 -> y;
    event_2 -> y;
}
"""

mmm = MMM(
    model_config=my_model_config,
    sampler_config=my_sampler_config,
    date_column="date_week",
    channel_columns=["x1", "x2"],
    control_columns=["event_1", "event_2"],
    yearly_seasonality=2,  # Disabled if 'yearly_seasonality' is not in minimal adjustment set
    dag=dag_str,
    outcome_column="y",
)
```

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1166.org.readthedocs.build/en/1166/


wd60622:
What is `z` in the 2nd body example? Would that be in the model?

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,190 @@
+import networkx as nx
+from dowhy.graph import get_backdoor_paths, get_instruments, build_graph_from_str, build_graph
+from dowhy.causal_identifier.auto_identifier import (
+    construct_backdoor_estimand,
+    construct_frontdoor_estimand
+)
+from typing import list, Set, Optional
```
wd60622, line 7: You don't need these

```suggestion
```

The alternative now is: 

```python
list[str]
set[str]
str | None
```

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,190 @@
...
+            A set of variable names representing the minimal adjustment set, or None if not identifiable.
+        """
+        try:
+            estimand = construct_backdoor_estimand(
+                self.graph, self.treatment[0], self.outcome[0]
+            )
+            return estimand.get_backdoor_variables()
+        except Exception as e:
+            print("Error identifying backdoor adjustment set:", e)
+            return None
```
wd60622, line 142: Just raise error. User can handle downstream

```suggestion
        estimand = construct_backdoor_estimand(
            self.graph, self.treatment[0], self.outcome[0]
        )
        return estimand.get_backdoor_variables()
```

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,190 @@
...
+        bool
+            True if frontdoor adjustment is possible, False otherwise.
+        """
+        try:
+            frontdoor_estimand = construct_frontdoor_estimand(
+                self.graph, self.treatment[0], self.outcome[0]
+            )
+            return frontdoor_estimand is not None
+        except Exception:
+            return False
```
wd60622, line 159: same

wd60622, line 159:
```suggestion
        try:
            frontdoor_estimand = construct_frontdoor_estimand(
                self.graph, self.treatment[0], self.outcome[0]
            )
        except Exception:
            return False
        else: 
            return frontdoor_estimand is not None
```

cetagostini:
> What is `z` in the 2nd body example? Would that be in the model?

Old example, I did the correction!

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,134 @@
...
+    causal_model : CausalModel
+        An instance of dowhy's CausalModel, representing the causal graph and its relationships.
+    treatment : list[str]
+        A list of treatment variable names.
+    outcome : str
+        The outcome variable name.
+    """
+
+    def __init__(
+        self, causal_model: CausalModel, treatment: list[str] | tuple[str], outcome: str
```
juanitorduz, line 39: Shall we type `treatment` as an iterator https://wiki.python.org/moin/Iterator?

juanitorduz: (applied to all below)

pymc_marketing/mmm/mmm.py
```diff
@@ -113,6 +114,17 @@ def __init__(
...
+            None,
+            description="Optional DAG provided as a string Dot format for causal identification.",
+        ),
+        treatment_nodes: list[str] | tuple[str] | None = Field(
+            None,
+            description="Column names of the variables of interest to identify causal effects on outcome.",
+        ),
+        outcome_node: str | None = Field(
+            None, description="Name of the outcome variable."
+        ),
```
juanitorduz, line 127: This is where I would like to discuss the API. Our MMM class is already a huge monolith of many components, and I would like us to start modularizing more or even making it a subclass. 

For instance, we can keep `BaseMMM` as it is and have an additional

`CausalMMM(BaseMMM)`, and if people want to use this class, they need to install `DoWhy`. I am personally against adding `DoWhy` as a required dependency, as in my experience, they sometimes hard-pin soma packages and can make it harder to resolve dependencies. WDYT?

Thoughts @wd60622 ?

pymc_marketing/model_builder.py
```diff
@@ -341,6 +341,16 @@ def default(x):
...
 
+        attrs["dag"] = (
+            json.dumps(self.dag, default=default) if hasattr(self, "dag") else "None"
+        )
+        attrs["treatment_nodes"] = (
+            self.treatment_nodes if hasattr(self, "treatment_nodes") else "None"
+        )
+        attrs["outcome_node"] = (
+            self.outcome_node if hasattr(self, "outcome_node") else "None"
+        )
```
juanitorduz, line 352: I do not think this **has** to be in the model builder class, which is much more general. For instance, this is not necessary for the CLV module.

pyproject.toml
```diff
@@ -63,6 +65,8 @@ docs = [
     "sphinxext-opengraph",
     "watermark",
     "mlflow>=2.0.0",
+    "networkx",
+    "dowhy",
```
juanitorduz, line 69: I think this should not be core dependencies but rather optional (like numpyro for sampling). See the comment above on abstraction.

juanitorduz (changes requested):
Cool stuff @cetagostini ! I gave a quick look into the code and I think it brings a lot of value. Nevertheless, I think we need to improve the right level of abstraction and do not include this feature in the `BaseMMM` class (see comment below)

juanitorduz:
@carlosagostini I really liked the notebook 🚀 ! 

I think if we improve the level of abstraction (see comments above) and update the notebooks, we can merge this one soon!

@drbenvincent It would be great if you could review the notebook to provide feedback if you have time 🙏 :)

pymc_marketing/mmm/mmm.py
cetagostini: I like this idea. I can work quickly on it, will wait for William comments 🙌🏻 Probably will have a meeting with him on Tuesday!

pymc_marketing/model_builder.py
wd60622: 100%

tests/mmm/test_causal.py
```diff
@@ -0,0 +1,171 @@
...
+            """
+            digraph {
+                X -> Y;
+            }
+            """,
+            ["X"],
+            "Y",
+            [],  # No adjustment is needed
+        ),
+        # Case 4: Multiple treatments
```
wd60622, line 67: Let's use ids in pytest for these labels

pymc_marketing/mmm/mmm.py
wd60622: If the dependencies are the issue then I think we can get away with having the dowhy and networkx only be required if the dag is specified. That would make models with backward compat not needing to add the new dependencies for the same model. Would only checking for these depends in the case of using this functionality solve your concerns? @juanitorduz 

I think going the route of subclassing could just add more code to manage 😢

wd60622, line 127: Do we have to specify this? We already have the `output_var`. Could that be leveraged?

wd60622 (commented):
Some comments and questions

pymc_marketing/mmm/mmm.py
juanitorduz: > I think going the route of subclassing could just add more code to manage 😢

true ... what would be your suggestion @wd60622 ?

pymc_marketing/mmm/mmm.py
```diff
@@ -178,6 +196,37 @@ def __init__(
...
+                    "No treatment nodes provided, using channel columns as treatment nodes.",
+                    stacklevel=2,
+                )
+            self.causal_graphical_model = CausalGraphModel.build_graphical_model(
+                graph=self.dag,
+                treatment=self.treatment_nodes,
+                outcome=self.outcome_node,
+            )
+
+            self.control_columns = self.causal_graphical_model.compute_adjustment_sets(
```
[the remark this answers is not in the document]
cetagostini: Correct, thats why I added here.

pymc_marketing/mmm/mmm.py
cetagostini: Let me check, not necessarily.

pymc_marketing/model_builder.py
cetagostini: Then where should I added? I got issues if I don't add them here, but let me try again.

pymc_marketing/mmm/mmm.py
wd60622: If we could simplify, that'd be great

pyproject.toml
cetagostini: I'm following @wd60622 idea and moving this to optional dependencies.

pymc_marketing/mmm/mmm.py
cetagostini: I think less code to manage its better and users still import the same MMM class. The amount of code lines is only 20, I don't see it as something crazy. Whats your opinion?

cetagostini: @wd60622 we can but then I'll need to restrict that output variable in the DAG is always `y` (The return from `output_var `). Use `sales` or `revenue` or `registrations` or any other would be possible. Not crazy, but then we avoid one parameter. What do you think?

cetagostini:
Hey guys, the PR is good. But I reckon, I need a hand with the tests.

@wd60622 @juanitorduz Happy if we can have a quick meeting to address them?

pymc_marketing/model_builder.py
cetagostini: ps: This is connect to the test, without this code the test fail.

cetagostini:
Ready for final review @juanitorduz

juanitorduz:
@cetagostini is the change in the MMM example relevant for this PR? I can't find a difference between the original and the change :)

cetagostini:
> @cetagostini is the change in the MMM example relevant for this PR? I can't find a difference between the original and the change :)

I need it to run to get a new `model.nc` with the latest changes, as well, I'm use to run it to be sure everything works. But indeed nothing change!

cetagostini:
Something else here pending? @wd60622 @juanitorduz

cetagostini:
Something blocking this merge? @juanitorduz @wd60622

pymc_marketing/model_builder.py
wd60622, line 352: Thought we moved this to the mmm class. This shouldn't be in the model_builder

wd60622 (changes requested):
We will need some mmm tests as well

pymc_marketing/model_builder.py
cetagostini: Good catch, you did and was not removed.

cetagostini:
Current error is:
```sh
writing output... [ 75%] api/generated/pymc_marketing.mmm.mmm.BaseMMM.get_errors
Command killed due to timeout or excessive memory consumption
```

juanitorduz:
> Something blocking this merge? @juanitorduz @wd60622

I will take another look next week 🙏 . In the meantime I agree we need to add more tests :)

cetagostini:
@juanitorduz @wd60622 

Hey guys I added a few test to the causal model initialization in the MMM class. Everything should be to merge, main behavior its on the `test_causal.py` and the initialization its under `test_mmm.py`

juanitorduz:
I will take a detailed review this week now that we finally pushed the awaiting customer choice model 🙏

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,149 @@
...
+    for backdoor adjustment between treatment and outcome variables.
+
+    Parameters
+    ----------
+    causal_model : CausalModel
+        An instance of dowhy's CausalModel, representing the causal graph and its relationships.
+    treatment : list[str]
+        A list of treatment variable names.
+    outcome : str
+        The outcome variable name.
```
juanitorduz, line 50: Can you please add https://github.com/py-why/dowhy a a Reference in the class description?

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,149 @@
...
+        list[list[str]]
+            A list of backdoor paths, where each path is represented as a list of variable names.
+        """
+        # Use DoWhy's internal method to get backdoor paths for all treatments combined
+        return self.causal_model._graph.get_backdoor_paths(
+            nodes1=self.treatment, nodes2=[self.outcome]
+        )
+
+    def get_unique_adjustment_nodes(self) -> list[str]:
+        """Compute the minimal adjustment set required for backdoor adjustment across all treatments.
```
juanitorduz, line 99: Can you expand more on the meaning of the minimal set (think about new comers)? Can you also also add references? I suggest

Causal Inference in Statistics
A Primer
By [Judea Pearl](https://www.google.de/search?sca_esv=dfa57c05b4e3cad6&hl=en&q=inauthor:%22Judea+Pearl%22&udm=36), [Madelyn Glymour](https://www.google.de/search?sca_esv=dfa57c05b4e3cad6&hl=en&q=inauthor:%22Madelyn+Glymour%22&udm=36), [Nicholas P. Jewell](https://www.google.de/search?sca_esv=dfa57c05b4e3cad6&hl=en&q=inauthor:%22Nicholas+P.+Jewell%22&udm=36) · 2016

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,149 @@
...
+            raise ImportError(msg)
+
+    CausalModel = LazyCausalModel
+
+
+class CausalGraphModel:
+    """Represent a causal model based on a Directed Acyclic Graph (DAG).
+
+    Provides methods to analyze causal relationships and determine the minimal adjustment set
+    for backdoor adjustment between treatment and outcome variables.
```
juanitorduz, line 41: Sometimes, external regressors are not in the minimal set but help decreasing variance; see https://matheusfacure.github.io/python-causality-handbook/07-Beyond-Confounders.html#good-controls

Concretely:

> Anytime we have a control that is a good predictor of the outcome, even if it is not a confounder, adding it to our model is a good idea. 

So I am hesitant to remove, for example seasonality, if it is not in the minimal set. WDYT?

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,149 @@
...
+
+        if unique_controls:
+            warnings.warn(
+                f"Columns {unique_controls} are not in the adjustment set. Controls are being modified.",
+                stacklevel=2,
+            )
+
+        control_columns = list(common_controls - set(channel_columns))
+
+        self.minimal_adjustment_set = control_columns + list(channel_columns)
```
juanitorduz, line 139: I am hesitant on this step because my comment on variance reduction above. Maybe we can have an additional parameter, something like `minimal` or `maximal` set . WDYD?

pymc_marketing/mmm/mmm.py
```diff
@@ -180,6 +198,37 @@ def __init__(
...
+                control_columns=self.control_columns,
+                channel_columns=self.channel_columns,
+            )
+
+            if "yearly_seasonality" not in self.causal_graphical_model.adjustment_set:
+                warnings.warn(
+                    "Yearly seasonality excluded as it's not required for adjustment.",
+                    stacklevel=2,
+                )
```
juanitorduz, line 229: This should be split into at least one or two functions and just called at initialization (+ unit test for each function)

juanitorduz (changes requested):
Hey @carlosagostini ! Sorry for the late review #shameonme 

I left some comments. One important observation is the missing opportunities on variance reduction with non-minima sets (please see comment below). There are other small comments and suggested changes regarding some minor code modularization.

juanitorduz:
<p>Add subtitle like: business problem</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2481416233'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Shall we remove the first data points which are generated by the natural fact that we can not adstock much for the initial point ?</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2481416242'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>+1</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2481416246'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
I think the previous comments on the notebooks have not been addressed yet ;)

pymc_marketing/mmm/causal.py
cetagostini: Done!

pymc_marketing/mmm/causal.py
cetagostini: Done!

pymc_marketing/mmm/causal.py
cetagostini: Based on my current understanding, this is correct **under the assumption that the variable is conditionally independent of the treatment** given the adjustment set and does not introduce bias. If thats the case, then including strong predictors of the outcome can reduce the residual variance in the outcome, leading to more precise estimates of the treatment effect on other nodes.

The example where it is shown with email and credit limit works, the other nodes do not generate a problem, but many times we have situations where the DAG is more complicated and if you do not remove it, you introduce bias. I think that assuming that whoever uses the module will always have an easy situation where they can leave everything would no longer be optimal.

As you say, it reduces the variance but by adding or not seasonal, if seasonal is not part of the minimum adjustment set then the mean estimation of the nodes of interest would not change. However, not removing it assuming it will be independent is a huge assumption that could affect the estimation of the node of interest, if thats not the case for a specific user.

pymc_marketing/mmm/causal.py
cetagostini: Regarding variance reduction, I replied. On the other hand, indeed we could implement something like that, should not be complicated, and I'm happy to work on it. 

Do we want this for initial release? Sounds like we could add later on, I think its well understood that minimal its minimal adjustment, only the necessary is needed. Maximal could come on a following PR to allow largest possible set of variables that can be adjusted for without introducing bias.

@juanitorduz

pymc_marketing/mmm/mmm.py
cetagostini: eh! 😅 I'm not sue how much more we can reduce, its 23 lines, and if you see, I'm calling only two methods to define two variables. I need those inputs from the init, any reduction will end in the same similar 23 lines, because those inputs are needed and pre-commit will create spaces between them.

The only reason why is 23 lines is for the pre-commit, if I could write this in long format, would be 3 lines, but the inputs get space between them. So, how to make it lower than 3 actual lines?

If statements doesn't count, the must be their anyways, we can't compile outside.  Same thing with warnings.

Or am I seeing something different?

cetagostini:
<p>Removed by mean plus noise instead of drop.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2481416242'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
<p>This will hold a bit by the adstock effects, always with the first points.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2481416246'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
<p>Yes, I explain about them later on in the text.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2481416250'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
<p>Got some latex issue when I tried, can you give me a hand? Maybe I did something wrong.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2481416256'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
<p>Done!</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2481416262'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
<p>Replaced.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2568279793'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
<p>Done!</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2568279803'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
<p>Must be x1t but maybe I'm running something wrong, can you give me the markdown to replace? Just the style, I can propagate. </p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1166/discussion/#comment-2568279813'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

pymc_marketing/mmm/mmm.py
juanitorduz: yeah, you right. Thanks for the feedback 🙇

pymc_marketing/mmm/causal.py
juanitorduz: Thanks @carlosagostini . I think we can provide a minimal set as you suggest but add a user warning that this feature is experimental and that the minimal set does not always lead to the best model, as we could be missing opportunities to reduce the variance of the estimate. We can release this initial experimental feature and collect feedback from the users, WDYT?

juanitorduz:
@wd60622 it would be great to have another pair of eyes to review this cool PR ;) (if possible)

docs/source/notebooks/index.md
```diff
@@ -7,6 +7,9 @@ Here you will find a collection of examples and how-to guides for using PyMC-Mar
 :caption: MMMs
 :maxdepth: 1
 
+mmm/mmm_example
+mmm/mmm_causal_identification
+mmm/mmm_budget_allocation_example
```
juanitorduz, line 12: we want to make sure these sections are ordered alphabetically, can you order them please 🙏 ?

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,158 @@
...
+        The outcome variable name.
+
+    References
+    ----------
+    .. [1] https://github.com/microsoft/dowhy
+    """
+
+    def __init__(
+        self, causal_model: CausalModel, treatment: list[str] | tuple[str], outcome: str
+    ):
```
juanitorduz, line 59: return None type hint

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,158 @@
...
+        self.causal_model = causal_model
+        self.treatment = treatment
+        self.outcome = outcome
+
+    @classmethod
+    def build_graphical_model(
+        cls, graph: str, treatment: list[str] | tuple[str], outcome: str
+    ):
```
juanitorduz, line 67: return type hint

pymc_marketing/mmm/causal.py
```diff
@@ -0,0 +1,158 @@
...
+            for node in path
+            if node not in self.treatment and node != self.outcome
+        )
+        return list(adjustment_nodes)
+
+    def compute_adjustment_sets(
+        self,
+        channel_columns: list[str],
+        control_columns: list[str] | None = None,
+    ):
```
juanitorduz, line 129: return type hint

pymc_marketing/mmm/mmm.py
```diff
@@ -672,6 +724,9 @@ def attrs_to_init_kwargs(cls, attrs) -> dict[str, Any]:
             "time_varying_media": json.loads(attrs.get("time_varying_media", "false")),
             "validate_data": json.loads(attrs["validate_data"]),
             "sampler_config": json.loads(attrs["sampler_config"]),
+            "dag": json.loads(attrs.get("dag", "null")),
+            "treatment_nodes": json.loads(attrs.get("treatment_nodes", "null")),
+            "outcome_node": json.loads(attrs.get("outcome_node", "null")),
```
juanitorduz, line 729: Can we test this in the load tests 🙏 ?

juanitorduz (changes requested):
Some additional code changes :)

juanitorduz:
The notebook looks 🚀 ! I think after the previous small code comments, we can merge this one (plus potential @wd60622 comments)

docs/source/notebooks/index.md
cetagostini: Done!

pymc_marketing/mmm/causal.py
cetagostini: Done!

pymc_marketing/mmm/causal.py
cetagostini: Done!

pymc_marketing/mmm/causal.py
cetagostini: Done!

cetagostini:
Thank you @juanitorduz, I follow your comments! Hope everything is looking a bit better.

pymc_marketing/mmm/mmm.py
cetagostini: I did an independent test for this only!

wd60622 (approved):
Looks good!

Can we provide a better PR name before the release. "Add Causal DAG support in MMM definition" or something

Issue #1148, opened 2024-11-02
`plot_grouped_contribution_breakdown_over_time` fails with negative values in group


wd60622:
While working on #1147 and running in notebook

<details>
<summary>Error Message</summary>

```python
ValueError                                Traceback (most recent call last)
Cell In[45], line 13
      1 groups = {
      2     "Base": [
      3         "intercept",
   (...)
     10     "Channel 2": ["x2"],
     11 }
---> 13 fig = mmm.plot_grouped_contribution_breakdown_over_time(
     14     stack_groups=groups,
     15     original_scale=True,
     16     area_kwargs={
     17         "color": {
     18             "Channel 1": "C0",
     19             "Channel 2": "C1",
     20             "Base": "gray",
     21             "Seasonality": "black",
     22         },
     23         "alpha": 0.7,
     24     },
     25 )
     27 fig.suptitle("Contribution Breakdown over Time", fontsize=16);

File ~/GitHub/pymc-eco/pymc-marketing/pymc_marketing/mmm/base.py:1027, in MMMModelBuilder.plot_grouped_contribution_breakdown_over_time(self, stack_groups, original_scale, area_kwargs, **plt_kwargs)
   1025 if area_kwargs is not None:
   1026     area_params.update(area_kwargs)
-> 1027 all_contributions_over_time.plot.area(**area_params)
   1028 ax.legend(title="groups", loc="center left", bbox_to_anchor=(1, 0.5))
   1029 return fig

File ~/micromamba/envs/pymc-marketing-dev/lib/python3.10/site-packages/pandas/plotting/_core.py:1603, in PlotAccessor.area(self, x, y, stacked, **kwargs)
   1526 def area(
   1527     self,
   1528     x: Hashable | None = None,
   (...)
   1531     **kwargs,
   1532 ) -> PlotAccessor:
   1533     """
   1534     Draw a stacked area plot.
   1535
   (...)
   1601         >>> ax = df.plot.area(x='day')
   1602     """
-> 1603     return self(kind="area", x=x, y=y, stacked=stacked, **kwargs)

File ~/micromamba/envs/pymc-marketing-dev/lib/python3.10/site-packages/pandas/plotting/_core.py:1030, in PlotAccessor.__call__(self, *args, **kwargs)
   1027             label_name = label_kw or data.columns
   1028             data.columns = label_name
-> 1030 return plot_backend.plot(data, kind=kind, **kwargs)

[...]

PR #1181, opened 2024-11-07
patch values stacked plot

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/1148

Simply clip the contributions and add a warning.

----
📚 Documentation preview 📚: https://pymc-marketing--1181.org.readthedocs.build/en/1181/

### DIFF
--- a/pymc_marketing/mmm/base.py
+++ b/pymc_marketing/mmm/base.py
@@ -1075,7 +1075,7 @@ class MMMModelBuilder(ModelBuilder):
         original_scale: bool = False,
         area_kwargs: dict[str, Any] | None = None,
         **plt_kwargs: Any,
-    ) -> plt.Figure | None:
+    ) -> plt.Figure:
         """Plot a time series area chart for all channel contributions.
 
         Since a chart like this can become quite crowded if you have many channels or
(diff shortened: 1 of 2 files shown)


pymc_marketing/mmm/base.py
```diff
@@ -1126,7 +1127,14 @@ def plot_grouped_contribution_breakdown_over_time(
...
             area_params.update(area_kwargs)
-        all_contributions_over_time.plot.area(**area_params)
+        try:
+            all_contributions_over_time.plot.area(**area_params)
+        except ValueError:
+            warnings.warn(
+                "Each contribution value must be either all positive or all negative",
+                stacklevel=2,
+            )
+            return None
```
juanitorduz, line 1137: I do not think there is an easy fix. It is a limitation of the plot. 

Any other ideas?

wd60622: This is what comes to my mind. We were basically just lucky that posterior estimates where positive for "Base" group.

wd60622: I would just return the figure still

pymc_marketing/mmm/base.py
```diff
@@ -1126,7 +1127,17 @@ def plot_grouped_contribution_breakdown_over_time(
         area_params = dict(stacked=True, ax=ax)
         if area_kwargs is not None:
             area_params.update(area_kwargs)
-        all_contributions_over_time.plot.area(**area_params)
+        try:
+            all_contributions_over_time.plot.area(**area_params)
```
wd60622, line 1131: Thoughts of check on frame values instead of try except?

wd60622 (approved):
Not blocker. Lets get fix through

pymc_marketing/mmm/base.py
juanitorduz: The cheks are done in the plot.area function itself and can have many edge cases. For now lets direct the responsibility to matplotlib 🙏

PR #1189, opened 2024-11-11
Run sBG with seed

wd60622:
The notebook runner has been failing on this notebook every so often. Trying to add a random seed which will then be passed even for the injected code.

----
📚 Documentation preview 📚: https://pymc-marketing--1189.org.readthedocs.build/en/1189/


juanitorduz (approved):
Great!

PR #1212, opened 2024-11-17
[FIX] Wrong legend in the posterior predictive plot then we set gradient=True

juanitorduz:
Fix wrong legend.

### Before 

image

### After

image

(see example notebook diff)

----
📚 Documentation preview 📚: https://pymc-marketing--1212.org.readthedocs.build/en/1212/


wd60622 (approved):
Easy enough. Looks good

Issue #726, opened 2024-06-08
Prepare for PyMC 5.16 release


juanitorduz:
This new release has breaking changes because of the latest version of PyTensor. We are "solving" this by having an upper bound on the PyMC version, see https://github.com/pymc-labs/pymc-marketing/pull/725

The updates we need to do are [(only) here (?)](https://github.com/pymc-labs/pymc-marketing/blob/6edc4ca4e9353d10804c8817252a7e45879c0827/pymc_marketing/clv/distributions.py)

juanitorduz:
Following https://github.com/pymc-labs/pymc-marketing/pull/725#discussion_r1634913083 we will unpin the upper bound version once we fix this

wd60622:
This PR might be helpful
https://github.com/pymc-devs/pymc-bart/pull/194

PR #1241, opened 2024-11-26
Bump minimal dependency PyMC >= 5.16

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/726

----
📚 Documentation preview 📚: https://pymc-marketing--1241.org.readthedocs.build/en/1241/


wd60622:
What's the reason for the upper bound at 5.19.0?

juanitorduz:
> What's the reason for the upper bound at 5.19.0?

Not strong opinions and we can remove it. I was thinking if 5.19.0 has other breaking changes that users are not faced with cryptic error messages. We can also make pymc-marketing releases more often that support every pymc major version update. WDYT?

wd60622:
> Not strong opinions and we can remove it. I was thinking if 5.19.0 has other breaking changes that users are not faced with cryptic error messages. We can also make pymc-marketing releases more often that support every pymc major version update. WDYT?

That'd be ideal so new pytensor can be used

wd60622:
And wouldn't the test suite catch if the latest version of pymc is not working?

juanitorduz:
Yes, but imagine another package or model in prod using pymc marketing. If they do not pin for a potential pymc breaking change then it will break their pipelines. Whereas if we have this is just a much more conservative approach. TBH, not strong opinion but this might motivate us to sync pathc releases with pymc

juanitorduz:
See https://github.com/pymc-labs/pymc-marketing/pull/1269

PR #1246, opened 2024-11-30
HSGP as component

### LINKED ISSUE #751: Pull out TVP as component

wd60622:
## Description

Adding `HSGP` component which acts a bit more like `Prior` class than MMM
components but with some `plot_curve` logic as well.

Bringing back the PR that @bwengals started for better priors for the 1d HSGP

```python
import numpy as np
import pandas as pd

from pymc_marketing.mmm import HSGP

n = 52
X = np.arange(n)

hsgp = HSGP.parameterize_from_data(X=X, dims="time")

dates = pd.date_range("2024-01-01", periods=n, freq="W-MON")
coords = {"time": dates}
prior = hsgp.sample_prior(coords=coords)
curve = prior["f"]
fig, axes = hsgp.plot_curve(curve)
ax = axes[0]
ax.set(xlabel="Time", ylabel="f", title="Out of the box HSGP prior")
plt.show()
```

hsgp

*EDIT:* I've implemented this with the `HSGP.parameterize_from_data` class method: 

We might want to refactor this to pull out the defaults instead of having them
as `property` of the class. Maybe like:

```python
eta: Prior = get_recommended_eta(X)
ell: Prior = get_recommended_ell(X)
hsgp = HSGP(dims="time", eta=eta, ell=ell)
```

## Related Issue

- [ ] Closes #751
- [ ] Related to #829

## Checklist

## Modules affected

- [x] MMM
- [ ] CLV

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1246.org.readthedocs.build/en/1246/


wd60622:
Checkout the documentation section here: 
https://pymc-marketing--1246.org.readthedocs.build/en/1246/api/generated/pymc_marketing.hsgp_kwargs.HSGP.html#pymc_marketing.hsgp_kwargs.HSGP

Thoughts on using the `create_variable` over `apply` as GP is thought to be a distribution over functions wheras `apply` is a bit more for transformation of data. 

One aspect that is tough is that the new parametrization is different than the `HSGPKwargs` class and conflicts a bit. This would be a breaking change and require a version increase for `MMM` class. However, this might be another argument for parameterizing the `HSGP` with distributions which might be more flexible in the long run. I think there are still some conflicts with this route due to the nature of this. 

```python

eta = Prior("HalfNormal", dims="channel")

hsgp = 

TODO: 

- [ ] Add to the components notebook
- [ ] Incorporate into the MMM directly after determining relationship between `HSGP` and `HSGPKwargs` currently used.

**Comment by wd60622** [comment]
Question for @bwengals :
Is there a benefit of centered vs non-centered since `sqrt_psd` will be a vector and not a **random** vector of size (m, )? hsgp_coefs will potentially have more dims (*other_dims, m)

**Comment by juanitorduz** [comment]
This looks very nice! In the long run, this seems like a nicer approach. As we have not reached a v1.0.0 version I think it's fine that the API changes from time to time

**Review by juanitorduz — COMMENTED** [review]
This looks great @wd60622 ! I really like the proposal! A couple of questions and comments:

- Do we need more tests for the `hsgp_kwargs`? What is missing in this PR?
- It would be great to update the GP MMM notebooks (maybe different PR) to showcase this component point of view.
- Can we also do HSGPPeriodic? I think modeling the media time-varying coefficients as a periodic GP is something we have been asked for.

File: pymc_marketing/hsgp_kwargs.py [hunk 1]
```diff
@@ -13,10 +13,576 @@
...
+
+    return Prior(
+        "Weibull",
+        alpha=0.5,
+        beta=1.0 / np.square(lam_ell),
+        transform="reciprocal",
+    )
+
+
+def approx_hsgp_hyperparams(
```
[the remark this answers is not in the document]
wd60622: Doesn't seem to be in pymc 5.13.0: https://github.com/pymc-devs/pymc/blob/v5.13.0/pymc/gp/hsgp_approx.py#L603

pymc_marketing/hsgp_kwargs.py
```diff
@@ -13,10 +13,576 @@
...
+        Tail probability.
+    lower : float
+        Lower bound for the lengthscale.
+
+    Returns
+    -------
+    Prior
+        Weibull prior for the lengthscale with the given properties.
+    """
+    lam_ell = -np.log(alpha) * (1.0 / np.sqrt(lower))
```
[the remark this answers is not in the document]
wd60622: probably not much as basically just weibull with constants as parameters. I don't have preference but can change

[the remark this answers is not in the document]
cetagostini: We are moving components, transformations and utility functions to `pytensor`, I think would be better to stick to pytensor if its not difficult to.

pymc_marketing/hsgp_kwargs.py
```diff
@@ -13,10 +13,576 @@
...
+from pytensor.tensor import TensorLike
+from pytensor.tensor.variable import TensorVariable
+from typing_extensions import Self
+
+from pymc_marketing.plot import SelToString, plot_curve
+from pymc_marketing.prior import Prior, create_dim_handler
+
+
+def pc_prior_1d(alpha: float = 0.1, lower: float = 1.0) -> Prior:
+    R"""Create a one-dimensional PC prior for GP lengthscale.
```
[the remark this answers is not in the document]
bwengals: Penalized complexity.  Here's the reference: https://arxiv.org/abs/1503.00256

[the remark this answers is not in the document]
bwengals: It'd be nice to have the PC prior available, but IMO a better default prior for PyMC marketing would be InverseGamma, and then have the user give a lower and upper bound, as described here: https://betanalpha.github.io/assets/case_studies/gaussian_processes.html

I think ideally:
- user specifies a lower bound and prior mass -> PC prior
- user specifies a lower, upper bound and mass -> InverseGamma

wd60622:
I've added a `HSGPPeriodic` class that will work the same way as the `HSGP` class. These stand alone from the `MMM` class. Shall we approach that in a different PR? Finish this one up with: 

- [ ] Tests
- [ ] Compoment notebook update

and then open a new PR for the `MMM` class integration?

Thoughts? @juanitorduz

wd60622:
Thoughts on the location of this logic? `hsgp_kwargs` doesn't seem like a good name and might be good in the `mmm` module.

pymc_marketing/mmm/tvp.py
```diff
@@ -92,7 +92,40 @@
 from pymc.distributions.shape_utils import Dims
 
 from pymc_marketing.constants import DAYS_IN_YEAR
-from pymc_marketing.hsgp_kwargs import HSGPKwargs
+from pymc_marketing.hsgp_kwargs import HSGP, CovFunc, HSGPKwargs
```
cetagostini, line 95: Why not to move `tvp` into components folder? Maybe with a better name. I'm thinking with HSGP we are modeling latent variables, maybe `latent.py` under components. Replacing `hsgp_kwargs` instead.

`from pymc_marketing.components.latent import HSGP, CovFunc, HSGPKwargs`

cetagostini:
> Thoughts on the location of this logic? `hsgp_kwargs` doesn't seem like a good name and might be good in the `mmm` module.

Indeed, maybe something related to latent makes more sense to me, or time. If this is a component already, should be under component folders, and the name should be related to what are we trying to model.

We want to model saturation, then we have the saturation component, if we want to model adstock, then we have the adstock component, same thing if we want to model time/latent effects.

cetagostini:
> Question for @bwengals : Is there a benefit of centered vs non-centered since `sqrt_psd` will be a vector and not a **random** vector of size (m, )? hsgp_coefs will potentially have more dims (*other_dims, m)

I think should be possible, but better to have @bwengals input here 😅

bwengals:
> Is there a benefit of centered vs non-centered since sqrt_psd will be a vector and not a random vector of size (m, )? hsgp_coefs will potentially have more dims (*other_dims, m)

Sorry i missed this.  Yes sometimes you might want to use the centered parameterization, though non-centered is probably the better default.

pymc_marketing/mmm/hsgp.py
```diff
@@ -0,0 +1,967 @@
...
+        Upper bound for the variance. Default is 1.0.
+
+    """
+    return Prior(
+        "Exponential",
+        lam=-pt.log(mass) / upper,
+    )
+
+
+def create_m_and_L_recommendations(
```
juanitorduz, line 208: Can't we use https://github.com/pymc-devs/pymc/blob/main/pymc/gp/hsgp_approx.py#L97 ?

pymc_marketing/mmm/hsgp.py
```diff
@@ -0,0 +1,967 @@
...
+    Prior
+        The prior for the lengthscale.
+
+    """
+    return Prior(
+        "InverseGamma",
+    ).constrain(lower=lower, upper=upper, mass=mass)
+
+
+def approx_hsgp_hyperparams(
```
juanitorduz, line 110: Is it the same as in https://github.com/pymc-devs/pymc/blob/main/pymc/gp/hsgp_approx.py#L97 ?

pymc_marketing/mmm/hsgp.py
juanitorduz: Can we add some references on where these heuristic come from (just the paper reference, ideally the section)

juanitorduz: (I think we can use the same ref as above)

pymc_marketing/mmm/hsgp.py
```diff
@@ -0,0 +1,967 @@
...
+class HSGPBase(BaseModel):
+    """Shared logic between HSGP and HSGPPeriodic."""
+
+    m: int = Field(..., description="Number of basis functions")
+    X: InstanceOf[TensorVariable] | InstanceOf[np.ndarray] | None = Field(
+        None,
+        description="The data to be used in the model",
+        exclude=True,
+    )
+    X_mid: float | None = Field(None, description="The mean of the data")
```
juanitorduz, line 261: I would add the mean of the **training** data

juanitorduz (approved):
I actually think this is great for a first iteration and I think we can improve this over time :)

I just left some comments on the references and some apparent dupplicated code from pymc

pymc_marketing/mmm/hsgp.py
wd60622: Not in the min pymc version that we support. How should we handle?

pymc_marketing/hsgp_kwargs.py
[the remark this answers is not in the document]
wd60622: @juanitorduz

pymc_marketing/mmm/tvp.py
wd60622: Moved to hsgp module

wd60622:
> > Is there a benefit of centered vs non-centered since sqrt_psd will be a vector and not a random vector of size (m, )? hsgp_coefs will potentially have more dims (*other_dims, m)
> 
> Sorry i missed this. Yes sometimes you might want to use the centered parameterization, though non-centered is probably the better default.

Need to double check which I am using as the default. I will set non-centered as default

wd60622:
What is the status on `drop_first`? I am getting warnings

juanitorduz:
> What is the status on `drop_first`? I am getting warnings

They will probably go after we support the latest versions (some important fixes were made)

juanitorduz:
@wd60622 what is the status on this one :) ?

wd60622:
> @wd60622 what is the status on this one :) ?

Ready to merge. Will add to example notebooks later

wd60622:
> > @wd60622 what is the status on this one :) ?
> 
> Ready to merge. Will add to example notebooks later

Later like in another PR

juanitorduz:
Merged ! Cool stuff @wd60622 !

PR #1304, opened 2024-12-20
Allow VariableFactory in FourierBase

aseyboldt:
----
📚 Documentation preview 📚: https://pymc-marketing--1304.org.readthedocs.build/en/1304/


wd60622 (commented):
Can you can line 328 to call to_dict if it exists instead of to_json

wd60622 (approved):

PR #1309, opened 2024-12-22
Add Censored wrapper for Prior class

### LINKED ISSUE #1238: Prior support for censored variables
Support for `pm.Censored` would require making unregistered variables via `Prior` API.

## API

What would the API look like?

### 1. Class that works like `Prior` which includes methods for: 

- `create_variable`
- `create_likelihood_variable`
- `sample_prior`

and attributes for: 

- dims

```python
from pymc_marketing.prior import Prior, Censored

# No dims
dist = Prior("Normal")
censored_dist = Censored(dist, upper=10)

# With dims
dist = Prior("Normal", dims="channel")
censored_dist = Censored(dist, upper=10, dims="channel")

# Hierarchical variables
dist = Prior("Normal", sigma=Prior("HalfNormal"), dims="channel")
censored_dist = Censored(dist, upper=10, dims="channel")
```

### 2. New / modified method to existing `Prior` class

This would have to be for both of the `create_variable` and `create_likelihood_variable` methods.

### 3. Add args at initialization of `Prior` class

I'm not fond of this and might clash with existing distribution parameters.

```python
dist = Prior("Normal", centered=True, lower=-10, upper=10)
```

## Logic to wrap

We ultimately need to wrap the `pm.Censored` class.

```python
with pm.Model():
    normal_dist = pm.Normal.dist(mu=0.0, sigma=1.0)
    censored_normal = pm.Censored("censored_normal", normal_dist, lower=-1, upper=1, dims=...)
```

## Details

Would child variables in hierarchical distributions be stored or be unregistered variables as well?

References: 

- PyMC Class: https://www.pymc.io/projects/docs/en/stable/api/distributions/censored.html

CC: @drbenvincent

wd60622:
Closes #1238

Related to #1296

```python
from pymc_marketing.prior import Prior, Censored

normal = Prior("Normal", dims=("channel", "geo"))
censored_normal = Censored(normal, lower=0)
```

With implemented methods: 

- `create_variable`
- `create_likelihood_variable`
- `sample_prior`
- `to_graph`
- `to_dict` 
- `from_dict`

And support from `deserialize` function

This allows it to be used in MMM. For example: 

```python

normal = Prior("Normal", sigma=Prior("HalfNormal"), dims="date")
censored_normal = Censored(normal, lower=0) 

model_config = {"likelihood": censored_normal}

mmm = MMM(
    ...,
    model_config=model_config,
)
```

CC: @drbenvincent

----
📚 Documentation preview 📚: https://pymc-marketing--1309.org.readthedocs.build/en/1309/


pymc_marketing/prior.py
```diff
@@ -1025,8 +1026,263 @@ def create_likelihood_variable(
...
+            break
+    else:
+        raise VariableNotFound(f"Variable {var.name!r} not found")
+
+    var.name = None
+    model.free_RVs.pop(index_to_remove)
+    model.named_vars.pop(name)
+
+
+@dataclass
```
juanitorduz, line 1052: What about https://docs.pydantic.dev/latest/api/base_model/ ? or even or even https://docs.pydantic.dev/latest/concepts/dataclasses/ for data validation as we are using it on other parts of the code?

juanitorduz (approved):
cool! just a suggestion about the class type :)

pymc_marketing/prior.py
wd60622: Switched in fe3d70c5

Issue #1333, opened 2025-01-03
Put `fit_result` property into ModelBuilder


wd60622:
This can be consolidated into ModelBuilder. Here it is for MMM:

https://github.com/pymc-labs/pymc-marketing/blob/10e1cbdefc0bec138177e39a5f17a1e3cb9186ef/pymc_marketing/mmm/base.py#L292-L297

And here it is for CLV:

https://github.com/pymc-labs/pymc-marketing/blob/10e1cbdefc0bec138177e39a5f17a1e3cb9186ef/pymc_marketing/clv/models/basic.py#L259-L264

This can be done like here: 

https://github.com/pymc-labs/pymc-marketing/blob/10e1cbdefc0bec138177e39a5f17a1e3cb9186ef/pymc_marketing/model_builder.py#L952-L968

sreekailash:
Hi @wd60622 - I can take it up since it was already a part of my earlier PR. Let me know if that is okay :D

sreekailash:
@wd60622  Clarification on your preference - I am planning to remove the redundant `fit_result` method from `base.py` and replace the name `posterior` with `fit_result`. Would that work or do you want me to replace `fit_result` with `posterior` everywhere? Do you have a preference?

wd60622:
> Hi @wd60622 - I can take it up since it was already a part of my earlier PR. Let me know if that is okay :D. 

Thanks @sreekailash 
Lets quickly get the other one through fit_result property should still stay. Dont replace with posterior despite both being available. The base class will implement for all child classes

wd60622:
The fit_replace is still used in many places

PR #1344, opened 2025-01-05
Move `fit_result` to `model_builder.py` and remove redundancies from CLV and MMM

sreekailash:
## Description

Have made the following changes:
1. Add `fit_result` similar to posterior in the `model_builder` since it is still used across many files
2. Added a property() setter to tackle the setter function in the `basic.py` in CLV
3.  Removed `fit_result` from MMM and CLV base 

Submitting a draft PR and appreciate any feedback. 

## Related Issue

- [ ] Closes #
- [x] Related to #1333 

## Checklist

## Modules affected

- [x] MMM
- [x] CLV
- [ ] Customer Choice

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1344.org.readthedocs.build/en/1344/

### DIFF
--- a/pymc_marketing/model_builder.py
+++ b/pymc_marketing/model_builder.py
@@ -72,16 +72,7 @@ def create_idata_accessor(value: str, message: str):
 
         return self.idata[value]
 
-    def setter(self, res: az.InferenceData) -> None:
-        if self.idata is None:
-            self.idata = res
-        elif "posterior" in self.idata:
-            warnings.warn("Overriding pre-existing fit_result", stacklevel=1)
-            self.idata.posterior = res
-        else:
-            self.idata.posterior = res
-
-    return property(accessor, setter)
+    return property(accessor)
 
 
 def create_sample_kwargs(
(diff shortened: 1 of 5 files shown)


sreekailash:
@wd60622 Created a draft PR. Let me know if these changes look good to you for this issue.

pymc_marketing/model_builder.py
```diff
@@ -72,7 +72,16 @@ def accessor(self) -> xr.Dataset:
...
 
-    return property(accessor)
+    def setter(self, res: az.InferenceData) -> None:
+        if self.idata is None:
+            self.idata = res
+        elif "posterior" in self.idata:
+            warnings.warn("Overriding pre-existing fit_result", stacklevel=1)
+            self.idata.posterior = res
+        else:
+            self.idata.posterior = res
```
wd60622, line 82: This doesnt seem general enough to have in the helper function. Id say just define once in the ModelBuilder after the helper is used for "fit_result"

wd60622 (changes requested):

pymc_marketing/model_builder.py
sreekailash: Ah! That makes sense. Let me change and send a PR!

pymc_marketing/model_builder.py
```diff
@@ -684,6 +684,42 @@ def fit(
...
+        """Get the posterior fit_result.
+
+        Returns
+        -------
+        InferenceData object.
+
+        """
+        return create_idata_accessor(
+            "posterior", "The model hasn't been fit yet, call .fit() first"
+        ).__get__(self)
```
wd60622, line 698: Why does it have to be like this? Does it not work if fit_result is using helper and the setter is defined after as normal setter?

sreekailash: It didn't work. I tried that. Which is why I did incorporate the setter within the helper function

wd60622 (changes requested):
Looking good in package. Could you move any related tests into the model_builder test suite now. For instance, if there are checks for warnings in clv tests, just move to model_builder now and remove from clv

sreekailash:
Thanks. Let me do that and also run all the unit tests once. Let me get back in a bit

sreekailash:
@wd60622 Did move two tests from CLV `test_basic.py` to `test_model_builder.py`. However, I did run unit tests and they seem to fail with an error `"pytensor.link.c.exceptions.CompileError: Compilation failed (return status=1)".`  I am sure this is related to my local dev environment and how it is setup. Seems like it is an open issue too - https://github.com/pymc-devs/pytensor/issues/991. 

TL;DR - I think the PR is okay and the errors are local env specific. Let me know if you want me to move any other test. I don't think other tests should be moved but I may be wrong. 

Screenshot 2025-01-08 at 1 17 31 PM

wd60622:
I am fixing the fixtures then the tests should pass

wd60622 (approved):
Thanks for making the change @sreekailash

PR #1357, opened 2025-01-08
Budget optimizer refactor

### LINKED ISSUE #1331: Making optimizer pymc model agnostic
With @ricardoV94 we were talking about make the budget optimizer model agnostic. Something like:

```python
from pymc_marketing.mmm.budget_optimizer import BudgetOptimizer

BudgetOptimizer(
pymc_model = model,
 . . . . 
 )
```

Ideally, we are able to hold the current work around custom constraints and utility functions, adding this feature to be used by any pymc model.

ricardoV94:
This PR refactors the budget optimizer to extract the response graph from the MMM model directly. The users don't have to pass the arguments needed to rebuild the response function, which simplifies things quite a lot and makes the optimizer able to handle arbitrarily complex models [citation needed]

There are some lingering issues. When transformations are done out-of model (as are the channel variables), we need to find them from the model. These would ideally be part of the model, so the optimizer doesn't have to worry about this.

## Breaking changes:
1. Budget optimizer API totally changed. Only need to specify mmm_model, and num_periods.
1. No more automatic summation of the response variable, so objective functions and constraints may need to be rewritten. Should be less messy after #1387 
1. In anticipation of models with more dimensions (such as hierarchical MMM), inputs and outputs to the optimizer are now xarray Datarray, instead of dictionaries as was the case before. For back-compat inputs can still be dictionaries, but outputs are not.

## TODO:
- [x] Tests of the BudgetOptimizer will need to be updated to have a hmm_model
- [x] Update NBs (?)
- [x] Update docs

Closes #1331 

----
📚 Documentation preview 📚: https://pymc-marketing--1357.org.readthedocs.build/en/1357/


juanitorduz:
yay! 🚀

ColtAllen:
Why does this have a CLV label?

wd60622:
> Why does this have a CLV label?

Seems like from file change bot. Then the force push made any clv files have no changes. I removed the clv label

ricardoV94:
Fixed the tests, NBs next

ricardoV94:
@carlosagostini do I need to worry about other functions in https://github.com/pymc-labs/pymc-marketing/blob/7bed11a61b2310835863a7d092b9b63a2ac29081/pymc_marketing/mmm/utility.py

Asking because we are no longer raveling samples / automatically summing budget dims inside the optimization code. Not sure if that could has any overlap with this change.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -109,35 +84,147 @@ class BudgetOptimizer(BaseModel):
         arbitrary_types_allowed=True,
     )
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
ricardoV94, line 87: This is not being used anywhere. Is it needed?

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -109,35 +84,147 @@ class BudgetOptimizer(BaseModel):
...
+    ) -> pt.TensorVariable:
+        """Extract the response graph, conditioned on the posterior draws and a placeholder budget variable."""
+        if not (isinstance(budgets, pt.TensorVariable)):  # and budgets.type.ndim == 1):
+            raise ValueError("budgets must be a TensorVariable")
+
+        num_periods = self.num_periods
+        model = self._pymc_model
+        posterior = self.mmm_model.idata.posterior  # type: ignore
+        max_lag = self.mmm_model.adstock.l_max
+        channel_scales = self.mmm_model.channel_transformer["scaler"].scale_
```
ricardoV94, line 129: This is very model specific, so we could revert to be passed by the user? Ideally it would be baked in in the model

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -281,16 +335,8 @@ def allocate_budget(
...
-        bounds = [
-            (
-                (budget_bounds[channel][0], budget_bounds[channel][1])
-                if channel in budget_bounds
-                else (0, total_budget)
-            )
-            for channel in self.parameters["channels"]
-        ]
+        budgets_size = np.prod(self._budget_shape)
+        initial_guess = np.ones(budgets_size) * total_budget / budgets_size
```
ricardoV94, line 339: TODO: Re-implement old initial_guess based on bounds? Hard to make it work for higher-dimensional budgets

tests/mmm/test_budget_optimizer.py
```diff
@@ -73,49 +101,7 @@
             {"channel_1": 50.0, "channel_2": 50.0},
             48.8,
         ),
-        # Add more test cases if needed
-    ],
-    ids=["default_minimizer_kwargs", "custom_minimizer_kwargs"],
-)
-def test_allocate_budget(
```
ricardoV94, line 80: I merged this with the previous test, since all the logic was the same

pymc_marketing/mmm/budget_optimizer.py
ricardoV94: The examples we have seem happy with this simple initial_guess

pymc_marketing/mmm/budget_optimizer.py
ricardoV94: This now uses `self.mmm_model._channel_scales`

ricardoV94:
99% confident tests will pass. This is ready for review

ricardoV94:
The notebook job stops at the first failure. Is there a way to have it try all the notebooks like pytest does with tests.

I thought I was done when I fixed the last failure but now it went and failed in the next one.

juanitorduz:
We will work on it 🙇!

In the meantime, i will review this one either tonight or tomorrow 🙏

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -109,35 +84,147 @@ class BudgetOptimizer(BaseModel):
...
+
+        num_periods = self.num_periods
+        model = self._pymc_model
+        posterior = self.mmm_model.idata.posterior  # type: ignore
+        max_lag = self.mmm_model.adstock.l_max
+        channel_scales = self.mmm_model._channel_scales
+
+        # Freeze all but date dims for a more succinct graph
+        coords = self._pymc_model._coords
+        model = freeze_dims_and_data(
```
ricardoV94, line 133: do this freeze only once in init

pymc_marketing/mmm/budget_optimizer.py
juanitorduz: This is to allow non-native Python types (like `mmm_model: MMM`)

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -109,35 +84,147 @@ class BudgetOptimizer(BaseModel):
...
+
+        response_variable = model[self.response_variable]
+
+        # Replicate the budget over num_periods and append zeros to also quantify carry-over effects
+        channel_data_dims = model.named_vars_to_dims["channel_data"]
+        date_dim_idx = list(channel_data_dims).index("date")
+
+        budgets_tiled_shape = list(tuple(budgets.shape))
+        budgets_tiled_shape.insert(date_dim_idx, num_periods)
+        # TODO: If scales become part of the model, we don't need to transform it here
```
juanitorduz, line 145: shall we create an issue?

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -109,35 +84,147 @@ class BudgetOptimizer(BaseModel):
...
+        budgets_tiled = pt.broadcast_to(
+            pt.expand_dims(budgets, date_dim_idx), budgets_tiled_shape
+        )
+
+        budget_full_shape = list(tuple(budgets.shape))
+        budget_full_shape.insert(date_dim_idx, num_periods + max_lag)
+        budgets_full = pt.zeros(budget_full_shape)
+        set_idxs = (*((slice(None),) * date_dim_idx), slice(None, num_periods))
+        budgets_full = budgets_full[set_idxs].set(budgets_tiled)
+        budgets_full.name = "budgets_full"
```
juanitorduz, line 156: these lines are a bit hard to read / digest . Shall we extract the to simple functions (small suggestion)

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -109,35 +84,147 @@ class BudgetOptimizer(BaseModel):
...
+    def _create_budget_variable(self) -> pt.TensorVariable:
+        size_budgets = np.prod(self._budget_shape)
+        budgets_flat = pt.tensor("budgets_flat", shape=(size_budgets,))
+        budgets = budgets_flat.reshape(self._budget_shape)
+        return budgets
+
+    def extract_response_distribution(
+        self, budgets: pt.TensorVariable
+    ) -> pt.TensorVariable:
+        """Extract the response graph, conditioned on the posterior draws and a placeholder budget variable."""
```
juanitorduz, line 121: I think a more explicit explanation about the response graph would be helpful (especially for people not very familiar with pytensor)

pymc_marketing/mmm/mmm.py
```diff
@@ -2409,15 +2381,13 @@ def optimize_budget(
         ----------
         budget : float or int
             The total budget to be allocated.
-        num_periods : float
+        num_periods : int
             The number of time units over which the budget is to be allocated.
-        budget_bounds : dict[str, list[Any]], optional
+        budget_bounds : dict[str, tuple[float, float]], optional
```
juanitorduz, line 2386: But this can be `DataArray` from the signature right?

tests/mmm/test_budget_optimizer.py
```diff
@@ -135,72 +120,97 @@ def test_allocate_budget(
...
-    with pytest.warns(UserWarning, match=match):
-        optimal_budgets, total_response = optimizer.allocate_budget(
-            total_budget, budget_bounds
-        )
-    assert optimal_budgets == pytest.approx(expected_optimal, rel=1e-2)
-    assert -total_response.fun == pytest.approx(expected_response, abs=1e-1)
+
+    # Allocate Budget
+    # match = "Using default equality constraint"
+    # with pytest.warns(UserWarning, match=match):
```
juanitorduz, line 166: do you want to remove these commented lines?

juanitorduz (commented):
Thanks @ricardoV94 ! This looks great! I left some small comments. 

I also see the budget notebook is failing maybe due to an error not being caught by the tests? 

There is another relevant notebook which uses the budget optimization: https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_case_study.html I can work ok it in a different PR :)

pymc_marketing/mmm/budget_optimizer.py
ricardoV94: I mean, do we need `model_config` in the BudgetOptimizer

pymc_marketing/mmm/budget_optimizer.py
ricardoV94: Done: https://github.com/pymc-labs/pymc-marketing/issues/1383

Will remove inline comment

pymc_marketing/mmm/budget_optimizer.py
juanitorduz: This is needed as we are using a Pydantic `BaseModel` to do type-checking

pymc_marketing/mmm/budget_optimizer.py
juanitorduz: seems areasonable

pymc_marketing/mmm/budget_optimizer.py
ricardoV94: Done

pymc_marketing/mmm/budget_optimizer.py
ricardoV94: Done

tests/mmm/test_budget_optimizer.py
ricardoV94: No, we can keep them

pymc_marketing/mmm/mmm.py
ricardoV94: Yes, fixed

juanitorduz:
Thanks for the changes @ricardoV94 ! 

I think the only things missing are the notebooks. Do you want to work on that or do you wanna support so that so can focus on doing your magic ;) ) ?

ricardoV94:
I'm on it! I will push soon (and revert the merge @cetagostini just did). I appreciate another deep review as I changed a couple things to make it more clean

juanitorduz:
> I'm on it! I will push soon (and revert the merge @cetagostini just did). I appreciate another deep review as I changed a couple things to make it more clean

Absolutely 💪 !

ricardoV94:
The logic used here to extract the `response_distribution` could be easily repurposed to simplify/speedup `get_channel_contributions_forward_pass_grid`.

I noticed this was rather slow while checking if the mmm case study notebook is now passing :D

ricardoV94:
@cetagostini I did something simple for the dict->DataArray, for `sample_response_distribution` which is user facing, so it still accepts a dict:

```python
        if isinstance(allocation_strategy, dict):
            # For backward compatibility
            allocation_strategy = DataArray(
                pd.Series(allocation_strategy), dims=("channel",)
            )
```
It seems like you are reinventing a way to define DataArray, perhaps better to push users to be comfortable with them, since those are the objects we work with outside of PyMC/PyTensor models?

ricardoV94:
@cetagostini I'm not against it though. That can possibly be done as a separate PR at this point, I guess?

cetagostini:
I think the approach you added works, and push users to get familiar. No against, I usually try to create helpers anyway, up to the users if want to get familiar and not use it, or use it. But I'm okay to address on maybe in other PR or #1538

ricardoV94:
Tests are passing!

juanitorduz (approved):

juanitorduz:
Wow! That was fast!!! Thanks!

PR #1414, opened 2025-01-20
Remove sample_posterior_predictive_patch

### LINKED ISSUE #1394: MLflow sample_posterior_predictive only works on in-sample data
This will break on out of sample since self.y is used. sample_posterior_predictive doesn't have a y argument. Likely will have to remove from autologging

https://github.com/pymc-labs/pymc-marketing/blob/ed354c3500185c4ef98bd04566f8dde9567e499c/pymc_marketing/mlflow.py#L994-L997

What are you thoughts on the behavior here? @juanitorduz

I am leaning the we remove the patched `sample_posterior_predictive` method and document the `log_mmm_evaluation_metrics` function specifically

wd60622:
## Description

Remove the patch of `sample_posterior_predictive` since it only works for in-sample. 

There is some docs added from previous PR: 

https://github.com/pymc-labs/pymc-marketing/blob/9ccb221e91518b4b4aa6f42557afe4e50cea9ecc/pymc_marketing/mlflow.py?plain=1#L499-L509

## Related Issue

- [ ] Closes #1394 
- [ ] Related to #

## Checklist

## Modules affected

- [ ] MMM
- [ ] CLV
- [ ] Customer Choice

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1414.org.readthedocs.build/en/1414/


tests/test_mlflow.py
```diff
@@ -583,18 +583,23 @@ def test_log_mmm_evaluation_metrics() -> None:
     """Test logging of summary metrics to MLflow."""
     y_true = np.array([1.0, 2.0, 3.0])
     y_pred = np.array([[1.1, 2.1, 3.1]]).T
-    custom_metrics = ["r_squared", "rmse", "mae", "mape", "nrmse", "nmae"]
+    custom_metrics = ["r_squared", "rmse"]
```
juanitorduz, line 586: why this change?

wd60622: I thought it was calculating all of the ones despite provides custom_metrics. Only the two makes it a bit more obvious that we are not using *all*. With the longer list, I was going back and forth

PR #1437, opened 2025-01-26
[DOC] improvement roas efficiency plot. mmm case study

juanitorduz:
Similar to https://github.com/pymc-labs/pymc-marketing/pull/1436

----
📚 Documentation preview 📚: https://pymc-marketing--1437.org.readthedocs.build/en/1437/


wd60622 (approved):
Looks good

Thought: what if we just change the csv file to have the sorted columns and nice column names so none of processing happen in notebook?

juanitorduz:
>Thought: what if we just change the csv file to have the sorted columns and nice column names so none of processing happen in notebook?

Good point! Still, I want to make sure the data is exactly the same as the original lightweight MMM example (pydata talk) I took the example from :)

PR #1450, opened 2025-02-01
migrate predict parameter X_pred to X

wd60622:
## Description

The predictive methods now use X instead of X_pred

Should use some deprecation warnings for the previous keyword arguments?

CC: @cetagostini

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -1902,7 +1902,7 @@ class MMM(
 
     def sample_posterior_predictive(
         self,
-        X_pred,
+        X,
         extend_idata: bool = True,
         combined: bool = True,
         include_last_observations: bool = False,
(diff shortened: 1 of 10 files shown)


pymc_marketing/mmm/mmm.py
```diff
@@ -1902,7 +1902,7 @@ def legend_title_func(channel):
 
     def sample_posterior_predictive(
         self,
-        X_pred,
```
cetagostini, line 1905: Hey hey! 

I like this more, `X_pred` its confusing but definitely we need deprecation warnings and continue supporting `X_pred` maybe for at least one extra release (?).

With that backward compatibility, would be enough for my side to approve changes!

cetagostini (changes requested):
Really like this, adding backwards compatible should be enough to merge!

juanitorduz (approved):

PR #1463, opened 2025-02-03
remove abstract requirement of data_setter

wd60622:
## Description

I've removed the implementation for the CLV classes since the `sample_posterior_predictive` methods are not used in those models.

CC: @ColtAllen

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1463.org.readthedocs.build/en/1463/


wd60622:
Looking at how the customer choice model uses this, I am second guessing. For instance, if there is in sample posterior predictive then this wouldn't really make sense. 

However, X is always required for the method at the moment. So maybe it makes sense... 

What are people thoughts?

ColtAllen:
> I've removed the implementation for the CLV classes since the `sample_posterior_predictive` methods are not used in those models.

`pymc.sample_posterior_predictive` is being called for the `plot_expected_purchases_ppc` function. I haven't tested the ModelBuilder variant yet, but given how it's currently handling X data, CLV support might be possible.

wd60622:
> `pymc.sample_posterior_predictive` is being called for the `plot_expected_purchases_ppc` function. I haven't tested the ModelBuilder variant yet, but given how it's currently handling X data, CLV support might be possible.

`sample_posterior_predictive` method and not the `pymc` function

juanitorduz (approved):

Issue #1476, opened 2025-02-07
broken pre-commit ci


williambdean:
I know there is an open source one that we can run like any other action

Thoughts @juanitorduz

juanitorduz:
You mean like https://github.com/pymc-labs/pymc-marketing/pull/1477 ?

This won't update the pre-commit versions

juanitorduz:
We just merged the pre-commit check so we can bypass the failing one. Let's wait a week and see if it gets fixed. We had a similar issue last year. Maybe we re evaluate the approach depending on this ?

juanitorduz:
Ahg! It is mypy being too slow https://github.com/pymc-labs/pymc-marketing/pull/1478/commits/c6026665e0a7758e63e6b16842d86457e6b06022!

wd60622:
The pre-commit hok for `no-commit-to-branch` fails upon with runs on the main. A solution might be to

Add `always_run: false` or to not run the action on the main branch since they got merged anyway. Thoughts?

juanitorduz:
yes! try what it works! the idea is to make sure it runs locally but not on a merge commit. I tried to fix it but failed :(

anevolbap:
Looks resolved: #1977 removed the `pre-commit/action` workflow that ran on every push, including `main`, and dropped `skip: [mypy]`. pre-commit.ci passes on every push to `main` over the last 60 commits and on the last 12 autoupdate PRs. OK to close?

PR #1480, opened 2025-02-07
skip mypy in pre-commit ci

juanitorduz:
Related to https://github.com/pymc-labs/pymc-marketing/issues/1476

MyPy takes too much and we can use the action in https://github.com/pymc-labs/pymc-marketing/pull/1477 to check for it.
We can still use pre-commit ci for auto fix and revision update.

----
📚 Documentation preview 📚: https://pymc-marketing--1480.org.readthedocs.build/en/1480/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,6 @@
 ci:
   autofix_prs: false
+  skip: [mypy]
 
 repos:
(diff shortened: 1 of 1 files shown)


wd60622 (approved):

wd60622:
Thanks for looking into!

PR #1498, opened 2025-02-13
excluded _prior suffix from CLV models

### LINKED ISSUE #1428: Exclude prior suffix from CLV model config
The MMM and customer choice models don't have this suffix. All CLV models have this convention. Here is an example:

https://github.com/pymc-labs/pymc-marketing/blob/d9f0c5140c681bd9d5214e3c8d552186016aac96/pymc_marketing/clv/models/beta_geo.py?plain=1#L169-L172

I think the "prior" is implied in the context of the model so I find it a bit redundant. Then the variable drops the suffix in the model.

What are your thoughts @ColtAllen?

arthurmello:
## Description
Removed the `_prior` suffix from parameters of CLV model config + adjacent files (tests, notebooks, etc.)

## Related Issue
- [x] Closes #1428 
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1498.org.readthedocs.build/en/1498/


wd60622 (commented):
Nice. Thanks for the PR @arthurmello 

We should provide a deprecation for this. Likely in the clv/basic class. Maybe we can check for any model_config keys with "_prior" suffix then remove while having a warning

arthurmello:
Sure, do you mean a test that ensures that models saved with the old configuration can still be loaded?
Something like:
```py
def test_backward_compatibility_with_old_config(self, mocker):
        old_model_config = {
            "alpha_prior": ...
            "r_prior": ...
        }
        model = CLVModelTest(model_config=old_model_config)

        mocker.patch("pymc.sample", mock_sample)
        model.fit(...)
        model.save("old_model_config_test")

        loaded_model = CLVModelTest.load("old_model_config_test")

        assert loaded_model.model_config == {
            "alpha": ...,
            "r": ...
        }

```

wd60622:
Yeah, something like that. Dont need to sample. Just load and make sure that a method that uses the posterior still works

tests/clv/models/test_basic.py
```diff
@@ -234,3 +234,19 @@ def test_model_config_warns(self) -> None:
         assert model.model_config == {
             "x": Prior("StudentT", mu=0, sigma=5, nu=15),
         }
+
+    def test_backward_compatibility_with_old_config(self):
+        old_model_config = {
+            "alpha_prior": Prior("Weibull", alpha=2, beta=10),
+            "r_prior": Prior("Weibull", alpha=2, beta=1),
+        }
+        model = CLVModelTest(model_config=old_model_config)
```
wd60622, line 243: Let's test that there is a warning

tests/clv/models/test_basic.py
```diff
@@ -234,3 +234,19 @@ def test_model_config_warns(self) -> None:
...
+        model.fit(tune=0, chains=2, draws=5)
+        model.save("old_model_config_test")
```
wd60622, line 245: This isn't testing for loading from an *old* saved model. The idata.posterior will have the suffix. We want to check for that suffix in the posterior will not break using the model in the future

wd60622 (approved):
Looks good to me. Thanks for the PR!

@ColtAllen Can you take a look as well

ColtAllen:
> Looks good to me. Thanks for the PR!
> 
> @ColtAllen Can you take a look as well

I've started looking at this; will post my review tomorrow.

tests/clv/models/test_basic.py
```diff
@@ -234,3 +234,32 @@ def test_model_config_warns(self) -> None:
...
+        assert "alpha_prior" not in loaded_model.idata.posterior
+
+        os.remove("test_model")
+
+    def test_deprecation_warning_on_old_config(self):
+        old_model_config = {
+            "x_prior": {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}}
+        }
+        with pytest.warns(
+            DeprecationWarning, match="The key 'x_prior' in model_config is deprecated"
```
ColtAllen, line 261: Looks like this test needs to be updated:

```python
with pytest.warns(
            DeprecationWarning, match="The key 'x_prior' in model_config is deprecated and will be removed in future versions. Use 'x' instead.",
```

pymc_marketing/clv/models/basic.py
```diff
@@ -279,12 +292,28 @@ def _build_with_idata(cls, idata: az.InferenceData):
...
+
         model.idata = idata
+        model._rename_posterior_variables()
+
         model.build_model()  # type: ignore
         if model.id != idata.attrs["id"]:
             raise ValueError(f"Inference data not compatible with {cls._model_type}")
         return model
 
+    def _rename_posterior_variables(self):
```
ColtAllen, line 304: Let's add a TODO to remove this in a future release. I'd say 2025-Q4 or early 2026 will give users enough time to update their codebases.

pymc_marketing/clv/models/basic.py
```diff
@@ -47,6 +47,19 @@ def __init__(
         non_distributions: list[str] | None = None,
     ):
         model_config = model_config or {}
+
+        deprecated_keys = [key for key in model_config if key.endswith("_prior")]
```
ColtAllen, line 51: Let's add a TODO to remove these codelines in a future release. I'd say 2025-Q4 or early 2026 will give users enough time to update their codebases.

ColtAllen (changes requested):
Looks good; just some suggestions to update a unit test and add some removal reminders in addition to @wd60622's requested changes.

pymc_marketing/clv/models/basic.py
wd60622: Let's do an issue instead 🙏

tests/clv/models/test_basic.py
arthurmello: I put "match" + only the first part of the message so that it's more robust to future changes: if the second part of the warning message has to change (ex.: we change 'x' to something else) the test still works, since it looks for a partial match.
What do you think?

pymc_marketing/clv/models/basic.py
arthurmello: I'd go with the issue too (I've created it here: #1516)

tests/clv/models/test_basic.py
ColtAllen: We'll be removing this test within a year, so it can proceed as-is.

ColtAllen (approved):
Thanks for creating the issue for follow-up work!

PR #1510, opened 2025-02-18
Use SoftPlusHSGP in multidimensional HSGP

wd60622:
## Description

This might still differ from the other implementation where the HSGP intercept has a baseline which is then multiplied by HSGP:

https://github.com/pymc-labs/pymc-marketing/blob/89c9e973df66beb7548708abf9f25079b4f6f342/pymc_marketing/mmm/mmm.py?plain=1#L482-L498

Another alternative is to build that part into the HSGP model itself. However, the mmm would require another prior for the intercept. Which I think is good to keep separate. We can always wrap that logic into a separate function like in `MuEffect`

https://github.com/pymc-labs/pymc-marketing/blob/89c9e973df66beb7548708abf9f25079b4f6f342/pymc_marketing/mmm/multidimensional.py?plain=1#L58-L68

Thoughts?

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1510.org.readthedocs.build/en/1510/


pymc_marketing/mmm/multidimensional.py
```diff
@@ -1057,7 +1057,7 @@ def build_model(
                     dims=("date", *self.dims, "channel"),
                 )
 
-                media_latent_process = HSGP.parameterize_from_data(
+                media_latent_process = SoftPlusHSGP.parameterize_from_data(
```
juanitorduz, line 1060: @wd60622 can you please explain the motivation behind the softplus transform? 🙏

wd60622: - Aligns the default hsgp with other MMM class
- remove negative channel contributions with time varying media = True

juanitorduz: Thanks @wd60622 ! Would you mind running the time-varying parameter notebooks affected by this change just to control the expected result? 🙏

wd60622: Is there a multidimensional mmm notebook already? This is just the new class

juanitorduz: Right, but in theory you could reproduce the results of the MMM basic class with this object right? The intention is not to update the docs but maybe have some experiments on the effect of this parametrization. I think you can just share some plots or experiments 9for the sake of illustrating the benefits) :) Just for future reference. It is not a blocker, but more to have a bit more intuition WDYT?

wd60622: Sure. I can update 
However, there is still slight difference in the intercept which I noted. Should that be consolidated? 

Thoughts @cetagostini

wd60622:
```python

import pandas as pd
import pymc as pm

from pymc_marketing.mmm import SoftPlusHSGP

n_dates = 52 * 2

dates = pd.date_range("2025-01-01", periods=n_dates, freq="W-MON")

X = (dates - dates[0]).days.values

hsgp = SoftPlusHSGP.parameterize_from_data(X, ls_lower=10, ls_upper=25, dims="date")

coords = {"date": dates}
with pm.Model(coords=coords) as model:
    name = "intercept"
    intercept_baseline = pm.Gamma(f"{name}_baseline", mu=3, sigma=1)
    intercept_moving = hsgp.create_variable(f"{name}_multiplier")

    intercept = pm.Deterministic(
        name,
        intercept_baseline * intercept_moving,
        dims=hsgp.dims,
    )

    idata = pm.sample_prior_predictive()


prior = idata.prior

from pymc_marketing.plot import plot_curve

prior["intercept"].pipe(plot_curve, {"date"})
plt.show()
```

juanitorduz (approved):
Thanks for the example @wd60622 !

wd60622:
Realizing that the current implementation does that have baseline and multiplier concept 

https://github.com/pymc-labs/pymc-marketing/blob/b9bbcd416eb57dc14deb51a71ffacda5e4b1dd76/pymc_marketing/mmm/multidimensional.py?plain=1#L1030-L1044

So merging this now!

PR #1562, opened 2025-03-14
Specify target scaling dimensions

wd60622:
## Description

This allows a user to specify the `scaling` parameter 

```python
from pymc_marketing.mmm.multidimensional import MMM

mmm = MMM(..., 
    dims=("DMA", "product"),
    scaling={
        "target": {
            # Take max through DMA (and date) 
            "method": "max",
            "dims": ("DMA", )
        },
    },
)
```

The dims are where the operation is done through. So specifying dims = DMA results in DMA not being included. Maybe the parameterization of this 
should be the opposite, the dims are the ones that will be result `mmm.scalers`

----
📚 Documentation preview 📚: https://pymc-marketing--1562.org.readthedocs.build/en/1562/

### DIFF
--- a/pymc_marketing/mmm/multidimensional.py
+++ b/pymc_marketing/mmm/multidimensional.py
@@ -196,7 +196,11 @@ def create_event_mu_effect(
 
 
 class VariableScaling(BaseModel):
-    """How to scale a variable."""
+    """How to scale a variable.
+
+    The scaling through the dimension of 'date' is assumed and doesn't need to be specified.
+
+    """
 
     method: Literal["max", "mean"] = Field(..., description="The scaling method.")
     dims: str | tuple[str, ...] = Field(
@@ -219,7 +223,25 @@ class VariableScaling(BaseModel):
 
 
 class Scaling(BaseModel):
-    """Scaling configuration for the MMM."""
+    """Scaling configuration for the MMM.
+
+    Examples
+    --------
+    Scale the target variable by max value by group of 'DMA'
+
+    .. code-block:: python
+
+        from pymc_marketing.mmm.multidimensional import Scaling
+
+        scaling = Scaling(**{
+            "target": {
+                "method": "max",
+                # Exclude 'DMA' from dims here.
+                "dims": (),
+            },
+        })
+
+    """
 
     target: VariableScaling = Field(
         ...,
--- a/pymc_marketing/mmm/multidimensional.py
+++ b/pymc_marketing/mmm/multidimensional.py
@@ -26,8 +26,10 @@ import pandas as pd
 import pymc as pm
 import pytensor.tensor as pt
 import xarray as xr
+from pydantic import BaseModel, Field, model_validator
 from pymc.model.fgraph import clone_model as cm
 from pymc.util import RandomState
+from typing_extensions import Self
 
 from pymc_marketing.mmm import SoftPlusHSGP
 from pymc_marketing.mmm.components.adstock import (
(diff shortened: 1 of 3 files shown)


wd60622:
Maybe something like

```python
MMM(
    scaling={
        "target": {
            "method": "max",
            "dims": ("DMA", )
        },
    }

**Comment by wd60622** [comment]
Should dims mean the dimensions that the scalers will be? For instance, (DMA, product) would mean that the scalers will be DMA and product . Thoughts on this?

**Comment by cetagostini** [comment]
> Should dims mean the dimensions that the scalers will be? For instance, (DMA, product) would mean that the scalers will be DMA and product . Thoughts on this?

xarray works in that way and I feel better to stick to xarray behavior to not cause confusion.

**Comment by juanitorduz** [comment]
Agree! :) (Nice feature!)

File: tests/mmm/test_multidimensional.py [hunk 1]
```diff
@@ -573,3 +573,34 @@ def test_check_for_incompatible_dims(adstock, saturation, dims) -> None:
...
+        saturation=LogisticSaturation(),
+        scaling=scaling,
+        date_column="date",
+        target_column="target",
+        channel_columns=["channel_1", "channel_2"],
+        dims=("country",),
+    )
+    mmm.fit(X, y)
+    assert mmm.xarray_dataset._target.dims == ("date", "country")
+    assert mmm.scalers._target.dims == ("country",)
```
juanitorduz, line 592: Could we explicitly test tthe scaling (e.g take max for the desired dims)

juanitorduz (commented):
Great feature! I suggest adding some additional tests to test the scaling effect explicitly 🙏

wd60622:
Sounds good!

tests/mmm/test_multidimensional.py
wd60622: I've added this in with 44087a63

cetagostini:
Default scalings are the same, right?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -264,6 +299,18 @@ def __init__(
         dims = dims if dims is not None else ()
         self.dims = dims
 
+        if isinstance(scaling, dict):
+            scaling = Scaling(**scaling)
+
+        self.scaling = scaling or Scaling(
+            target=VariableScaling(method="max", dims=self.dims)
```
wd60622, line 306: This was the default before, right?

wd60622: This is is the global max of the target. Regardless of the number of groups

cetagostini: Indeed

cetagostini: Do you have date under the hood? If so, should be the same, before was `self.xarray_dataset.max(dim=["date", *self.dims])`

wd60622: Yes, date is used under the hood. User doesn't specify

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1013,9 +1090,7 @@ def build_model(
             _target = pm.Data(
                 name="target_data",
                 value=(
-                    self.xarray_dataset._target.sum(dim="target")
-                    .transpose("date", *self.dims)
-                    .values
+                    self.xarray_dataset._target.transpose("date", *self.dims).values
```
cetagostini, line 1093: Dimension target is delete or I can access to it under `_target` variable in `xarray_dataset`?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1210,29 +1291,31 @@ def _posterior_predictive_data_transformation(
...
                 np.zeros(
                     (
                         X[self.date_column].nunique(),
                         *[len(self.xarray_dataset.coords[dim]) for dim in self.dims],
-                        1,
                     ),
                     dtype=np.int32,
                 ),
-                dims=("date", *self.dims, "target"),
+                dims=("date", *self.dims),
```
cetagostini, line 1315: Same question here about target dim.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -193,6 +195,60 @@ def set_data(self, mmm: MMM, model: pm.Model, X: xr.Dataset) -> None:
     return Effect()
 
 
+class VariableScaling(BaseModel):
```
cetagostini, line 198: No need to do it now, but should we move this to another module? Maybe `scalers.py` or similar?

cetagostini (approved):
Everything looks good, a few questions but still nice for merge!

pymc_marketing/mmm/multidimensional.py
wd60622: Did doesnt exist anymore because it was a dummy single value

pymc_marketing/mmm/multidimensional.py
wd60622: Done

PR #1565, opened 2025-03-16
Permit `BudgetOptimizer.allocate_budget()` to take `x0` as an argument

### LINKED ISSUE #1564: `BudgetOptimizer.allocate_budget()` doc string doesn't specify default value of x0
The `BudgetOptimizer.allocate_budget()` doc string includes the argument `minimize_kwargs` which is described as consisting of "Extra kwargs for `scipy.optimize.minimize`".  The "extra" is because the `BudgetOptimizer.allocate_budget()` argument `budget_bounds` gets passed to `scipy.optimize.minimize` (though it is processed first. No mention of this in the docstring).  But the `x0` argument is also passed into `scipy.optimize.minimize` but there is no mention of this.  As a result, the following fails:

```py
BudgetOptimizer.allocate_budget(total_budget=100_000, x0=[50_000, 50_000])` 
```
with the following error:

```py
TypeError: scipy.optimizer_minimize.minimize() got multiple values for keyword argument `x0`
```

Docstring should be updated or `BudgetOptimizer.allocate_budget()` should handle `x0` as an argument.

cluhmann:
## Description
`BudgetOptimizer.allocate_budget()` now accepts `x0` as an argument.

## Related Issue

- [x] Closes #1564 
- [ ] Related to #

## Checklist

`x0` is now accepted via the `minimize_kwargs` (as the docstring implies is possible).  I looked into adding a corresponding test, but these lines in `tests/mmm/test_budget_optimizer::test_allocate_budget_custom_minimize_args()` confuse me.  Any guidance?

```py
kwargs = minimize_mock.call_args_list[0].kwargs

np.testing.assert_array_equal(x=kwargs["x0"], y=np.array([50.0, 50.0]))
assert kwargs["bounds"] == [(0.0, 50.0), (0.0, 50.0)]
assert kwargs["method"] == minimize_kwargs["method"]
assert kwargs["options"] == minimize_kwargs["options"]
```

----
📚 Documentation preview 📚: https://pymc-marketing--1565.org.readthedocs.build/en/1565/


pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -466,21 +474,19 @@ def allocate_budget(
...
-        initial_guess = np.ones(budgets_size) * (total_budget / budgets_size)
-        initial_guess = initial_guess.astype(self._budgets_flat.type.dtype)
-
-        if minimize_kwargs is None:
-            minimize_kwargs = self.DEFAULT_MINIMIZE_KWARGS.copy()
-        else:
-            # Merge with defaults (preferring user-supplied keys)
-            minimize_kwargs = {**self.DEFAULT_MINIMIZE_KWARGS, **minimize_kwargs}
+        # 5. Construct the initial guess (x0) if not provided
+        if "x0" not in minimize_kwargs:
```
ricardoV94, line 478: I don't think x0 should be hidden in minimize kwargs, I would make it an explicit argument of `allocate_budget`

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -466,21 +474,19 @@ def allocate_budget(
...
+            minimize_kwargs["x0"] = np.ones(budgets_size) * (
+                total_budget / budgets_size
+            )
+            minimize_kwargs["x0"] = minimize_kwargs["x0"].astype(
```
ricardoV94, line 482: I would check if it has the expected shape `(budgets_size,)`, as the compiled functions have `trust_input=True` and won't check if x0 has the promised shape

cetagostini: very good point here!

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -409,8 +413,16 @@ def allocate_budget(
         MinimizeException
             If the optimization fails for any reason, the exception message will contain the details.
         """
+        # set total budget
         self._total_budget.set_value(np.asarray(total_budget, dtype="float64"))
 
+        # coordinate user-provided and default minimize_kwargs
+        if minimize_kwargs is None:
+            minimize_kwargs = self.DEFAULT_MINIMIZE_KWARGS.copy()
```
ricardoV94, line 421: Nit: no real need to copy
```suggestion
            minimize_kwargs = self.DEFAULT_MINIMIZE_KWARGS
```

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -391,8 +392,11 @@ def allocate_budget(
             - If None, default bounds of [0, total_budget] per channel are assumed.
             - If a dict, must map each channel to (low, high) budget pairs (only valid if there's one dimension).
             - If an xarray.DataArray, must have dims (*budget_dims, "bound"), specifying [low, high] per channel cell.
+        x0 : DataArray, optional
```
ricardoV94, line 395: I think it should be a np.ndarray not a DataArray

cluhmann:
Just need to write some tests.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -466,21 +478,26 @@ def allocate_budget(
...
+            x0 = np.ones(budgets_size) * (total_budget / budgets_size).astype(
+                self._budgets_flat.type.dtype
+            )
+        # if x0 arg is provided, validate shape
+        elif x0.shape != (budgets_size,):
+            raise ValueError(
+                f"""The shape of 'x0' {x0.shape} does not match the expected shape {(budgets_size,)}."""
+            )
+        # if x0 arg is provided, validate dtype
+        elif np.issubdtype(x0.dtype, np.number):
```
ricardoV94, line 492: No specifically the `x0.dtype` must match `self._budgets_flat.type.dtype`. I think there's a shortcut you can use for the shape and dtype validation: `x0 = self.budget_flat.type.filter(x0)`, which will do any allowed casting or raise if things cannot be safely cast (and also checks shape)

cluhmann:
Thank you!  If no one else wants to merge this, I will at some point soon.

williambdean:
> Thank you! If no one else wants to merge this, I will at some point soon.

do the honors!

PR #1521, opened 2025-02-25
MMM Multi-Dimensional Example

juanitorduz:
Intro notebook to the new (experimental) `MMM` class.

----
📚 Documentation preview 📚: https://pymc-marketing--1521.org.readthedocs.build/en/1521/

PR #1581, opened 2025-03-24
Allow original scales in saturation_curves_scatter

PabloRoque:
## Description

Allows original scales in `MMMPlotSuite.saturation_curves_scatter`
- Introduces a boolean in `saturation_curves_scatter(..., original_scale: bool)`.

## Related Issue

- [ ] Closes #
- [x] Related to #1521 

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1581.org.readthedocs.build/en/1581/


pymc_marketing/mmm/multidimensional.py
```diff
@@ -922,6 +922,31 @@ def get_scales_as_xarray(self) -> dict[str, xr.DataArray]:
             "target_scale": self.scalers._target,
         }
 
+    def get_scales_as_dataset(self) -> dict[str, xr.DataArray]:
```
williambdean, line 925: Can we just use a Dataset here?

PabloRoque: Messed the type hint, sorry.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -922,7 +922,7 @@ def get_scales_as_xarray(self) -> dict[str, xr.DataArray]:
             "target_scale": self.scalers._target,
         }
 
-    def get_scales_as_dataset(self) -> dict[str, xr.DataArray]:
+    def get_scales_as_dataset(self) -> xr.Dataset:
```
williambdean, line 925: Is this different than `self.scalers`?

PabloRoque: Just different labels: `_channel` -> `channel_scale` and `_target` -> `target_scale`. 

Happy to use directly `self.scalers` and avoid clogging the new API already 😅

cetagostini: Sorry, maybe I'm late to the party but if the user add the variable original scale to the model, maybe its better to allow for `original_scale=True` and then replace by `channel_contribution_original_scale` adding the last by string concatenation.

This feels more like the way to go based on the API construction instead of share scalers, which are not needed because the model is scaling everything.

pymc_marketing/mmm/plot.py
```diff
@@ -372,6 +377,13 @@ def saturation_curves_scatter(self, **kwargs) -> tuple[Figure, NDArray[Axes]]:
         # Create subplots
         fig, axes = self._init_subplots(n_subplots=n_rows, ncols=n_columns, **kwargs)
 
+        # Channel in original_scale if selected
+        channel_contribution = (
+            "channel_contribution_original_scale"
+            if original_scale
+            else "channel_contribution"
```
cetagostini, line 384: Yes, I feel this looks like the way. Almost none adjustment to add the original, scale!

pymc_marketing/mmm/plot.py
```diff
@@ -381,8 +393,8 @@ def saturation_curves_scatter(self, **kwargs) -> tuple[Figure, NDArray[Axes]]:
 
                 # Select X data (constant_data)
                 x_data = self.idata.constant_data.channel_data.sel(**indexers)
-                # Select Y data (posterior contributions)
-                y_data = self.idata.posterior.channel_contribution.sel(**indexers)
+                # Select Y data (posterior contributions) and scale if needed
+                y_data = self.idata.posterior[channel_contribution].sel(**indexers)
```
cetagostini, line 397: Could we add a unit test to check this? With a minor plot test should be enough.

PabloRoque: Added. The method also raises ValueError if the deterministic is not present, and advices the user to introduce it.

tests/mmm/test_mmmplotsuite.py
```diff
@@ -0,0 +1,110 @@
+#   Copyright 2022 - 2025 The PyMC Labs Developers
```
cetagostini, line 1: Amazing, thanks for creating the test, all good from my side!

cetagostini (approved):
Amazing, thanks @PabloRoque

PR #1590, opened 2025-03-25
minor improvements

juanitorduz:
- Typos
- Remove redundant t feature from the sample dataframe

----
📚 Documentation preview 📚: https://pymc-marketing--1590.org.readthedocs.build/en/1590/


PabloRoque (commented):
Diffs are not rendering, but it everything looks great.

Only minor typo on `# Posterior Predictive Checks`. Notice: 

"""
We can now inspect the posterior predictive distribution. **AS** before, we need to scale the posterior predictive to the original scale to make it comparable to the data.
"""

The AS should be changed to **As**

PabloRoque (approved):

juanitorduz:
Thank you @PabloRoque !

PR #1603, opened 2025-03-29
Fix test to suppress warning on allocate_budget_to_maximize_response

caiquefortunato:
## Description
This PR updates a test in test_mmm.py to suppress a warning emitted by allocate_budget_to_maximize_response when noise_level is not set appropriately. 

## Related Issue

- [ ] Closes #
- [ ] Related to #1598 

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [x] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [x] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--1603.org.readthedocs.build/en/1603/

Issue #1598, opened 2025-03-29
Use `optimize_budget` method instead in tests


williambdean:
This warns and should be removed from tests unless it is trying to check for the warning

https://github.com/pymc-labs/pymc-marketing/blob/2d8bd16564b11bb4857fbad0cb8c5e0edb31c591/tests/mmm/test_mmm.py?plain=1#L583

LeonxLJX:
Hi! I'd like to work on this issue. Please let me know if I can be assigned. Thanks!

LeonxLJX:
Hi! I'd like to update the tests to use the optimize_budget method. Thanks!

LeonxLJX:
Hi! I'd like to replace the deprecated `optimize_budget` calls in `tests/mmm/test_mmm.py` (around line 583) with the current API, unless the test is explicitly asserting the deprecation warning. Small test cleanup. May I be assigned?

anevolbap:
The deprecated call at the linked line was removed together with its tests in #1637, and the MMM tests now call `optimize_budget` (12 calls in `tests/mmm/test_budget_optimizer_mmm.py`). Close as completed?

PR #1604, opened 2025-03-29
Fix/replace deprecated budget method

caiquefortunato:
## Description
This PR updates the test test_allocate_budget_to_maximize_response_bad_noise_level by replacing the usage of the deprecated method allocate_budget_to_maximize_response with the recommended optimize_budget method.

This aligns the codebase with the current API and avoids suppressing DeprecationWarnings, as discussed in [issue #1598](https://github.com/pymc-labs/pymc-marketing/issues/1598) and the feedback on [PR #1603](https://github.com/pymc-labs/pymc-marketing/pull/1603).

## Related Issue

- [ ] Closes #
- [ ] Related to #1598 

## Checklist

- [ x ] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [ x ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ x ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

### DIFF
--- a/tests/mmm/test_mmm.py
+++ b/tests/mmm/test_mmm.py
@@ -19,7 +19,6 @@ import pandas as pd
 import pymc as pm
 import pytest
 import xarray as xr
-import warnings
 from matplotlib import pyplot as plt
 
 from pymc_marketing.mmm.budget_optimizer import optimizer_xarray_builder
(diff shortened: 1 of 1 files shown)


cetagostini (approved):
Hey @caiquefortunato thank you for the PR. Everything looks good 👍

cetagostini:
Let's wait for the test to pass before merge but code looks okay!

tests/mmm/test_mmm.py
```diff
@@ -571,21 +571,17 @@ def test_allocate_budget_to_maximize_response_bad_noise_level(
...
-        time_granularity = "weekly"
         budget_bounds = optimizer_xarray_builder(
             value=[[0.5, 1.2], [0.5, 1.5]],
             channel=["channel_1", "channel_2"],
             bound=["lower", "upper"],
         )
-        noise_level = "bad_noise_level"
 
         with pytest.raises(ValueError, match="noise_level must be a float"):
-            mmm_fitted.allocate_budget_to_maximize_response(
```
cetagostini, line 583: Sorry, reading this better. Here we are doing the wrong fix. We should:
1. Replace all the suite (not only this test) to work with the new class.
2. We need to test the same scenarios, meaning, you need to test `optimize budget` and `sample_response_distribution`
3. No parameters should be delete, the test should be handle differently.
4. We need test for the method to be deprecated, and the new one. This because we still will hold the old method for at least one more release. So, best way should be to have test for both, and be sure non of them is doing something funny. IF we don't want to do this then we should deprecated already the old method. @williambdean Whats your take?

cetagostini (changes requested):
Sorry, I came back from work and see that the current solutions its not enough. Would be great, and we can support you if you need a hand to work on it.

PR #1611, opened 2025-04-01
Add Example Gallery

juanitorduz:
Add an example gallery to harmonize the example layout in PyMC and Pytensor. I followed https://github.com/pymc-devs/pytensor/pull/1150 from @jessegrabowski (thank you!) and some help from Sonnet ;) 

I also added some subsections for the MMM docs to make it easy to access the material.

This is a snapshot: 

image


williambdean:
Are thumbnails required? What happens if one isnt included?

scripts/generate_gallery.py
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env python3
+"""
+Sphinx plugin to generate a gallery for notebooks
+
+Modified from the PyTensor project, which was modified from the pymc project,
+which modified the seaborn project, which modified the mpld3 project.
```
jessegrabowski, line 6: You don't know how much it means to me that you kept and updated this

twiecki:
This is great! Is there no NB on geo?

juanitorduz:
> This is great! Is there no NB on geo?

There is! MMM Multidimensional Example Notebook :)

twiecki:
Should we give that a better name?

williambdean:
This looks awesome. Do you know if "Section Navigation" on the left side can be removed. It is currently empty

williambdean:
> Show we give that a better name? 

Specifically Geo MMM Example? What do you have in mind?

juanitorduz:
The dependency issue is independent of this PR, it seems as other builds are failing https://app.readthedocs.org/projects/pymc-marketing/builds/27719823/

williambdean:
The "How-to" is now pushed into the "More" dropdown menu. At least with my screen size. Maybe we can link at the top of the Gallery section. Or specify that these are all the same notebooks, just different way to organize them

scripts/generate_gallery.py
```diff
@@ -0,0 +1,292 @@
...
+GALLERY_IMG_DIR = GALLERY_DIR / "images"
+
+# Create gallery images directory if it doesn't exist
+GALLERY_IMG_DIR.mkdir(exist_ok=True, parents=True)
+
+# Default image in case we can't extract one from a notebook
+DEFAULT_IMG_LOC = PROJECT_ROOT / "docs" / "source" / "_static" / "flat_logo.png"
+
+# Map folder names to display titles
+folder_title_map = {
```
williambdean, line 41: Is the secretly used? I don't see a renference

juanitorduz: good catch! removed in https://github.com/pymc-labs/pymc-marketing/pull/1611/commits/f24602e59ee43f64eaabc0787a59cacc58b5400b

williambdean (approved):
Think this looks great. 

What is the process for adding a new notebook now? Can we outline that and raise if we can detect?

jessegrabowski:
> I think the workflow is very similar as before: add the notebook into the notebooks folder and add it in gallery.md (instead of howto.md)

You didn't like the auto-generated gallery the other script did?

juanitorduz:
Oh! I need to double-check! I kind of wanted to have the freedom to define the gallery structure but ill revise it 👀

juanitorduz:
I will takle this in a following PR https://github.com/pymc-labs/pymc-marketing/issues/1617 The current notebook list is just very ugly, so let's work on iterations 🙏

twiecki:
> > Show we give that a better name?
> 
> Specifically Geo MMM Example? What do you have in mind?

Anything that makes someone that is looking for Geo MMM click on that NB. I just think multi-dimensional is not that. Whether that's a rename or new NB with that name.

twiecki:
Is howto superseded by this? In that case I think we should consider just removing it.

williambdean:
> Is howto superseded by this? In that case I think we should consider just removing it.

Would the previous links still work? Would that break any references in social posts?

juanitorduz:
> Is howto superseded by this? In that case I think we should consider just removing it.

It is now removed

juanitorduz:
> > Is howto superseded by this? In that case I think we should consider just removing it.
> 
> Would the previous links still work? Would that break any references in social posts?

I think so, still, I had gotten feedback that it was not clear for many people that "howto" contained the example notetbooks.

juanitorduz:
@twiecki I added some info regarding geo-mmm

image

It is not the main title, as many users are using other types of dimensions (e.g., outlets)

juanitorduz:
@williambdean can we merge this one? 🙏

williambdean:
> @williambdean can we merge this one? 🙏

Let's do it!

PR #1636, opened 2025-04-21
Fix casting bug when x0 is not supplied to optimizer

cluhmann:
This is a quick fix on a very small but problematic bug introduced in https://github.com/pymc-labs/pymc-marketing/pull/1565 . I am unclear how the tests added in that PR did not trigger this bug. If anyone has time to figure out that mystery and/or sanity-check this fix, I would very much appreciate it.

----
📚 Documentation preview 📚: https://pymc-marketing--1636.org.readthedocs.build/en/1636/


juanitorduz (approved):

juanitorduz:
@cluhmann do you need a patch release?

cluhmann:
That would be helpful!

PR #1640, opened 2025-04-22
use float type from pytensor config

### LINKED ISSUE #1633: RuntimeError: Incorrect output dtype for return value #0: Expected: int64, Actual: int32
`fit` results in the following error. Tried with 0.11.0, 0.12.0, 0.13.0. 
`XlaRuntimeError: FAILED_PRECONDITION: Buffer Definition Event: Error dispatching computation: %sCpuCallback error: Traceback (most recent call last):
  File "C:\Users\kbope\anaconda3\envs\mmm\Lib\site-packages\jax\_src\interpreters\mlir.py", line 2973, in _wrapped_callback
RuntimeError: Incorrect output dtype for return value #0: Expected: int64, Actual: int32`

The only workaround is to set `pytensor.config.floatX = 'float32'`. But `optimize_budget` doesn't work with that option.

williambdean:
## Related Issue

- [ ] Closes #1633
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1640.org.readthedocs.build/en/1640/

### DIFF
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -20,6 +20,7 @@ from typing import Any, ClassVar, Protocol, runtime_checkable
 
 import arviz as az
 import numpy as np
+import pytensor
 import pytensor.tensor as pt
 import xarray as xr
 from arviz import InferenceData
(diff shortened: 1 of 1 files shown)


williambdean:
Not sure if this is the correct fix. 

It seems that this would already be done since it is not provided

https://github.com/pymc-devs/pytensor/blob/e98cbbcfb28a3c7061c5138b236666c383e6ea15/pytensor/tensor/type.py?plain=1#L807-L808

juanitorduz (approved):

PR #1673, opened 2025-05-06
Marginal effects + counterfactual sweeps for MMM insights

drbenvincent:
## Description
This PR provides extra functionality to extract insights from MMM's. It adds a new class which allows the MMM to be evaluated at a grid of parameters (in what I'm calling a sweep). We can use this functionality to plot marginal effects (where the sweep defines the absolute value of a driver) or where we run a counterfactual sweep. In this latter case, we can run either a multiplicative or additive sweep. This mode retains the time-varying driver values but modifies them multiplicatively or additively. 

## Checklist


cetagostini:
Hey @drbenvincent very nice work, I really like the functionality here. Nevertheless, after a first quick scan, a few suggestion thinking in the future:

1. All related work to estimation of marginal effects would be nice to have it in the `.py` file created, but all related plots would be better to have in the [plot suite](https://github.com/pymc-labs/pymc-marketing/blob/marginal-effects/pymc_marketing/mmm/plot.py). It's going slowly but we are trying to move all plots there (Consider this plot suite must be pymc.model agnostic, take as reference other plots were we infer dimensions or shapes automatically).
2. I'll use the multidimensional class for the example, we are moving away from the previous class slowly. You can create the same model, the only you need to do is `from pymc_marketing.mmm.multidimensional import MMM` all the signature will be the same, except you need to define the `target_column`. That's it :) 

Doing this hopefully we can aim to a signature like:
```py
sweep_values = mmm.counterfactual_sweep(
    predictors=["influencer_spend"],
    sweep_values=np.linspace(0, 2, 12),
    sweep_type="absolute",
)

optimizable_model.plot.marginal_effects(
    samples=sweep_values,
);
```

This will be consistent with how other plots and methods are being used, for example, optimizer have very similar way of working. How does that sound?

On the other hand, I left a few minor questions!

> [!NOTE]
> I saw an issue with drivers being scaled (Ensure the sweep values are being applied in the original space), if you are using the new MMM class, you can make any variable in the graph original scale. Meaning, you can create the original scale var, and probably change your signature from `predictors=["influencer_spend"]` to `var_names=["influencer_spend_original_scale"]`. Users could use any var to plot this effects, and you can probably control that it's not able to run arbitrary, adding a raise error if the `var_name` it's not in the coords for channel data (?) or similar.

drbenvincent:
<p>The top plot starts negative on the left because we've gone from some influencer spend to zero influencer spend. The y-axis is <em>change</em> in uplift relative to the actual scenario, so it answers the question of "what would have happened to our total sales if we had spent zero on influencers. </p><p> </p><p>Second plot - I think it makes sense to keep the "marginal effects" terminology because that is quite popular in stats circles. There's certainly scope for changing the y-axis label, but I think it makes sense to describe the marginal effects as the derivative when I flesh out the notebook text. </p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1673/discussion/#comment-2865565177'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Good point. Check the y-axis scale. The top plot is essentially linear, so the derivative is actually a constant value. But when the axes are scaled so tightly around the values it magnifies numerical imprecision. I will see what I can do with y-axis scaling to better visually portray that the marginal effect here is flat as a function of the variable being manipulated on the x-axis</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1673/discussion/#comment-2865565341'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Ah, so this will be better explained as I flesh out the notebook text. The additive, multiplicative, or absolute options are how we are manipulating the target driver. So setting the sweep values to have a multiplicative effect keeps the time course of their values but literally just scales them up or down. Nothing about the model is changed other than how we are manipulating the target driver.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1673/discussion/#comment-2865565475'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
> 1. All related work to estimation of marginal effects would be nice to have it in the `.py` file created, but all related plots would be better to have in the [plot suite](https://github.com/pymc-labs/pymc-marketing/blob/marginal-effects/pymc_marketing/mmm/plot.py). It's going slowly but we are trying to move all plots there (Consider this plot suite must be pymc.model agnostic, take as reference other plots were we infer dimensions or shapes automatically).

So it's very worth having this discussion now. Bear in mind that the counterfactual sweep returns a more complex idata - a set of uplifts but with an additional sweep dimension. At the moment this is stored as it's own thing, but I guess I could store it as a new DataArray in the idata.

Happy to be guided by you guys about the best way to incorporate this into the codebase. I think there are a few options:

1. Make the counterfactual sweep a method of the `MMM` class
  * Either, return a results object with it's own plot methods
  * Or, add a new sweep DataArray to the `MMM`'s idata and add new plot methods to the `MMMPlotSuite`
2. Keep plot functions in the `CounterfactualSweep` but these just point to the real plot functionality which lives in `MMMPlotSuite`
3. Keep it as it is. Probably not what you want.

I'm guessing (1) might be your favourite? Any thoughts from @williambdean on this? Would be ideal to pin down the API and general approach early, to avoid having to do any reimplementation/restructuring.

pymc_marketing/mmm/marginal_effects.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+        return self.marginal_effects
+
+    def plot_marginal_effects(self, hdi_prob: float = 0.94, ax=None) -> plt.Axes:
+        """Plot the marginal effects curve."""
+        if not hasattr(self, "marginal_effects"):
+            self.compute_marginal_effects()
+
+        if ax is None:
+            fig, ax = plt.subplots(figsize=(10, 6))
```
williambdean, line 147:
```suggestion
            _, ax = plt.subplots(figsize=(10, 6))
```

pymc_marketing/mmm/marginal_effects.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+        self.marginal_effects = xr.DataArray(
+            marginal_effects,
+            dims=self.results.dims,
+            coords=self.results.coords,
+        )
+        return self.marginal_effects
+
+    def plot_marginal_effects(self, hdi_prob: float = 0.94, ax=None) -> plt.Axes:
+        """Plot the marginal effects curve."""
```
williambdean, line 142: Add the Parameters and Returns in docstring

pymc_marketing/mmm/marginal_effects.py
```diff
@@ -0,0 +1,173 @@
...
+
+class CounterfactualSweep:
+    """CounterfactualSweep class is used to perform counterfactual analysis on MMM's."""
+
+    def __init__(
+        self,
+        mmm,
+        X: pd.DataFrame,
+        predictors: list[str],
+        sweep_values: np.ndarray,
```
williambdean, line 32: Think the npt.NDArray typing would be best to use

https://numpy.org/doc/stable/reference/typing.html

pymc_marketing/mmm/marginal_effects.py
```diff
@@ -0,0 +1,173 @@
...
+        - sweep_values (np.ndarray): Array of sweep values.
+        - sweep_type (str): 'multiplicative', 'additive', or 'absolute'.
+            - 'multiplicative': Multiply the original predictor values by each sweep value.
+            - 'additive': Add each sweep value to the original predictor values.
+            - 'absolute': Set the predictor values directly to each sweep value (ignoring original values).
+        """
+        if sweep_type not in ["multiplicative", "additive", "absolute"]:
+            raise ValueError(
+                "sweep_type must be 'multiplicative', 'additive', or 'absolute'."
+            )
```
williambdean, line 52: Let's remove and user `Literal["multiplicative", "additive", "absolute"]` instead. The editor will display incorrect input from the user

Another alternative is to use the validate_call from pydantic. Here is example: 
https://github.com/pymc-labs/pymc-marketing/blob/ae215f001eebbee5705953b409c571d9c81bea12/pymc_marketing/mmm/mmm.py?plain=1#L80-L81

pymc_marketing/mmm/marginal_effects.py
```diff
@@ -0,0 +1,173 @@
...
+
+        self.mmm = mmm
+        self.X = X
+        self.predictors = predictors
+        self.sweep_values = sweep_values
+        self.sweep_type = sweep_type
+
+        # Run sweep and store results
+        self.run_sweep()
```
williambdean, line 61: Not a fan of the running with initialization. In favor of it being two step process

hiding behind the method call like @cetagostini mentioned

pymc_marketing/mmm/marginal_effects.py
drbenvincent: done in https://github.com/pymc-labs/pymc-marketing/pull/1673/commits/00e43099e81cd4221e7ab89e1ce53888516938bf

pymc_marketing/mmm/marginal_effects.py
drbenvincent: Done

drbenvincent:
@cetagostini I've had luck switching to `pymc_marketing.mmm.multidimensional.MMM`, however my sweep code calls a method

```python
actual = self.mmm._get_group_predictive_data(
    group="posterior_predictive", original_scale=True
)["y"]
```

which does not seem to have made it from the original `MMM` class into the new multidimensional one.

https://github.com/pymc-labs/pymc-marketing/blob/6d71a2015995a91d29e9546379eb7475a17e2d70/pymc_marketing/mmm/base.py#L275-L295

Any chance you could add this add this method in, or otherwise let me know how to easily access the posterior predictions in the original scale? I can see how you can do it in the [multidimensional example notebook](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_multidimensional_example.html#posterior-predictive-checks), but it would be useful if a user doesn't have to remember to manually rescale

drbenvincent:
<p>Resolved in <a href="https://github.com/pymc-labs/pymc-marketing/pull/1673/commits/fb51cc4ac5cd1152a23d25dc06ed79c0027d855a" target="_blank">fb51cc4</a></p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1673/discussion/#comment-2865565341'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini:
> @cetagostini I've had luck switching to `pymc_marketing.mmm.multidimensional.MMM`, however my sweep code calls a method
> 
> ```python
> actual = self.mmm._get_group_predictive_data(
>     group="posterior_predictive", original_scale=True
> )["y"]
> ```
> 
> which does not seem to have made it from the original `MMM` class into the new multidimensional one.
> 
> https://github.com/pymc-labs/pymc-marketing/blob/6d71a2015995a91d29e9546379eb7475a17e2d70/pymc_marketing/mmm/base.py#L275-L295
> 
> Any chance you could add this add this method in, or otherwise let me know how to easily access the posterior predictions in the original scale? I can see how you can do it in the [multidimensional example notebook](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_multidimensional_example.html#posterior-predictive-checks), but it would be useful if a user doesn't have to remember to manually rescale

Sure I'll give you a hand in a few hours, should be easy 🙌🏻

drbenvincent:
Oh, I've already switched to use this

```python
mmm.idata["posterior_predictive"]["y"]* scalers["target_scale"].item()
```

I could leave it and file a feature request in the future - but just saying now that I assume users would prefer a way of getting predictions back in the original scale without having to remember and do manual rescaling themselves?

But I'm not blocked by this currently :)

cetagostini:
Hey @drbenvincent 

I have been going through the code and trying to think in the most efficient way to make the API and align to our current code base.

1. The first thing, I think, is to **work with Pytensor**. That would make things much simpler and faster. You can add to your class the same function from the optimizer `extract_response_distribution` and then grab any variable from your pytensor graph and compute deterministically the posterior. This will allow you then use the automatic methods `.grad` and `.jacobian` from `pytensor`. Your class `CounterfactualSweep` will get only a pymc model, from it extract the pytensor model `mmm.model` and then triggers the computation of the variable of interest (We don't need to compute `y` directly, we can only compute `channel_contributions` and the change there will be the same in `y` but we save computation). This is similar to how optimizer works.
2. Using the `.jacobian` actually save us from do a loop over the `predict` method. We should be able to compute this in a vectorized way, and get the insight. 
3. Once everything is compute in the class then will access to the `mmm` object and add two groups to the `idata` the `posterior_sweep`, and `posterior_sweep_marginal` (Or any other name).
4. Then we are able to share the idata only to the plot suite, and everything should be okay for the plots.


**I can write down an small example around the pytensor process**, and give you a hand with it. It's very similar to what we have in the optimizer, and again, I really think this is the way to avoid hardcodes and things into the new class (multidimensional) which are not flexible enough around how the library have been design, and I honestly don't think its a lot of work, and will make things very elegant as well.

I'll come back to help with this on 21st and 22nd.

cetagostini:
Hey hey, quick update:

We can borrow from the _pymc-marketing_ optimizer the following function:
```py
# We use pytensor to grab any var in the model.
from pymc.pytensorf import rewrite_pregrad
from pytensor import clone_replace, function
from pytensor.compile.sharedvalue import SharedVariable, shared
from pytensor.graph import rewrite_graph, vectorize_graph
from pytensor.graph.basic import ancestors
import pytensor.tensor as pt

from pymc.logprob.utils import rvs_in_graph
from pymc.model.transform.optimization import freeze_dims_and_data

def extract_response_distribution(
    _pymc_model: pm.Model, response_variable: str, idata: az.InferenceData
) -> pt.TensorVariable:
    """Extract the response distribution graph, conditioned on posterior parameters.
    """
    model = _pymc_model
    # Convert InferenceData to a sample-major xarray
    posterior = az.extract(idata).transpose("sample", ...)  # type: ignore

    # The PyMC variable to extract
    response_var = model[response_variable]

    # Identify which free RVs are needed to compute `response_var`
    free_rvs = set(model.free_RVs)
    needed_rvs = [
        rv for rv in ancestors([response_var], blockers=free_rvs) if rv in free_rvs
    ]
    placeholder_replace_dict = {
        model[rv.name]: pt.tensor(name=rv.name, shape=rv.type.shape, dtype=rv.dtype)
        for rv in needed_rvs
    }

    [response_var] = clone_replace(
        [response_var],
        replace=placeholder_replace_dict,
    )

    if rvs_in_graph([response_var]):
        raise RuntimeError("RVs found in the extracted graph, this is likely a bug")

    # Cleanup graph
    response_var = rewrite_graph(response_var, include=("canonicalize", "ShapeOpt"))

    # Replace placeholders with actual posterior samples
    replace_dict = {}
    for placeholder in placeholder_replace_dict.values():
        replace_dict[placeholder] = pt.constant(
            posterior[placeholder.name].astype(placeholder.dtype),
            name=placeholder.name,
        )

    # Vectorize across samples
    response_distribution = vectorize_graph(response_var, replace=replace_dict)

    # Final cleanup
    response_distribution = rewrite_graph(
        response_distribution,
        include=(
            "useless",
            "local_eager_useless_unbatched_blockwise",
            "local_useless_unbatched_blockwise",
        ),
    )

    return response_distribution
```

Then creating the following function should allow us to run the sweep analysis.

```py
import numpy as np
import pytensor.tensor as pt
from pytensor import function
from pytensor.compile.sharedvalue import SharedVariable
import xarray as xr

def make_marginal_effect_fn(
    pymc_model: pm.Model,
    idata: az.InferenceData,
    original_X: np.ndarray,            # shape (n_dates, n_channels)
    sweep_type: str = "multiplicative",# or "additive", "absolute"
    response_variable: str = "channel_contribution",
    posterior_sample_batch: int = 10
):
    # 1) Extract the Pytensor graph for your channel_contribution
    #    This gives a TensorVariable of shape (n_samples, n_dates, n_channels).
    resp_graph = extract_response_distribution(
        pymc_model, response_variable, idata.isel(draw=slice(None, None, posterior_sample_batch))
    )  # shape (samples, date, channel)

    # 2) Sum over date & channel to get a scalar-per-sample
    f_total = resp_graph.sum(axis=(1, 2))  # shape (n_samples,) -> TO DO: Must sum across all dims, no matter what, we nee to take out the hardcoded 1,2

    # 3) Grab the SharedVariable you created with pm.Data(...)
    channel_data_shared: SharedVariable = pymc_model["channel_data"] # TO DO: This can be input, so we can set input and output for any arbitrary variable

    # 4) Compute the Jacobian ∂f_total / ∂channel_data_shared
    #    Shape: (n_samples, n_dates, n_channels)
    J = pt.jacobian(f_total, channel_data_shared)

    # 5) Sum out the date axis → (n_samples, n_channels)
    J_per_channel = J.sum(axis=1)

    # 6) Compile a single function that reads the current shared value
    jac_fn = function(inputs=[], outputs=J_per_channel)

    # 7) Return a Python wrapper that applies each sweep
    def marginal_fn(sweep_values: np.ndarray) -> np.ndarray:
        outs = []
        for s in sweep_values:
            if sweep_type == "multiplicative":
                channel_data_shared.set_value(original_X * s)
            elif sweep_type == "additive":
                channel_data_shared.set_value(original_X + s)
            elif sweep_type == "absolute":
                # reshape s if needed to broadcast across (date,channel)
                channel_data_shared.set_value(pt.full_like(original_X, s))
            else:
                raise ValueError(f"Unknown sweep_type {sweep_type!r}")
            outs.append(jac_fn())  # (n_samples, n_channels)
        # Stack → (n_samples, n_sweeps, n_channels)
        return pt.stack(outs, axis=1)

    return marginal_fn
```

This two can exist in the class as private methods, and then we can have a property which returns the class. Ideally, we initialize with the model, idata and X data. Then we can run the analysis, something like `mmm.sweep` or maybe `mmm.marginal` which calls `run`. So, the whole thing can look like:
```py
mmm.marginal.run(
extend_idata=True, # or false to return an xarray dataset object
sweep_type="absolute",
varname="channel_contribution",
varinput="channel_data",
posterior_sample_batch=2
)
```

This will return or extend the idata object, adding the `posterior_sweep`, and `posterior_sweep_marginal`. This objects will have the marginals for all channel or other vectors in the deterministic used in the `varname` By consequence in the plot, you can decide if to show one single channel or variable (predictor) or many.

This will allow us to:
- Make the method similar to the current api.
- Separate the plots and calculations in different places. Having a two step process.
- Apply the same analysis over any variable in the model graph (linear or not). Not only to controls or channels, can be applied to any arbitrary `mu_effects` as well.
- Make the class functional to any PyMC model, as soon as they share the idata, and proper variables names. We are working hard in make components open to any PyMC model, so people can build their custom models and plug and play all our stuff easily.

Add the same time:
- We avoid to compute the counterfactual to get the jacobian, and estimate directly the jacobian using the methods from pytensor.
- We avoid the loop over `sample_posterior_predictive`. Instead we compile the jac once, and join the function given a sweep, running everything once.
- Allow to control to the user if wants to use the full posterior or not for the marginal estimations, its possible to reduce the posterior to get speed on the calculations.

There is a few to-do's in the functions in order to make them fully generalizable for any PyMC model, but should be easy to apply.

A [reference notebook is here](https://colab.research.google.com/drive/16xn_lGYflxcyKZilSEGr0A6ShJuKSStK#scrollTo=2t3bV-pbV89s)!

> [!IMPORTANT]
> We are duplicating a existing function in the optimizer. Maybe all of this pytensor utils functions custom made, could exist in some `pytensor_utils.py` and being used by all other methods. So, for the future we can plan this.

juanitorduz:
Hi! Now that I am back I will look into this PR this week 🙏

drbenvincent:
FYI I've already started to adapt the PR based on feedback from @cetagostini. I'm focusing on getting the broad structure and API of the code right first. Will then tackle the implementation issues.

drbenvincent:
FYI. At this point of my refactor, the API is:

```python
results:xr.Dataset = CounterfactualSweep(
    mmm=mmm,
    predictors=["influencer_spend"],
    sweep_values=np.linspace(0, 2, 12),  # Set spend directly from 0 to $100k
    sweep_type="absolute",
).run_sweep()
CounterfactualSweep.plot_uplift(results);
CounterfactualSweep.plot_marginal_effects(results);
```

The changes are:
* No longer auto-computes, it needs to be triggered with the run_sweep method,
* Results are now self contained in an `xr.Dataset`, not inside the class.
* Plot methods only require this `xr.Dataset`.
* No longer have to input `X` into the `CounterfactualSweep` class because it takes it directly from `mmm`.

This is not the end goal. Just keeping track of progress because I'm only able to work on this in bursts.

drbenvincent:
Note to self: next focus should be making the sweep intervention in the original data space. Once this is done, we can verify it is working correctly by checking the uplift for a multiplicative change of 1 is zero (because the counterfactual scenario is equal to the actual scenario).

drbenvincent:
Notes for when I pick this up again. Current status...

```python
mmm.sensitivity_analysis(
    predictors=["influencer_spend"],
    sweep_values=np.linspace(0, 2, 12),  # Set spend directly from 0 to $100k
    sweep_type="multiplicative",
)
```

This now adds a new group called "sensitivity_analysis" to the `mmm.idata`.

Plotting can be done entirely from this idata group. Right now it's done like this

```python
SensitivityAnalysis.plot(mmm.idata.sensitivity_analysis)
SensitivityAnalysis.plot(mmm.idata.sensitivity_analysis, marginal=True);
```

but the next step is to move the plot code into the `MMMPlotSuite`. That will get us to the desired API.

pymc_marketing/mmm/sensitivity_analysis.py
```diff
@@ -0,0 +1,114 @@
...
+import xarray as xr
+
+
+class SensitivityAnalysis:
+    """SensitivityAnalysis class is used to perform counterfactual analysis on MMM's."""
+
+    def __init__(
+        self,
+        mmm,
+        predictors: list[str],
```
cetagostini: Can we called `vars`? or `var_names`? Just for consistency with all other methods, thats how we refer to any variable in the graph.

pymc_marketing/mmm/plot.py
```diff
@@ -945,3 +945,108 @@ def allocated_contribution_by_channel_over_time(
 
         fig.tight_layout()
         return fig, axes
+
+    def plot_sensitivity_analysis(
+        self,
+        hdi_prob: float = 0.94,
+        ax: plt.Axes | None = None,
+        marginal: bool = False,
+        percentage: bool = False,
```
cetagostini, line 1252: What happen if marginal and percentage are both `True`? If thats not allow, the we should have some warning or raise error.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1388,6 +1389,20 @@ def sample_posterior_predictive(
 
         return posterior_predictive_samples
 
+    def sensitivity_analysis(self, *args, **kwargs) -> None:
+        """Run sensitivity analysis on the model. The results are added to the
+        InferenceData object in a new group, `sensitivity_analysis`.
+        """
+        results: xr.DataSet = SensitivityAnalysis(
```
cetagostini: Why not make it a property?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1388,6 +1389,20 @@ def sample_posterior_predictive(
...
+            mmm=self,
+            *args,
+            **kwargs,
+        ).run_sweep()
+
+        if hasattr(self.idata, "sensitivity_analysis"):
+            delattr(self.idata, "sensitivity_analysis")
+        self.idata.add_groups({"sensitivity_analysis": results})
```
cetagostini: I think this:
```py
if hasattr(self.idata, "sensitivity_analysis"):
  delattr(self.idata, "sensitivity_analysis")
  self.idata.add_groups({"sensitivity_analysis": results})
```

Can exist in the `sensitivity_analysis.py` instead. That way this become a property, and the signature should be like `mmm.sensitivity.run_sweep()`

pymc_marketing/mmm/plot.py
```diff
@@ -945,3 +945,108 @@ def allocated_contribution_by_channel_over_time(
...
+            The Axes object with the plot.
+        """
+        if ax is None:
+            _, ax = plt.subplots(figsize=(10, 6))
+
+        if percentage and marginal:
+            raise ValueError("Not implemented marginal effects in percentage scale.")
+
+        # grab sensitivity analysis results from idata
+        results = self.idata.sensitivity_analysis
```
cetagostini, line 1290: Please add a verification to check if exist, otherwise Raise an error with recommendation to run the sweep.

juanitorduz:
This looks great! I agree with @cetagostini ! Let's add these last fixes (and the pre-commit) and we should be good to go.

juanitorduz:
@drbenvincent do you need any support to bring this one to the finish line? :)

pymc_marketing/mmm/plot.py
drbenvincent: Done in https://github.com/pymc-labs/pymc-marketing/pull/1673/commits/2be3268636d69f70d74086b62b582c9a194accc1

pymc_marketing/mmm/multidimensional.py
drbenvincent: done in https://github.com/pymc-labs/pymc-marketing/pull/1673/commits/69f27a10e97bb082fdb1b04c1d206930186951f3 Let me know if that's what you were thinking

pymc_marketing/mmm/multidimensional.py
drbenvincent: I think this is now done in https://github.com/pymc-labs/pymc-marketing/commit/69f27a10e97bb082fdb1b04c1d206930186951f3

pymc_marketing/mmm/sensitivity_analysis.py
drbenvincent: Good idea. Done in https://github.com/pymc-labs/pymc-marketing/pull/1673/commits/316f4e6e402ae51a4008d29eb9b7187944f212db

pymc_marketing/mmm/plot.py
drbenvincent: Already done :)

```
if percentage and marginal:
    raise ValueError("Not implemented marginal effects in percentage scale.")
```

drbenvincent:
Wrong Ben tag by the way

drbenvincent:
I did already add a whole bunch of tests. For whatever reason, all tests are now passing.

So I'll focus on edits to the new notebook and ping you when I'm done.

cetagostini:
@drbenvincent if you want, you can open a new PR for the notebook, and keep the notebook in a draft hidden folder. So, we merge and continue working on it.

drbenvincent:
I don't have that much more to do. Will aim to get some updates today

juanitorduz:
Do you need to work with `multidimensional_model.nc` explicitly?

drbenvincent:
It's showing up as a changed file. I don't think I changed that file, I don't want to commit it. I can't even revert changes.

juanitorduz:
You can do something like `git checkout origin/master -- ./abc1.py` from https://stackoverflow.com/questions/38743912/remove-a-file-from-a-git-pull-request 

To remove it. I will create a PR that comments on writing this file in the notebook

juanitorduz:
`git checkout origin/master -- ./docs/source/notebooks/mmm/multidimensional_model.nc` ?

drbenvincent:
`git checkout origin/master -- ./docs/source/notebooks/mmm/multidimensional_model.nc` gives me

```
fatal: invalid reference: origin/master
```

drbenvincent:
> probably master -> main?

same...

```
Encountered 1 file that should have been a pointer, but wasn't:
        docs/source/notebooks/mmm/multidimensional_model.nc
```

juanitorduz:
We can merge it even with this conflict and I will fix in a following PR

drbenvincent:
Ok. So I'll then just re-do my notebook updates in a follow up PR?

That's fine with me.

When I build the docs locally, I see the notebook in the example gallery, but when I look at the online docs preview https://pymc-marketing--1673.org.readthedocs.build/en/1673/gallery/gallery.html I don't see it. Just letting you know. 

Happy if you want to merge at this point.

juanitorduz:
Thanks @drbenvincent ! You can add the notebook improvements on a next PR (create an issue if needed :))

juanitorduz (approved):

juanitorduz:
@carlosagostini  can you please give us a stamp here ? 🙏

PR #1679, opened 2025-05-10
Change documentation of multi-dimensional

### LINKED ISSUE #1665: Change documentation of multi-dimensional.
Currently our documentation says exist a pooling between countries/geos but by giving to “channel-level” parameters a geo-specific hyperprior, the model induce partial pooling within each geo but no pooling across geos. I found a [example here](https://www.pymc.io/projects/docs/en/v5.10.2/learn/core_notebooks/model_comparison.html) in our page!

This means we are misleading a bit.

gdeleva:
Correct the prior of the lambda parameter of the logistic saturation function so that partial pooling across geos within the same channel is enforced.

## Description
Correct the dim argument of the prior of the lambda parameter of the logistic saturation function.

## Related Issue

- [ ] Closes #1665

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1679.org.readthedocs.build/en/1679/


docs/source/notebooks/mmm/mmm_multidimensional_example.ipynb
```diff
@@ -357,12 +357,12 @@
...
...e are ready to specify the priors for the saturation function. For the beta parameters, we use a half-normal distribution with a sigma parameter that is computed as a function of the spend shares. Beta values are allowed to vary across both channel and geo, but they are drawn from the same underlying distribution, which applies partial pooling via a shared sigma. For the lambda parameters, we add a a channel-level hierarchical structure through the hyperparameters of the gamma prior (mu and sigma). These hyperparameters vary by channel, while the lambda values themselves vary across both channel and geo. This setup enables partial pooling across geos within the same channel, allowing the model to share information across geos. If you need an introduction on Bayesian hierarchical models, check out the comprehensive example [\"A Primer on Bayesian Methods for Multilevel Modeling\"](https://www.pymc.io/projects/examples/en/latest/generalized_linear_models/multilevel_modeling.html) in the PyMC documentation. Please note the `dims` argument in the priors is used to specify the dimensions of the distribution. Here you can control the dimensions along which the hierarchies are defined.\n"
```
TeemuSailynoja: > Beta values are allowed to vary across both channel and geo, but they are drawn from the same underlying distribution, which applies partial pooling via a shared sigma.

If this shared sigma is not a model parameter, there is no partial pooling happening here. The beta parameters are fully separate.

TeemuSailynoja: I know this is a comment to an older commit -- gh didn't show me the diff for a newer one due to size limits...

I checked the raw file and the text for the beta parameters is still:
> For the beta parameters, we use a half-normal distribution with a sigma parameter that is computed as a function of the spend shares. Beta values are allowed to vary across both channel and geo, but they are drawn from the same underlying distribution, which applies partial pooling via a shared sigma.

Although the betas have the same prior, there is no information shared between them -- the observations in channel X1 do not affect the beta for channel X2.

Also the graph shows that betas are fully separate:
image

gdeleva: @TeemuSailynoja  Thanks for the comment. 

Indeed, they are independent. I've updated the comment to clarify this, so users can see how both a no-sharing parameter design (for beta) and a partial pooling parameter design (for lambda) can be implemented.

docs/source/notebooks/mmm/mmm_multidimensional_example.ipynb
```diff
@@ -357,7 +357,7 @@
    "source": [
     "Now we are ready to specify the priors for the saturation function.\n",
     "\n",
-    "For the beta parameters, we use a half-normal distribution with a sigma parameter that is computed as a function of the spend shares. Beta values are allowed to vary across both channel and geo, but they are drawn from the same underlying distribution, which applies partial pooling via a shared sigma.\n",
+    "For the beta parameters, we use a half-normal distribution with a sigma parameter that is computed as a function of the spend shares. For illustrative purposes, we define the beta values to be independent across both channels and geos.\n",
     "\n",
     "For the lambda parameters, we add a a channel-level hierarchical structure through the hyperparameters of the gamma prior (mu and sigma). These hyperparameters vary by channel, while the lambda values themselves vary across both channel and geo. This setup enables partial pooling across geos within the same channel, allowing the model to share information across geos.\n",
```
TeemuSailynoja: > we add a a channel-level...
typo, unfortunately, I don't have permissions for editing the PR, so I have to ask you to fix it.

docs/source/notebooks/mmm/mmm_multidimensional_example.ipynb
TeemuSailynoja, line 1: At the top with other imports, we should add
```python
import warnings
```
and then
```python
warnings.filterwarnings("ignore", category=UserWarning)
```
This would hide the warnings that show your username.
like this one:
> /Users/giacomodeleva/Library/Caches/pypoetry/virtualenvs/pymc-marketing-bZefcsde-py3.11/lib/python3.11/site-packages/seaborn/axisgrid.py:123: UserWarning: The figure layout has changed to tight

@williambdean would you agree?

Also, the last cell with the computation environment is not ran. Could you run that still?

docs/source/notebooks/mmm/mmm_multidimensional_example.ipynb
gdeleva: @TeemuSailynoja  Thanks for the comment. I fixed the typo and run the last cell. Warnings with my user name should not be visible but I didn't commit the code that suppress the warnings.

docs/source/notebooks/mmm/mmm_multidimensional_example.ipynb
TeemuSailynoja: 8688565 looks good. Thank you!

TeemuSailynoja (approved):

PR #1702, opened 2025-05-21
Fix Bass model derivative

### LINKED ISSUE #1701: Bass asymptotic behavior is unexpected.
It seems that the Bass model does not work as expected.  Specifically, the total adoption does not approach `m` in the limit of large `t`.  @williambdean suggested that the issue may lie with the accuracy of [the derivative](https://github.com/pymc-labs/pymc-marketing/blob/7659d7102433bc1b7ef59d15e7c6c248f1fc248c/pymc_marketing/bass/model.py?plain=1#L215).   It seems like the code here is wrong and that one correct version would be:

$$
\dfrac{dF}{dt}=\frac{p (p + q)^2 e^{t (p + q)}}{\left(p e^{t (p + q)} + q\right)^2}
$$

If that's correct (someone should double-check), then I think the derivative should be:

```py
(p * pt.square(p + q) * pt.exp(t * (p + q)) ) / pt.square(p * pt.exp(t * (p + q)) + q)
```

williambdean:
## Description

The derivative is off

## Related Issue

- [ ] Closes  #1701
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1702.org.readthedocs.build/en/1702/

### DIFF
--- a/pymc_marketing/bass/model.py
+++ b/pymc_marketing/bass/model.py
@@ -212,7 +212,9 @@ def f(
 
     The peak adoption rate occurs at time :math:`t^* = \frac{\ln(q/p)}{p+q}`
     """
-    return (p + q) * pt.exp(-(p + q) * t) / (1 + (q / p) * pt.exp(-(p + q) * t)) ** 2
+    return (p * pt.square(p + q) * pt.exp(t * (p + q))) / pt.square(
+        p * pt.exp(t * (p + q)) + q
+    )
 
 
 def create_bass_model(
--- a/pymc_marketing/mmm/plot.py
+++ b/pymc_marketing/mmm/plot.py
@@ -125,17 +125,6 @@ class MMMPlotSuite:
 
         return data
 
-    def _compute_ci(
-        self, data: xr.DataArray, ci: float = 0.85, sample_dim: str = "sample"
-    ) -> tuple[xr.DataArray, xr.DataArray, xr.DataArray]:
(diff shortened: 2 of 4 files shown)


williambdean:
Do we need to change anything in the notebook example because of this change?

cluhmann:
> Do we need to change anything in the notebook example because of this change?

Depends on whether the fix works :wink:

cluhmann:
Ok, looks like the asymptotic behavior is correct now.  But I suspect that you will want to loosen the sigmas now, because the plots suggest that the priors are extremely tight.  So much so that the posterior rarely mirrors the observed data.

williambdean:
I've made that change in the notebook and I've added a plot for the estimated `m` parameter which compares observed cumulative at T=51 (despite some having not fully matured) ... but it seems that they are roughly working out.

cluhmann:
Looks better.  At some point (not necessarily now), I might suggest using 2 different y-axes for the "Innovators vs Imitators" plots.  Because the counts and the cumulative counts are on such different scales, it might make sense to put them on 2 different axes so that you can actually see the innovator/imitator counts.  Right now only the cumulative plot is effectively visible.

williambdean:
Okay. I will make an issue

cluhmann (approved):
Sounds good.

williambdean:
Thanks @cluhmann

PR #1713, opened 2025-05-26
Add LinearRegression wrapper

### LINKED ISSUE #1554: Wrapper LinearRegression MMM for baseline analysis
Special case MMM which no adstock and linear saturation transformation. For baseline model but allows for all of the features that MMM allows.

Wrapper around:

```python
from pymc_marketing.mmm import AdstockTransformation, SaturationTransformation, MMM

class NoAdstock(AdstockTransformation): 
    lookup_name: str = "no_adstock"

    def transform(self, x): 
        return x

class Linear(SaturationTransformation): 
    lookup_name: str = "linear"
    
    def transform(self, x, beta): 
        return beta * x

mmm = MMM(adstock=NoAdstock(), saturation=Linear(), **kwargs)
```

CC: @cetagostini

williambdean:
## Description

Adds a wrapper for linear regression. The components can be used in the multidimensional MMM as well

## Related Issue

- [ ] Closes #1554
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1713.org.readthedocs.build/en/1713/


pymc_marketing/mmm/__init__.py
```diff
@@ -27,6 +28,7 @@
     HillSaturation,
     HillSaturationSigmoid,
     InverseScaledLogisticSaturation,
+    LinearSaturation,
```
cetagostini-wise, line 31: I think its better to say NoSaturation in the same way than NoAdstock, because LinearSaturation its a bit confusing in my view, but up to you!

cetagostini-wise (approved):

pymc_marketing/mmm/__init__.py
williambdean: Done. Name is now NoSaturation

PR #1737, opened 2025-06-05
Update pyproject.toml

juanitorduz:
Remove pymc 5.23 upper bound from deps.


pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
     "pandas",
     "pydantic>=2.1.0",
     # NOTE: Used as minimum pymc version with test.yml `OLDEST_PYMC_VERSION`
-    "pymc>=5.22.0,<5.23.0",
+    "pymc>=5.22.0",
     "pytensor>=2.30.3",
```
ricardoV94: This way users won't face that bug?
```suggestion
    "pytensor>=2.31.3",
```

PR #1738, opened 2025-06-06
Implement add_lift_test_measurements for MMMM

### LINKED ISSUE #1667: add_lift_test_measurements method is not available in pymc_marketing.mmm.multidimensional
We dont have a default way to calibrate multidimensional model with lift test results. 
Is there hack that we can use (instead of rewriting everything) ?

context: 
I have read through https://www.pymc-marketing.io/en/latest/api/generated/pymc_marketing.mmm.lift_test.add_saturation_observations.html, but using that I have to rewrite every default functionality of pymc_marketing.mmm.multidimensional

Code:
from pymc_marketing.mmm.multidimensional import MMM

# Base MMM model specification
mmm = MMM(
    date_column="date",
    target_column="y",
    channel_columns=["x1", "x2"],
    control_columns=["event_1", "event_2"],
    dims=("geo",),
    scaling={
        "channel": {"method": "max", "dims": ()},
        "target": {"method": "max", "dims": ()},
    },
    adstock=adstock,
    saturation=saturation,
    yearly_seasonality=2,
    model_config=model_config,
)

Issue:
MMM.add_lift_test_measurements ?

add_lift_test_measurements  is not found

TeemuSailynoja:
## Description
Extend the relevant functions in `mmm/lift_test.py` to enable adding lift tests for multidimensional MMMs.
In this implementation, the lift test results need to specify which dimensions each of the measurements applies to.

## Related Issue

- [x] Closes #1667

## Checklist

    ~~Still need to check these.~~ The documentation example works now.

----
📚 Documentation preview 📚: https://pymc-marketing--1738.org.readthedocs.build/en/1738/


juanitorduz:
> Should this have a connected notebook? The only difference in the usage is that the lift tests need to include information on which model dimensions the measurements are connected.

This is very cool! The notebook is not necessary, but is a nice-to-have (even for a following PR).

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1426,6 +1433,157 @@ def sample_posterior_predictive(
...
+                "channel": ["x1", "x1"],
+                "geo": ["FIN", "SWE"],
+                "x": [1, 1],
+                "delta_x": [0.1, 0.2],
+                "delta_y": [0.1, 0.1],
+                "sigma": [0.1, 0.1],
+            })
+
+            model.add_lift_test_measurements(df_lift_test)
+
```
TeemuSailynoja: I have now checked that this example, modified from the 1D MMM, works. Added some Finland vs. Sweden flavour.

TeemuSailynoja: Forgot to fix the imports, but further fixed this in 2bf7c0888138edf9a38d2b0317dc981b59767761.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1426,6 +1433,158 @@ def sample_posterior_predictive(
...
+            if dim not in df_lift_test.columns:
+                raise KeyError(
+                    f"The {dim} column is required to map the lift measurements to the model."
+                )
+
+        # Define a target transformer from the target scaler.
+        def target_transform(target):
+            return target / self.scalers._target.to_numpy()
+
+        # Define a channel transformer from the channel scalker.
```
williambdean:
```suggestion
        # Define a channel transformer from the channel scaler.
```

TeemuSailynoja:
> Overall looks good! I've added some tests for different scaling configurations other than the defaults which fail. Might be good to make use of xarray broadcasting here

Good catch! I love xarray more and more every day :) I think I have good solution. I'll continue on this later today and hope to have the PR updated soon.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1426,6 +1433,158 @@ def sample_posterior_predictive(
...
+                raise KeyError(
+                    f"The {dim} column is required to map the lift measurements to the model."
+                )
+
+        # Define a target transformer from the target scaler.
+        def target_transform(target):
+            return target / self.scalers._target.to_numpy()
+
+        # Define a channel transformer from the channel scaler.
+        def channel_transform(channels):
```
williambdean: Based on the latest comment, can't you just wrap these in xarray, do the operation, then transform back to numpy? We will know all of the dimensions of these variables from the class and can use them in this local function

However, we can change the functions in lift_tests to be better structured if needed

TeemuSailynoja: I can make it so. I'll make sure to document it more descriptively, as it will possibly look weird.
The transform functions passed to the `scale_lift_measurements` function will still have the signature np.array -> np.array, but as we know at the moment of constructing the transform, which coordinates the np.array dimensions correspond to, we can make it an xarray within the transform function.

tests/mmm/test_multidimensional.py
```diff
@@ -1020,6 +1020,213 @@ def test_time_varying_media_with_lift_test(multi_dim_data, df_lift_test) -> None
...
+    with pytest.raises(
+        KeyError,
+        match="The product column is required to map the lift measurements to the model.",
+    ):
+        mmm.add_lift_test_measurements(df_lift_test_missing_product)
+
+
+def test_add_lift_test_measurements_no_dimensions_success(multi_dim_data) -> None:
+    """Test that lift test measurements work correctly when no dimensions are specified."""
+    X, y = multi_dim_data
```
williambdean: Shouldn't this be the single dim data then?

williambdean: I think this is that silient error that was raised by Ben #1753

tests/mmm/test_multidimensional.py
williambdean, line 1: Can we add mock_pymc_sample fixture and call mmm.fit in order to see if there are some shape issues that pop up?

tests/mmm/test_multidimensional.py
TeemuSailynoja: Fixed.

pymc_marketing/mmm/multidimensional.py
williambdean: nice. Thanks

tests/mmm/test_multidimensional.py
TeemuSailynoja: Added in 4bb798f6e2b824a7d24a8a3aa3f86d028842cfa8. Thank you.

williambdean: Would @juanitorduz be able to take a look?

juanitorduz:
> With
> 
> ```python
> def target_transform(df_lift_tests: pd.DataFrame, target_col: str, channel_col: str, , dim_cols: str) -> np.ndarray:
>     index_cols = dim_cols + [channel_col]
>     re_indexed = df_lift_tests.set_index(
>             index_cols
>         )
>     scaled = re_indexed.loc[:,[target_col]].to_xarray() / mmm.scalers._target
>     return scaled.to_dataframe().loc[re_indexed.index,:].reset_index().loc[:, target_col]
> ```
> 
> and
> 
> ```python
> def channel_transform(df_lift_tests: pd.DataFrame, channel_col: str, dim_cols: str) -> np.ndarray:
>     index_cols = dim_cols + [channel_col]
>     re_indexed = df_lift_tests: .set_index(
>             index_cols
>         )
>     scaled = re_indexed.loc[:, ["x", "delta_x"]].to_xarray() / mmm.scalers._channel
>     return scaled.to_dataframe().loc[re_indexed.index,:].reset_index()
> ```
> 
> the `scale_lift_measurements()` function could be implemented for MMMM as:
> 
> ```python
> def scale_lift_measurements(
>    df_lift_test: pd.DataFrame,
>    channel_col: str,
>    dim_cols: list[str] | None = None,
> ) -> pd.DataFrame:
> if "date" in df_lift_test().columns:
>    return pd.concat(
>        [
>            channel_transform(df_lift_test, channel_col, dim_cols),
>            target_transform(df_lift_test, "delta_y", channel_col, dim_cols),
>            target_transform(df_lift_test, "sigma", channel_col, dim_cols),
>            df_lift_test["date"],
>        ],
>        axis=1,
>    )
> else:
>    return pd.concat([
>            channel_transform(df_lift_test, channel_col, dim_cols),
>            target_transform(df_lift_test, "delta_y", channel_col, dim_cols),
>            target_transform(df_lift_test, "sigma", channel_col, dim_cols),
>    ], axis=1)
>    
> ```
> 
> Would it make more sense to make an implementation, that would make use of the existing `mmm.lift_test.scale_lift_measurements` function with the function signature
> 
> ```python
> def scale_lift_measurements(
>     df_lift_test: pd.DataFrame,
>     channel_col: str,
>     channel_columns: list[str | int],
>     channel_transform: Callable[[np.ndarray], np.ndarray],
>     target_transform: Callable[[np.ndarray], np.ndarray],
>     dim_cols: list[str] | None = None,
> ) -> pd.DataFrame:
> ```
> 
> or to wrap the above scaling with xarrays into a `_scale_lift_measurements()` method for the multidimensional class? At least in the later case, this could be further simplified.

From a user's perspective, I think it would be better to "make use of the existing mmm.lift_test.scale_lift_measurements function with the function signature", still balancing that the code dos not become cumbersome to maintain.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1490,6 +1497,240 @@ def sample_posterior_predictive(
...
+        input_df = (
+            df_lift_test.loc[:, [*index_cols, "x", "delta_x"]]
+            .set_index(index_cols, append=True)
+            .stack()
+            .unstack(level=-2)
+            .reindex(self.channel_columns, axis=1)  # type: ignore
+            .fillna(0)
+        )
+
+        def transform(input: np.ndarray) -> np.ndarray:
```
juanitorduz: can we add docstrings please 🙏

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1490,6 +1497,240 @@ def sample_posterior_predictive(
...
+        Returns
+        -------
+        Callable[[np.ndarray], np.ndarray]
+            The function for scaling the target data.
+        """
+        # These are the same order as in the original lift test measurements.
+        index_cols = [*list(self.dims), "channel"]
+        input_idx = df_lift_test.set_index(index_cols, append=True).index
+
+        def transform(input: np.ndarray) -> np.ndarray:
```
juanitorduz: can we add a test for this function ?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1490,6 +1497,240 @@ def sample_posterior_predictive(
...
+                .stack()
+                .unstack(level=-2)
+                .loc[input_df.index, :]
+                .values
+            )
+
+        # Finally return the scaled data as a np.ndarray corresponding to the input index order.
+        return transform
+
+    def _make_target_transform(
```
juanitorduz, line 1567: can we add a test for this function ?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1490,6 +1497,240 @@ def sample_posterior_predictive(
 
         return posterior_predictive_samples
 
+    def _make_channel_transform(
```
juanitorduz, line 1515: can we add a test for this function ?

juanitorduz (changes requested):
Hi! I gave it a quick look and looks very nice! I suggest we add more tests for the individual functions 🙏 .

I will try to take a deeper look and test it with the mmm_multi_dim example during the weekend or early nextr week (busy days for me)

pymc_marketing/mmm/multidimensional.py
TeemuSailynoja: added test in 977096bb7dc48d7a24a0cafe147ca73bbf0ba806

pymc_marketing/mmm/multidimensional.py
TeemuSailynoja: added test in 977096bb7dc48d7a24a0cafe147ca73bbf0ba806

PR #1739, opened 2025-06-07
Fix LinearTrend changepoints in `plot_curve` method

gdeleva:
We corrected the formula for the change point time positions in the docstrings and aligned the vertical lines in the plot_curve function with the actual change point positions.

## Description

Correct the formula of the position of the change points;
Align the vertical lines of plot_curve to the position of the change points.

## Related Issue

- [ ] Closes #1672 
- [ ] Related to #

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--1739.org.readthedocs.build/en/1739/

PR #1741, opened 2025-06-08
Fix dim issue for default priors of linear trend and replace linear trend related doc pictures

### LINKED ISSUE #1693: LinearTrend curve workflow fails with scalar parameters dims but extra dims
The prior, curve, plot workflow fails with scalar dims and extra dims specified on the LinearTrend component.

```python
from pymc_marketing.mmm import LinearTrend

trend = LinearTrend(n_changepoints=4, dims=("geo", "product"))

# This is fine
coords = dict(geo=["A", "B"], product=["X", "Y", "Z"])
prior = trend.sample_prior(coords=coords)

# This fails...
curve = trend.sample_curve(prior)

trend.plot_curve(curve)
```

Here is the traceback: 

<details>
<summary>Trackback</summary>

```

---------------------------------------------------------------
ValueError                    Traceback (most recent call last)
Cell In[7], line 2
      1 # This fails...
----> 2 curve = trend.sample_curve(prior)

File ~/GitHub/pymc-eco/pymc-marketing/pymc_marketing/mmm/linear_trend.py:400, in LinearTrend.sample_curve(self, parameters, max_value)
    398 with pm.Model(coords=coords):
    399     name = "trend"
--> 400     pm.Deterministic(
    401         name,
    402         self.apply(t),
    403         dims=("t", *cast(Dims, self.dims)),
    404     )
    406     return pm.sample_posterior_predictive(
    407         parameters,
    408         var_names=[name],
    409     ).posterior_predictive[name]

File ~/mamba/envs/pymc-marketing-dev/lib/python3.10/site-packages/pymc/model/core.py:2263, in Deterministic(name, var, model, dims)
   2261 var = var.copy(model.name_for(name))
   2262 model.deterministics.append(var)
-> 2263 model.add_named_variable(var, dims)
   2265 from pymc.printing import str_for_potential_or_deterministic
   2267 var.str_repr = types.MethodType(
   2268     functools.partial(str_for_potential_or_deterministic, dist_name="Deterministic"), var
   2269 )

File ~/mamba/envs/pymc-marketing-dev/lib/python3.10/site-packages/pymc/model/core.py:1476, in Model.add_named_variable(self, var, dims)
   1474 for dim in dims:
   1475     if dim not in self.coords and dim is not None:
-> 1476         raise ValueError(f"Dimension {dim} is not specified in `coords`.")
   1477 if any(var.name == dim for dim in dims if dim is not None):
   1478     raise ValueError(f"Variable `{var.name}` has the same name as its dimension label.")

ValueError: Dimension geo is not specified in `coords`.

```

</details>

### LINKED ISSUE #1740: t Parameter of LinearTrend.apply
Shall we specify in 

https://www.pymc-marketing.io/en/latest/api/generated/classmethods/pymc_marketing.mmm.linear_trend.LinearTrend.apply.html#pymc_marketing.mmm.linear_trend.LinearTrend.apply

that t must be a strictly increasing tensor with first element equals to 0?

These constraints are ensured when a MMM is built, but users may use this component independently.

gdeleva:
The default prior is currently created using only the 'changepoint' dimension, but it should also include all additional dimensions used to construct the LinearTrend.

## Description

- Define the dimensions of the default priors as the union of "changepoint" and all dimensions used in LinearTrend.
- Replace static images in the documentation that were generated using plot_curve before PR #1739.
- Make minor edits to the docstrings.

## Related Issue

- [ ] Closes #1740 
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1741.org.readthedocs.build/en/1741/


pymc_marketing/mmm/linear_trend.py
```diff
@@ -82,6 +82,7 @@ class LinearTrend(BaseModel):
 
     where:
 
+    - :math:`0 \le t \le T`,
```
williambdean: Out of sample predictions are possible with t>T

pymc_marketing/mmm/linear_trend.py
```diff
@@ -276,7 +277,7 @@ def default_priors(self) -> dict[str, Prior]:
                 "Laplace",
                 mu=0,
                 b=0.25,
-                dims="changepoint",
+                dims=tuple(d for d in {"changepoint"}.union(cast(Dims, self.dims))),
```
williambdean: None of the other transformations work like this. They all require specifying dims if desired. So I suggest we keep that behavior

pymc_marketing/mmm/linear_trend.py
gdeleva: @williambdean  Thanks for the comment. Before I edit this line, could you please confirm the following?
- There is a change point at time T (or max(t), if you prefer).
- The posterior distribution of the parameters of this change point will be equal to the prior distribution (because there is no observed data after T).
- Out of sample prediction is therefore done using the prior distribution of the change point at time T. If the user wants to extract the trend from the last period (the period between the last two change points), the it is expected that the prior distribution for the slop change at time T is set to zero.

If this is true, it may be better to document all this.

pymc_marketing/mmm/linear_trend.py
gdeleva: @williambdean  Sorry, I am afraid I did not understand your comment. You can construct priors with desired dimensions. Here we are only modifying the default prior in case this is not defined (as in your example in #1693). If you build a LinearTrend object with specific (desired) dimensions without specifying the prior, shall the prior not have the same dimensions as default?

williambdean: The user would specify the dims they want. They default is that each parameter is a scalar just like with the SaturationTransformation and AdstockTransformation

The plotting would likely have to consider this. Potentially using expand_dims
https://docs.xarray.dev/en/stable/generated/xarray.DataArray.expand_dims.html

pymc_marketing/mmm/linear_trend.py
gdeleva: @williambdean  Hi. I made some corrections. Please let me know what you think.

pymc_marketing/mmm/linear_trend.py
gdeleva: @williambdean I reverted to the previous line of code and took the fix of #1693 out of the scope of this PR.

pymc_marketing/mmm/linear_trend.py
```diff
@@ -290,7 +292,7 @@ def apply(self, t: pt.TensorLike) -> TensorVariable:
         Parameters
         ----------
         t : pt.TensorLike
-            Input values for the trend.
+            # 1D array of strictly increasing time values for the trend starting from 0.
```
williambdean:
```suggestion
            1D array of strictly increasing time values for the trend starting from 0.
```

PR #1763, opened 2025-06-16
Standardize usage of singular|plural for posterior params

PabloRoque:
Resolve inconsistent naming of model variables between `mmm.MMM` and `multidimensional.MMM`.
The original uses plural, whereas the multidimensional uses singular in variable names. 

The inconsistency is problematic when coupled with the BudgetOptimizer, since we don't have a consistent way to access posterior variables. 

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist


pymc_marketing/mmm/mmm.py
```diff
@@ -1026,7 +1026,7 @@ class MMM(
     _model_type: str = "MMM"
     version: str = "0.0.2"
```
williambdean: Can we up the version so that the load fails

PabloRoque: Yessir!

PabloRoque:
Do we have a script to generate the `model.nc` file?

Tried to patch the idata var_names but now I am getting a `DifferentModelError` when loading it in `mmm_budget_allocation_example.ipynb`

PR #1809, opened 2025-07-03
Narwhals support for CLV aggregation

williambdean:
## Description

Still a work in progress. 

The LazyFrame like libraries will require a provided observation_end_date. However, that can be found outside of the 

Still building out the functionality for the: 

- remove first observation - mean with nans might differ in the backends

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/clv/utils.py
+++ b/pymc_marketing/clv/utils.py
@@ -14,11 +14,13 @@
 """Utilities for the CLV module."""
 
 import warnings
-from datetime import date, datetime
+from datetime import date, datetime, timedelta
 
+import narwhals as nw
 import numpy as np
 import pandas
 import xarray
+from narwhals.typing import IntoFrameT
 from numpy import datetime64
 
 __all__ = [
@@ -156,6 +158,30 @@ def customer_lifetime_value(
     return clv.transpose("chain", "draw", "customer_id")
 
 
+def _find_first_transactions_alternative(
+    transactions: IntoFrameT,
+    customer_id_col: str,
+    datetime_col: str,
+    monetary_value_col: str | None = None,
+    datetime_format: str | None = None,
+) -> IntoFrameT:
+    transactions = nw.from_native(transactions)
+
+    first_date = transactions.group_by(customer_id_col).agg(
+        first_date=nw.col(datetime_col).min()
+    )
+
+    agg_cols = [] if monetary_value_col is None else [nw.col(monetary_value_col).sum()]
+    agg = transactions.group_by([customer_id_col, datetime_col]).agg(*agg_cols)
+
+    return (
+        agg.join(first_date, on=customer_id_col)
+        .with_columns(first=nw.col(datetime_col) == nw.col("first_date"))
+        .drop("first_date")
+        .to_native()
+    )
+
+
 def _find_first_transactions(
     transactions: pandas.DataFrame,
     customer_id_col: str,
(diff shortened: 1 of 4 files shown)


juanitorduz:
OMG! 100% yes! 🥳

ColtAllen:
> OMG! 100% yes! 🥳

Indeed, thanks for starting this!

Do you think the current pandas functions should still be retained for a time even after this is merged?

williambdean:
> Do you think the current pandas functions should still be retained for a time even after this is merged? Also, it seems like the PR description in the original message requires more details.

I was just doing some comparisons of the two at the moment. However, I think that the new one should just take it's place.

williambdean:
Maybe @ColtAllen is interested in taking this over?

pymc_marketing/clv/utils.py
```diff
@@ -264,6 +290,58 @@ def _find_first_transactions(
...
+    datetime_format: str | None = None,
+    observation_period_end: str | pandas.Period | datetime | None = None,
+    time_scaler: float = 1.0,
+) -> IntoFrameT:
+    transactions = nw.from_native(transactions)
+
+    date = nw.col(datetime_col).cast(nw.Datetime)
+
+    if observation_period_end is None:
+        observation_period_end = transactions[datetime_col].cast(nw.Datetime).max()
```
FBruzzesi, line 307: I am not sure if this would work/is supported, but you might try to do: 

```suggestion
    if observation_period_end is None:
        observation_period_end = pl.col("max").max()
```

to get the global max datetime value.

This might also help to avoid this requirement:

> The LazyFrame like libraries will require a provided observation_end_date. However, that can be found outside of the

pymc_marketing/clv/utils.py
```diff
@@ -264,6 +290,58 @@ def _find_first_transactions(
...
+    customer_id_col: str,
+    datetime_col: str,
+    monetary_value_col: str | None = None,
+    datetime_format: str | None = None,
+    observation_period_end: str | pandas.Period | datetime | None = None,
+    time_scaler: float = 1.0,
+) -> IntoFrameT:
+    transactions = nw.from_native(transactions)
+
+    date = nw.col(datetime_col).cast(nw.Datetime)
```
FBruzzesi, line 304: This is very tempting, but consider creating a new column between operations - I would be afraid that for pandas the casting happens multiple times instead of once

pymc_marketing/clv/utils.py
```diff
@@ -264,6 +290,58 @@ def _find_first_transactions(
...
+    # TODO: Support the various units
+    divisor = timedelta(days=1) * time_scaler
+
+    additional_cols = (
+        [] if monetary_value_col is None else [nw.col(monetary_value_col).mean()]
+    )
+
+    customers = (
+        nw.from_native(repeated_transactions)
+        .group_by(customer_id_col)
```
FBruzzesi, line 326: For some time now, it should be possible to pass an expression so that you can avoid the renaming down in the pipeline, but it's definitely more of a personal preference 😇

```suggestion
        .group_by(nw.col(customer_id_col).alias("customer_id"))
```

ColtAllen:
> Maybe @ColtAllen is interested in taking this over?

We're still roadmapping what will be included in the [version 1.0](https://github.com/pymc-labs/pymc-marketing/issues/2108) release, but I think this could be a great addition to it!

PR #1432, opened 2025-01-24
Pass arbitrary HSGP to MMM class

williambdean:
## Description

WIP but the API would look like this:

```python
hsgp = SoftPlusHSGP.parameterize_from_data(
    np.arange(X.shape[0]),
    dims=("date", "channel"),
)

mmm = MMM(
    adstock=GeometricAdstock(l_max=8),
    saturation=LogisticSaturation(),
    date_column="date_week",
    channel_columns=["x1", "x2"],
    control_columns=[
        "event_1",
        "event_2",
        "t",
    ],
    time_varying_media=hsgp,
    yearly_seasonality=2,
)
mmm.fit(X, y)
```

Anything that follows the HSGPLike protocol can be used. This can be GPs or other time varying functions

```python
@runtime_checkable
class HSGPLike(Protocol):
    """Quacks like a HSGP."""

    dims: tuple[str, ...]

    def create_variable(self, name: str) -> TensorVariable:
        """Create a variable."""

    def register_data(self, X: TensorLike) -> Self:
        """Register the data."""
        ...
```

CC: @cetagostini

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--1432.org.readthedocs.build/en/1432/

PR #1832, opened 2025-07-16
Deprecate and disable budget optimization method

cetagostini:
## Description

Marked the budget optimization method as deprecated and replaced its implementation with a NotImplementedError. Users are advised to migrate to the `Multidimensal.MMM` class.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -2336,9 +2336,9 @@ class MMM(
         tuple[DataArray, OptimizeResult]
         | tuple[DataArray, OptimizeResult, list[dict[str, Any]]]
     ):
-        """Optimize the given budget based on the specified utility function over a specified time period.
+        """Optimize the given budget based on the specified utility function over a specified time period (DEPRECATED).
 
-        This function optimizes the allocation of a given budget across different channels
+        DEPRECATED: This function optimizes the allocation of a given budget across different channels
         to maximize the response, considering adstock and saturation effects. It scales the
         budget and budget bounds, performs the optimization, and generates a synthetic dataset
         for posterior predictive sampling.
@@ -2390,22 +2390,9 @@ class MMM(
         ValueError
             If the noise level is not a float.
         """
-        from pymc_marketing.mmm.budget_optimizer import BudgetOptimizer
-
-        allocator = BudgetOptimizer(
-            num_periods=num_periods,
-            utility_function=utility_function,
-            response_variable=response_variable,
-            custom_constraints=constraints,
-            default_constraints=default_constraints,
-            model=self,
-        )
-
-        return allocator.allocate_budget(
-            total_budget=budget,
-            budget_bounds=budget_bounds,
-            callback=callback,
-            **minimize_kwargs,
+        raise NotImplementedError(
+            "This method is deprecated and no longer available. "
+            "Please migrate to the `Multidimensal.MMM` class."
         )
 
     def plot_budget_allocation(


pymc_marketing/mmm/mmm.py
```diff
@@ -2390,22 +2390,9 @@ def optimize_budget(
...
-            default_constraints=default_constraints,
-            model=self,
-        )
-
-        return allocator.allocate_budget(
-            total_budget=budget,
-            budget_bounds=budget_bounds,
-            callback=callback,
-            **minimize_kwargs,
+        raise NotImplementedError(
```
williambdean: can we warn instead of raise?

pymc_marketing/mmm/mmm.py
```diff
@@ -2336,9 +2336,9 @@ def optimize_budget(
         tuple[DataArray, OptimizeResult]
         | tuple[DataArray, OptimizeResult, list[dict[str, Any]]]
     ):
-        """Optimize the given budget based on the specified utility function over a specified time period.
+        """Optimize the given budget based on the specified utility function over a specified time period (DEPRECATED).
 
-        This function optimizes the allocation of a given budget across different channels
+        DEPRECATED: This function optimizes the allocation of a given budget across different channels
```
williambdean: There is a specific way to call out in numpydocs

https://numpydoc.readthedocs.io/en/latest/format.html#deprecation-warning

pymc_marketing/mmm/mmm.py
cetagostini: I think better raise, only because the current method even if it works will return incorrect results. No sense to keep it, but what do you think?

pymc_marketing/mmm/mmm.py
cetagostini: I'll do this!

cetagostini:
This complements https://github.com/pymc-labs/pymc-marketing/pull/1849 deprecating method, meanwhile 1849 remove and update notebooks and dependencies.

juanitorduz:
@cetagostini i think its better to warn on the deprecation and remove in in 0.17.0

People are running these models in production and this is a key method. We will remove it for sure, but I think we need to do it in two steps.

Thoughts?

juanitorduz:
In the meantime… while warning. Can we somehow use the optimizer from the mmm multi clas I the old mmm class?

PabloRoque:
> @cetagostini i think its better to warn on the deprecation and remove in in 0.17.0
> 
> People are running these models in production and this is a key method. We will remove it for sure, but I think we need to do it in two steps.
> 
> Thoughts?

Users can always pin to 0.15.1. Maybe raise the error and tell them to use 0.15.1?

cetagostini:
> In the meantime… while warning. Can we somehow use the optimizer from the mmm multi clas I the old mmm class?

So, I could work on a PR to modify MMM base model to work as multidimensional, and be okay with the optimizer again. Nevertheless, probably I agree with Pablo that users who wants to continue with MMM base can always pin `15.1`. But I follow your lead here @juanitorduz 

You dictate the course and I'll follow capitan!

juanitorduz:
I mean, I think we can wait a week (or two), there is really no rush. It would be great to modify MMM base model to work as multidimensional, if possible, in view that I know many people looking into the end-to-end notebook:  https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_case_study.html. Also, the MMM multidim still does not fully support the HSGP (see https://github.com/pymc-labs/pymc-marketing/pull/1432). Hence, if we remove the optimizer, users won't be able to have a single model to fit and optimize. 

Hence, I suggest we modify MMM base model to work as multidimensional. No rush. Let me know if I can help. 

Does it make sense?

cetagostini:
Okay, I'll move this to draft, and will try to crack a PR for that during the next few days!

juanitorduz:
> I need to get #1862 merge before continue here.

I just merged https://github.com/pymc-labs/pymc-marketing/pull/1862 ! Thanks 🚀

cetagostini:
This looks ready, but the notebook get issue. I feel it's related to the `nc` loaded from my env, can you give me a hand here? and push the one you have @juanitorduz ? That should be all.

Thank you!

juanitorduz:
@carlosagostini  can you please comment out the line `mmm.save("multidimensional_model.nc")` in the https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_multidimensional_example.html#save-model notebook? I think this might be causing the problem as the model is saved every time the notebook tests are run (we had a similar issue in the past with the normal mmm notebook)

PR #1841, opened 2025-07-23
Add support for time-based budget distribution in optimizer

cetagostini:
## Description

Introduces the `budget_distribution_over_period` parameter to BudgetOptimizer and MultiDimensionalBudgetOptimizerWrapper, allowing users to specify how budgets are distributed across time periods for each budget dimension. Adds validation, processing, and application logic for these factors, and includes comprehensive tests for correct and incorrect usage, integration, and multidimensional scenarios.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1841.org.readthedocs.build/en/1841/

### DIFF
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -123,6 +123,10 @@ class BudgetOptimizer(BaseModel):
         Custom constraints for the optimizer.
     default_constraints : bool, optional
         Whether to add a default sum constraint on the total budget. Default is True.
+    budget_distribution_over_period : xarray.DataArray, optional
+        Distribution factors for budget allocation over time. Should have dims ("date", *budget_dims)
+        where date dimension has length num_periods. Values along date dimension should sum to 1 for
+        each combination of other dimensions. If None, budget is distributed evenly across periods.
     """
 
     num_periods: int = Field(
@@ -163,6 +167,15 @@ class BudgetOptimizer(BaseModel):
     default_constraints: bool = Field(
         default=True,
         description="Whether to add a default sum constraint on the total budget.",
+    )
+
+    budget_distribution_over_period: DataArray | None = Field(
+        default=None,
+        description=(
+            "Distribution factors for budget allocation over time. Should have dims ('date', *budget_dims) "
+            "where date dimension has length num_periods. Values along date dimension should sum to 1 for "
+            "each combination of other dimensions. If None, budget is distributed evenly across periods."
+        ),
     )
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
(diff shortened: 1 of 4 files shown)


pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -268,6 +291,125 @@ def set_constraints(self, constraints, default=None) -> None:
...
+
+        # Get the optimized budget values
+        bool_mask = np.asarray(self.budgets_to_optimize).astype(bool)
+        budgets_optimized = budgets[bool_mask]  # Shape: (num_optimized_budgets,)
+
+        # Now multiply budgets by time factors
+        budgets_expanded = pt.expand_dims(
+            budgets_optimized, 0
+        )  # Shape: (1, num_optimized_budgets)
+        repeated_budgets_flat = (
```
daniel-saunders-phil, line 399: Maybe related to the duplicate over periods / divide over periods point we were discussing Monday. If I'm reading this code correctly, this would divide the budget out across periods. So if before I was going to spend 10 dollars a day for 3 days, i get this:

[10, 10, 10] * [1, 1, 1] ->  [10, 10, 10]

if I put a flighting pattern, I would get 

[10, 10, 10] * [0.6, 0.3, 0.1] ->  [6, 3, 1]

so less total money is flowing through the optimizer if I specify the flighting pattern. Is that the correct read on this code? 

If we want to keep the behaviour consistent between the two ways of spreading a budget out over time, we could multiply the flighting pattern by the number of periods. Otherwise, I think we should divide out the budget in the base case as well. Dividing it out is more intuitive but it would be a surprising change too for existing users.

tests/mmm/test_budget_optimizer_multidimensional.py
```diff
@@ -53,31 +53,46 @@ def dummy_df():
...
 
     mmm = MMM(
         adstock=GeometricAdstock(l_max=4),
         saturation=LogisticSaturation(),
         **df_kwargs,
     )
+
     mmm.build_model(X=X_dummy, y=y_dummy)
+
+    # Fit the model once
```
daniel-saunders-phil, line 69: Does this library have some magic so this just causes a mock sample? Or does it run true MCMC?

tests/mmm/test_budget_optimizer.py
```diff
@@ -630,3 +630,275 @@ def test_mmm_optimize_budget_callback_parametrized(dummy_df, dummy_idata, callba
     assert hasattr(opt_result, "success")
     assert hasattr(opt_result, "x")
     assert hasattr(opt_result, "fun")
+
+
+@pytest.mark.parametrize(
```
daniel-saunders-phil, line 635: Do we test for correctness of the optimization result in this library? Probably too heavy of a lift if that's not been the practice. But it would be really nice to have a test that demonstrates the optimizer arrives at the correct result with and without flighting.

tests/mmm/test_budget_optimizer.py
```diff
@@ -630,3 +630,275 @@ def test_mmm_optimize_budget_callback_parametrized(dummy_df, dummy_idata, callba
...
+    "budget_distribution_over_period, num_periods, should_error, error_message",
+    [
+        # Valid case: uniform distribution
+        (
+            {
+                "channel_1": [0.25, 0.25, 0.25, 0.25],
```
daniel-saunders-phil, line 641: Slightly different angle is to just test that the optimizer arrives at the same result when you put in a uniform flight and when you don't pass a flight.

pymc_marketing/mmm/budget_optimizer.py
cetagostini-wise: I'll make a check, and you are probably right. Let me think about this.

tests/mmm/test_budget_optimizer_multidimensional.py
cetagostini-wise: Fitting a model by now, got issues doing mock `idata` with different dimensions, it generates to much troubles. Could be a PR to improve the tests.

tests/mmm/test_budget_optimizer.py
cetagostini-wise: Agree, I'll keep it separate from this PR, given the fact we'll need mock idata. Creating issue for it.

tests/mmm/test_budget_optimizer.py
cetagostini-wise: Yes, I just wanna to make quite evident here, we are using flat.

pymc_marketing/mmm/budget_optimizer.py
cetagostini: I applied your first recommendation, good catch! That should do the trick by now, and validate with a test :)

tests/mmm/test_budget_optimizer.py
daniel-saunders-phil: I got an issue opened #1846

pymc_marketing/mmm/budget_optimizer.py
daniel-saunders-phil: Okay sounds good.

daniel-saunders-phil (approved):

Issue #1835, opened 2025-07-20
Predicted Target ≠ Sum of Contributions OOS period (Case Study Notebook)


RamiFisherTW:
(Reopen of issue 1826)

I'm experiencing a discrepancy between direct predictions and reconstructed predictions from model components when using sample_posterior_predictive.
Steps to Reproduce:

1. Generate test predictions with component decomposition:

```
y_pred_test = mmm.sample_posterior_predictive(
    X_pred=X_test[0:1],
    include_last_observations=True,
    original_scale=True,
    var_names=["y", "channel_contribution", "intercept", "fourier_contribution", 
               "control_contribution", "yearly_seasonality_contribution"],
    extend_idata=False,
    progressbar=True,
    random_seed=rng
)
```
2. Decompose components and reconstruct total prediction:
```

# Extract component contributions
control_contrib_pred_test = y_pred_test['control_contribution'].mean(dim=["sample"]).sum()
channel_contrib_pred_test = y_pred_test['channel_contribution'].mean(dim=["sample"]).sum()
intercept_pred_test = y_pred_test['intercept'].mean(dim=["sample"]) * mmm.get_target_transformer()["scaler"].scale_[0]
seasonality_pred_test = y_pred_test['yearly_seasonality_contribution'].mean(dim=["sample"]).sum()

# Reconstruct total prediction
total_prediction_test = control_contrib_pred_test + channel_contrib_pred_test + intercept_pred_test + seasonality_pred_test
```
Issue:

Reconstructed sum: 67,879,827.31
Direct prediction (y_pred_test['y'].mean(dim=["sample"]).sum()): 72,591,532.16
Difference: ~4,711,704.85

When using sample_response_distribution with the same allocation strategy:
```

response = mmm.sample_response_distribution(
    allocation_strategy=allocation_strategy,  # Using X_test[0] budget
    time_granularity=model_granularity,
    num_periods=num_periods,
    noise_level=0.0,
)
```
I get the same channel contribution value (27,098,727.90), which seems consistent.
Questions:

Why is there a discrepancy between direct prediction and the sum of components?
Are there any missing components in the reconstruction? - (It seems to be the intercept)
Why does `budget_optimize` only return channel contribution as the optimized value when I set 'total_contribution' as the response?

Any insights would be appreciated!

juanitorduz:
Can you please try with 

y_out_of_sample_with_adstock = mmm.sample_posterior_predictive(
[...]

PR #1845, opened 2025-07-23
MMM: Fix Scaling Intercept

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/1835

The problem comes from 

https://github.com/pymc-labs/pymc-marketing/blob/1f259a8c1c9e247f138cc6f8aee0f1e7a7f0cb06/pymc_marketing/mmm/mmm.py#L1986-L1991

as the intercept component does not have a "date" coordinate.

----
📚 Documentation preview 📚: https://pymc-marketing--1845.org.readthedocs.build/en/1845/

### DIFF
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -1983,12 +1983,34 @@ class MMM(
                 date=slice(self.adstock.l_max, None)
             )
 
+        intercept_condition = (
+            "intercept" in posterior_predictive_samples.data_vars
+            and not self.time_varying_intercept
+        )
+
         if original_scale:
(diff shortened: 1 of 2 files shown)


juanitorduz:
> @juanitorduz Didn't have the time to check in detail #1835, but every time I see `sample_posterior_predictive` with `var_names` different from the y/obs I am always extra careful. You are basically resampling the posterior conditioning on parent nodes, and that tends to be tricky if not handled with care.

Thanks @PabloRoque indeed. This PR at least makes the intercept calculation more reasonable. It is clearly broken at the moment :( (I also checked the other components like `channel_contribution`, ...)

williambdean (approved):

PR #1851, opened 2025-07-25
fix(clv): Improve validation for missing columns in CLVModel

aaron1-z:
## Description

This PR improves the column validation check in the base `CLVModel`. Previously, the model would raise an error for the first missing column it found. This change modifies the logic to find all missing columns at once and raises a single, informative `ValueError` that lists all of them. This provides a much better user experience when the input data is incomplete.

**Note for maintainers:** I've implemented the fix and added a corresponding test for the improved validation. I encountered significant local C++ compiler issues on my Windows machine which prevented the full test suite from passing on my end, but the Python changes and the new test itself are correct and should pass in the CI environment.

----
📚 Documentation preview 📚: https://pymc-marketing--1851.org.readthedocs.build/en/1851/

### DIFF
--- a/tests/clv/models/test_basic.py
+++ b/tests/clv/models/test_basic.py
@@ -274,14 +274,16 @@ class TestCLVModel:
             model = CLVModelTest(model_config=old_model_config)
 
         assert model.model_config == {"x": Prior("Normal", mu=0, sigma=1)}
-   
+
     def test_validate_cols_reports_all_missing_columns(self):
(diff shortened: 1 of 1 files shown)


juanitorduz:
pre-commit autofix

juanitorduz:
pre-commit.ci autofix

juanitorduz (approved):
Thank you @aaron1-z

PR #1894, opened 2025-08-18
Support HSGP instance for time_varying_intercept in MMM

### LINKED ISSUE #1874: HalfNormal prior on intercept does not guarantee positive time-varying intercept values due to internal reparametrization
The pymc-marketing documentation recommends using a HalfNormal prior on the intercept to ensure positive values. However, this does not work as expected when time_varying_intercept=True due to internal model transformations. This creates confusion for users trying to enforce business-logical constraints.
### The Issue:
The documentation suggests:
``` mmm.model_config["intercept"] = Prior("HalfNormal", sigma=5)```
However, when using time_varying_intercept=True, the reported intercept values can still be negative. According to my reaserch this happens because:

1. The HalfNormal prior is placed on an unconstrained internal parameter
2.  A log transform is applied internally
3. The model is centered such that the mean of all linear components is zero
4. The derived intercept that gets reported to users is shifted by other model components

So while the underlying parameter respects the HalfNormal constraint, the actual intercept values users see and work with can be negative.
Why This Matters:

**Documentation is misleading:** Users following the documentation expect positive intercepts but don't get them
**Business logic violation:** In media mix modeling, baseline sales (intercept) should be non-negative

### Minimal Example:
```
model_config = {
    "intercept": Prior(HalfNormal, sigma=2)
}

mmm = MMM(
    date_column="date",
    channel_columns=source_columns,
    adstock=adstock,
    saturation=saturation,
    model_config=model_config,
    time_varying_intercept=True,
)

# Fit model
mmm.fit(X, y)

# Check intercept values - can contain negative values despite HalfNormal prior
posterior_intercept = mmm.posterior["intercept"].values
print(f"Min intercept value: {posterior_intercept.min()}")  # Can be negative!
```

### Current Workaround:
* Post-process results (not ideal for uncertainty quantification)
* Write custom models (defeats the purpose of the convenient API)
* Accept negative intercepts (violates business logic)
* **I tried also adding addidional transform, such as `softplus` or `exp` but it also does not work.**

Documentation Reference:
[[Link to the specific documentation section that recommends using HalfNormal for positive intercepts]
](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_tvp_example.html)

Environment:
pymc-marketing version: 0.14 / 0.15.1

### PR DESCRIPTION
**Pull request by cetagostini**
## Description

Allow passing an HSGP instance to the time_varying_intercept parameter in the MMM class, enabling custom dimensions and priors for the intercept's latent process. Added corresponding tests to verify correct behavior for both single and multi-dimensional cases.

## Related Issue

- [X] Closes #1874 
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1894.org.readthedocs.build/en/1894/

### DIFF
--- a/pymc_marketing/mmm/multidimensional.py
+++ b/pymc_marketing/mmm/multidimensional.py
@@ -508,9 +508,7 @@ class MMM(ModelBuilder):
         names = []
         if self.time_varying_intercept:
             names.extend(
-                SoftPlusHSGP.deterministics_to_replace(
-                    "intercept_temporal_latent_multiplier"
-                )
+                SoftPlusHSGP.deterministics_to_replace("intercept_latent_process")
             )
         if self.time_varying_media:
             names.extend(
@@ -1139,7 +1137,7 @@ class MMM(ModelBuilder):
                 # Register internal time index and build latent process
                 self.time_varying_intercept.register_data(time_index)
                 intercept_latent_process = self.time_varying_intercept.create_variable(
-                    "intercept_temporal_latent_multiplier"
+                    "intercept_latent_process"
                 )
 
                 intercept = pm.Deterministic(
--- a/tests/mmm/test_multidimensional.py
+++ b/tests/mmm/test_multidimensional.py
@@ -565,7 +565,7 @@ def test_time_varying_intercept_with_custom_hsgp_single_dim(single_dim_data, hsg
 
     mmm.build_model(X, y)
 
-    var_name = "intercept_temporal_latent_multiplier"
+    var_name = "intercept_latent_process"
     assert var_name in mmm.model.named_vars
     latent_dims = mmm.model.named_vars_to_dims[var_name]
     assert latent_dims == hsgp_dims
@@ -634,7 +634,7 @@ def test_time_varying_intercept_with_custom_hsgp_single_dim_kernels(
 
     mmm.build_model(X, y)
 
-    var_name = "intercept_temporal_latent_multiplier"
+    var_name = "intercept_latent_process"
     assert var_name in mmm.model.named_vars
     assert mmm.model.named_vars_to_dims[var_name] == ("date",)
 
@@ -712,7 +712,7 @@ def test_time_varying_intercept_with_custom_hsgp_multi_dim(df, hsgp_dims):
 
     mmm.build_model(X, y)
 
-    var_name = "intercept_temporal_latent_multiplier"
+    var_name = "intercept_latent_process"
     assert var_name in mmm.model.named_vars
     latent_dims = mmm.model.named_vars_to_dims[var_name]
     assert latent_dims == hsgp_dims
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -268,9 +268,7 @@ class BaseMMM(BaseValidateMMM):
         names = []
         if self.time_varying_intercept:
             names.extend(
-                SoftPlusHSGP.deterministics_to_replace(
-                    "intercept_temporal_latent_multiplier"
-                )
+                SoftPlusHSGP.deterministics_to_replace("intercept_latent_process")
             )
         if self.time_varying_media:
             names.extend(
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: pymc_marketing/mmm/multidimensional.py [hunk 1]
```diff
@@ -1106,17 +1111,36 @@ def build_model(
...
                     dims=("date", *self.dims),
                     **self.model_config["intercept_tvp_config"],
                 ).create_variable("intercept_latent_process")
 
+                intercept = pm.Deterministic(
+                    name="intercept_contribution",
+                    var=intercept_baseline[None, ...] * intercept_latent_process,
+                    dims=("date", *self.dims),
+                )
+            elif isinstance(self.time_varying_intercept, HSGPBase):
```
juanitorduz: nit : add space row before `elif` :D

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1106,17 +1111,36 @@ def build_model(
...
+                intercept_baseline = self.model_config["intercept"].create_variable(
+                    "intercept_baseline"
+                )
+
+                # Register internal time index and build latent process
+                self.time_varying_intercept.register_data(time_index)
+                intercept_latent_process = self.time_varying_intercept.create_variable(
+                    "intercept_temporal_latent_multiplier"
```
juanitorduz: Is it a `latent multiplier`? I thought this was a simple additive GP 🤔 ?

tests/mmm/test_multidimensional.py
```diff
@@ -539,6 +539,38 @@ def test_time_varying_media_with_custom_hsgp_single_dim(single_dim_data, hsgp_di
...
+    "hsgp_dims",
+    [
+        pytest.param(("date",), id="hsgp-dims=date"),
+    ],
+)
+def test_time_varying_intercept_with_custom_hsgp_single_dim(single_dim_data, hsgp_dims):
+    """Ensure passing an HSGP instance to time_varying_intercept works (single/multi-dim)."""
+    X, y = single_dim_data
+
+    hsgp = SoftPlusHSGP.parameterize_from_data(
```
juanitorduz, line 552: Can we also add custom kernels, say Marten5/2?

pymc_marketing/mmm/multidimensional.py
cetagostini: Yes, have been always multiplying the intercept. You can check the previous!

tests/mmm/test_multidimensional.py
cetagostini: Sure!

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1106,17 +1111,37 @@ def build_model(
...
+
+            elif isinstance(self.time_varying_intercept, HSGPBase):
+                intercept_baseline = self.model_config["intercept"].create_variable(
+                    "intercept_baseline"
+                )
+
+                # Register internal time index and build latent process
+                self.time_varying_intercept.register_data(time_index)
+                intercept_latent_process = self.time_varying_intercept.create_variable(
+                    "intercept_temporal_latent_multiplier"
```
PabloRoque: I like better `intercept_temporal_latent_multiplier`, but the naming is inconsistent with `intercept_latent_process` in the case in which we don't pass a custom HSGP. 

Shouldn't we aim to have the same names, so analysis is consistent downstream?

PabloRoque (approved):
LGTM!

Left minor comment on the naming consistency of the variable.

pymc_marketing/mmm/hsgp.py
```diff
@@ -1428,5 +1429,6 @@ def create_variable(self, name: str) -> TensorVariable:
         f_mean_name = f"{name}_f_mean"
         f_mean = pm.Deterministic(f_mean_name, f.mean(axis=0), dims=f_mean_dims)
 
-        centered_f = f - f_mean + 1
+        # Multiplicative centering to preserve positivity and enforce mean 1
+        centered_f = f / f_mean
```
PabloRoque, line 1433: Indeed

PabloRoque (approved):

pymc_marketing/mmm/multidimensional.py
juanitorduz: Yes, let's aim for consistency :)

cetagostini: Done!

PR #1899, opened 2025-08-19
Refactor and enhance SensitivityAnalysis for MMM

cetagostini:
## Description

Expanded SensitivityAnalysis to support arbitrary PyMC models, improved documentation with usage examples, and added flexible sweep and filtering logic. The class now works directly with PyMC models and InferenceData, supports multi-dimensional inputs, and provides Jacobian-based marginal effects with optional xarray output and idata extension. 

Legacy code using pandas was removed in favor of a more general and robust approach.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/mmm/sensitivity_analysis.py
+++ b/pymc_marketing/mmm/sensitivity_analysis.py
@@ -12,121 +12,315 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-"""Counterfactual sweeps for Marketing Mix Models (MMM)."""
+"""Counterfactual sweeps for Marketing Mix Models (MMM).
+
+Quick start
+-----------
+
+SensitivityAnalysis can work with any PyMC model whose response depends on an
+input `pm.Data` tensor. The functional relationship between inputs and the
+response can be arbitrary (e.g., adstock, saturation, splines, custom
+PyTensor ops). You only need to provide:
+
+- the name of the input data variable (`varinput`, e.g., `"channel_data"`),
+- the name of the deterministic/response variable you want to analyze
+  (`var_names`, e.g., `"channel_contribution"`),
+- a grid of `sweep_values`, and the sweep type.
+
+Example
+-------
+The example below uses an arbitrary saturation transformation, but any
+PyTensor-compatible transformation will work the same way.
+
+.. code-block:: python
+
+    import numpy as np
+    import pymc as pm
+    from pymc_marketing.mmm.sensitivity_analysis import SensitivityAnalysis
+
+
+    def saturation(x, alpha, lam):
+        return (alpha * x) / (x + lam)
+
+
+    coords = {"date": np.arange(52), "channel": list("abcde")}
+    with pm.Model(coords=coords) as m:
+        X = pm.Data(
+            "channel_data",
+            np.random.Gamma(13, 12, size=(52, 5)),
+            dims=("date", "channel"),
+        )
+        alpha = pm.Gamma("alpha", 1, 1, dims="channel")
+        lam = pm.Gamma("lam", 1, 1, dims="channel")
+        contrib = pm.Deterministic(
+            "channel_contribution", saturation(X, alpha, lam), dims=("date", "channel")
+        )
+        mu = pm.Normal("intercept", 0.0, 1.0) + contrib.sum(axis=-1)
+        pm.Normal("likelihood", mu=mu, sigma=1.0, dims=("date",))
+        idata = pm.sample(1000, tune=500, chains=2, target_accept=0.95, random_seed=42)
+
+    sa = SensitivityAnalysis(m, idata)
+    sweeps = np.linspace(0.2, 3.0, 8)
+    # Jacobian-based effects with respect to the input; shape: (sample, sweep, channel)
+    result = sa.run_sweep(
+        sweep_values=sweeps,
+        varinput="channel_data",
+        var_names="channel_contribution",
+        sweep_type="multiplicative",
+    )
+    # Optional: finite-difference marginals along the sweep axis
+    me = SensitivityAnalysis.compute_marginal_effects(result, sweeps)
+
+Notes
+-----
+- Arbitrary models: As long as the response graph depends (directly or indirectly)
+  on the `pm.Data` provided via `varinput`, and you pass the name of the
+  deterministic/response via `var_names`, the class builds the Jacobian and
+  evaluates it across sweeps automatically.
+- Multi-dimensional inputs: If `varinput` has dims like `(date, country, channel)`,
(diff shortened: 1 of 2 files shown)


pymc_marketing/mmm/sensitivity_analysis.py
```diff
@@ -12,121 +12,315 @@
...
+        """
+        Run sweeps and compute marginal effects.
 
         Parameters
         ----------
-        var_names : list[str]
-            List of variable names to intervene on.
+        varinput : str
+            Name of the pm.Data variable (e.g., "channel_data").
+            Expected shape: (date, *dims, channel)
```
cetagostini: I'll change here, should say: `(date, *dims, arbitrary_dim)` that match `varinput` dims.

pymc_marketing/mmm/sensitivity_analysis.py
```diff
@@ -12,121 +12,315 @@
...
+        stacked = np.stack(outs, axis=1)  # (n_samples, n_sweeps, *dims, channel)
+        xr_result = self._transform_output_to_xarray(
+            stacked,
+            sweep_values=sweep_values,
+            dims_order=dims_order,
+            var_names_filter=var_names_filter,
+        )
+        if extend_idata:
+            self._add_to_idata(xr_result)
+            return stacked
```
cetagostini: Here we should return either the xr result or none. I'll change this.

pymc_marketing/mmm/sensitivity_analysis.py
```diff
@@ -12,121 +12,315 @@
...
+                    indices.append(int(coord_to_index[req]))
+            return [base_labels[i] for i in indices]
+
+        # Assemble coords aligned with dims
+        coords: dict[str, np.ndarray | list] = {}
+        coords["sample"] = np.arange(result.shape[0])
+        coords["sweep"] = np.asarray(sweep_values)
+
+        # Map remaining dims using model coords and filters
+        # Compute axis offsets: after (sample, sweep), axes align with dims_order
```
PabloRoque, line 257: I tend to dislike `combined=True` and `sample` instead of the usual `('chain', 'draw')` because at times there are issues down the line if you want to merge xarrays.

pymc_marketing/mmm/sensitivity_analysis.py
```diff
@@ -12,121 +12,317 @@
...
-        if self.sweep_type == "multiplicative":
-            for var_name in self.var_names:
-                X_new[var_name] *= sweep_value
-        elif self.sweep_type == "additive":
-            for var_name in self.var_names:
-                X_new[var_name] += sweep_value
-        elif self.sweep_type == "absolute":
-            for var_name in self.var_names:
-                X_new[var_name] = sweep_value
+        # 2) Sum over date dimension → shape (samples, *dims, channel)
```
ErikRingen: It looks like the summation from steps 2 and 4 can be combined no? I don't see why you would need to first sum the date axis

pymc_marketing/mmm/sensitivity_analysis.py
```diff
@@ -12,121 +12,317 @@
...
+            dims_order=dims_order,
+            var_names_filter=var_names_filter,
+        )
+        if extend_idata:
+            self._add_to_idata(xr_result)
+            return None
         else:
-            raise ValueError(f"Unsupported sweep_type: {self.sweep_type}")
-        return X_new
+            return xr_result
```
ErikRingen, line 438: So there has been a shift where previous "results" represented the counterfactual values of the response distribution, and then `compute_marginal_effects` gave the finite differences of those counterfactuals along the sweep values. But now you use pt jacobian to get the differences, which means that results are already the marginal effects, right? So the additional call to compute_marginal_effects() gives you now the 2nd derivative, which I don't think is as intended.
 
I think some of @drbenvincent's plots rely on having the counterfactual values that used to be in results.

cetagostini: Good catch, I'll do that change!

pymc_marketing/mmm/sensitivity_analysis.py
cetagostini: Good point, I'll do that.

drbenvincent (changes requested):
Just bear in mind that I am very very beginner when it comes to pytensor.

Here's a question - originally raised by @ErikRingen - but that I'm picking up on as well. As far as I can tell, this is going straight for the marginal effect with the Jacobian right? The original functionality was first computing what the outcome is for different sweep values _then_ computed the gradient as a function of the sweep values (the marginal effect).

In discussion with an LLM, we need something like this?

```python
# Add this to get actual values
resp_fn = function(inputs=[], outputs=resp_graph)

# Then in the sweep loop:
for s in sweep_values:
    data_shared.set_value(original_X * s)
    actual_values = resp_fn()  # Actual outcome values
    marginal_effects = jac_fn()  # Marginal effects (current implementation)
```

My only other thought (at this time) is that there seems to be some assumptions about dimensions etc. We don't have the useful named dimensions that we do in xarray, so perhaps we need some extra validation here?

Actually, another thought. This seems _very general_. So in theory this could feasibly be ported to `pymc-extras` or `pymc` to enable generic sensitivity / marginal effects for arbitrary models? Not saying we should do that now, but it's perhaps worth considering, and thinking about the feasibility?

pymc_marketing/mmm/plot.py
```diff
@@ -1479,213 +1480,444 @@ def allocated_contribution_by_channel_over_time(
         fig.tight_layout()
         return fig, axes
 
-    def plot_sensitivity_analysis(
+    def sensitivity_analysis(
```
williambdean, line 1483: Breaking change. Can we provide previous method as a wrapper to this and and warning

pymc_marketing/mmm/sensitivity_analysis.py
```diff
@@ -12,125 +12,509 @@
...
+            self.idata.add_groups({"sensitivity_analysis": dataset})
 
     def run_sweep(
         self,
-        var_names: list[str],
+        var_input: str,
         sweep_values: np.ndarray,
+        *,
+        var_names: str = "channel_contribution",
         sweep_type: Literal[
```
williambdean, line 295: Another breaking change

cetagostini:
> Various breaking changes here. How long have they existed / were they in releases?

A single release, sensitivity was release on 0.16 (currently on 0.17) but after this change we can patch an release a 0.17.1 - I don't see it mega critical, given is a single release probably not hardly adopted. I was trying to get last approval for 0.17.0 but this have been sitting for a while.

drbenvincent (approved):

PR #1904, opened 2025-08-21
Streamlit Dependency Fix

timbo112711:
## Description
The last dependency update fixed the TVP page, but introduced a new error on the Bayesian Prior Distribution Demonstrator page 😅. Decided to tarn down my env and build a new one to test the changed made here. All should be working now!! 

## Checklist


juanitorduz:
we can use pixi which is like UV with conda (as we need to install PyMC) 👍

PR #1911, opened 2025-08-29
Add more Basis implementations for use with EventEffect

TeemuSailynoja:
Add extra Basis functions for the EventEffect.

## Description
Implemented in this PR:

### HalfGaussianBasis
- Add a Gaussian on one side of an event.
- Parameter for creating the effect before/after the event.
- Parameter to control whether the event days should be included or excluded from the effect.

image

### AsymmetricGaussianBasis
- Adds a Gaussian for both sides of an event
- Parameter for controlling widths and amplitudes before/after the event

asym-example
asym-example2

## Related Issue
I jumped straight to implementation.

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1911.org.readthedocs.build/en/1911/

### DIFF
--- a/pymc_marketing/mmm/events.py
+++ b/pymc_marketing/mmm/events.py
@@ -355,7 +355,7 @@ class AsymmetricGaussianBasis(Basis):
 
     .. plot::
         :context: close-figs
-        
+
         import matplotlib.pyplot as plt
         from pymc_marketing.mmm.events import AsymmetricGaussianBasis
         from pymc_extras.prior import Prior


cetagostini:
Great, @timbo112711 nice if you can add tests and example to the notebook of components!

tests/mmm/test_events.py
```diff
@@ -489,3 +490,109 @@ def test_event_effect_dim_validation(sigma_dims, effect_dims) -> None:
 
     with pytest.raises(ValueError):
         EventEffect(basis=basis, effect_size=effect_size, dims="event")
+
+
+@pytest.mark.parametrize(
+    "mode, include_event",
+    [
+        pytest.param("after", True, id="after_include_event_true"),
+        pytest.param("before", True, id="before_include_event_true"),
```
cetagostini, line 502: Way better, thanks @TeemuSailynoja

timbo112711:
@cetagostini looks like @TeemuSailynoja beat me to the tests, much appreciated! I'll add an example to the notebook of components then we should be good here.

pymc_marketing/mmm/events.py
```diff
@@ -268,6 +268,146 @@ def function(self, x: pt.TensorLike, sigma: pt.TensorLike) -> TensorVariable:
...
+    def function(
+        self,
+        x: pt.TensorLike,
+        sigma_before: pt.TensorLike,
+        sigma_after: pt.TensorLike,
+        a_before: pt.TensorLike,
+        a_after: pt.TensorLike,
+    ) -> pt.TensorVariable:
+        """Asymmetric Gaussian bump function."""
+        match self.event_in:
```
timbo112711, line 407: @TeemuSailynoja nice modification of the cases here 🙌🏻!

juanitorduz:
Nice work @TeemuSailynoja ! Is this one ready for review ?

TeemuSailynoja:
Yes. I would like to add a usage example  that would be rendered on the docs page, but the implementation and tests should be ready.

TeemuSailynoja:
I just realized two things:

1. As these basis functions are intended to be wrapped in an EventEffect (correct me if this is not the case), which includes an `effect_size` parameter, we should probably have AsymmetricGaussianBasis have one of the amplitudes fixed to one.

2. the `drop` boolean is redundant, as the amplitudes can just be given priors with opposite signs. I could remove that from the class.

timbo112711:
@TeemuSailynoja 

> 1. As these basis functions are intended to be wrapped in an EventEffect (correct me if this is not the case), which includes an `effect_size` parameter, we should probably have AsymmetricGaussianBasis have one of the amplitudes fixed to one. This would avoid a three way unidentifiability between the parameters.

That is correct, this is intended to be wrapped in an `EventEffect`. This is a good point, if we leave it how it is we pose the rise of multiple parameter settings give the same likelihood. Setting one of the amplitudes to 1 will help by

- Having basis just encode the shape asymmetry before vs. after.

- The `effect_size` parameter in `EventEffect` governs the overall scale of the effect.

I agree with you in that having one of the amplitudes fixed to 1, not sure it matters which one however. Maybe we just set the `a_before` to 1. 

> 2. the `drop` boolean is redundant, as the amplitudes can just be given priors with opposite signs.

This makes sense! I'm cool with dropping this arg.

juanitorduz:
@TeemuSailynoja, anything to change in this one in view of @PabloRoque 's enhancement https://github.com/pymc-labs/pymc-marketing/pull/1912 ?

TeemuSailynoja:
> @TeemuSailynoja, anything to change in this one in view of @PabloRoque 's enhancement #1912 ?

@juanitorduz I changed these implementations to align with #1912 . I have come to agree that that is a better way to implement these :)

I don't see the example plots I added to the new Basis classes in the docs rendered by the readthedocs check/action. Should it be visible there?

pymc_marketing/mmm/events.py
```diff
@@ -321,7 +320,8 @@ def __init__(
 
     def function(self, x: pt.TensorLike, sigma: pt.TensorLike) -> TensorVariable:
         """One-sided Gaussian bump function."""
-        out = pm.math.exp(-0.5 * (x / sigma) ** 2)
+        rv = pm.Normal.dist(mu=0.0, sigma=sigma)
+        out = pm.math.exp(pm.logp(rv, x))
```
TeemuSailynoja, line 324: Here, I opted to have the HalfGaussian be literally half of a Gaussian, that is, it integrates to 1/2 instead of 1.
I figured as there is in any case an option to exclude the event date from the effect, and the values are only evaluated at discrete places, the total effect would most often not sum to `effect_size` anyway, so this is maybe the more intuitive implementation.

juanitorduz:
Amazing @TeemuSailynoja !

The plots are not render indeed. Can you try generating the docs locally? Or seeing if there is any little syntax detail that prevents the run? (I can also check tomorrow) :)

pymc_marketing/mmm/events.py
```diff
@@ -278,10 +278,10 @@ class HalfGaussianBasis(Basis):
 
         import matplotlib.pyplot as plt
         from pymc_marketing.mmm.events import HalfGaussianBasis
-        from pymc_extras.priors import Prior
+        from pymc_extras.prior import Prior
```
TeemuSailynoja, line 281: @juanitorduz, I found the issue. A bit of a facepalm moment. Excellent point to call it a night. Fingers crossed we see the plots in the docs now.

PR #1922, opened 2025-09-03
Pydantic `MuEffect`

PabloRoque:
## Description

Provides serialization/deserialization of MuEffect using Pydantic BaseModel

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/mmm/additive_effect.py
+++ b/pymc_marketing/mmm/additive_effect.py
@@ -104,12 +104,13 @@ Tips for custom components
   - In `set_data`, update the data variables when dates/dims change.
 """
 
-from typing import Any, Protocol
+from abc import ABC, abstractmethod
+from typing import Annotated, Any, Protocol
 
 import pandas as pd
 import pymc as pm
 import xarray as xr
-from pydantic import BaseModel, InstanceOf
+from pydantic import BaseModel, Field, InstanceOf, PlainValidator, WithJsonSchema
 from pymc_extras.prior import create_dim_handler
 from pytensor import tensor as pt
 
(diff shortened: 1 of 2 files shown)


pymc_marketing/mmm/additive_effect.py
```diff
@@ -131,35 +132,31 @@ def model(self) -> pm.Model:
...
-        fourier : FourierBase
-            The FourierBase instance to use for the effect.
-        date_dim_name : str, optional
-            The name of the date dimension in the model, by default "date".
-
-        """
-        self.fourier = fourier
-        self.date_dim_name: str = date_dim_name
+    fourier: InstanceOf[FourierBase]
+    date_dim_name: str = Field("date")
```
TeemuSailynoja, line 248: Do we need to supply the `date_dim_name` here, as the mutltidimensional MMM has the `date_column` attribute?

pymc_marketing/mmm/additive_effect.py
```diff
@@ -259,6 +263,8 @@ class LinearTrendEffect:
         The LinearTrend instance to wrap.
     prefix : str
         The prefix to use for variables in the model.
+    date_dim_name : str
```
TeemuSailynoja, line 361: Same `date_dim_name` question as with FourierEffect.

pymc_marketing/mmm/additive_effect.py
PabloRoque: yes, because `MuEffect` is just an interface that has nothing to do with an MMM class. 
So either we have this explicit `date_dim_name` or we need to introduce coupling with the MMM class.

TeemuSailynoja: I see, sorry, I was confused with the `mmm: Model` argument, but I do understand that it isn't restricted to MMMs.

TeemuSailynoja (approved):

pymc_marketing/mmm/multidimensional.py
```diff
@@ -205,6 +206,149 @@
 from pymc_marketing.model_config import parse_model_config
 from pymc_marketing.model_graph import deterministics_to_flat
 
+# ============================================================================
+# MuEffect Serialization Registry
+# ============================================================================
+# Uses singledispatch for serialization (type-based dispatch)
+# and a simple dict registry for deserialization (string-based dispatch)
```
juanitorduz: Maybe we remove these lines?

PabloRoque: You got it

juanitorduz (changes requested):
@PabloRoque, we are almost there :) Can you please look into the pre-commit issues (I think mypy is complaining)

PR #1937, opened 2025-09-10
Enable JAX backend in `BudgetOptimizer`

PabloRoque:
## Description

Enables computation of `BudgetOptimizer.objective_and_grad_func` via JAX backend. 
This enables GPU based computation if present (and enabled) through JAX envvars.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -363,6 +363,11 @@ class BudgetOptimizer(BaseModel):
         ),
     )
 
+    jax_backend: bool = Field(
+        default=False,
+        description="Boolean flag controlling jax backend. Needed for GPU support",
+    )
+
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     DEFAULT_MINIMIZE_KWARGS: ClassVar[dict] = {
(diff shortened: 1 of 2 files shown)


pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -684,10 +689,24 @@ def _compile_objective_and_grad(self):
         )
         objective_grad = pt.grad(rewrite_pregrad(objective), budgets_flat)
 
-        objective_and_grad_func = function([budgets_flat], [objective, objective_grad])
+        if self.compile_kwargs and (self.compile_kwargs["mode"]).lower() == "jax":
+            # Use PyMC's JAX infrastructure for robust compilation
+            from pymc.sampling.jax import get_jaxified_graph
```
ricardoV94, line 694: Why did you need to do this instead of just compiling the same pytensor function with jax backend?

PyMC does that to pass a pure jax function for jax-based samplers, but here we're just using Scipy minimize, so nothing special should be needed?

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -684,10 +689,24 @@ def _compile_objective_and_grad(self):
...
+
+            objective_and_grad_func = get_jaxified_graph(
```
ricardoV94, line 696: I don't think the JAX function returned by this helper is even jitted

pymc_marketing/mmm/budget_optimizer.py
cetagostini: Will pick this comments in a new PR!

PR #1939, opened 2025-09-12
Special priors

daniel-saunders-phil:
This is the code implementation of that positive-constrained prior that works well for hierarchies that I was telling a few folks about. Wrote up the rough draft in [this post](https://daniel-saunders-phil.github.io/imagination_machine/posts/a-positive-constrained-non-centered-prior-that-sparks-joy/) and finally got a production-ready version coded up.

We have to implement as it's own class instead of using the Prior class because it doesn't correspond to any pymc distribution and you cannot build it out of the transformations permitted by the Prior class.

I'm curious what people think of the strategy of introducing `special_priors.py`. I think there are quite a few odd distributions (Non-centered laplace for seasonality is sometimes way more efficient than centered laplace) that don't match the logic of the Prior class very well but would be nice to have in marketing projects. If we want to keep push that route, it might be nice to set up a base class with some abstract methods so we know what is required of all special priors.

In a future PR, I'll update the multi-dimensional notebook to use this class.

----
📚 Documentation preview 📚: https://pymc-marketing--1939.org.readthedocs.build/en/1939/

### DIFF
--- a/pymc_marketing/special_priors.py
+++ b/pymc_marketing/special_priors.py
@@ -39,7 +39,7 @@ class LogNormalPositiveParam:
 
     The other advantage of this prior is in constructing hierarchical models. It allows users to toggle
     between centered and non-centered parameterizations. This enables rapid iteration when searching
-    for a parameterization samples efficiently.
+    for a parameterization that samples efficiently.
 
     Parameters
     ----------
@@ -55,8 +55,8 @@ class LogNormalPositiveParam:
     Examples
     --------
     Build a non-centered hierarchical model where information is shared across geos.
-    .. code-block:: python
 
+    .. code-block:: python
         from pymc_marketing.special_priors import LogNormalPositiveParam
 
         normal = LogNormalPositiveParam(


juanitorduz (changes requested):
Just add this new module to the docs :)

daniel-saunders-phil:
> My two recommendations would be:
> 
> 1. Maybe better say "Custom" instead of "Special"?
> 2. If you want (Can be one issue and another PR), I feel separate to dict, from dict, sample prior should be easy in a class, then inherit, and only define `create_variable`.
> 
> I feel we could get from `pymc-extras` all the base class from prior even, and only define the custom logic in create variable, or not? @daniel-saunders-phil
> 
> cc: @williambdean

Thanks Carlos!

1. I think "custom" might add a bit of confusion. In other places on the docs, we use custom prior to refer to anything other than default priors. The ["How we compare" page](https://www.pymc-marketing.io/en/latest/guide/mmm/comparison.html) lists custom priors as a feature we already have. So I won't want new users to come clicking on custom priors, expecting something else.

2. Yeah it's a good idea but it should be follow up PR. We need to refactor on pymc-extras a bit before we could inherit methods like `to_dict`. The LogNormalPositiveParam method is not exactly the same as the Prior class's method.

PR #1943, opened 2025-09-16
Add ModelSamplerEstimator for PyMC Marketing model benchmarking

cetagostini:
## Description

Introduces the ModelSamplerEstimator utility to estimate computational time of a PyMC models using JAX/NumPyro, including logp/gradient evaluation time and NUTS step counts. The output should be the minimum sample time from their models. Doing so, users could compare the minimum amount of time that their given model will take.

Adds corresponding tests to validate estimator output and schema for simple and multidimensional models.

Example:
```python
est = ModelSamplerEstimator(
    tune=1000, draws=1000, chains=4, sequential_chains=1, seed=1
)
df = est.run(model)
```

> [!IMPORTANT]  
> Only working with JAX at the moment.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/pytensor_utils.py
+++ b/pymc_marketing/pytensor_utils.py
@@ -15,6 +15,7 @@
 """PyTensor utility functions."""
 
 import arviz as az
+import pandas as pd
 import pytensor.tensor as pt
 from arviz import InferenceData
 from pymc import Model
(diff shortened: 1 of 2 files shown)


cetagostini:
I have a few PRs, I'll add examples in the notebook on another PR :)

pymc_marketing/pytensor_utils.py
```diff
@@ -106,3 +107,236 @@ def extract_response_distribution(
     )
 
     return response_distribution
+
+
+class ModelSamplerEstimator:
```
juanitorduz, line 245: The name is confusing because it is not an estimator right? Also, the name must contain something referring to JAX or numpyro?

pymc_marketing/pytensor_utils.py
```diff
@@ -106,3 +107,236 @@ def extract_response_distribution(
...
+        sequential_chains: int = 1,
+        seed: int | None = None,
+    ) -> None:
+        self.tune = int(tune)
+        self.draws = int(draws)
+        self.chains = int(chains)
+        self.sequential_chains = int(sequential_chains)
+        self.seed = seed
+
+    def estimate_model_eval_time(self, model: Model, n: int | None = None) -> float:
```
juanitorduz: can we give `n` a more descriptive namr?

pymc_marketing/pytensor_utils.py
```diff
@@ -106,3 +107,236 @@ def extract_response_distribution(
...
+        num_samples = int(self.draws if draws is None else draws)
+
+        initial_point = list(model.initial_point().values())
+        logp_fn = get_jaxified_logp(model, negative_logp=False)
+        nuts_kernel = NUTS(
+            potential_fn=logp_fn,
+            target_accept_prob=0.8,
+            adapt_step_size=True,
+            adapt_mass_matrix=True,
+            dense_mass=False,
```
juanitorduz: can we pass these as kwargs?

pymc_marketing/pytensor_utils.py
```diff
@@ -106,3 +107,236 @@ def extract_response_distribution(
...
+        )
+
+        mcmc = MCMC(
+            nuts_kernel,
+            num_warmup=num_warmup,
+            num_samples=num_samples,
+            num_chains=1,
+            postprocess_fn=None,
+            chain_method="sequential",
+            progress_bar=False,
```
juanitorduz: same here (kwargs)

pymc_marketing/pytensor_utils.py
```diff
@@ -106,3 +107,236 @@ def extract_response_distribution(
...
+            "estimated_sampling_time_seconds": sampling_time_seconds,
+            "estimated_sampling_time_minutes": sampling_time_seconds / 60.0,
+            "estimated_sampling_time_hours": sampling_time_seconds / 3600.0,
+            "tune": int(self.tune),
+            "draws": int(self.draws),
+            "chains": int(self.chains),
+            "seed": int(self.seed) if self.seed is not None else None,
+            "timestamp": pd.Timestamp.utcfromtimestamp(int(time.time())),
+        }
+        df = pd.DataFrame([data])
```
juanitorduz: I think `pd.DataFrame(data)` still work as well (or data=data)?

pymc_marketing/pytensor_utils.py
cetagostini: What do you mean? We are using this to estimate the sampling time 👀

pymc_marketing/pytensor_utils.py
cetagostini: Done!

pymc_marketing/pytensor_utils.py
cetagostini: Done!

pymc_marketing/pytensor_utils.py
cetagostini: Done!

pymc_marketing/pytensor_utils.py
cetagostini: I'm keeping name open because this could get expand and not only use JAX, we can do the same with any backend. I don't want to close the name for it, and then do a "deprecation" or keep a new name warning or so.

pymc_marketing/pytensor_utils.py
cetagostini: Nop, with all scalars will raise “If using all scalar values, you must pass an index.”

cetagostini:
@juanitorduz feedback apply!

PR #1944, opened 2025-09-16
Add model merging utilities and tests for multi-model optimization

cetagostini:
## Description

Introduces functions and a wrapper class in `pytensor_utils.py` to merge multiple PyMC models with variable prefixing and shared variable support, enabling multi-model budget optimization. Updates `budget_optimizer.py` to use delayed imports to avoid circular dependencies. Adds tests for model merging and the new wrapper in `test_pytensor_utils.py`.
Combines multiple independent PyMC Models into a single Model.

### Example
As we move forward in causality, and discover new DAGs, a way to collect direct or indirect effects could be create different regressions, every one adjusting by the minimum set of variables. _Each regression will respond to the same DAG but will uncover different effects per channel depending on the adjustment_.

Nevertheless, we probably want to optimize all regressions together or we want to join the mode to make several operations. But how to? Well let's remember PyMC models are graphical generative model so, each model can be represent as a function graph, and their nodes/variables could be prefixed (e.g., model1_, model2_). Doing so, we could decide which variable or data will be the one to merge_on="some_var" and the variable is kept **unprefixed** and shared across the merged models. All outputs and coordinates are merged into a single function graph.

Imagine you have the following DAG:
```
digraph G {
  rankdir=LR; splines=true; nodesep=0.4; ranksep=0.5;
  node [shape=ellipse, fontsize=12];

  C1 -> X1; C1 -> Y;
  C2 -> X2; C2 -> Y;
  C3 -> X3; C3 -> Y;

  X1 -> X2; X2 -> X3;
  X1 -> Y;  X2 -> Y;  X3 -> Y;

  // styling
  {rank=same; C1; C2; C3}
  {rank=same; X1; X2; X3}
  Y [shape=doublecircle];
}
```

This structure makes the minimal sufficient sets different for each channel’s **direct** effect on Y:

- **For x1:** backdoor via C1; mediated paths via X2 to X3.  
  **Adjust:** C1,X2 (conditioning on X2 also blocks the X1 to X2 to X3 to Y path).

- **For X2:** backdoors via C2 and via X1 (since X1 to X2 and X1 to Y); mediated path via X3.  
  **Adjust:** C2, X1, X3.

- **For X3:** backdoors via C3 and via X2 (since X2 to X3 and X2 to Y).  
  **Adjust:** C_3, X_2.

Meaning, we need three models to uncover the effects:
```python
# Get x1 over Y
mask = np.array([True, False, False])
mmm_x1_effect = MMM(  # (Y ~ X1 + C1 + X2)
    date_column="date",
    target_column="Y",
    channel_columns=[
        "X1",
        "X2",
        "X3",
    ],  # always use all channels, but mask the ones that are not used
    control_columns=["X2", "C1"],
    adstock=NoAdstock(l_max=1),
    saturation=NoSaturation(
        priors={
            "beta": MaskedDist(
                Prior("HalfNormal", sigma=1.0, dims="channel"), mask=mask
            )
        }
    ),
)

# get x2 over Y
# make mask for channels making x1,x2 false
mask2 = np.array([False, True, False])
mmm_x2_effect = MMM(  # (Y ~ X2 + C2 + X1 + X3)
    date_column="date",
    target_column="Y",
    channel_columns=[
        "X1",
        "X2",
        "X3",
    ],  # always use all channels, but mask the ones that are not used
    control_columns=["X1", "C2", "X3"],
    adstock=NoAdstock(l_max=1),
    saturation=NoSaturation(
        priors={
            "beta": MaskedDist(
                Prior("HalfNormal", sigma=1.0, dims="channel"), mask=mask2
            )
        }
    ),
)

# get x3 over Y
# make mask for channels making x1,x2 false
mask3 = np.array([False, False, True])
mmm_x3_effect = MMM(  # (Y ~ X3 + C3 + X2)
    date_column="date",
    target_column="Y",
    channel_columns=[
        "X1",
        "X2",
        "X3",
    ],  # always use all channels, but mask the ones that are not used
    control_columns=["X2", "C3"],
    adstock=NoAdstock(l_max=1),
    saturation=NoSaturation(
        priors={
            "beta": MaskedDist(
                Prior("HalfNormal", sigma=1.0, dims="channel"), mask=mask3
            )
        }
    ),
)
```

As you can see we should be able to _zero out with mask the distributions to avoid calculate the contributions for the channels we don't want to adjust in the regression_. Nevertheless, even when channels are zero out given the priors and not consider during the sampling, the data input is the same for all.

We could merge based on it.

```python
optimizable_model1 = MultiDimensionalBudgetOptimizerWrapper(
    model=mmm_x1_effect, start_date="2026-01-01", end_date="2026-01-31"
)

optimizable_model2 = MultiDimensionalBudgetOptimizerWrapper(
    model=mmm_x2_effect, start_date="2026-01-01", end_date="2026-01-31"
)

optimizable_model3 = MultiDimensionalBudgetOptimizerWrapper(
    model=mmm_x3_effect, start_date="2026-01-01", end_date="2026-01-31"
)

multi_model = BuildMergedModel(
    models=[optimizable_model1, optimizable_model2, optimizable_model3],
    prefixes=["model1", "model2", "model3"],
    merge_on="channel_data",
)
```

And finally, optimize.

```python
optimizer = BudgetOptimizer(
    num_periods=multi_model.num_periods,
    model=multi_model,
    response_variable="total_media_contribution_original_scale",
)
```

This opens the door to make optimization of different effects (total, indirects or directs) together. Allows for multi-objective optimization. Every model could have a different target variable but the target depends on same inputs, and consequently we want to optimize A and B together.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/pytensor_utils.py
+++ b/pymc_marketing/pytensor_utils.py
@@ -14,18 +14,324 @@
 
 """PyTensor utility functions."""
 
+from typing import Any, cast
+
 import arviz as az
+import pytensor
 import pytensor.tensor as pt
+import xarray as xr
 from arviz import InferenceData
 from pymc import Model
+from pymc.model.fgraph import (
+    FunctionGraph,
+    ModelVar,
+    extract_dims,
+    fgraph_from_model,
+    model_from_fgraph,
+)
+from pymc.model.transform.optimization import freeze_dims_and_data
 from pytensor import clone_replace
 from pytensor.graph import rewrite_graph, vectorize_graph
 from pytensor.graph.basic import ancestors
 
 try:
+    # Prefer importing the Protocol when available without triggering circular import
+    from pymc_marketing.mmm.budget_optimizer import OptimizerCompatibleModelWrapper
+except Exception:  # pragma: no cover - during circular import bootstrap
+    from typing import Protocol as OptimizerCompatibleModelWrapper  # type: ignore
+
+try:
     from pymc.pytensorf import rvs_in_graph
 except ImportError:
     from pymc.logprob.utils import rvs_in_graph
+
+
+def _prefix_model(f2, prefix: str, exclude_vars: set | None = None):
+    """Prefix variable and dimension names in a FunctionGraph.
+
+    Variables listed in ``exclude_vars`` (e.g., a shared variable like ``channel_data``)
+    are kept unprefixed, and their dims/coords are also preserved without prefix.
+    """
+    if exclude_vars is None:
+        exclude_vars = set()
+
+    # First, collect dimensions that belong to excluded variables
+    exclude_dims = set()
+    for v in f2.outputs:
+        if v.name in exclude_vars:
+            v_dims = extract_dims(v)
+            for dim in v_dims:
+                exclude_dims.add(dim.data)
+
+    dims = set()
+    for v in f2.outputs:
+        # Only prefix if not in exclude_vars
+        if v.name not in exclude_vars:
+            new_name = f"{prefix}_{v.name}"
+            v.name = new_name
+            if isinstance(v.owner.op, ModelVar):
+                rv = v.owner.inputs[0]
+                rv.name = new_name
+        dims.update(extract_dims(v))
+
+    # Don't rename dimensions that belong to excluded variables
+    dims_rename = {
+        dim: pytensor.as_symbolic(f"{prefix}_{dim.data}")
+        for dim in dims
+        if dim.data not in exclude_dims
+    }
+    if dims_rename:
+        f2.replace_all(tuple(dims_rename.items()))
+
+    # Don't prefix coordinates for excluded dimensions
+    new_coords = {}
+    for k, v in f2._coords.items():  # type: ignore[attr-defined]
+        if k not in exclude_dims:
+            new_coords[f"{prefix}_{k}"] = v
+        else:
+            new_coords[k] = v
+    f2._coords = new_coords  # type: ignore[attr-defined]
+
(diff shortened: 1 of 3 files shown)


juanitorduz:
Thanks @cetagostini ! Im looking into the codecov report, and it seems we are still not testing all of the edge cases :) Would you mind adding tests for these ones please :)  Thanks!

pymc_marketing/pytensor_utils.py
```diff
@@ -14,20 +14,326 @@
...
+    # Merge coordinates from all models
+    merged_coords: dict = {}
+    for fg in fgraphs:
+        merged_coords.update(fg._coords)  # type: ignore[attr-defined]
+    f._coords = merged_coords  # type: ignore[attr-defined]
+
+    return model_from_fgraph(f, mutate_fgraph=True)
+
+
+class BuildMergedModel(OptimizerCompatibleModelWrapper):
```
williambdean: If this is causing circular imports, it should be somewhere else. Why not in the budget optimizer module since it is specific to that?

cetagostini: I was thinking it could be use for more than optimizer, you could actually merge models and them sample the full merge model, nevertheless, make sense to have the class there. I'll be moving!

cetagostini:
@juanitorduz or @williambdean ready for review!

juanitorduz (approved):
Thanks @cetagostini ! Do you plan to add a example notebook showcasing this ?

PR #1945, opened 2025-09-16
Add MaskedPrior class for masked variable creation

cetagostini:
## Description

Introduces the experimental MaskedPrior class to create variables and likelihoods over only the active entries of a boolean mask, expanding back to the full shape with zeros at inactive positions. Includes serialization/deserialization support and comprehensive tests for 1D, multidimensional, and integration with saturation functions and MMM fitting.

**What it does (in short)**
1. Wraps a Prior that is defined over named dims.
2. Applies a boolean mask to those dims to:
3. Build an “active” 1D RV only for unmasked positions (under a new active dim).
4. Scatter the active RV back to the original full grid and return a deterministic with the original dims.
5. Can also create masked likelihoods so only active positions contribute to the log-likelihood.

```python
import numpy as np
import pymc as pm
from pymc_marketing.pytensor_utils import MaskedPrior

# Define a 3x4 grid
coords = {"row": [0, 1, 2], "col": [0, 1, 2, 3]}

# Create mask - only activate positions (0,0), (1,2), (2,3)
mask = np.array(
    [
        [True, False, False, False],
        [False, False, True, False],
        [False, False, False, True],
    ]
)

# Define prior over full grid
prior = Prior("Normal", mu=0, sigma=1, dims=("row", "col"))
masked_dist = MaskedPrior(prior, mask)

with pm.Model(coords=coords):
    # Create masked distribution
    coeff = masked_dist.create_variable("coeff")
```

Users can combine this to modify the params into their functions, such as `saturation` or `adstock`.

```python
LogisticSaturation(
    priors={
        "lam": MaskedPrior(
            Prior("HalfNormal", sigma=1.0, dims=("country", "region")),
            mask=mask,
        ),
        "beta": MaskedPrior(
            Prior("HalfNormal", sigma=1.0, dims=("country", "region")),
            mask=mask,
        ),
    }
)
```

If from the full grid, some panels are missing they can mask the likelihood as well to not sample on dates or combos dates-region-city which doesn't exist.

```python
mu_prior = Prior("Normal", mu=0.0, sigma=1.0, dims=("date", "geo"))
masked_mu = MaskedPrior(mu_prior, mask=mask)

observed = np.zeros((len(coords["date"]), len(coords["geo"])))

likelihood_prior = Prior(
    "Normal",
    dims=("date", "geo"),
)
masked_lik = MaskedPrior(likelihood_prior, mask=mask)

with pm.Model(coords=coords):
    # Masked deterministic mean over full grid (zeros where mask is False)
    mu_full = masked_mu.create_variable("mu_full")

    # Build likelihood as a Prior and mask it, so both mu and observed are masked
    masked_lik.create_likelihood_variable(
        name="y",
        mu=mu_full,
        observed=observed,
    )
```

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

### DIFF
--- a/pymc_marketing/pytensor_utils.py
+++ b/pymc_marketing/pytensor_utils.py
@@ -27,6 +27,14 @@ try:
 except ImportError:
     from pymc.logprob.utils import rvs_in_graph
 
+import warnings
+
+import numpy as np
+import pymc as pm
+import xarray as xr
+from pymc_extras.deserialize import deserialize, register_deserialization
+from pymc_extras.prior import Prior
+
 
 def extract_response_distribution(
     pymc_model: Model,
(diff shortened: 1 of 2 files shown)


cetagostini:
Documentation and notebook example for this will come with https://github.com/pymc-labs/pymc-marketing/pull/1944 #1943 

Should be ready for review otherwise!

cetagostini:
All PRs are separate @williambdean

pymc_marketing/pytensor_utils.py
```diff
@@ -106,3 +114,370 @@ def extract_response_distribution(
     )
 
     return response_distribution
+
+
+class MaskedPrior:
```
williambdean: Would the special_priors modules from Daniel's PR be a good location for this class?

cetagostini: I see this as another use, but could be? I feel Daniel's goal is different, and by consequence should be in other place.

williambdean: Why is it a pytensor_util then? prior_util?

williambdean: This seems like a special prior, no?

williambdean: Seems more related to pymc and pymc-extras than pytensor

cetagostini: I can move it under pymc_utils then?

williambdean: How about `pymc_marketing.prior` where the Prior class used to be

cetagostini: I move it to special Priors after thinking about it. Based on Daniel's description fit well:
> this module contains a collection of priors that do not inherit from the Prior class but have many of the same methods.

pymc_marketing/special_priors.py
```diff
@@ -207,3 +209,370 @@ def _is_lognormalpositiveparam_type(data: dict) -> bool:
...
+        -------
+        pt.TensorVariable
+            Deterministic variable over the full dims with observed RV on active entries.
+        """
+        model = pm.modelcontext(None)
+        flat_mask = self.mask.values.ravel().astype(bool)
+        n_active = int(flat_mask.sum())
+
+        if n_active == 0:
+            return pm.Deterministic(name, pt.zeros(self.mask.shape), dims=self.dims)
```
williambdean, line 574: Does this use the observed variable at all?

pymc_marketing/special_priors.py
```diff
@@ -207,3 +209,370 @@ def _is_lognormalpositiveparam_type(data: dict) -> bool:
...
+        # Remap dims on a deep copy so nested parameter priors match the active subset
+        reduced = self._remap_dims(self.prior.deepcopy())
+
+        # Broadcast mu/observed to full mask shape via arithmetic broadcasting, then select active entries
+        mu_tensor = pt.as_tensor_variable(mu)
+        mu_full = mu_tensor + pt.zeros(self.mask.shape, dtype=mu_tensor.dtype)
+        mu_active = mu_full.reshape((self.mask.size,))[flat_mask]
+
+        obs = observed.values if hasattr(observed, "values") else observed
+        obs_tensor = pt.as_tensor_variable(obs)
```
williambdean, line 593: `pt.as_tensor_variable` can already handle pandas and numpy. Is this for xarray case?

cetagostini: No, doesn't work directly with xarray! You can [check it here](https://colab.research.google.com/drive/19rj0FXzjBPlwXw2VdI05tVV1HY2OK394#scrollTo=NoYjmVMzvO9e) at the end under "**things didn't work out**"

pymc_marketing/special_priors.py
cetagostini: Indeed, It does not read or use any observed data, but thats the objective. The goal is when `n_active == 0` (everything masked), skip creating a stochastic RV and instead return a zero-valued placeholder of the expected shape/dims, so downstream code that references this variable still works and its contribution is identically zero. It adds no logp and simply shows up in the trace as a constant zero array. If you wanted it to mirror an observed array’s shape, you’d need to build it from that (e.g., pt.zeros_like(observed)), but here it’s keyed to the mask/dims instead. This mostly avoids adding empty coords/RVs and keeps array ops simple.

I'm not sure, if I have a use case to fully masks all, but prefer to have it there, if someone masks all and thats not what they want, will saw output just zeros and will know. I can add a warning though?

pymc_marketing/special_priors.py
```diff
@@ -207,3 +209,370 @@ def _is_lognormalpositiveparam_type(data: dict) -> bool:
...
+        warnings.warn(
+            "This class is experimental and its API may change in future versions.",
+            stacklevel=2,
+        )
+
+    def _validate_mask(self):
+        if tuple(self.mask.dims) != tuple(self.dims):
+            raise ValueError("mask dims must match prior.dims order")
+
+    def _remap_dims(self, factory):
```
juanitorduz: missing type hint :)

pymc_marketing/special_priors.py
```diff
@@ -207,3 +209,370 @@ def _is_lognormalpositiveparam_type(data: dict) -> bool:
...
+        if hasattr(factory, "dims"):
+            dims = factory.dims
+            if isinstance(dims, str):
+                dims = (dims,)
+            if tuple(dims) == tuple(self.dims):
+                factory.dims = (self.active_dim,)
+
+        return factory
+
+    def create_variable(self, name: str):
```
juanitorduz: missing return type hint

pymc_marketing/special_priors.py
```diff
@@ -207,3 +209,370 @@ def _is_lognormalpositiveparam_type(data: dict) -> bool:
...
+        mask_vals = payload.get("mask")
+        # Fallback to provided dims or infer from prior if available
+        mask_dims = payload.get("mask_dims") or (getattr(prior, "dims", None) or ())
+        mask_da = xr.DataArray(np.asarray(mask_vals, dtype=bool), dims=tuple(mask_dims))
+        active_dim = payload.get("active_dim")
+        return cls(prior=prior, mask=mask_da, active_dim=active_dim)
+
+    def create_likelihood_variable(
+        self, name: str, *, mu: pt.TensorLike, observed: pt.TensorLike
+    ):
```
juanitorduz: missing return type

tests/test_special_priors.py
```diff
@@ -13,13 +13,18 @@
...
 import pytest
 import xarray as xr
 from pymc_extras.prior import Prior
+from pytensor import function
 
+from pymc_marketing.mmm import GeometricAdstock, LogisticSaturation
+from pymc_marketing.mmm.multidimensional import MMM
 from pymc_marketing.special_priors import (
     LogNormalPositiveParam,
+    MaskedPrior,
```
juanitorduz, line 27: Do we have a test showing that to_dict and then from_dict give the same thing?

tests/test_special_priors.py
```diff
@@ -79,3 +84,323 @@ def test_the_deserializer_can_distinguish_between_types_of_prior_classes():
     assert not _is_lognormalpositiveparam_type(
         Prior("Normal", mu=1.0, sigma=1.0).to_dict()
     )
+
+
+def test_masked_prior_simple_1d():
```
juanitorduz, line 85: Do we want a test in higher dimensions?

cetagostini: Other tests like `test_masked_prior_with_logistic_saturation_prior_sampling` already check higher dimensions.

pymc_marketing/special_priors.py
```diff
@@ -234,3 +244,372 @@ def _is_LogNormalPrior_type(data: dict) -> bool:
...
+        from pymc_marketing.mmm import LogisticSaturation
+        from pymc_marketing.pytensor_utils import MaskedPrior
+
+        coords = {
+            "country": ["Colombia", "Venezuela"],
+            "channel": ["x1", "x2", "x3", "x4"],
+        }
+
+        mask_excluded_x4_colombia = xr.DataArray(
+            [[True, False, True, False], [True, True, True, True]],
```
TeemuSailynoja, line 353: Is this excluding x2 and x4 form Colombia?

pymc_marketing/special_priors.py
```diff
@@ -234,3 +244,372 @@ def _is_LogNormalPrior_type(data: dict) -> bool:
...
+        with pm.Model(coords=coords):
+            masked = MaskedPrior(prior, mask)
+            zeros = masked.create_variable("intercept")
+
+    Apply over a saturation function priors:
+
+    .. code-block:: python
+
+        from pymc_marketing.mmm import LogisticSaturation
+        from pymc_marketing.pytensor_utils import MaskedPrior
```
TeemuSailynoja: Should be
`from pymc_marketing.special_priors import MaskedPrior`

TeemuSailynoja:
> expanding back to the full shape with zeros at inactive positions.

What if we have a variable, where zero has a special meaning?
Say I have the location parameter for a lognormal. Then I would like the fill value to possibly be -Inf, or some very negative value. This might be a future extension to the MaskedPrior class, but at least a thing to keep in mind.

pymc_marketing/special_priors.py
cetagostini: Yes!

pymc_marketing/special_priors.py
```diff
@@ -234,3 +244,372 @@ def _is_LogNormalPrior_type(data: dict) -> bool:
...
+        self.mask = mask
+        self.dims = prior.dims
+        self.active_dim = active_dim or f"non_null_dims:{'_'.join(self.dims)}"
+        self._validate_mask()
+
+        warnings.warn(
+            "This class is experimental and its API may change in future versions.",
+            stacklevel=2,
+        )
+
```
williambdean, line 434: Think it is fine to remove this. This is suppose to be a VariableFactory, right?
If the implementation changes but the methods are kept within that protocol, then that is fine

juanitorduz: @carlosagostini would you like to push a patch for this :?

cetagostini: Agree but I'm not sure if `methods are kept within that protocol` depending on what we want to do, that could not be the case, so I would prefer to keep it, so if at some point we need to go out of the protocol people is not surprise.

ps: I don't have any case in my head to go out of protocol, but as soon as I start to use it more and more, could appear (Maybe, maybe not?). So, my view is just stay safe and flexible.

I'll follow any of your thoughts anyways @juanitorduz @williambdean

PR #1948, opened 2025-09-19
Rename LogNormalPositiveParam to LogNormalPrior

### LINKED ISSUE #1946: Change name of LogNormalPositiveParam
Hey, I was checking this in https://github.com/pymc-labs/pymc-marketing/pull/1945

I feel we should rename to something more descriptive, and align to other class. I feel the original name `LogNormalExp` was better, and as consequence something like `LogNormalExpPrior` will sound and look nicer.

Plus, I was thinking on make the docstring a bit verbose, showing the formula under the hood already. That way users don't need to go to the blog to find out whats different in this LogNormal:

```python
class LogNormalExpPrior:
    """
    A lognormal prior parameterized by mean and standard deviation
    on the positive domain, with optional centered or non-centered
    parameterization.

    This prior differs from the standard ``LogNormal`` distribution,
    which takes log-scale parameters (``mu_log``, ``sigma_log``).
    Instead, it is parameterized directly in terms of the mean and
    standard deviation on the positive scale, making it more intuitive
    and suitable for hierarchical modeling.

    To achieve this, the lognormal parameters are computed internally
    from the positive-domain parameters:

    .. math::

        \mu_{\log} &= \ln \left( \frac{\mu^2}{\sqrt{\mu^2 + \sigma^2}} \right) \\
        \sigma_{\log} &= \sqrt{ \ln \left( 1 + \frac{\sigma^2}{\mu^2} \right) }

    where :math:`\mu > 0` and :math:`\sigma > 0`.

    The prior is then defined as:

    .. math::

        \phi &\sim \text{LogNormal}(\mu_{\log}, \sigma_{\log})

    This construction ensures that the resulting random variable
    has approximately the intended mean and variance on the positive scale,
    even when :math:`\mu` and :math:`\sigma` are themselves random variables.

    Parameters
    ----------
    mu : Prior, float, int, array-like
        The mean of the distribution on the positive scale.
    sigma : Prior, float, int, array-like
        The standard deviation of the distribution on the positive scale.
    dims : tuple[str, ...], optional
        The dimensions of the distribution, by default None.
    centered : bool, optional
        Whether to use the centered parameterization, by default True.

    Examples
    --------
    Build a non-centered hierarchical model where information is shared across groups:

    .. code-block:: python

        from pymc_marketing.special_priors import LogNormalExpPrior

        prior = LogNormalExpPrior(
            mu=Prior("Gamma", mu=1.0, sigma=1.0),
            dims=("geo",),
            centered=False,
        )
    
    References
    ----------
    - D. Saunders, *A positive constrained non-centered prior that sparks joy*.
    - Wikipedia, *Log-normal distribution — Definitions*.
    """
. . . . 
````

Should I make a PR @daniel-saunders-phil ?

### PR DESCRIPTION
**Pull request by cetagostini**
## Description

Refactored the LogNormalPositiveParam class to LogNormalPrior for clarity and consistency. Updated parameter names from 'mu' and 'sigma' to 'mean' and 'std', revised docstrings, and changed all references in code and tests to match the new naming convention.

## Related Issue

- [x] Closes #1946 
- [x] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1948.org.readthedocs.build/en/1948/

### DIFF
--- a/.github/workflows/install-conda-env.yml
+++ b/.github/workflows/install-conda-env.yml
@@ -18,6 +18,6 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v5
     - name: setup-micromamba
-      uses: mamba-org/setup-micromamba@v2.0.6
+      uses: mamba-org/setup-micromamba@v2.0.7
       with:
         environment-file: environment.yml
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -204,7 +204,7 @@ Notes
 
 import warnings
 from collections.abc import Sequence
-from typing import Any, ClassVar, Protocol, runtime_checkable
+from typing import Any, ClassVar, Protocol, cast, runtime_checkable
 
 import numpy as np
 import pytensor.tensor as pt
@@ -225,7 +225,9 @@ from pymc_marketing.mmm.constraints import (
     compile_constraints_for_scipy,
 )
 from pymc_marketing.mmm.utility import UtilityFunctionType, average_response
-from pymc_marketing.pytensor_utils import extract_response_distribution
+from pymc_marketing.pytensor_utils import merge_models
+
+# Delayed import inside methods to avoid circular dependency on pytensor_utils
 
 
 def optimizer_xarray_builder(value, **kwargs):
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by cetagostini** [comment]
Let's check and approve but not merge until https://github.com/pymc-labs/pymc-marketing/pull/1945 is merge. Just to avoid to get conflict's there, would be easier to spot things on this branch!

**Comment by cetagostini** [comment]
Failing test related to: https://github.com/pymc-labs/pymc-marketing/pull/1949 not the PR

**Comment by cetagostini** [comment]
Failing test related to: https://github.com/pymc-labs/pymc-marketing/pull/1949 not the PR

File: pymc_marketing/special_priors.py [hunk 1]
```diff
@@ -29,42 +29,69 @@
...
 
     Parameters
     ----------
-    mu : Prior, float, int, array-like
-        The mean of the distribution.
-    sigma : Prior, float, int, array-like
-        The standard deviation of the distribution.
+    mean : Prior, float, int, array-like
+        The mean of the distribution on the positive scale.
+    std : Prior, float, int, array-like
```
williambdean, line 69: Why this change? PyMC distributions use mu and sigma

daniel-saunders-phil: To distinguish this parameterization from the mu/sigma parameterization of the LogNormal. If we left it as the mu and sigma, it's hard to tell how this distribution differs from the usual LogNormal. What's special about this distribution is that you are specifying the actual mean and standard deviation.

williambdean: I'm still for mu sigma 😄 
Think the name of the class will make that distinction

cetagostini: We discussed here: #1946 I feel mean and std would be nice, that's very minor. I'll not stop for the merge for this.

PR #1955, opened 2025-09-23
Adds Optional `dims` arg for subsetting by dimensions when plotting a multidimensional model

### LINKED ISSUE #1952: Support for subsetting by dimensions when plotting a multidimensional model
@TeemuSailynoja and I have ran into a case where the model has too many dimensions (for example - channel, geo, user type) and plotting things like `contributions_over_time` can either be not visible or throw a warning `WARNING: Output too large, the following mime types were removed from the output: image/png` (on Databricks).

We will need add arguments to the function(s) to support subsetting by dimensions, so that the plots can be easily read and rendered.

Places that this will be needed:

- in `contributions_over_time`
- in `plot.budget_allocation`

timbo112711:
## Description
@TeemuSailynoja and I have ran into a case where the model has too many dimensions (for example - channel, geo, user type) and plotting things like `contributions_over_time` can either be not visible or throw a warning `WARNING: Output too large, the following mime types were removed from the output: image/png` (on Databricks).

This PR will add an optional `dim` arg to the plotting methods listed below.

1. `contributions_over_time` ✅
2. `budget_allocation` ✅
3. `saturation_scatterplot` ✅
4. `saturation_curves`✅

## Related Issue

- [x] Closes #1952
- [ ] Related to #

## Checklist

- [x] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes) 🤷🏻‍♂️

----
📚 Documentation preview 📚: https://pymc-marketing--1955.org.readthedocs.build/en/1955/


pymc_marketing/mmm/plot.py
```diff
@@ -464,11 +468,21 @@ def contributions_over_time(
...
         }
 
-        # Identify combos
+        # Apply user-specified filters (`dims`)
+        if dims:
+            for key, val in dims.items():
+                if key not in all_dims:
+                    raise ValueError(f"Dimension '{key}' not found in posterior dims.")
+                if val not in self.idata.posterior.coords[key].values:
+                    raise ValueError(f"Value '{val}' not found in dimension '{key}'.")
```
juanitorduz: This code is also below. Would it make sense to have a helper function 

```python
def _validate_dims( . . .) -> None
    . . .
```

timbo112711: Good call out, will add this little block (and the one below it) into a helper function

timbo112711:
> View / edit / reply to [this conversation](https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/1955/discussion/#comment-3325343272) on ReviewNB
> 
> _juanitorduz commented on 2025-09-23T19:54:53Z_ ----------------------------------------------------------------
> 
> where do these divergences come from ?

Not quite sure, will have to rerun the notebook to see if that resolves.

pymc_marketing/mmm/plot.py
```diff
@@ -321,6 +321,19 @@ def _add_median_and_hdi(
...
+    def _validate_dims(
+        self,
+        dims: dict[str, str | int],
+        all_dims: list[str],
+    ) -> None:
+        """Validate that provided dims exist in the model's dimensions."""
+        if dims:
+            for key, val in dims.items():
+                if key not in all_dims:
+                    raise ValueError(f"Dimension '{key}' not found in data dimensions.")
```
TeemuSailynoja: perhaps "... not found in idata dimensions." So we are even more clear where we the dimension should be.

pymc_marketing/mmm/plot.py
```diff
@@ -419,6 +432,7 @@ def contributions_over_time(
         self,
         var: list[str],
         hdi_prob: float = 0.85,
+        dims: dict[str, str | int] | None = None,
```
TeemuSailynoja: Could we also allow for a list of values here? I.e. {"country": ["US", "UK"]}

pymc_marketing/mmm/plot.py
```diff
@@ -464,11 +481,17 @@ def contributions_over_time(
             for key, value in self.idata.posterior[var].coords.items()
         }
 
-        # Identify combos
+        # Apply user-specified filters (`dims`)
+        if dims:
+            self._validate_dims(dims=dims, all_dims=all_dims)
+            # Remove filtered dims from the combinations
+            additional_dims = [d for d in additional_dims if d not in dims]
```
TeemuSailynoja, line 521: if we allow for subsetting by a list of values for some dim, then here we would only take away those keys in dims, where there is only a single value.

pymc_marketing/mmm/plot.py
```diff
@@ -493,14 +519,20 @@ def contributions_over_time(
...
                 }
                 data = data.expand_dims(**missing_coords)
-                data = data.sel(**indexers)  # type: ignore
+                data = data.sel(**indexers)  # apply slice
                 data = self._reduce_and_stack(
                     data, dims_to_ignore={"date", "chain", "draw", "sample"}
                 )
                 ax = self._add_median_and_hdi(ax, data, v, hdi_prob=hdi_prob)
 
+            # Title includes both fixed and combo dims
```
TeemuSailynoja, line 573: The subsetting by list would need to have here
```python
title_dims = (
    [key for key, val in dims.items() if not isinstance(val, list) or len(val) == 1] +
    additional_dims if dims else additional_dims
)
```

And then below
```python
title_combo = (
    (val for val in dims.values() if not isinstance(val, list) or len(val) == 1)) +
    combo if dims else combo
)
```

pymc_marketing/mmm/plot.py
```diff
@@ -898,6 +930,7 @@ def budget_allocation(
         figsize: tuple[float, float] = (12, 6),
         ax: plt.Axes | None = None,
         original_scale: bool = True,
+        dims: dict[str, str | int] | None = None,
```
TeemuSailynoja: Same thoughts on subsetting by a list of values for some dimension as above.

tests/mmm/test_plot.py
TeemuSailynoja, line 1: If we decide to add an option to subset the plots by a list of values for some dimension, that would also need a tests.

pymc_marketing/mmm/plot.py
TeemuSailynoja, line 1: What about `saturation_scatterplot` and `saturation_curves`? Perhaps we leave those for another PR, but those could plausibly have the same argument added.

TeemuSailynoja (commented):
I mainly want to have a conversation on whether we want to also allow a dims argument subsetting by a list of desired values, like:

`dims={"Country": ["US", "UK"]}`

timbo112711:
> I mainly want to have a conversation on whether we want to also allow a dims argument subsetting by a list of desired values, like:
> 
> `dims={"Country": ["US", "UK"]}`

Expanding the usability to allow for lists in the dims arg makes sense to me.

pymc_marketing/mmm/plot.py
timbo112711: > What about `saturation_scatterplot` and `saturation_curves`? Perhaps we leave those for another PR, but those could plausibly have the same argument added.

 I would say let's add the dims arg for those in this PR. I've added them to the PR description!!

PR #1961, opened 2025-09-25
Add lognormalprior to the multidimensional notebook

daniel-saunders-phil:
I wanted to showcase the new prior in the notebook because I think it ought to be new default approach when building hierarchical MMMs. It is much easier to reason about because everything stays on the positive scale. It gives us a small (10 seconds) speed up for this notebook. This is a follow up to #1939

I also removed the discussion of VI. We should be careful about directing new users to VI. If people hear this other thing is faster, they are inclined to switch over. But VI will give an inaccurate description of the MMM posterior when you have a logistic saturation function (the interaction between parameters is non-linear, so VI cannot deal with that). This might be fine if you have a way of measuring the cost of inaccuracy but that's a big can of worms. 

Initially, I started expanding the discussion to explain the above point and realized it was growing into a tangent that disrupted the flow of the discussion. I think it might be nice to have a separate notebook that explains how to use VI responsibly with MMMs.

----
📚 Documentation preview 📚: https://pymc-marketing--1961.org.readthedocs.build/en/1961/


cetagostini:
@daniel-saunders-phil Why are you removing the advi part?

cetagostini (approved):

PR #1967, opened 2025-09-30
Align y-axes in mmm.plot.budget_allocation

TeemuSailynoja:
## Description

Ensure that y=0 is aligned between the primary and secondary y-axis in `mmm.plot.budget_allocation`.
## Solution
If one axis includes negative, values the other is extended down so that the y=0 is aligned.

If all contributions would be fully negative (I don't know when this would happen...), the y=0 level is centered vertically on both axes.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1967.org.readthedocs.build/en/1967/


TeemuSailynoja:
still needs tests to go through the combinations of negative positive and ensure that the axis limits are correct after the call.

pymc_marketing/mmm/plot.py
```diff
@@ -243,6 +243,40 @@ def _build_subplot_title(
...
+            if ax1_yratio < ax2_yratio:
+                ax2.set_ylim(bottom=ylims2[1] * ax1_yratio)
+                if ax1_yratio == -1:
+                    # if the axis is fully negative, center zero.
+                    ax.set_ylim(top=-ylims1[0])
+            elif ax2_yratio < ax1_yratio:
+                ax.set_ylim(bottom=ylims1[1] * ax2_yratio)
+                if ax2_yratio == -1:
+                    # if the axis is fully negative, center zero.
+                    ax2.set_ylim(top=-ylims2[0])
```
williambdean: Can you use NamedTuple or something. Quite confusing to read

TeemuSailynoja: Looking at this now, I fully agree. I also think that this is not capturing every possible combination the axes can get odd. I'll move this to a draft and rework it.

TeemuSailynoja: I completely rewrote the function, and added tests that go through the possible cases.
Not the most compact solution, but I hope it exposes its logic better.

pymc_marketing/mmm/plot.py
```diff
@@ -243,6 +244,83 @@ def _build_subplot_title(
             return ", ".join(title_parts)
         return fallback_title
 
+    def _align_y_axes(self, ax_left, ax_right) -> None:
```
juanitorduz: @TeemuSailynoja can you please add type hinds and complete docstrings to this function so that we can bring it to the finish line :) Thanks!

juanitorduz (changes requested):

pymc_marketing/mmm/plot.py
TeemuSailynoja: @juanitorduz I'm sorry for the delay. Work got in the way 😝 

I hope what I added is good. Thank you for pingin me on this!

juanitorduz (approved):

PR #1802, opened 2025-06-26
Compute deterministics after sampling

PabloRoque:
## Description

Compute `pm.Deterministic`(s) after sampling.

Follows the implementation discussed in https://github.com/pymc-devs/pymc/issues/7461

## Related Issue

- [x] Related to #1792 

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--1802.org.readthedocs.build/en/1802/

PR #1968, opened 2025-10-01
Add `FrequencyReachAdditiveEffect`

PabloRoque:
## Description

Add Frequency/Reach support via the `MuEffect` interface. 
Similar, but slightly different (see below) to the implementation described [here](https://developers.google.com/meridian/docs/advanced-modeling/reach-frequency)

Notes:

~~1) The formulation is not strictly equivalent. Notice that our `frequency_sat: Tensorvariable` already applies the multiplicative `Beta` coefficient. If we want a fully equivalent implementation we need to call , but then we cannot rely on `.apply` to do dim_handling.~~

1) Introduced `HillShapeSaturation` to make the formulation equivalent

2) This is a very drafty feature. No support for optimization whatsoever.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--1968.org.readthedocs.build/en/1968/


pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
williambdean, line 1: Shall we change additive_effect.py into additive_effects/ submodule and has this as a file?

williambdean: All can added to `__init__` so that the import is kept

cetagostini: I like this idea!

PabloRoque: I am happy to move things around, but for me the main question of this PR is:

- Should we somehow modify `SaturationTransformation` to offer strong separation between saturation asymptote and lambda, c, slope, etc?

juanitorduz:
Amazing!

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
```diff
@@ -0,0 +1,416 @@
...
+
+    Parameters
+    ----------
+    df_frequency_reach : pd.DataFrame
+        Long format DataFrame with columns at least: date, channel, frequency, reach.
+        Additional columns matching `mmm.dims` (e.g. geo) are optional.
+    saturation : SaturationTransformation
+        Saturation transformation applied after adstock.
+    adstock : AdstockTransformation
+        Adstock transformation applied first to the effective exposure signal.
```
TeemuSailynoja: Ordering here is reversed in relation to the code. Code is correct – first saturation, then adstock.

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
```diff
@@ -0,0 +1,416 @@
...
+* Saturation acts only on frequency to avoid stacking nonlinearities on both
+    multiplicative terms (improves identifiability of beta).
+
+Edge Cases & Validation
+-----------------------
+* Negative frequency or reach outside [0,1] raises ValueError.
+* Duplicate rows for the same (date, dims, channel) are aggregated (mean).
+* If channels present in the DataFrame are not a subset of the MMM channel
+  coordinate, a ValueError is raised.
+
```
TeemuSailynoja: Why is this required? I would assume the opposite. Isn't reach + frequency an alternative to using impressions or spend, and having both in the model would effectively double count the contribution of the involved channels?
The estimated contributions are also separate from the `channel_contribution` deterministic in the parent MMM, so this could lead to confusion.

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
```diff
@@ -0,0 +1,416 @@
...
+    frequency_sat (reach remains linear / unsaturated).
+3. Form effective_exposure_raw = reach_raw * frequency_sat (element-wise).
+4. Apply the adstock transformation to effective_exposure_raw producing
+    effective_exposure_adstocked.
+5. Draw per-channel scaling coefficients beta[channel] and compute
+    channel_contribution = beta * effective_exposure_adstocked.
+6. Aggregate over channel to obtain total_effect added to model mean.
+7. Expose intermediate deterministics for diagnostics:
+    frequency_sat, effective_exposure_raw, effective_exposure_adstocked,
+    channel_contribution, total_effect.
```
TeemuSailynoja: For larger models, perhaps creating the intermediate deterministics could be optional?

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
```diff
@@ -0,0 +1,416 @@
...
+            dims=(self.date_dim_name, *mmm.dims, self.channel_coord_name),
+        )
+
+        # 4. Per-channel beta scaling
+        beta = pm.Normal(
+            f"{self.prefix}_beta",
+            mu=0.0,
+            sigma=1.0,
+            dims=(self.channel_coord_name,),
+        )
```
TeemuSailynoja: perhaps expose to user as an optional argument of the Prior class, and have this as a default value.

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
```diff
@@ -0,0 +1,416 @@
...
+        dim_handler = create_dim_handler((self.date_dim_name, *mmm.dims))
+        return dim_handler(total_effect, (self.date_dim_name, *mmm.dims))
+
+    def set_data(self, mmm: Model, model: pm.Model, X: xr.Dataset) -> None:
+        """Update reach & frequency raw data for prediction dates.
+
+        Extends the registered pm.Data nodes `<prefix>_frequency_raw` and
+        `<prefix>_reach_raw` to any new dates in the model coordinates, filling
+        unseen future dates with zeros. Does not accept a modified DataFrame;
+        construct a new effect instance if the historical data themselves change.
```
TeemuSailynoja: Unfortunate that no predictions to future unless they were provided when hte class was initially created.
After #2188, we could actually supply future values when calling `sample_posterior_predictive`.

Just thought I had when reading though.

TeemuSailynoja (changes requested):
The order of adstock and saturation is reversed in the documentation.
Should we reference [the source paper](https://research.google/pubs/bayesian-hierarchical-media-mix-model-incorporating-reach-and-frequency-data/) instead of Meridian documentation?

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
PabloRoque: I would not worry about the presence of Deterministics in build_model. 
They are computed after sampling after #1802

TeemuSailynoja: Oh that's a cool PR!

TeemuSailynoja:
@PabloRoque this is cool! Need any help? @juanitorduz wants this 🤣

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
PabloRoque: Thanks for catching this — you're right that the code is correct. To confirm the intended order: Meridian applies Hill saturation to frequency first, then adstock to the effective exposure signal (reach x saturated_frequency). This is consistent with the Meridian source code in adstock_hill.py, where AdstockTransformer.forward explicitly notes: the Hill function can be applied before or after Adstock, and Hill-first is the intended path for R&F channels. The current docstrings already reflect the correct order (saturation then adstock), so no code change is needed here.

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
PabloRoque: You're right — R&F is an *alternative* to impressions/spend for those channels, not an addition. Using both would double-count their contribution.\n\nMeridian enforces this the same way: `InputData._validate_media_channels` raises a `ValueError` if any channel name appears in more than one of `media_channel`, `rf_channel`, etc.\n\nWe've added the same guard in `create_data` — it now raises with a clear message if any R&F channel name overlaps with `mmm.channel_columns`. The docstring has also been updated to make the alternative-not-additive semantics explicit.

pymc_marketing/mmm/additive_effect/frequency_reach_effect.py
PabloRoque: Good call. Changed the default to `HalfNormal(sigma=1)` — semantically correct for an R&F scaling coefficient (contributions should be non-negative). Users can still pass any `Prior` they like at construction time, e.g. `beta_prior=Prior("Gamma", alpha=2, beta=1)`.

PR #1977, opened 2025-10-06
Remove redundant pre-commit action

juanitorduz:
Removing the redundat pre-commit action we added some time ago because mypy was too slow. It seems it's better now.

----
📚 Documentation preview 📚: https://pymc-marketing--1977.org.readthedocs.build/en/1977/


williambdean (approved):
Works well. Maybe we explore `ty` down the line. Think it is still experimental.

Issue #2009, opened 2025-10-17
Bug with TestShiftedBetaGeometric


cetagostini:
```sh
FAILED tests/clv/test_distributions.py::TestShiftedBetaGeometric::TestRandomVariable::test_random_edge_cases - assert np.float64(0.0) > 0
[...]

Issue #167, opened 2023-02-21
Add Additional Expressions to sBG Model


ColtAllen:
**UPDATED 2025-OCT-18**

There is a follow-up research paper for `ShiftedBetaGeoModel` that introduces additional predictive methods:

https://faculty.wharton.upenn.edu/wp-content/uploads/2012/04/Fader_hardie_contractual_mksc_10.pdf

Specifically, expected discounted residual lifetime and expected retention elasticity.

ricardoV94:
Probably relevant: #169

ColtAllen:
I think adding `customer_lifetime_value` support for this model is a good idea, per https://github.com/pymc-labs/pymc-marketing/issues/527.

PR #2010, opened 2025-10-17
Add `expected_residual_lifetime` predictions to `ShiftedBetaGeoModel`

ColtAllen:
## Description

This began as a bug fix for prior/posterior predictive checks in `ShiftedBetaGeoModel`, but I decided to expand this PR & add another predictive method while waiting for unrelated CI checks to be resolved.

## Related Issue

- [ ] Closes https://github.com/pymc-labs/pymc-marketing/issues/2009
- [ ] Related to https://github.com/pymc-labs/pymc-marketing/issues/167

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2010.org.readthedocs.build/en/2010/


ColtAllen:
```shell
pymc_marketing/mmm/transformers.py:24: in <module>
    from pytensor.npy_2_compat import normalize_axis_index
E   ImportError: cannot import name 'normalize_axis_index' from 'pytensor.npy_2_compat' 
```

cetagostini-wise:
> ```shell
> pymc_marketing/mmm/transformers.py:24: in <module>
>     from pytensor.npy_2_compat import normalize_axis_index
> E   ImportError: cannot import name 'normalize_axis_index' from 'pytensor.npy_2_compat' 
> ```

@ricardoV94 can you give us a hand here?

ricardoV94:
Now you import directly from numpy. It existed because the import was different in numpy 2 and before, now we only support numpy 2

pymc_marketing/clv/distributions.py
```diff
@@ -908,12 +908,12 @@ def rng_fn(cls, rng, alpha, beta, size):
         alpha = np.broadcast_to(alpha, size)
         beta = np.broadcast_to(beta, size)
 
-        p = rng.beta(a=alpha, b=beta, size=size)
+        p_samples = rng.beta(a=alpha, b=beta, size=size)
 
-        samples = rng.geometric(p, size=size)
-
-        return samples
+        # prevent log(0) by clipping small p samples
```
ricardoV94, line 913: I wouldn't do this. If the only reason is the test, relax it or seed it. I can easily make a test that still fails with this hack.

ColtAllen:
```shell
FAILED tests/customer_choice/test_nestedlogit.py::test_counterfactual 
- ValueError: Runtime broadcasting not allowed. 
One input had a distinct dimension length of 1, but was not marked as broadcastable: 
(input[%i].shape[%i] = %lld, input[%i].shape[%i] = %lld). 
If broadcasting was intended, use `specify_broadcastable` on the relevant input.
```

pymc_marketing/clv/distributions.py
ColtAllen: PPCs occasionally fail during notebook testing, and it's no different when reformulating the RV with the  `np.ceil(np.log(U) / np.log(1 - p)))` trick.

That said, most users will be doing MAP fits with this model and evaluating with the post-inference predictive methods. Ardent Bayesians may be annoyed by the lack of PPCs, but this bug isn't something that would break a typical workflow.

ricardoV94: I'm no longer opposed to this truncation, because numpy does an even more aggressive truncation so that it can return integer (as those can't represent infinity): https://github.com/numpy/numpy/blob/4992eaf12a8c7a3eeb0f6d13f832a05f21e3834c/numpy/random/src/distributions/distributions.c#L996-L997

I would open an issue to consider returning floats, in which case both infinity can be returned or larger values if  one writes the geometric as `geom = np.ceil(np.log(u) / np.log1p(-p))` without casting to integer

ricardoV94: In fact you can truncate as high as 1e-35 without changing the behavior, as that will still be higher than the cutoff point for any value drawn by numpy:

```python
numpy_cutoff = 9.223372036854776e+18
largest_uniform = np.nextafter(1, 0)
beta_clip = 1e-35
assert np.log(largest_uniform) / np.log1p(-beta_clip)  >= numpy_cutoff
```

ricardoV94: maybe just leave a comment in the code about this discussion?

cetagostini (approved):
Code looks good to me :)

Issue #650, opened 2024-04-27
Speed up sphinx builds


wd60622:
It would be awfully convenient to be able to build a single notebook, only the notebooks, or specific module docstrings. 

The `make html` command is very helpful. However, the build takes some time

https://github.com/pymc-labs/pymc-marketing/blob/91564cfbcce3d341d3d9bdab91a5b25766764b5a/Makefile#L20-L22

For notebooks, the sphinx rendering is a bit different than the jupyter notebook which can cause render issues not seen.

Not sure if this is possible though!

OriolAbril:
It is possible to some extent. It should work to quickly check some rendering issues, but many cross-references might break, as sphinx needs to generate the documentation as a whole for all cross-references to be resolvable. 

<details open><summary>Full background and examples (foldable)</summary>

For example, if I use `sphinx-build docs/source docs/build -b html -D include_patterns=index.md,notebooks/mmm/mmm_example.ipynb` I get a 2 page build that takes a couple seconds and renders the notebook page:

"fast" build page:

![imatge](https://github.com/pymc-labs/pymc-marketing/assets/23738400/97f32e8b-58dd-4771-95a1-f5e8bf2252a7)


whole/regular build page (from website):

![imatge](https://github.com/pymc-labs/pymc-marketing/assets/23738400/e91bf3db-7bf9-4aff-921f-080d184072b8)

Save the sidebar and navbar that are mostly missing (given there are now no other pages in the website so they aren't being populated) everything looks good. If we take a bit more in depth look however, to the "tip" box:

"fast" build page:

![imatge](https://github.com/pymc-labs/pymc-marketing/assets/23738400/49bb909e-0567-473a-8e8a-7365ac565fd0)

whole/regular builg page:

![imatge](https://github.com/pymc-labs/pymc-marketing/assets/23738400/999b777d-8cd9-4cd2-a3f7-a1107578b48f)

The other notebooks are not part of the "fast" build, so the cross-references pointing to them can't be resolved and therefore don't result in any link, but that doesn't mean their syntax is wrong, they are in fact valid cross-references.

If we modify a bit the build command to include these referenced pages `sphinx-build docs/source docs/build -b html -D include_patterns=index.md,notebooks/mmm/mmm_example.ipynb,notebooks/mmm/mmm_lift_test.ipynb,notebooks/mmm/mmm_budget_allocation_example.ipynb` now we get:

[...]

PR #2013, opened 2025-10-17
Make fast html

lfiaschi:
Available Commands Now

  # Fast build 
  make fasthtml       # ~30-60 seconds

  # Alternative builds
  make fasthtml-nb    # Skip notebooks only (~5-10 min)
  make fasthtml-api   # Skip API only (~15-25 min)

  # Full build
  make html           # Complete build with cache

  # Cleanup
  make cleandocs      # Clean everything
  make cleancache     # Clean cache only


pymc_marketing/mmm/transformers.py
```diff
@@ -21,7 +21,17 @@
...
+# Import normalize_axis_index - handle compatibility
+try:
+    from pytensor.npy_2_compat import normalize_axis_index
+except ImportError:
+    # PyTensor 2.35+ no longer exports this, use numpy directly
+    try:
+        from numpy._core.numeric import normalize_axis_index
+    except (ImportError, AttributeError):
+        # Older numpy versions
+        from numpy.core.numeric import normalize_axis_index
```
juanitorduz, line 34: This was "fixed" in https://github.com/pymc-labs/pymc-marketing/pull/2015 We want to keep up to datw with PyMC latest versions, where they dropped NumPy < 2 support, see https://github.com/pymc-labs/pymc-marketing/pull/2010#issuecomment-3414928489

williambdean:
pre-commit.ci autofix

williambdean:
Related to #650

docs/source/conf.py
```diff
@@ -87,15 +131,41 @@
...
+nb_execution_allow_errors = False
+nb_execution_raise_on_error = True
+nb_execution_excludepatterns = [
+    # Heavy notebooks that take too long - execute manually when needed
+    "notebooks/mmm/mmm_case_study.ipynb",
+    "notebooks/mmm/mmm_multidimensional_example.ipynb",
+    "notebooks/mmm/mmm_tvp_example.ipynb",
+    "notebooks/mmm/mmm_time_varying_media_example.ipynb",
+    "notebooks/clv/dev/*.ipynb",  # Development notebooks
+]
```
OriolAbril, line 152: TL; DR. Before this PR, ipynb notebooks were never executed because we had this combination of settings:

```python
nb_execution_mode = "auto"
nb_execution_excludepatterns = ["*.ipynb"]
```

The reason including the notebooks as part of the build slows it down isn't so much execution as it is processing and rendering them. There are a lot of notebooks which are usually long and have a lot of code and images, so processing them and converting to html is already somewhat expensive (but on a whole other scale than executing them).

---

Full context:

Looking at the blame of these two lines, I see I added that 3 years ago when setting up the docs and it hasn't been updated since. I probably copied it from ArviZ because it is the pattern we use there, but without a comment it isn't very clear what is happening. The sphinx extension we use to process and render notebooks supports multiple formats for notebooks, not only `.ipynb`. This pattern will ensure non ipynb notebooks are executed and therefore have outputs included in the rendered docs while at the same time skipping execution of ipynb notebooks because they already have the outputs stored in them.

We then use `.ipynb` format for the more expensive notebooks and `.md` format for lighter notebooks so we don't need to run them manually and they are always kept in sync. For example [this page](https://python.arviz.org/en/stable/user_guide/plotting_with_bokeh.html) comes from [one of these lighter notebooks in markdown format](https://github.com/arviz-devs/arviz/blob/main/doc/source/user_guide/plotting_with_bokeh.md?plain=1).

After the change, only a handful of the notebooks are excluded, with the rest being executed fully. This makes much more expensive the first time we build the docs. After that first time the notebook outputs are cached so it will be faster, but I don't think we want executing notebooks to be a part of doc building at all.

If we do want to have notebook execution be a step in building the html docs, we should make sure this doesn't happen on readthedocs. On readthedocs we build on their servers and unless we subscribed to one of their paid plans, builds have a time limit limited resources. MCMC sampling is beyond the compute power of our readthedocs plan.

Note relevant to the discussion. There is already a [GitHub Action](https://github.com/pymc-labs/pymc-marketing/blob/main/.github/workflows/test_notebook.yml) that [runs all the notebooks with mock sampling](https://github.com/pymc-labs/pymc-marketing/tree/main/scripts/run_notebooks) to check for potential errors, so there is already infrastructure in place to ensure notebooks always run.

PR #2055, opened 2025-11-02
Preparing plots and notebook for old API deprecation

cetagostini:
## Description

Kick-off API deprecation. Made a new notebook, in dev not visible to replicate current MMM example but with multidimensional. Only a few plots are missing.

Did additions in the plot API, and adjust some docs.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2055.org.readthedocs.build/en/2055/

### DIFF
--- a/.github/workflows/bot-prs.yml
+++ b/.github/workflows/bot-prs.yml
@@ -11,7 +11,7 @@ jobs:
   label-no-releasenotes:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Label no releasenotes
         run: |
           number=${{ github.event.pull_request.number }}
--- a/.github/workflows/cc-gha-implement.yml
+++ b/.github/workflows/cc-gha-implement.yml
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
 
--- a/.github/workflows/cc-gha-iterate.yml
+++ b/.github/workflows/cc-gha-iterate.yml
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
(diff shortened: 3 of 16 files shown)


pymc_marketing/mmm/multidimensional.py
```diff
@@ -1679,8 +1688,10 @@ def sample_posterior_predictive(
                 self.idata, **sample_posterior_predictive_kwargs
             )
 
-            if extend_idata:
-                self.idata.extend(post_pred, join="right")  # type: ignore
+            if extend_idata and self.idata is not None:
```
cetagostini, line 1707: If we don't do it in this way, the posterior predictive overwrite the constant data.

williambdean: Don't we want that though? It's the predicted_constant_data, right?

pymc_marketing/mmm/plot.py
```diff
@@ -462,6 +501,429 @@ def posterior_predictive(
 
         return fig, axes
 
+    def prior_predictive(
+        self,
+        var: list[str] | None = None,
```
williambdean: Why not just a string if you will take the first value?

williambdean (commented):
First pass

pymc_marketing/mmm/multidimensional.py
cetagostini-wise: Nope, if we want `predicted_constant_data` then we need to save it as that group otherwise overwrite the regular constant data from training, and we don't want that. But what I'll do is add those then as well the same way I added posterior predictive :)

So, we'll have `predicted_*` constant data or other element.

pymc_marketing/mmm/plot.py
cetagostini-wise: Good point, will do.

pymc_marketing/mmm/multidimensional.py
cetagostini: Done, instead of predicted I'm thinking better `posterior_predictive_constant_data`

PR #2093, opened 2025-11-17
Add special Laplace prior

daniel-saunders-phil:
The Laplace prior we have access to through the prior class doesn't have the option to be non-centered. There is a way of rewriting the density of Laplace in terms of a normal and an exponential distribution (check out the references, especially the stan doc if you are curious why that works). When you use that alternative parameterization, then you can also non-center it. None of this fit very naturally into the way the Prior class works at PyMC-extras. It would be very weird for that class to accept a laplace kwarg but turn it into a normal underneath. So I thought this was another good candidate for the SpecialPriors.

I added a base class to collect methods common to Laplace and LogNormal.

Finally, some minor edits to the LogNormal docs.

----
📚 Documentation preview 📚: https://pymc-marketing--2093.org.readthedocs.build/en/2093/

### DIFF
--- a/pymc_marketing/special_priors.py
+++ b/pymc_marketing/special_priors.py
@@ -158,7 +158,7 @@ class LogNormalPrior(SpecialPrior):
 
     .. math::
 
-        \phi & \sim \text{LogNormal}(\mu_{\text{log}}, \sigma_{\text{log}})
+        \phi \sim \text{LogNormal}(\mu_{\text{log}}, \sigma_{\text{log}})
 
     This construction ensures that the resulting random variable
     has approximately the intended mean and variance on the positive scale,


pymc_marketing/special_priors.py
```diff
@@ -156,93 +254,91 @@ def create_variable(self, name: str) -> TensorVariable:
...
-        kwargs = data.get("kwargs", {})
+    def __init__(self, dims: tuple | None = None, centered: bool = True, **parameters):
+        self.parameters = parameters
+        self.dims = dims
+        self.centered = centered
 
-        def handle_value(value):
-            if isinstance(value, dict):
-                return deserialize(value)
+        self._checks()
```
williambdean: Why redefine?

pymc_marketing/special_priors.py
```diff
@@ -93,8 +198,8 @@ class LogNormalPrior:
 
     References
     ----------
-    - D. Saunders, *A positive constrained non-centered prior that sparks joy*.
-    - Wikipedia, *Log-normal distribution — Definitions*.
+    - D. Saunders, `A positive constrained non-centered prior that sparks joy <https://daniel-saunders-phil.github.io/imagination_machine/posts/a-positive-constrained-non-centered-prior-that-sparks-joy/>`_.
+    - Wikipedia, `Log-normal distribution — Definitions <https://en.wikipedia.org/wiki/Log-normal_distribution#Definitions>`_.
     """
 
     def __init__(self, dims: tuple | None = None, centered: bool = True, **parameters):
```
williambdean, line 199: Call super here

daniel-s-tccc: Oh nice, thanks for the tip.

pymc_marketing/special_priors.py
daniel-s-tccc: Fixed.

PR #2116, opened 2025-12-02
[DOCS] Improve MMM Case Study Notebook

juanitorduz:
Improve mmm case study notebook

----
📚 Documentation preview 📚: https://pymc-marketing--2116.org.readthedocs.build/en/2116/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -147,3 +147,6 @@ dmypy.json
 # Gallery images
 docs/source/gallery/images/
 docs/gettext/
+
+# Sandbox folder
+sandbox/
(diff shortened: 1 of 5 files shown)


bhauryal-eliya:
Hello @juanitorduz , Currently, the optimize_budget method in the MMM class appears to use a static allocation approach. When optimizing over num_periods, the solver calculates a single optimal spend value per channel that is applied constantly across all time steps. Is it possible to make it dynamic, so the solver assigns different values per channel on different periods? Thank you!

juanitorduz:
> Hello @juanitorduz , Currently, the optimize_budget method in the MMM class appears to use a static allocation approach. When optimizing over num_periods, the solver calculates a single optimal spend value per channel that is applied constantly across all time steps. Is it possible to make it dynamic, so the solver assigns different values per channel on different periods? Thank you!

Hey! Sorry I missed this! Could you please open a GitHub issue so that we can track this request?

PR #2130, opened 2025-12-10
Fix plot_curve documentation examples not rendering

### LINKED ISSUE #2127: `plot_curve` examples not showing up
These are not showing up. Can see this here: https://www.pymc-marketing.io/en/latest/api/generated/pymc_marketing.plot.plot_curve.html

https://github.com/pymc-labs/pymc-marketing/blob/021b29b8f5743dd087153b2bdade19652111c1a1/pymc_marketing/plot.py?plain=1#L687-L715

Eshaan-byte:
Fixed the `plot_curve` documentation examples that were not rendering on the API documentation page.

## Changes
- Changed the second example's Sphinx plot directive from `:context: reset` to `:context: close-figs` in `pymc_marketing/plot.py:687`

## Root Cause
The second example was using `:context: reset` which clears all previous variables from the Sphinx plot directive context. This caused the example to fail because it tried to reference `curve` and `rng` variables that were defined in the first example but were cleared by the reset directive.

## Solution
By using `:context: close-figs` instead, the example now:
- Closes previous figures to avoid display issues
- Preserves all variables and imports from the first example
- Allows the second and third examples to properly reference `curve` and `rng`

## Testing
The fix ensures that all three plot examples in the `plot_curve` docstring will now render correctly in the documentation build.

Fixes #2127

----
📚 Documentation preview 📚: https://pymc-marketing--2130.org.readthedocs.build/en/2130/

### DIFF
--- a/pymc_marketing/plot.py
+++ b/pymc_marketing/plot.py
@@ -684,7 +684,7 @@ def plot_curve(
 
     .. plot::
         :include-source: True
-        :context: reset
+        :context: close-figs
 
         fig, axes = plot_curve(
             curve,


williambdean:
Looks great! Thanks
https://pymc-marketing--2130.org.readthedocs.build/en/2130/api/generated/pymc_marketing.plot.plot_curve.html?readthedocs-diff=true&readthedocs-diff-chunk=1

williambdean (approved):

PR #2135, opened 2025-12-14
Run whole suite with C and Numba linkers

ricardoV94:
Relies on next PyTensor/PyMC release

Testing compatibility in anticipation of default switch in PyTensor

----
📚 Documentation preview 📚: https://pymc-marketing--2135.org.readthedocs.build/en/2135/

### DIFF
--- a/pymc_marketing/mmm/transformers.py
+++ b/pymc_marketing/mmm/transformers.py
@@ -20,7 +20,7 @@ import numpy as np
 import numpy.typing as npt
 import pymc as pm
 import pytensor.tensor as pt
-from numpy.core.multiarray import normalize_axis_index
+from numpy.lib.array_utils import normalize_axis_index
 from pymc.distributions.dist_math import check_parameters
 from pytensor.tensor.variable import TensorVariable
 
--- a/pymc_marketing/pytensor_utils.py
+++ b/pymc_marketing/pytensor_utils.py
@@ -18,25 +18,21 @@ from collections import Counter
 
 import arviz as az
 import pandas as pd
-import pytensor
 import pytensor.tensor as pt
 from arviz import InferenceData
-from pymc import Model
+from pymc.model.core import Model
 from pymc.model.fgraph import (
-    FunctionGraph,
     ModelVar,
     extract_dims,
     fgraph_from_model,
     model_from_fgraph,
 )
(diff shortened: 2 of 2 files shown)


tests/mmm/test_transformers.py
```diff
@@ -692,10 +692,10 @@ def test_logistic_saturation_geometric_adstock_composition(self, x, alpha, lam):
         "x, alpha, lam, theta, l_max",
         [
             (np.ones(shape=(100)), 0.5, 1.0, 0, 1),
-            (np.ones(shape=(100)), 0.2, 19.0, 1, 2),
-            (np.zeros(shape=(100)), 0.6, 5.0, 3, 4),
-            (np.ones(shape=(100)), 0.99, 10.0, 0, 5),
-            (np.linspace(start=0.0, stop=1.0, num=50), 0.001, 0.01, 4, 5),
+            # (np.ones(shape=(100)), 0.2, 19.0, 1, 2),
```
ricardoV94: No idea what happened here, I'll revert

.github/workflows/test.yml
```diff
@@ -72,17 +75,20 @@ jobs:
...
           - "--ignore tests/clv/models/test_beta_geo_beta_binom.py --ignore tests/clv/models/test_pareto_nbd.py"
+    env:
+      PYTENSOR_FLAGS: linker=${{ matrix.linker }}
     steps:
       - uses: actions/checkout@v6
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.11"
+          python-version: "3.13"
```
ricardoV94, line 91: newer versions of python are getting faster, so better running slow test on those?

ricardoV94:
There are loads of splits in the CI, some that are like only 2-3 minute each. Doesn't matter much, but now everything is doubled as well with the linkers. If this is a concern we can run cvm on one of the python versions and numba on the other, so only notebook tests would be "duplicated"

.github/workflows/test.yml
juanitorduz: makes sense!

juanitorduz:
Thanks @ricardoV94 ! I agree. Let's merge this one and then we can pus them together :) 

Thanks!

PR #2222, opened 2026-01-26
Fix FutureWarning

ricardoV94:
Showed up in newer version of xarray/arviz in #2219

----
📚 Documentation preview 📚: https://pymc-marketing--2222.org.readthedocs.build/en/2222/

### DIFF
--- a/pymc_marketing/mmm/multidimensional.py
+++ b/pymc_marketing/mmm/multidimensional.py
@@ -1667,9 +1667,6 @@ class MMM(RegressionModelBuilder):
             ).to_dataset()
 
         dataarrays.append(y_xarray)
-        self.dataarrays = dataarrays
-        self._new_internal_xarray = xr.merge(dataarrays).fillna(0)
-
         return xr.merge(dataarrays).fillna(0)
 
     def _set_xarray_data(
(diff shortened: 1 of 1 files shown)


pymc_marketing/mmm/multidimensional.py
```diff
@@ -1667,10 +1667,7 @@ def _posterior_predictive_data_transformation(
             ).to_dataset()
 
         dataarrays.append(y_xarray)
-        self.dataarrays = dataarrays
```
ricardoV94, line 1670: these inner variables seemed odd?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1667,10 +1667,7 @@ def _posterior_predictive_data_transformation(
...
-        self._new_internal_xarray = xr.merge(dataarrays).fillna(0)
-
-        return xr.merge(dataarrays).fillna(0)
+        return xr.merge(dataarrays, join="outer", compat="no_conflicts").fillna(0)
```
ricardoV94, line 1670: AFAICT this is the old default, there were FutureWarnings that this would change. The notebook fails without `join="outer"`, so that's needed (or the function was broken).

Not sure whether we want something else for `compat`

pymc_marketing/mmm/multidimensional.py
williambdean: yea

williambdean (approved):
LGTM

PR #2224, opened 2026-01-26
Add narwhals support to Fivetran data adaptors

### LINKED ISSUE #1925: Use narwhals for fivetran data adaptor logic
The fivetran adaptors can be easy use case for narwhals and opens support to polars, pyspark, etc

Related to #1691

williambdean:
This adds polars.DataFrame and LazyFrame support via narwhals. There is limited pyspark testing here but that functionality could be included here. Please raise another issue if issues are encountered.

Implements **narwhals** support for Fivetran data processing functions (`process_fivetran_ad_reporting()` and `process_fivetran_shopify_unique_orders()`), enabling multi-backend DataFrame compatibility (pandas, Polars, PyArrow) while maintaining 100% backwards compatibility with the existing pandas API.

Resolves #1925


juanitorduz (commented):
Thanks @williambdean ! Can we add some tests in https://github.com/pymc-labs/pymc-marketing/blob/main/tests/data/test_fivetran.py 🙏 :)

williambdean:
Yes. Noticing that the LazyFrame support of pivot is limited from narwhals

I will explore route around. 

My goal is to support: 

- pandas
- polars DataFrame LazyFrame
- pyspark (Databricks notebooks)

pymc_marketing/data/fivetran.py
```diff
@@ -139,43 +147,84 @@ def _aggregate_and_pivot(
...
     """
+    # Group and aggregate
+    # Map aggregation string to narwhals aggregation method
+    agg_map = {
+        "sum": lambda c: nw.col(c).sum(),
+        "mean": lambda c: nw.col(c).mean(),
+        "min": lambda c: nw.col(c).min(),
+        "max": lambda c: nw.col(c).max(),
+        "count": lambda c: nw.col(c).count(),
+    }
```
FBruzzesi, line 584: We don't have `nw.count` (if you don't expect nulls, you could use `nw.len`), but for all the others aggregations we have top level functionalities, so there is no need to create lambdas:

```suggestion
    agg_map = {
        "sum": nw.sum,
        "mean": nw.mean,
        "min": nw.min,
        "max": nw.max,
        "count": lambda c: nw.col(c).count(),
    }
```

pymc_marketing/data/fivetran.py
```diff
@@ -139,43 +147,84 @@ def _aggregate_and_pivot(
...
+    if agg not in agg_map:
+        raise ValueError(
+            f"Unsupported aggregation: {agg}. Supported: {list(agg_map.keys())}"
+        )
+
+    agg_fn = agg_map[agg]
+    agg_exprs = [agg_fn(c) for c in value_columns]
```
FBruzzesi: Polars/Narwhals syntax would allow for:

```suggestion
    agg_exprs = agg_fn(value_columns)
```

pymc_marketing/data/fivetran.py
```diff
@@ -139,43 +147,84 @@ def _aggregate_and_pivot(
...
     grouped = (
-        df[[date_col, platform_col, *value_columns]]
-        .groupby([date_col, platform_col], dropna=False)
-        .agg(agg)
+        df.select([date_col, platform_col, *value_columns])
+        .group_by([date_col, platform_col], drop_null_keys=False)
+        .agg(agg_exprs)
     )
```
FBruzzesi: The select is possibly redundant - for pandas/pyarrow we end up running some computation eagerly:

```suggestion
        df
        .group_by([date_col, platform_col], drop_null_keys=False)
        .agg(agg_exprs)
    )
```

pymc_marketing/data/fivetran.py
```diff
@@ -189,37 +238,58 @@ def _finalize_wide_output(
...
+    narwhals.DataFrame
         Finalized wide-format dataframe with standardized columns and first column as the date.
     """
     if include_missing_dates:
-        full_index = pd.date_range(
-            start=wide.index.min(), end=wide.index.max(), freq=freq
+        # FBruzzesi's workaround for reindex: use join with full date range
+        # Get min/max dates (may trigger computation for lazy frames)
+        min_date = wide.select(nw.col(date_col).min()).item()
+        max_date = wide.select(nw.col(date_col).max()).item()
```
FBruzzesi: You can save some typing (and maybe it's more intuitive)

```suggestion
        date_series = wide.get_column(date_col)
        min_date, max_date = date_series.min(), date_series.max()
```

pymc_marketing/data/fivetran.py
```diff
@@ -189,37 +238,58 @@ def _finalize_wide_output(
...
     -------
-    pandas.DataFrame
+    narwhals.DataFrame
         Finalized wide-format dataframe with standardized columns and first column as the date.
     """
     if include_missing_dates:
-        full_index = pd.date_range(
-            start=wide.index.min(), end=wide.index.max(), freq=freq
+        # FBruzzesi's workaround for reindex: use join with full date range
+        # Get min/max dates (may trigger computation for lazy frames)
```
FBruzzesi: Yes this definitely requires the dataframe to be eager

pymc_marketing/data/fivetran.py
```diff
@@ -189,37 +238,58 @@ def _finalize_wide_output(
...
+        min_date = wide.select(nw.col(date_col).min()).item()
+        max_date = wide.select(nw.col(date_col).max()).item()
+
+        # Generate full date range using pandas (temporary)
+        full_dates_pd = pd.DataFrame(
+            {date_col: pd.date_range(start=min_date, end=max_date, freq=freq)}
         )
```
FBruzzesi: I guess related to https://github.com/narwhals-dev/narwhals/discussions/3425

---

The current implementation would have issue for polars: `full_dates` will be a narwhals datarame backend by pandas no matter what. The later join will raise as it tries to join a pandas with a polars dataframe.

The current workaround would be to have some extra utility that branches based on the original implementation. 
Here is an unpolished version:

```py
def date_range(
    frame: nw.DataFrame,
    date_col: str,
    freq: str,
) -> nw.DataFrame:
    date_series = frame.get_column(date_col)
    min_date, max_date = date_series.min(), date_series.max()
    
    if (impl:=frame.implementation).is_polars():
        import polars as pl
        dates = pl.date_range(start=min_date, end=max_date, freq=freq)
    else:  # fallback to pandas as pyarrow does not have a date_range functionality
        dates = pd.date_range(start=min_date, end=max_date, freq=freq)

    dates_frame = nw.from_native(dates, series_only=True).rename(date_col).to_frame()
    return dates_frame.lazy().collect(backend=impl)  # This guarantees to remap pandas to pyarrow
```

The last line is really a workaround if you want to support pyarrow as well. We have a related issue for how to better align dataframe backends but it's till under discussion https://github.com/narwhals-dev/narwhals/issues/2193

pymc_marketing/data/fivetran.py
```diff
@@ -189,37 +238,58 @@ def _finalize_wide_output(
...
+        # Left join: keep all dates from full range
+        wide = full_dates.join(wide, on=date_col, how="left")
+
+    # Fill nulls if specified
     if fill_value is not None:
-        wide = wide.fillna(fill_value)
+        # Fill null values in all columns except the date column
+        cols_to_fill = [c for c in wide.columns if c != date_col]
+        fill_exprs = [nw.col(c).fill_null(fill_value).alias(c) for c in cols_to_fill]
+        wide = wide.with_columns(fill_exprs)
```
FBruzzesi: If you want to embrace the polars syntax fully, the following should be enough:

```suggestion
        fill_expr = nw.exclude(date_col).fill_null(fill_value)
        wide = wide.with_columns(fill_expr)
```

FBruzzesi (commented):
Hey @williambdean 👋🏼

I added a few comments - mostly related to ergonomics that polars/narwhals enable.

There is a  _critical_ issue I see when the `full_dates_pd` dataframe is created as the backend might differ unless it's pandas in the first place. I added an unpolished suggestion for how I would address it, I hope it helps

pymc_marketing/data/fivetran.py
williambdean: Thanks for the suggestion. Went with it!

juanitorduz (approved):
🙌

FBruzzesi:
Thanks @williambdean - If there is a straightforward way to run the part of the CI that uses narwhals, we can consider adding pymc-marketing in our [downstream tests](https://github.com/narwhals-dev/narwhals/blob/main/.github/workflows/downstream_tests.yml)

williambdean:
> Thanks @williambdean - If there is a straightforward way to run the part of the CI that uses narwhals, we can consider adding pymc-marketing in our [downstream tests](https://github.com/narwhals-dev/narwhals/blob/main/.github/workflows/downstream_tests.yml)

Good thinking. I will look into that and let y'all know

PR #2232, opened 2026-01-28
Port MMM Quickstart Notebook to use Multidimensional MMM

ricardoV94:
Closes #2180

----
📚 Documentation preview 📚: https://pymc-marketing--2232.org.readthedocs.build/en/2232/

Issue #2230, opened 2026-01-28
UX issues with plot.waterfall_components_decomposition(original_scale=True)


ricardoV94:
* If there are no `original_scale` variables it silently returns results on internal scale

* Users have to manually call `add_original_scale_contribution_variable`

* If only some components are available in `original_scale`, it silently ignores the remaining ones.

All this seems wrong. Users shouldn't have to know what contributions variable there are, or add them manually. We should compute those things automatically (fine to be on-demand, to not impose a memory cost everywhere).

ricardoV94:
Plotting method is detached from the model, so it doesn't know what are all the `_contribution` variables...

PR #2234, opened 2026-01-29
Small fixes to multidimensional MMM

### LINKED ISSUE #2213: Better documentation and validation for `dims` argument in multidimensional mmm
First if the argument is for extra dims, should it be called `extra_dims` ?.

The docstrings should state what are the default dims. control, channel, date, and fourier_mode (others?)?

dims argument should be validated:
 * `date` seems to be validated explicitly
 * I can pass `channel` without an error (even though the model will be non-sensical).  
 * If I pass `control` and `fourier_mode` I get an error that "dims must be unique", which may be too low-level for users

Also where are the coords/dimension of these extra dims coming from? `X/Y` when we pass xarray / pandas MultiIndex?

### LINKED ISSUE #2229: Default `target_column` for Multidimensonal MMM`?
Would it be too "magical" to leave the default be `y` like for regular `MMM`?

Without more context, seems like a fine default?

ricardoV94:
Cherry picked some commits from #2232 

~~NoLongerCloses #2230~~
Closes #2229
Closes #2213

Commits are logically organized

### DIFF
--- a/pymc_marketing/mmm/multidimensional.py
+++ b/pymc_marketing/mmm/multidimensional.py
@@ -156,7 +156,7 @@ import json
 import warnings
 from collections.abc import Callable, Sequence
 from copy import deepcopy
-from typing import Annotated, Any, Literal, cast
+from typing import Annotated, Any, cast
 
 import arviz as az
 import numpy as np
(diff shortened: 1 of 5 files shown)


pymc_marketing/mmm/multidimensional.py
```diff
@@ -751,15 +757,15 @@ def default_model_config(self) -> dict:
         }
 
     @property
-    def output_var(self) -> Literal["y"]:
+    def output_var(self) -> str:
```
williambdean: I think the DataFrame column in is different from the variable name of the Likelihood

ricardoV94: You're right, but should it? Or taking a step back, do we even need this argument? It seems like `y` has to be passed separately all the time anyway?

ricardoV94: Anyway, I did misunderstand the point of this variable, I'll revert these changes and clarify the docstrings

juanitorduz (approved):

Issue #2247, opened 2026-02-03
Port total contribution components plots


juanitorduz:
See https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_time_varying_media_example.html

to port into other examples

PR #2250, opened 2026-02-03
Fix Plot Contributions Over Time

juanitorduz:
Related to https://github.com/pymc-labs/pymc-marketing/issues/2247 , see https://github.com/pymc-labs/pymc-marketing/issues/2247#issuecomment-3843395335

I am also running the relevant notebooks.

## Summary

**Root Cause**: The `contributions_over_time` function was incorrectly inflating values for variables that don't have certain dimensions. Specifically:

1. The `coords` dictionary was built from ALL variables passed to the function, creating a superset of all coordinates
2. Each variable was then expanded with `expand_dims()` to include ALL coordinates from other variables (e.g., `intercept` was expanded to include `control` dimension from `control_contribution`)
3. The `_reduce_and_stack()` function then summed over these artificially-added dimensions, effectively **multiplying values by the size of those dimensions** (e.g., intercept × 3 controls = 3× inflation)

**Fix Applied** (in both `combine_dims=True` and `combine_dims=False` branches):

1. **Only expand `date` dimension**: Changed from expanding ALL missing coordinates to only expanding `date` if missing (needed for time series plotting)
2. **Filter indexers per variable**: Only apply dimension selectors (`sel()`) for dimensions that actually exist in each variable
3. **Skip duplicate plotting** (for `combine_dims=True`): Variables without indexed dimensions are only plotted once, avoiding overlapping identical lines

The fix ensures each variable is processed using only its native dimensions, preventing artificial inflation from dimensions belonging to other variables.

----
📚 Documentation preview 📚: https://pymc-marketing--2250.org.readthedocs.build/en/2250/

### DIFF
--- a/pymc_marketing/mmm/plot.py
+++ b/pymc_marketing/mmm/plot.py
@@ -1061,6 +1061,8 @@ class MMMPlotSuite:
             # Single subplot with all dimension combinations overlaid
             fig, axes = self._init_subplots(1, ncols=1, figsize=figsize)
             ax = axes[0][0]
+            # Track variables without indexed dims to avoid duplicate plotting
+            plotted_vars_without_indexed_dims: set[str] = set()
 
             for dims_combo, addl_combo in total_combos:
                 # Build indexers for dims and additional_dims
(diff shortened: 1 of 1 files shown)


cetagostini (approved):
Amazing catch!

PR #2258, opened 2026-02-04
Fix SpecialPrior serialization and deterministic model IDs

PabloRoque:
## Summary

This PR fixes serialization/deserialization issues for models using `SpecialPrior` subclasses (like `LogNormalPrior`, `LaplacePrior`) and ensures model IDs are deterministic across save/load cycles.

## Problem

1. **SpecialPrior deserialization failed**: Models with `LogNormalPrior` or `LaplacePrior` in saturation functions couldn't be loaded, throwing `DeserializableError: Couldn't deserialize {'special_prior': 'LogNormalPrior', ...}`
2. **Non-deterministic model IDs**: The `id` property used `str(self.model_config.values())` which included memory addresses like `<LogNormalPrior object at 0x...>`, causing different IDs for identical configs after save/load
3. **Non-serializable config**: `_serializable_model_config` returned dicts containing non-serializable objects instead of truly JSON-serializable data

### 1. Generic SpecialPrior Deserializer (`special_priors.py`)
- Added `__eq__` method to `SpecialPrior` for proper comparison with numpy array support
- Implemented generic deserializer using `__subclasses__()` auto-discovery
- New SpecialPrior subclasses automatically get deserialization support without manual registration
- Registered centrally in `pymc_marketing/__init__.py` (similar to fivetran import pattern)

### 2. Deterministic Model IDs (`model_builder.py`)
- Changed `id` property to use `json.dumps(self._serializable_model_config, sort_keys=True)` instead of `str()`
- Ensures consistent IDs across save/load cycles

### 3. Truly Serializable Config (`multidimensional.py`)
- Rewrote `_serializable_model_config` to recursively serialize all values
- Handles numpy arrays, nested dicts, objects with `to_dict()` or `model_dump()` methods

### 4. Comprehensive Tests (`test_special_priors.py`)
- `test_special_prior_serialization_round_trip()` - Tests deserialize() works
- `test_special_prior_equality()` - Tests `__eq__` method with various cases
- `test_mmm_with_special_prior_save_load_round_trip()` - Full integration test with fit/save/load

## Testing

All new tests pass, and pre-commit hooks (ruff, mypy) pass cleanly.

## Breaking Changes

None. This is a bug fix that makes the existing functionality work correctly.

----
📚 Documentation preview 📚: https://pymc-marketing--2258.org.readthedocs.build/en/2258/

### DIFF
--- a/pymc_marketing/mmm/multidimensional.py
+++ b/pymc_marketing/mmm/multidimensional.py
@@ -768,19 +768,28 @@ class MMM(RegressionModelBuilder):
 
     @property
     def _serializable_model_config(self) -> dict[str, Any]:
-        def ndarray_to_list(d: dict) -> dict:
-            new_d = d.copy()  # Copy the dictionary to avoid mutating the original one
-            for key, value in new_d.items():
-                if isinstance(value, np.ndarray):
-                    new_d[key] = value.tolist()
-                elif isinstance(value, HSGPKwargs):
-                    new_d[key] = value.model_dump()
-                elif isinstance(value, dict):
-                    new_d[key] = ndarray_to_list(value)
-            return new_d
+        def serialize_value(value):
+            """Recursively serialize values to JSON-compatible types."""
+            if isinstance(value, np.ndarray):
+                return value.tolist()
+            elif isinstance(value, dict):
+                return {k: serialize_value(v) for k, v in value.items()}
+            elif isinstance(value, (list, tuple)):
+                return [serialize_value(v) for v in value]
+            elif hasattr(value, "to_dict"):
+                # Handle any object with a to_dict method (Prior, SpecialPrior, etc.)
+                return value.to_dict()
+            elif hasattr(value, "model_dump"):
+                # Handle Pydantic models
+                return value.model_dump()
+            else:
+                return value
 
-        serializable_config = self.model_config.copy()
-        return ndarray_to_list(serializable_config)
+        serializable_config = {}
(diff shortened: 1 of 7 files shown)


juanitorduz:
Amazing @PabloRoque ! I will look into it! In the meantime, I think there are some tests failing :)

pymc_marketing/mmm/multidimensional.py
```diff
@@ -768,19 +768,28 @@ def add_events(
 
     @property
     def _serializable_model_config(self) -> dict[str, Any]:
-        def ndarray_to_list(d: dict) -> dict:
-            new_d = d.copy()  # Copy the dictionary to avoid mutating the original one
-            for key, value in new_d.items():
-                if isinstance(value, np.ndarray):
-                    new_d[key] = value.tolist()
-                elif isinstance(value, HSGPKwargs):
```
juanitorduz, line 776: Can we still serialize `HSGPKwargs`?

juanitorduz (commented):
@PabloRoque LGTM, but I have a question:

pymc_marketing/mmm/multidimensional.py
PabloRoque: Yes.

1) `HSGPKwargs` is a pydantic model
2) This gets serialized as 

```
    if hasattr(obj, "model_dump"):   
        return obj.model_dump(mode="json")
```

juanitorduz (approved):
Thanbk you @PabloRoque

PR #2273, opened 2026-02-09
chore: POC for backward compatible checks of models

williambdean:
## Description

The model_version in the ModelBuilder is a bit brittle. 

This puts some checks that a PRs changes would be able to be loaded in again.

Current checks might also be brittle as well. Open to suggestions.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2273.org.readthedocs.build/en/2273/


williambdean:
@copilot Please add tqdm to the conda enviroment so that it will be installed for these backward compatible checks in the CI/CD

ricardoV94:
What about building a hash of the generated model (more precisely fgraph_from_model), and having that hard-coded in the CI. Then there is no need to try and run in older versions of the library.

Want something dumb? `mmm.model.logp().dprint(print_type=True)` If the logp hasn't changed then the model likely neither? Maybe add `model.named_vars_to_dims` for good measure

williambdean:
Would you recommend changing this: 

https://github.com/pymc-labs/pymc-marketing/blob/beb622450820ebac678819aac3fb4be7995b4a29/pymc_marketing/model_builder.py?plain=1#L175-L223

Currently, it doesn't need data or build for an id.

PR #2293, opened 2026-02-14
Default to unpooled transformation priors in Multidimensional MMM

### LINKED ISSUE #2214: Default adstock/saturation parameters dims in multidimensional mmm
The default dims are `()`, implying complete pooling over `(*self.dims, channel)`.

Should these be completely unpooled by default, i.e., `dims=(*self.dims, channel)`? 

For reference the default dims for the control variates are `(*self.dims, control)`, and for measurement noise (in the likelihood) is `(*self.dims)`, doesn't it make sense to be the same for the adstock/saturation parameters?

It would make the transition from regular MMM to Multidimensional MMM easier as well, as the default was completely unpooled in the old class.

The method `set_dims_for_all_priors` seems a bit clunky too, and I suspect wouldn't be needed at all with the unpooled default?

ricardoV94:
Closes https://github.com/pymc-labs/pymc-marketing/issues/2214

Depends on https://github.com/pymc-devs/pymc-extras/pull/649

----
📚 Documentation preview 📚: https://pymc-marketing--2293.org.readthedocs.build/en/2293/


juanitorduz (approved):
This looks great! thanks for the great work :D

juanitorduz:
@williambdean It would be great if we could get your feedback :)

williambdean (approved):
LG!

PR #2304, opened 2026-02-18
fix(mmm): allow constant tensors in adstock sample_prior (fixes #1749)

### LINKED ISSUE #1749: plot curve workflow with constant parameters
The plot curve workflow 

```python
import pytensor.tensor as pt
from pymc_marketing.mmm import GeometricAdstock

alpha = pt.as_tensor_variable([0.5, 0.3, 0.2])
adstock = GeometricAdstock(priors={"alpha": alpha})

# No variable to sample so PyMC fails
coords = dict(channel=["A", "B", "C"])
prior = adstock.sample_prior(coords=coords)
```

Luckily this shouldn't cause an issue when being pass to a MMM and there is the work around
of using the pm.DiracDelta distribution to sample a constant value if really needed.

shivamlalakiya:
## Description
Fixes #1749

Previously, `sample_prior` would fail if an adstock transformation (e.g., `GeometricAdstock`) was initialized with constant tensors instead of PyMC distributions, as `pm.sample_prior_predictive` requires free random variables.

This PR adds a fallback mechanism: if no free random variables are detected in the model, it evaluates the constant tensors deterministically and returns a valid `InferenceData` object.

## Changes
- Modified `sample_prior` in `transformers.py` to handle `ValueError` from `sample_prior_predictive`.
- Added regression test `tests/mmm/test_issue_1749.py` to verify constant inputs work.

## Related Issue
- [x] Closes #1749 
- [ ] Related to #

## Checklist
- [x] If you are a pro: each commit corresponds to a [relevant logical change]

----
📚 Documentation preview 📚: https://pymc-marketing--2304.org.readthedocs.build/en/2304/

### DIFF
--- a/pymc_marketing/mmm/components/base.py
+++ b/pymc_marketing/mmm/components/base.py
@@ -425,10 +425,24 @@ class Transformation:
         xr.Dataset
             The dataset with the sampled priors.
 
+        Notes
+        -----
+        Parameters supplied as constant tensors (instead of a :class:`Prior`)
+        are not random variables, so ``pm.sample_prior_predictive`` has nothing
+        to sample. They are wrapped in a ``pm.DiracDelta`` so their constant
+        value is returned for every draw (fixes #1749).
+
         """
         coords = coords or {}
         with pm.Model(coords=coords):
-            self._create_distributions()
+            variables = self._create_distributions()
+            for parameter_name, variable_name in self.variable_mapping.items():
+                # ``_create_distributions`` returns constants unchanged (no
+                # random variable was registered for them), so wrap them in a
+                # DiracDelta to give ``sample_prior_predictive`` something to
+                # sample. Distributions return a freshly created variable.
+                if variables[parameter_name] is self.function_priors[parameter_name]:
+                    pm.DiracDelta(variable_name, variables[parameter_name])
             prior_pred = pm.sample_prior_predictive(**sample_prior_predictive_kwargs)
             return prior_pred["/prior"].to_dataset()
 
--- a/tests/mmm/components/test_adstock.py
+++ b/tests/mmm/components/test_adstock.py
@@ -117,6 +117,28 @@ def test_repr() -> None:
     )
 
 
+def test_geometric_adstock_sample_prior_constant_alpha() -> None:
+    """Regression test for GitHub issue #1749.
+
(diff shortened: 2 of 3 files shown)


pymc_marketing/mmm/components/base.py
```diff
@@ -434,8 +441,20 @@ def sample_prior(
...
+
+            # Register any constant tensors (non-distribution priors) as
+            # pm.Deterministic nodes so they are captured in the prior output.
+            # This handles both the case where all priors are constants and
+            # mixed cases where some priors are distributions and others are
+            # constant tensors.
+            for param_name, var_name in self.variable_mapping.items():
+                prior = self.function_priors[param_name]
+                if not hasattr(prior, "create_variable"):
+                    pm.Deterministic(var_name, variables[param_name])
```
williambdean: Don't think we should go this route.

shivamlalakiya:
Rebased onto `v1.0.0`; CI is now fully green (the earlier `test_mlflow…[blackjax]` failure is resolved here since `v1.0.0` pins `blackjax<1.6`, and a one-off `docs` failure was a transient `pandas.pydata.org` 503 during the intersphinx fetch — green on re-run).

Recap of how the feedback was addressed:
- **@ricardoV94** — agreed there's no true prior to sample; this uses `pm.DiracDelta` (the workaround noted in #1749) so `sample_prior_predictive` returns an `InferenceData` carrying the constant values.
- **@williambdean** — dropped the earlier `pm.Deterministic` route in favour of `pm.DiracDelta`.
- **cursor** — constants are wrapped per-parameter (identity check vs `function_priors`), so the mixed constant/`Prior` case is covered (`test_sample_prior_mixed_constant_and_prior`); duplicate regression tests were consolidated into `test_adstock.py` / `test_base.py`.

One scope note: this fixes `sample_prior`. The full `plot_curve` workflow works for scalar constants and `Prior` vectors, but a constant **vector** (the exact `alpha=[0.5, 0.3, 0.2]` example) still fails in `sample_curve → geometric_adstock` because a bare constant tensor carries no named dims for `as_xtensor`. Happy to keep this PR scoped to `sample_prior` and open a follow-up for the curve path, or address it here — whichever you prefer.

Does this look good?

Issue #2270, opened 2026-02-09
Remove lifetimes from dependencies


juanitorduz:
We should not rely on this old package as it already started breaking things. For example: https://github.com/pymc-labs/pymc-marketing/pull/2269

As pointed out by Ricardo

> Eventually we have to move away from lifetimes. Just hard code the results we were getting from them in the tests and have some notes on how to reproduce them (will need older versions of python at some point)

We need to do it asap.

We should also revert https://github.com/pymc-labs/pymc-marketing/pull/2269

PR #2305, opened 2026-02-18
Add Input Validation for CLV Models

### LINKED ISSUE #175: Add `_check_inputs()` Utility to `clv` Module?
`lifetimes` has a utility function for validating inputs prior to model fitting, which I expanded on in my `btyd` fork for `GammaGammaModel` input validation:

https://github.com/ColtAllen/btyd/blob/main/btyd/models/__init__.py#L254

(note that L294-L297 should be removed because I've since realized this is not a modeling assumption. In fact, rescaling time can help with model convergence when fitting to many years of data).

**`logp` in `BetaGeoModel` and the `clv.distribution` classes are already checking for many of these constraints, but I'm not sure how informative the error messages are. Should we add an additional validation check utility?**

Usually these constraints won't be violated if data is preprocessed with `clv_summary()` prior to model fitting, but that function is in written in `pandas` and will not be adequate for massive datasets exceeding local RAM size, particularly for data scientists using PySpark:

https://www.databricks.com/notebooks/Customer%20Lifetime%20Value%20Virtual%20Workshop/02%20The%20BTYD%20Models.html

Or just querying directly from a database, which is why the documentation for my `btyd` fork provides a SQL template:

https://btyd.readthedocs.io/en/latest/Generate%20RFM%20Summaries%20From%20a%20SQL%20Store.html

Could be worth referencing both of those in the documentation.

shivamlalakiya:
## Description
Addresses issue #175.

Following maintainer feedback, this PR implements a shared `_check_inputs` utility method in the base `CLVModel` rather than limiting it to `BetaGeoModel`. This centralizes data validation (`frequency`, `recency`, `T`) for all models that share identical input requirements (BG/NBD, Pareto/NBD, MBG/NBD). 

It also includes parameter handling to support the specific condition variations required by models like `ShiftedBetaGeoModel`.

Previously, invalid data (e.g., negative inputs or `recency > T`) would cause obscure errors downstream during model fitting or sampling. Now, the base model catches these early and raises a clear, descriptive `ValueError`.

## Changes
* Added `_check_inputs` to the base `CLVModel` to validate `frequency`, `recency`, and `T`.
* Added parameter support to accommodate specific edge cases (e.g., for `ShiftedBetaGeoModel`).
* Updated relevant CLV models to utilize this shared base validation.
* Updated unit tests in `tests/clv/` covering:
    * Negative values for frequency, recency, and T.
    * The invalid case where `recency > T`.
    * Ensure base model validation logic covers multiple CLV models.

## Related Issue
- [x] Closes #175 
- [ ] Related to #

## Checklist
- [x] If you are a pro: each commit corresponds to a [relevant logical change]

----
📚 Documentation preview 📚: https://pymc-marketing--2305.org.readthedocs.build/en/2305/

### DIFF
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -18,6 +18,7 @@ from collections.abc import Sequence
 from typing import Literal, cast
 
 import arviz as az
+import numpy as np
 import pandas as pd
 import pymc as pm
 import xarray as xr
@@ -77,6 +78,41 @@ class CLVModel(ModelBuilder):
             if col in must_be_homogenous:
                 if data[col].nunique() != 1:
                     raise ValueError(f"Column {col} has non-homogeneous entries")
+
+    def _validate_frequency(self, data: pd.DataFrame) -> None:
+        """Validate frequency column values shared across BTYD and Gamma-Gamma models."""
+        if (data["frequency"] < 0).any():
+            raise ValueError("Column frequency has negative values")
+
+        if not (np.mod(data["frequency"], 1) == 0).all():
+            raise ValueError("frequency column must contain only integer values")
+
+    def _validate_rfm_data(self, data: pd.DataFrame) -> None:
+        """Validate frequency, recency, and T columns shared by BTYD transaction models."""
+        self._validate_cols(data, required_cols=["frequency", "recency", "T"])
+        self._validate_frequency(data)
+
+        if (data["recency"] < 0).any():
+            raise ValueError("Column recency has negative values")
+        if (data["T"] < 0).any():
+            raise ValueError("Column T has negative values")
+
+        if ((data["frequency"] == 0) & (data["recency"] > 0)).any():
+            raise ValueError("recency cannot be greater than 0 if frequency is 0")
+
+        if (data["recency"] > data["T"]).any():
+            raise ValueError("recency cannot be greater than T")
+
+        if (data["T"] == 0).any():
+            warnings.warn(
+                "T=0 is mathematically valid but practically useless.",
+                UserWarning,
+                stacklevel=2,
+            )
+
+    def _validate_data(self, data: pd.DataFrame) -> None:
+        """Validate model input data. Child classes should override and call super()."""
+        self._validate_rfm_data(data)
 
     def __repr__(self) -> str:
         """Representation of the model."""
--- a/pymc_marketing/clv/models/beta_geo.py
+++ b/pymc_marketing/clv/models/beta_geo.py
@@ -193,9 +193,9 @@ class BetaGeoModel(CLVModel):
         """All covariate column names."""
         return self.purchase_covariate_cols + self.dropout_covariate_cols
 
-    # TODO: This placeholder will be superceded by https://github.com/pymc-labs/pymc-marketing/pull/2305
     def _validate_data(self, data: pd.DataFrame) -> None:
         """Validate BG/NBD-specific data requirements."""
+        super()._validate_data(data)
         self._validate_cols(
             data,
             required_cols=[
--- a/pymc_marketing/clv/models/beta_geo_beta_binom.py
+++ b/pymc_marketing/clv/models/beta_geo_beta_binom.py
@@ -168,9 +168,9 @@ class BetaGeoBetaBinomModel(CLVModel):
             "kappa_dropout": Prior("Pareto", alpha=1, m=1),
         }
 
-    # TODO: This placeholder will be superceded by https://github.com/pymc-labs/pymc-marketing/pull/2305
     def _validate_data(self, data: pd.DataFrame) -> None:
         """Validate BG/BB-specific data requirements."""
+        super()._validate_data(data)
         self._validate_cols(
             data,
             required_cols=[
--- a/pymc_marketing/clv/models/gamma_gamma.py
+++ b/pymc_marketing/clv/models/gamma_gamma.py
@@ -13,6 +13,8 @@
 #   limitations under the License.
 """Gamma-Gamma Model for expected future monetary value."""
 
+import warnings
+
 import numpy as np
 import pandas
 import pymc as pm
@@ -182,7 +184,8 @@ class BaseGammaGammaModel(CLVModel):
         Parameters
         ----------
         transaction_model : ~CLVModel
-            Predictive model for future transactions. `BetaGeoModel` and `ParetoNBDModel` are currently supported.
+            Predictive model for future transactions. `BetaGeoModel`,
+            `ModifiedBetaGeoModel`, and `ParetoNBDModel` are currently supported.
         data : ~pandas.DataFrame
             DataFrame containing the following columns:
 
@@ -322,7 +325,6 @@ class GammaGammaModel(BaseGammaGammaModel):
     ):
         super().__init__(model_config=model_config, sampler_config=sampler_config)
 
-    # TODO: This placeholder will be superceded by https://github.com/pymc-labs/pymc-marketing/pull/2305
     def _validate_data(self, data: pandas.DataFrame) -> None:
         """Validate Gamma-Gamma-specific data requirements."""
         self._validate_cols(
@@ -330,6 +332,13 @@ class GammaGammaModel(BaseGammaGammaModel):
             required_cols=["customer_id", "monetary_value", "frequency"],
             must_be_unique=["customer_id"],
         )
+        self._validate_frequency(data)
+        if (data["monetary_value"] <= 0).any():
+            warnings.warn(
+                "Non-positive monetary values are practically problematic.",
+                UserWarning,
+                stacklevel=2,
+            )
 
     def build_model(self, data: pandas.DataFrame) -> None:  # type: ignore[override]
         """Build the model.
@@ -468,6 +477,14 @@ class GammaGammaModelIndividual(BaseGammaGammaModel):
         self._validate_cols(
             data, required_cols=["customer_id", "individual_transaction_value"]
         )
+        if (data["individual_transaction_value"] < 0).any():
+            raise ValueError("Column individual_transaction_value has negative values")
+        if (data["individual_transaction_value"] <= 0).any():
+            warnings.warn(
+                "Non-positive individual transaction values are practically problematic.",
+                UserWarning,
+                stacklevel=2,
+            )
 
     def build_model(self, data: pandas.DataFrame) -> None:  # type: ignore[override]
         """Build the model.
--- a/pymc_marketing/clv/models/pareto_nbd.py
+++ b/pymc_marketing/clv/models/pareto_nbd.py
@@ -236,9 +236,9 @@ class ParetoNBDModel(CLVModel):
         """All covariate column names."""
         return self.purchase_covariate_cols + self.dropout_covariate_cols
 
-    # TODO: This placeholder will be superceded by https://github.com/pymc-labs/pymc-marketing/pull/2305
     def _validate_data(self, data: pd.DataFrame) -> None:
         """Validate Pareto/NBD-specific data requirements."""
+        super()._validate_data(data)
         self._validate_cols(
             data,
             required_cols=[
--- a/pymc_marketing/clv/models/shifted_beta_geo.py
+++ b/pymc_marketing/clv/models/shifted_beta_geo.py
@@ -277,10 +277,12 @@ class ShiftedBetaGeoModel(CLVModel):
             must_be_unique=["customer_id"],
         )
 
-        if np.any(
-            (data["recency"] < 1) | (data["recency"] > data["T"]) | (data["T"] < 2)
-        ):
-            raise ValueError("Model fitting requires 1 <= recency <= T, and T >= 2.")
+        if (data["recency"] < 1).any():
+            raise ValueError("Column recency must be at least 1")
+        if (data["recency"] > data["T"]).any():
+            raise ValueError("recency cannot be greater than T")
+        if (data["T"] < 2).any():
+            raise ValueError("Column T must be at least 2")
 
         self._validate_cohorts(data, check_param_dims=("alpha", "beta"))
 
--- a/pymc_marketing/clv/utils.py
+++ b/pymc_marketing/clv/utils.py
@@ -62,7 +62,8 @@ def customer_lifetime_value(
     Parameters
     ----------
     transaction_model : ~CLVModel
-        Predictive model for future transactions. `BetaGeoModel` and `ParetoNBDModel` are currently supported.
+        Predictive model for future transactions. `BetaGeoModel`,
+        `ModifiedBetaGeoModel`, and `ParetoNBDModel` are currently supported.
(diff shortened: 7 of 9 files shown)


test_model_multi
ColtAllen, line 1: Let's remove this.

ColtAllen (changes requested):
Hey @shivamlalakiya,

Thanks for opening this PR! Can we move this method into `CLVModel` so that all CLV models can use it? Supporting the data requirements of all models will require some modification, but the end result will look very similar to this [legacy code](https://github.com/ColtAllen/btyd/blob/main/btyd/models/__init__.py#L254).

I'm nearly done with a major PR involving changes to the CLV API. For this PR we can keep things simple and just make changes to `CLVModel`, and I'll update my PR with the model-specific validation calls.

ColtAllen (changes requested):
Keep this in `CLVModel` despite what the Bugbot says. Yes, some changes are needed  to support `ShiftedBetaGeoModel`, but we can add a new parameter for those conditions. The BG/NBD, Pareto/NBD, and MBG/NBD all have identical requirements.

ColtAllen (changes requested):
Thanks @shivamlalakiya - this PR is on the right track! Can we add a monetary_value parameter for the Gamma-Gamma model as well? Tests will also need to be added for models other than BGNBD and sBG.

test_model_multi
ColtAllen: Curious how this file keeps getting re-created? Is this due to an MMM test?

shivamlalakiya:
Thank you @ColtAllen , I have updated the PR to address all the feedback. Let me know if there's anything else you'd like me to tweak before we merge!

pymc_marketing/clv/models/modified_beta_geo.py
```diff
@@ -283,6 +284,29 @@ def build_model(self) -> None:  # type: ignore[override]
...
+            data["monetary_value"] = monetary_value
+
+        return customer_lifetime_value(
+            transaction_model=self,
+            data=data,
+            future_t=future_t,
+            discount_rate=discount_rate,
+            time_unit=time_unit,
+        )
+
```
PabloRoque: Why do we need this here? Are we overwriting anything on `BetaGeoModel`? Or is it the same method?

shivamlalakiya: @PabloRoque Thank you, ModifiedBetaGeoModel inherits directly from CLVModel, not from BetaGeoModel, so nothing is being overridden here. This method is needed — without it, ModifiedBetaGeoModel would have no way to call expected_customer_lifetime_value directly. That said, since the implementation is identical to ParetoNBDModel's, would you prefer I move it up to CLVModel as a shared base method?

ColtAllen: > ModifiedBetaGeoModel would have no way to call expected_customer_lifetime_value directly

None of the transaction models call `expected_customer_lifetime_value` directly. It is only called from the `GammaGammaModel` for monetary value, which itself is a wrapper for `clv.utils.customer_lifetime_value`, which users can call directly if they want to use monetary predictions from an external model.

PabloRoque: @shivamlalakiya I am having a look at this line `class ModifiedBetaGeoModel(BetaGeoModel)` so indeed inherits from `BetaGeoModel` and that method is described in that class if I recall correctly.

ColtAllen (changes requested):
Remove the CLV method from the MBG/NBD model.

I noticed the docstrings for the CLV estimation functions are outdated, which might be the source of the confusion. While you're at it, do you mind quickly updating these two lines to indicate the MBG/NBD model is supported as well?

https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/models/gamma_gamma.py#L175

https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/utils.py#L65

pymc_marketing/clv/models/basic.py
```diff
@@ -90,6 +91,66 @@ def _validate_cols(
...
+
+        if np.isnan(recency).any() or np.isnan(T).any():
+            raise ValueError("Missing values (NaN/pd.NA) found in recency or T.")
+
+        if frequency is not None:
+            frequency = np.asarray(frequency).astype(float)
+            if np.isnan(frequency).any():
+                raise ValueError("Missing values (NaN/pd.NA) found in frequency.")
+            if np.any(frequency < 0):
+                raise ValueError("Frequency must be >= 0.")
```
PabloRoque: This is not the case for the `ModifiedBetaGeoModel`. That model can produce outputs without repeated purchases.

pymc_marketing/clv/models/beta_geo.py
```diff
@@ -178,6 +178,11 @@ def __init__(
             ],
             must_be_unique=["customer_id"],
         )
+        self._check_inputs(
+            recency=data["recency"],
+            T=data["T"],
+            frequency=data["frequency"],
+        )
```
PabloRoque: Maybe we can move `_check_inputs` to `CLVModel.__init__`? We are copying boiler plate in a number of models, and I think this could be handled there at once.

pymc_marketing/clv/models/beta_geo_beta_binom.py
```diff
@@ -170,6 +170,11 @@ def __init__(
             must_be_unique=["customer_id"],
             must_be_homogenous=["T"],
         )
+        self._check_inputs(
+            recency=data["recency"],
+            T=data["T"],
+            frequency=data["frequency"],
+        )
```
PabloRoque: `CLVModel.__init__`?

pymc_marketing/clv/models/gamma_gamma.py
```diff
@@ -190,18 +191,27 @@ def expected_customer_lifetime_value(
...
+
+        if monetary_value is None:
+            # Use Gamma-Gamma estimates for the expected_spend values
+            predicted_monetary_value = self.expected_customer_spend(data=data)
+            data["future_spend"] = predicted_monetary_value.mean(
+                ("chain", "draw")
+            ).copy()
+        else:
+            # Use the explicitly provided monetary_value
+            data["future_spend"] = monetary_value
```
PabloRoque: I don't undestand the logic here. Do you want to overide what the model provides? Then I think you can do it externally to the library.

pymc_marketing/clv/models/modified_beta_geo.py
PabloRoque: @shivamlalakiya I understand now. `expected_customer_lifetime_value` is an external helper, not attached to any of the classes. Since that method accepts `transaction_model` then I don't see the need to add this.

pymc_marketing/clv/models/pareto_nbd.py
```diff
@@ -225,6 +225,11 @@ def __init__(
             ],
             must_be_unique=["customer_id"],
         )
+        self._check_inputs(
+            recency=data["recency"],
+            T=data["T"],
+            frequency=data["frequency"],
+        )
```
PabloRoque: `CLVModel.__init__`?

PabloRoque (commented):
@shivamlalakiya Thanks for the contribution, and sorry for the confusion around `expected_customer_lifetime_value`. I left some comments on the PR. Let me know your thoughts.

ColtAllen (commented):
@PabloRoque what do you think the code logic would look like if this were in `__init__.py`? I agree the current approach of parametrizing specific requirements in each model is a code smell.

Another possibility could be to use structured pattern matching with the `_model_type` attribute of the model child classes. This should require only a single parameter for `_check_inputs`, and would look something like this pseudocode:

```python

# declaration in CLVModel
def check_inputs(self, model_type):
     match model_type:
         case ["BG/NBD", "MBG/NBD", "Pareto/NBD":
             # self.data["column"] checks
         case ""Gamma-Gamma Model (Individual Transactions)"":
             # frequency and monetary checks
         case "BG/BB":
             # input checks
         case "sBG":
             # input checks
         case _:
            raise ValueError("Invalid Model Type") # this will probably never happen, but we still need a wildcard case

# usage in child class
self._check_inputs(self._model_type).
``` 

This way validation logic for all models is hard-coded and lives in this single method.

pymc_marketing/clv/models/basic.py
ColtAllen: This looks like a non-negativity check, but yes - `frequency >= 1` is only required for `GammaGammaModel`.

tests/clv/models/test_pareto_nbd.py
```diff
@@ -89,6 +88,74 @@ def setup_class(cls):
...
+            ParetoNBDModel(data=invalid_data)
+
+    def test_expected_customer_lifetime_value_removed(self):
+        """ParetoNBDModel no longer exposes expected_customer_lifetime_value."""
+        dummy_data = pd.DataFrame(
+            {"customer_id": [1], "frequency": [2], "recency": [5], "T": [10]}
+        )
+        model = ParetoNBDModel(data=dummy_data)
+        assert not hasattr(model, "expected_customer_lifetime_value")
+
```
ColtAllen: This test should be removed.

ColtAllen (changes requested):
@shivamlalakiya we need to address the points raised by @PabloRoque and myself in https://github.com/pymc-labs/pymc-marketing/pull/2305#pullrequestreview-3859128343. The "Ask" mode in whatever AI you're coding with will give some pointers.

**Before writing any more code** We have a [high-priority issue](https://github.com/pymc-labs/pymc-marketing/issues/2270) and a CLV API change PR about to be posted, both of which may create merge conflicts with your PR. Let's get both of those merged first, but in the meantime let us all strategize the most Pythonic path forward.

ColtAllen:
Hey @shivamlalakiya, sorry this fell off my radar. https://github.com/pymc-labs/pymc-marketing/issues/2270 is merged now and work on this can continue. Let me know if you need help resolving the merge conflicts.

> Location: Should this match/case block live directly inside CLVModel._check_inputs(), or would you both prefer it somewhere else in the module (like init.py)?

Let's keep it in `CLVModel._check_inputs()`, though maybe give it a more descriptive name like `_validate_rfm_data`.

> Scope: Should I apply this match/case refactor to all models right now (including Gamma-Gamma and ShiftedBetaGeoModelIndividual), or just start with the primary transaction models (BG/NBD, MBG/NBD, Pareto/NBD, sBG, BG/BB)?

Let's apply it to all models because the additional effort required is minor. Let me know if you need help with the data requirements of each model.

> Syntax: I plan to match on the exact _model_type strings (e.g., "Shifted Beta-Geometric"). Does that work for you?

Yes.

shivamlalakiya:
Includes an unrelated CI fix for test_log_model_graph_no_graphviz (MLflow 3.x log ordering flake) to unblock test groups 4/5. Let me know if we need to revert this.

shivamlalakiya:
CI note: the only failing job is `tests/test_mlflow.py::test_autolog_pymc_model[blackjax]` (`TypeError: build_kernel.<locals>.kernel() got an unexpected keyword argument 'progress_bar'`). This is unrelated to this PR — it's blackjax dependency drift that also fails on other open PRs (e.g. #2742) and would fail on `main` as well (`main` was last green on 2026-07-05, the exact commit this branch is rebased onto). This PR touches no mlflow/sampling code, and all CLV tests — including the slow-sampling jobs — pass. Should clear once blackjax is pinned/fixed on `main`.

ColtAllen:
Thanks for resurrecting this @shivamlalakiya ! 

> CI note: the only failing job is `tests/test_mlflow.py::test_autolog_pymc_model[blackjax]` (`TypeError: build_kernel.<locals>.kernel() got an unexpected keyword argument 'progress_bar'`). 

This was resolved in the `v1.0.0` branch, which is the branch for all PRs included in the 1.0 release. Can you switch from `main` to that branch and see if there are any merge conflicts?

shivamlalakiya:
@ColtAllen thanks — rebasing onto `v1.0.0` did the trick. I've retargeted the PR base to `v1.0.0`, and **all 54 checks are green**. The blackjax NUTS failure is gone now that we're on `v1.0.0` (`blackjax<1.6`).

The only merge conflict was the removed `ShiftedBetaGeoModelIndividual` (#2734) plus two shifted-model tests that needed updating to the new `build_model(data=...)` API — both resolved.

Quick recap of how this now addresses the earlier review from you and @PabloRoque:

- **Centralized in `CLVModel`**: `_validate_rfm_data` / `_validate_frequency` are shared by BG/NBD, Pareto/NBD, and MBG/NBD; Gamma-Gamma adds its monetary-value + frequency checks; ShiftedBetaGeo keeps its `recency`/`T` constraints. The shared frequency check is non-negativity only, so MBG's zero-frequency customers are still valid.
- Filled the per-model `_validate_data` placeholders `v1.0.0` left behind (the `# ... superceded by #2305` ones) via `super()._validate_data()`; validation runs inside `build_model` (that's where the data lives now).
- Removed the redundant `ModifiedBetaGeoModel` override, removed the stray `test_model_multi`, dropped the old per-model validation tests, and updated the "supported models" docstrings in `gamma_gamma.py` and `clv/utils.py` to include MBG/NBD.
- Consolidated the tests into `tests/clv/test_models.py`, covering every model with `match=` on the specific error messages.

One thing I'd like your call on: I implemented the dispatch as per-model `_validate_data` overrides (matching the placeholders already in `v1.0.0`) rather than the earlier `match/case` on `_model_type`. Happy to switch to `match/case` if you'd prefer.

Does this look good to you, or is there anything else you'd like changed before merge?

shivamlalakiya:
Hi @ColtAllen — this PR was closed automatically on 2026-07-28, not by a review decision. The `v1.0.0` branch was deleted right after #2677 was squash-merged into `main`, and GitHub auto-closes every PR whose base branch is removed. Seven PRs went out the same way, including #2735, #2724, #2630, #2754 and #2742.

Some context on how it ended up targeting `v1.0.0`: it was originally against `main`, and I retargeted it on 2026-07-18 per your suggestion to get around the `blackjax` NUTS failure. After the rebase all 54 checks were green and the last outstanding item was the `match/case` placement, which we settled on keeping in `CLVModel`.

The branch is intact at `209f84be`. Since `v1.0.0` was squash-merged rather than merged, I'd rebase with `git rebase --onto upstream/main <old-v1.0.0-head>` so only my commits replay onto `main`.

Could you reopen and retarget the base to `main`? GitHub disables the reopen button while the base ref is missing, so it needs to be switched first. If you'd rather have a clean PR against `main` instead, happy to open one — just let me know which you prefer.

Same message works for 2304 with @ricardoV94/@williambdean, branch 3e5b363e, and "the constant-tensor DiracDelta approach ricardoV94 signed off on" swapped in for the match/case line. Say the word and I'll write that one too.

ColtAllen:
Sorry about that @shivamlalakiya , deleting the remote `v1.0.0` branch impacted a PR I was working on as well.

Can you reopen a clean PR from `main`? I'd like for you to have credit for this work.

Issue #2309, opened 2026-02-20
Support `cost_per_unit` for channels with non-monetary data (impressions, clicks, etc.) - part 1


isofer:
This ticket is for the implementation of part 1 of #2210
It covers:
1) core infrastructure (which automatically affect ROAs calculation in plot_interactive)
2) budget optimizer

Part 2 will cover:
 - MMMPlotSuite
 - saturation curves plots in plot_interactive

---

# Support `cost_per_unit` for channels with non-monetary data (impressions, clicks, etc.)

## Problem

When channel data is provided in non-monetary units (impressions, GRPs, clicks), all downstream calculations that assume monetary units produce meaningless results:

- **ROAS** becomes revenue-per-impression instead of revenue-per-dollar
- **Budget optimization** allocates "dollars" that are actually impression counts, making the budget constraint and optimal allocation nonsensical
- **Summary reports** mix units across channels (e.g. one channel in dollars, another in impressions)

Currently, users must manually convert all channel data to spend before fitting, which is error-prone and loses the original unit information. There is no first-class way to tell the model "this channel's data is in impressions and costs $X per unit."

## Proposed Solution

Add a `cost_per_unit` parameter that stores per-channel, per-date conversion factors (e.g. cost per impression) and applies them at strategic access points — **without modifying the model fitting or saturation/adstock curves**, which continue to operate on raw units.

**Phase 1 — Core Infrastructure (reporting)**:
- Accept `cost_per_unit` as a `pd.DataFrame` at init or post-hoc via `mmm.set_cost_per_unit()`
- Store in `idata.constant_data` alongside `channel_data`
- `get_channel_spend()` multiplies `channel_data × cost_per_unit` to return spend
- ROAS and summaries automatically use converted spend
- Missing channels default to 1.0 (assumed already in spend units)

**Phase 2 — Budget Optimization**:
- `optimize_budget()` accepts forward-looking `cost_per_unit` for the optimization window
- Converts dollar budgets to original units internally: `budget / cost_per_unit`
- Each time period uses its own rate (no averaging) — supports seasonal pricing, contract changes
- Budget inputs/outputs remain in monetary units for the user

## Example Usage

```python
# Channel data in impressions — specify cost per impression
cost_per_unit = pd.DataFrame({
    "date": dates,
    "TV": [0.01] * len(dates),      # $0.01 per impression
    "Radio": [0.02] * len(dates),   # $0.02 per impression
})

mmm = MMM(
[...]

PR #2320, opened 2026-02-23
Support cost_per_unit for channels with non-monetary data (impressions, clicks, etc.) - part 1

### LINKED ISSUE #2210: converting between impressions and spend
Users would like to model using impressions but create saturation curve and budget optimization based on spend/cost (e.g. #1136). This feature is supported in in other packages, like simba, that allows users to link impressions to spend for each channel.
We need to figure out how we want to deal with that in pymc-marketing. 
We need somehow to tell the model that the data for some of the channels is provided as cost (so 1000 = $1000), and some of the channels are impressions (so 1000 = 1000 impressions). 
We want the user to provide the value of an impression (e.g. 1000 imperssions from channel x cost y) to the model. I'm not sure even if this only a scalar information or is it time-dependent array. We might need to support both. 

This would not affect modeling, but plots and budget optimization (and maybe other thing) will use this information to convert impressions to cost. for instance, saturation curves should be computed based on cost.

isofer:
## Description

**Goal:** Support channels measured in non-monetary units (impressions, clicks) while keeping ROAS and budget optimization in spend terms.

This PR introduces `cost_per_unit` support for two independent purposes that share only the DataFrame parsing logic (`_parse_cost_per_unit_df`):

### 1. Historical cost-per-unit — for ROAS calculation

Uses the cost-per-unit from the **training period** to convert `channel_data` (in original units) into `channel_spend` (in monetary units). This precomputed spend is stored in `idata.constant_data` and used by `get_channel_spend()` for ROAS calculations in `incrementality` and `plot_interactive`.

- `cost_per_unit` can be provided at init or post-fit via `set_cost_per_unit()`.
- After fit: `channel_spend = channel_data × cost_per_unit` is injected into `idata.constant_data`.
- `get_channel_spend()` returns `channel_spend` when present, falls back to `channel_data` (backward compatible).
- `get_channel_data()` always returns raw data in original units (needed by the model evaluator).
- `cost_per_unit` property on the wrapper derives the rate on-the-fly as `channel_spend / channel_data`.

### 2. Future cost-per-unit — for budget optimization

Uses the cost-per-unit for the **optimization window** to convert monetary budgets into the model's native units before optimization. This is completely independent of the historical cost-per-unit above.

- `cost_per_unit` parameter added to `optimize_budget()` and `BudgetOptimizer`.
- Conversion pipeline: monetary budget → time distribution → divide by `cost_per_unit[t]` → channel scaling → model.
- Each time period uses its own cost-per-unit rate (no averaging).
- Output `optimal_budgets` remain in monetary units for user convenience.

### Shared: DataFrame parsing

Both use cases share `MMM._parse_cost_per_unit_df()`, which converts a wide-format DataFrame (rows = date × custom_dims, columns = channels) into an `xr.DataArray`. Missing channels default to 1.0 (already in spend units).

### Why we store channel_spend instead of cost_per_unit in idata?
1. Our computations (ROAS, budget optimization, plots) need `channel_spend`, not the rate. There is no need to store `cost_per_unit`.
2. `aggregate_idata_time` (used in incrementality and for interactive plots) sums every variable with a `date` dimension. If we stored the rate, we would get aggregated `channel_data` and aggregated `cost_per_unit` after time aggregation, but there is no way to derive total channel spend from those two. `channel_spend` is additive, so summing it over time yields the correct total spend.

## Related Issue
- [x] Closes #2210
- [x] Related to #2309

## Screenshots / Examples

```python
# Historical cost_per_unit — for ROAS
mmm = MMM(
    ...,
    cost_per_unit=cost_per_unit_df,  # wide-format: date, *dims, channel columns
)
idata = mmm.fit(X, y)
# idata.constant_data.channel_spend = channel_data * cost_per_unit

# Or set post-fit
mmm.set_cost_per_unit(cost_per_unit_df, overwrite=False)

# Future cost_per_unit — for budget optimization
result = optimizer.optimize_budget(
    budget=100_000,
    cost_per_unit=future_cost_per_unit_df,  # for optimization window
)
```

## Changes Made
- **`_parse_cost_per_unit_df()`** (static, in `multidimensional.py`): Shared parser that converts a wide-format DataFrame into an `xr.DataArray` aligned to model coordinates. Used by both historical and optimizer paths.
- **`mmm_wrapper.py`**: `get_channel_spend()` prefers `channel_spend` over `channel_data`; add `get_channel_data()` for raw units; add `cost_per_unit` property (derived on-the-fly).
- **`schema.py`**: Add `channel_spend` VariableSchema to constant_data (optional).
- **`multidimensional.py`**: `cost_per_unit` init param; fit-time injection of `channel_spend`; `set_cost_per_unit()`; serialize/deserialize `_cost_per_unit_input`; `_parse_cost_per_unit_for_optimizer()` on the budget wrapper; `cost_per_unit` param on `optimize_budget()`.
- **`budget_optimizer.py`**: Add `cost_per_unit` field and `_validate_and_process_cost_per_unit()`; convert budgets from monetary to native units in `_replace_channel_data_by_optimization_variable` (after time distribution, before channel scaling).
- **`incrementality.py`**: Use `get_channel_data()` instead of `get_channel_spend()` for `baseline_array` in evaluator (baseline must be in original units, not spend).

## Breaking Changes
- None. `get_channel_spend()` falls back to `channel_data` when `channel_spend` is absent (backward compatible).


pymc_marketing/mmm/multidimensional.py
```diff
@@ -3037,6 +3238,45 @@ def build_from_idata(self, idata: az.InferenceData) -> None:
...
+                "Model must be fitted before setting cost_per_unit. "
+                "Call mmm.fit() first."
+            )
+
+        if not hasattr(self.idata, "constant_data"):
+            raise ValueError("InferenceData missing constant_data group")
+
+        cost_per_unit_array = self._build_cost_per_unit_array(cost_per_unit)
+        channel_data = self.idata.constant_data.channel_data
+        self.idata.constant_data["channel_spend"] = channel_data * cost_per_unit_array
```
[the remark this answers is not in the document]
isofer: fixed

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

cetagostini (commented):
_(cleaned up — replaced by proper review)_

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -926,6 +1022,18 @@ def _replace_channel_data_by_optimization_variable(self, model: Model) -> Model:
                 shape=repeated_budgets_shape,
             )
 
+        # Convert from monetary units to original units using date-specific rates.
+        # Applied AFTER time distribution so each period uses its own cost rate.
+        if self._cost_per_unit_tensor is not None:
+            if date_dim_idx != 0:
```
cetagostini: This contradicts the pattern in `_apply_budget_distribution_over_period` which already handles arbitrary `date_dim_idx` by accepting it as a parameter. Either:

- **A)** Handle non-zero `date_dim_idx` correctly here too (transpose `_cost_per_unit_tensor` accordingly), or
- **B)** Move this validation to `__init__` / `__model_post_init__` so it fails fast instead of at optimization time.

Right now users discover this limitation deep inside `_replace_channel_data_by_optimization_variable`, and the error message exposes internal implementation details.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -838,6 +884,56 @@ def _validate_and_process_budget_distribution(
         # Store only the masked tensor
         return pt.constant(time_factors_masked, name="budget_distribution_over_period")
 
+    def _validate_and_process_cost_per_unit(
```
cetagostini, line 889: Let's use the `@staticmethod` decorator here -- `self` is never accessed inside this method body. That way we'll be able to test it properly instead of hacking a workaround, no? If it were a `@staticmethod`, it could be called directly as `BudgetOptimizer._validate_and_process_cost_per_unit(cost_per_unit, num_periods, budget_dims)` without the fragile `__get__` binding trick in the tests.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -728,6 +750,134 @@ def _check_compatible_media_dims(self) -> None:
...
+            dims=(*dim_order, "channel"),
+            coords={**coord_dict, "channel": channels},
+        )
+
+        return result.transpose("date", *custom_dims, "channel")
+
+    def _build_cost_per_unit_array(self, df: pd.DataFrame) -> xr.DataArray:
+        """Parse cost_per_unit DataFrame using coordinates from the fitted model."""
+        custom_dims = tuple(self.data.custom_dims)
+        return MMM._parse_cost_per_unit_df(
```
cetagostini: If `_parse_cost_per_unit_df` is a `@staticmethod`, why do we call it as `MMM._parse_cost_per_unit_df(...)` instead of `self._parse_cost_per_unit_df(...)`? Option two is more sustainable (respects subclass overrides) unless I'm missing something?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -708,6 +721,15 @@ def __eq__(self, other: object) -> bool:
         if self.sampler_config != other.sampler_config:
             return False
 
+        # cost_per_unit
```
cetagostini: The comment below calls `id` "a content-based hash that validates the entire config". If `cost_per_unit` is part of model identity, it should be included in the `id` hash. If it's not in the hash, two models with different `cost_per_unit` will have the same `id` but return `False` from `__eq__`, which is logically inconsistent. Worth either adding it to the hash or adding a comment explaining why it's intentionally excluded.

pymc_marketing/data/idata/mmm_wrapper.py
```diff
@@ -471,6 +536,36 @@ def get_elementwise_roas(self, original_scale: bool = True) -> xr.DataArray:
 
         return contributions / spend_safe
 
+    def get_roas(self, original_scale: bool = True) -> xr.DataArray:
```
cetagostini: Adding an automatically deprecated method? `get_roas()` didn't exist on `MMMIDataWrapper` before this PR, so there are no existing callers to deprecate for. If backward compat is the goal, a changelog migration note would be cleaner than landing dead-on-arrival deprecated code. If `get_roas()` is meant to be the user-facing name, let's not deprecate it.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -838,6 +884,56 @@ def _validate_and_process_budget_distribution(
...
+        Raises
+        ------
+        ValueError
+            If dimensions or date length don't match expectations.
+        """
+        if cost_per_unit is None:
+            return None
+
+        expected_dims = ("date", *budget_dims)
+        if set(cost_per_unit.dims) != set(expected_dims):
```
cetagostini, line 924: The `set()` comparison here is fine for validation (since `.transpose()` fixes order next), but the error message says `"must have dims {expected_dims}"` which reads like the order matters. Let's tweak the message to say something like "must have exactly the dims" to avoid confusion about order expectations.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -901,6 +1058,11 @@ def attrs_to_init_kwargs(cls, attrs: dict[str, str]) -> dict[str, Any]:
             "dag": json.loads(attrs.get("dag", "null")),
             "treatment_nodes": json.loads(attrs.get("treatment_nodes", "null")),
             "outcome_node": json.loads(attrs.get("outcome_node", "null")),
+            "cost_per_unit": (
+                pd.read_json(io.StringIO(attrs["cost_per_unit"]), orient="split")
```
cetagostini: `pd.read_json(orient="split")` doesn't preserve timezone-aware date dtypes from the original DataFrame. If someone uses tz-aware dates, the round-trip through `to_json`/`read_json` could silently strip the timezone and cause `_parse_cost_per_unit_df` reindex mismatches after a save/load cycle. Might want to add explicit `convert_dates` or dtype handling here.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -3039,6 +3241,49 @@ def build_from_idata(self, idata: az.InferenceData) -> None:
...
+            raise RuntimeError(
+                "Model must be fitted before setting cost_per_unit. "
+                "Call mmm.fit() first."
+            )
+
+        if not hasattr(self.idata, "constant_data"):
+            raise ValueError("InferenceData missing constant_data group")
+
+        cost_per_unit_array = self._build_cost_per_unit_array(cost_per_unit)
+        channel_data = self.idata.constant_data.channel_data
```
cetagostini, line 3360: Edge case: if `constant_data` exists but `channel_data` is absent (only `channel_spend` or nothing), this will blow up with a cryptic AttributeError from `_build_cost_per_unit_array`. A quick guard like in the `get_channel_data()` pattern would give a clearer error.

tests/mmm/test_cost_per_unit.py
```diff
@@ -0,0 +1,750 @@
...
+
+    @pytest.fixture
+    def optimizer_instance(self):
+        """Bare BudgetOptimizer instance just for calling the validation method."""
+
+        class Stub:
+            pass
+
+        stub = Stub()
+        stub._validate_and_process_cost_per_unit = (
```
cetagostini: This unbound-method binding trick is fragile and non-idiomatic. Per the comment on `_validate_and_process_cost_per_unit` -- if we make it a `@staticmethod`, this entire stub becomes unnecessary and we can just call `BudgetOptimizer._validate_and_process_cost_per_unit(cost_per_unit, num_periods, budget_dims)` directly. Much cleaner.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -642,6 +670,17 @@ class BudgetOptimizer(BaseModel):
...
+            "monetary units (dollars) to original units (impressions, clicks). "
+            "Must have dims (date, *budget_dims) where date has length "
+            "num_periods. If None, budgets are assumed to already be in "
+            "the model's native units (no conversion applied)."
+        ),
+    )
+
     compile_kwargs: dict | None = Field(
         default=None,
         description="Keyword arguments for the model compilation. Specially usefull to pass compilation mode",
```
cetagostini: Typo: `"Specially usefull"` -> `"Especially useful"`

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -926,6 +1022,18 @@ def _replace_channel_data_by_optimization_variable(self, model: Model) -> Model:
...
+                    "cost_per_unit conversion assumes date is the first dimension "
+                    f"in channel_data_dims, but date_dim_idx={date_dim_idx}. "
+                    "If channel_data_dims ordering has changed, update "
+                    "_cost_per_unit_tensor transpose accordingly."
+                )
+            repeated_budgets = repeated_budgets / self._cost_per_unit_tensor
+
         repeated_budgets.name = "repeated_budgets"
 
         # Pad the repeated budgets with zeros to account for carry-over effects
```
cetagostini, line 1042: Typo on next line: `"repeated budgehts"` -> `"repeated budgets"`

cetagostini (commented):
Great work on this feature -- the design is solid and the coverage is really thoughtful. Just needs a small round of polish before we merge. I left inline comments with specific suggestions on each file. Most are quick fixes (typos, a missing `@staticmethod`, a `sorted()` return type). The two meatier items are the hardcoded `date_dim_idx == 0` assumption and the untested `fit()` override path. Happy to discuss any of these -- please take another look and re-request review when ready!

pymc_marketing/mmm/multidimensional.py
```diff
@@ -3143,6 +3388,46 @@ def _set_predictors_for_optimization(self, num_periods: int) -> pm.Model:
...
+
+        custom_dim_coords = None
+        if custom_dims:
+            idata = self.model_class.idata
+            custom_dim_coords = {
+                dim: idata.constant_data.coords[dim].values  # type: ignore[union-attr]
+                for dim in custom_dims
+            }
+
+        return MMM._parse_cost_per_unit_df(
```
cetagostini: Same note as `_build_cost_per_unit_array` above -- if this is a `@staticmethod`, calling it via `MMM._parse_cost_per_unit_df(...)` instead of `self.model_class._parse_cost_per_unit_df(...)` breaks subclass overrideability. Consistency-wise, pick one pattern.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1596,6 +1758,46 @@ def add_original_scale_contribution_variable(self, var: list[str]) -> None:
                     dims=deterministic_dims,
                 )
 
+    def fit(  # type: ignore[override]
```
cetagostini, line 1837: The `fit()` override here is never exercised in the test suite -- all fixtures use `mock_fit()` which bypasses this path entirely. Could we add a test that creates the MMM with `cost_per_unit=cpu_df` in the constructor and then calls `mock_fit` (or a lightweight version)? Even a mock fit with this parameter in the initialization should be sufficient to cover this branch, no?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -3143,6 +3388,46 @@ def _set_predictors_for_optimization(self, num_periods: int) -> pm.Model:
...
+
+        Returns
+        -------
+        xr.DataArray
+            Dims ``("date", *custom_dims, "channel")`` with ``date`` length
+            ``num_periods``.
+        """
+        channels = self.model_class.channel_columns
+        custom_dims = tuple(self.model_class.dims)
+        dates = sorted(df["date"].unique())
```
cetagostini: Let's replace this by `pd.DatetimeIndex(sorted(df["date"].unique()))` or `pd.DatetimeIndex(df["date"].unique()).sort_values()`. `sorted()` returns a plain Python `list`, but `_parse_cost_per_unit_df` type-hints `dates: pd.DatetimeIndex | np.ndarray`. The list could bring hidden bugs downstream (e.g. xarray alignment mismatches, timezone handling).

pymc_marketing/mmm/budget_optimizer.py
isofer: I removed the check entirely because it is validated before anyway

pymc_marketing/mmm/budget_optimizer.py
isofer: done

pymc_marketing/mmm/multidimensional.py
isofer: done

pymc_marketing/mmm/multidimensional.py
isofer: I decided to remove cost_per_unit from __eq__ entirely. for the same reason idata is not checked.

pymc_marketing/data/idata/mmm_wrapper.py
isofer: done

pymc_marketing/mmm/multidimensional.py
isofer: done

tests/mmm/test_cost_per_unit.py
isofer: done

pymc_marketing/mmm/multidimensional.py
isofer: done

pymc_marketing/mmm/multidimensional.py
isofer: done

pymc_marketing/mmm/multidimensional.py
isofer: done

pymc_marketing/mmm/multidimensional.py
ricardoV94: very minor sidenote: but two things can have the same hash and evaluate unequal, not the other way around, so this is strictly speaking fine.

hash is a low resolution equality check.

The ask may still be fine

cetagostini: Good point, totally fine then. Seeing this properly, `set_cost_per_unit()` is a post-hoc method, no? If so, then no need to be part of model identity, still a bit confusing. So, I think a short comment explaining that `cost_per_unit` is intentionally excluded (it's a unit-conversion factor, not a structural model param) would be enough to close this. Opinions? @isofer

isofer: done

cetagostini:
I'm happy with the PR state, let's make all related test pass and update the branch, but looks good now! No further input from now, but will take a final look once the small details from bugbot are done, thanks Imri!

pymc_marketing/mmm/multidimensional.py
```diff
@@ -3153,6 +3464,7 @@ def optimize_budget(
         default_constraints: bool = True,
         budgets_to_optimize: xr.DataArray | None = None,
         budget_distribution_over_period: xr.DataArray | None = None,
+        cost_per_unit: pd.DataFrame | xr.DataArray | None = None,
```
PabloRoque, line 3521: Do we really want to allow `None` here if the MMM contains cost_per_unit data?
I would expect at least a fat warning.

PabloRoque (approved):
Great PR, looks trivial but there are lots of tiny pieces that need to fit together. 

Minor comment on my side.

pymc_marketing/mmm/multidimensional.py
isofer: done

PR #2290, opened 2026-02-12
Incrementally, ROAs, CAC, etc...

isofer:
## Description

The research for this feature has been split into two PRs to reduce complexity. This first PR focuses on the core `Incrementality` module and its tests; a follow-up PR will handle shared infrastructure extraction, summary integration, and documentation.

Adds a new `Incrementality` module (`pymc_marketing/mmm/incrementality.py`) that computes **incremental channel contributions** using counterfactual analysis, properly accounting for adstock carryover effects.

The existing ROAS computation (`MMMIDataWrapper.get_roas()`) uses naïve element-wise division (contribution / spend), which ignores the temporal attribution from adstock transformations — spend at time *t* affects outcomes at *t, t+1, ..., t+l_max*. This PR implements the counterfactual approach from the Google MMM paper (Section 4.1), which compares the model's prediction under actual spend vs. a counterfactual where channel spend is removed or perturbed, summing over the full carryover window.

### Technical approach

The `Incrementality` class uses `extract_response_distribution()` + `vectorize_graph()` to compile a single batched evaluator that processes all counterfactual scenarios at once. For each period defined by the requested frequency:

1. Build a zero-padded counterfactual window covering `[t0 - l_max, t1 + l_max]` to capture carry-in history and carry-out effects
2. Apply the `counterfactual_spend_factor` only to spend within `[t0, t1]`, keeping surrounding spend intact for correct adstock context
3. Evaluate baseline on the full dataset once, then batch-evaluate all counterfactual windows
4. Sum the contribution difference over each period's evaluation window (including carryover when enabled)

The module exposes four public methods:
- `compute_incremental_contribution()` — core function, supports total (factor=0) and marginal (factor=1.01) incrementality
- `contribution_over_spend()` — ROAS (or customers/dollar, etc.)
- `spend_over_contribution()` — CAC (reciprocal of above)
- `marginal_contribution_over_spend()` — marginal ROAS at current operating point

> [!NOTE]
> **Key review focus**: The most important thing to review is `test_compute_incremental_contribution_matches_ground_truth` and its helper functions (`evaluate_channel_contribution` and `compute_ground_truth_incremental_by_period`). These validate the vectorized implementation against a completely independent oracle that uses `sample_posterior_predictive` — if the ground truth logic is correct, the rest follows. see more details below. 

## Related Issue

- Getting us closer to close #2211

## Differences from Research Architecture

[...]

PR #2325, opened 2026-02-24
BUG FIX: int dtype in budget optimization gradient bug

isofer:
## Description

When `channel_data` in the PyMC model has an integer dtype, `pt.cast` produces zero gradients during budget optimization. This causes the SLSQP optimizer to see a flat objective and stay at the initial equal-split allocation, regardless of the actual channel response curves.

This PR fixes the issue by casting `channel_data` to `float64` on the cloned optimization model before graph replacement, so that `pt.grad` can differentiate through the `do()` substitution.

**Root cause:** `_set_xarray_data` preserves the original dtype of `channel_data` (including integer). When the optimizer later replaces `channel_data` via `do()`, the graph contains a `pt.cast(float_budget, int)` node whose gradient is zero everywhere, making the objective appear constant.

**Approach:** A new static method `_ensure_float_channel_data` is called in `BudgetOptimizer.__init__` immediately after obtaining the cloned model. It checks whether `channel_data` has an integer dtype and, if so, mutates the shared variable's type and value to `float64` in-place. The mutation is scoped to the cloned optimization model — the user's original model is not affected.

To be merged: after #2290 is merged. 

## Changes Made

- **`pymc_marketing/mmm/budget_optimizer.py`**:
  - Added `_ensure_float_channel_data()` static method that converts integer `channel_data` to `float64` on the optimization model clone.
  - Called the method in `BudgetOptimizer.__init__` right after `_set_predictors_for_optimization`, before graph replacement.

- **`tests/mmm/test_budget_optimizer.py`**:
  - Added `_run_budget_optimization` helper to build and run the optimizer from a fitted MMM.
  - Added `test_int_and_float_channel_data_produce_same_allocation` that verifies integer and float `channel_data` models produce the same budget allocation
## How to Test

1. Run the new test directly:
   ```bash
   pytest tests/mmm/test_budget_optimizer.py::test_int_and_float_channel_data_produce_same_allocation -v
   ```
2. Verify the full budget optimizer test suite still passes:
   ```bash
   pytest tests/mmm/test_budget_optimizer.py -v
   ```
3. To reproduce the original bug, remove the `_ensure_float_channel_data` call in `__init__` and run the test — the integer model should stay at equal split while the float model does not.

----
📚 Documentation preview 📚: https://pymc-marketing--2325.org.readthedocs.build/en/2325/

### DIFF
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -661,6 +661,10 @@ class BudgetOptimizer(BaseModel):
             self.num_periods
         )  # TODO: Once multidimensional class becomes the main class.
 
+        # Ensure channel_data is float so pt.grad can differentiate through
+        # the do() replacement. Integer dtypes have zero gradient for pt.cast.
+        self._ensure_float_channel_data(pymc_model)
+
         # 2. Shared variable for total_budget: Use annotation to avoid type checking
         self._total_budget: SharedVariable = shared(
             np.array(0.0, dtype="float64"), name="total_budget"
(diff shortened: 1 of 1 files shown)


juanitorduz:
Thank you @isofer

PR #2337, opened 2026-02-25
Raise explicitly when trying to optimize or measure incrementality of discrete channel data

### LINKED ISSUE #2336: Changing the type of a variable is not a valid graph manipulation
https://github.com/pymc-labs/pymc-marketing/blob/23b9ee3b258116e48b5e393b0e37028529ce535a/pymc_marketing/mmm/budget_optimizer.py#L899-L913

ricardoV94:
Closes #2336 

If we want this to be supported we either:
1. Create / cast original channel_data always to float type
2. Use discrete optimizer without gradients
3. Do an unsafe `clone_replace(rebuild_strict=False)` and hope it does the right thing. 

----
📚 Documentation preview 📚: https://pymc-marketing--2337.org.readthedocs.build/en/2337/

### DIFF
--- a/pymc_marketing/mmm/incrementality.py
+++ b/pymc_marketing/mmm/incrementality.py
@@ -352,10 +352,14 @@ class Incrementality:
         # Use float64 for evaluation to avoid integer truncation when
         # counterfactual_spend_factor produces fractional values (e.g. 1.01).
         data_shared = self.model.model["channel_data"]
-        eval_dtype = "float64"
+        data_dtype = data_shared.dtype
+        if np.dtype(data_dtype).kind != "f":
+            raise ValueError(
+                f"Incrementality requires channel data of float type, got {data_dtype}"
+            )
         batched_input = pt.tensor(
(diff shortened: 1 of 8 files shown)


pymc_marketing/mmm/incrementality.py
```diff
@@ -352,10 +352,14 @@ def compute_incremental_contribution(
...
-        eval_dtype = "float64"
+        data_dtype = data_shared.dtype
+        if np.dtype(data_dtype).kind != "f":
+            raise ValueError(
+                f"Incrementality requires channel data of float type, got {data_dtype}"
+            )
         batched_input = pt.tensor(
             name="channel_data_batched",
-            dtype=eval_dtype,
+            dtype=data_shared.dtype,
```
ricardoV94, line 362: This was also unsafe, if the data was float32 originally it should be evaluated in float32 as well

juanitorduz (approved):
Thanks!

PR #2341, opened 2026-02-26
Cast channel_data to float at construction time

### LINKED ISSUE #2340: Cast channel_data to float at construction time
## Problem

The budget optimizer fails when `channel_data` has `dtype=int`. A hacky fix was merged that changed the node type to float ([related discussion](https://github.com/pymc-labs/pymc-marketing/issues/2336)), but this is not a valid graph manipulation. Ricardo opened [PR #2337](https://github.com/pymc-labs/pymc-marketing/pull/2337) to revert that fix and raise an error instead.

## Agreed Solution

After team discussion, we agreed to **always cast `channel_data` to float at construction time** (option 1). Reasons:

- The default priors and other components eventually become floats anyway, so casting early is consistent.
- Having `channel_data` as int can silently cast floats to ints elsewhere, causing subtle bugs (observed during incrementality work).
- Discrete optimization (option 2) would be surprising — the optimization algorithm should not change based on mostly hidden dtype info.
- Unsafe `clone_replace` with `rebuild_strict=False` (option 3) is too risky and could affect other parts of the graph.

## Implementation

- Cast `channel_data` to float at model construction time.
- Consider adding a warning/info message when the cast happens, so users are aware their integer data was converted.

## References

- Issue: #2336
- PR: #2337

isofer:
## Summary

- Cast `_channel` data to float in `MMM._generate_and_preprocess_model_data()` immediately after building the xarray dataset, so all downstream consumers (scalers, `pm.Data`, budget optimizer, incrementality, sensitivity analysis) always see float data.
- Adds a test verifying that integer channel input is cast to float at construction time.

Closes #2340

----
📚 Documentation preview 📚: https://pymc-marketing--2341.org.readthedocs.build/en/2341/


cetagostini (approved):

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1417,6 +1417,11 @@ def _generate_and_preprocess_model_data(
 
         self.xarray_dataset = xr.merge(dataarrays).fillna(0)
 
+        if not np.issubdtype(self.xarray_dataset["_channel"].dtype, np.floating):
```
ricardoV94: can't you just do `astype(float)` regardless?

isofer: done

Issue #2355, opened 2026-03-02
ExperimentDesigner: Posterior-aware experiment design for lift tests


drbenvincent:
# ExperimentDesigner: Posterior-aware experiment design for lift tests

## Summary

Add an `ExperimentDesigner` to PyMC-Marketing that recommends which marketing experiment to run — which channel, at what spend level, for how long — based on a fitted MMM's posterior uncertainty about channel response functions. The v1 scope is **national-level experiments** on multiple spend channels, analysed via Interrupted Time Series (ITS).

**Project home:** PyMC-Marketing (not CausalPy). The `ExperimentDesigner` consumes a fitted multidimensional `MMM` object (from `pymc_marketing.mmm.multidimensional`) directly — it reads posterior samples from the model's `InferenceData`, accesses the saturation and adstock function objects, and derives current spend and residual noise from the training data. No cross-repo protocol or intermediate representation is needed. Post-experiment analysis (ITS, Synthetic Control) remains in CausalPy.

The core innovation: GeoLift asks the user to guess the expected effect size (`effect_size = seq(0, 0.25, 0.05)`). The `ExperimentDesigner` replaces that guess with the posterior distribution of the predicted lift from a fitted model. Instead of "what's my power if the lift is 10%?", it answers "what's my power given what the model believes the lift will be?" — for every channel, at every candidate spend level.

This same principle extends to power: instead of computing power at one guessed effect size, the tool computes **posterior-predictive power** (Bayesian assurance) — the expected power averaged over the posterior distribution of the true effect.

**Relationship to CausalPy [#759](https://github.com/pymc-labs/CausalPy/issues/759):** Issue #759 tracks GeoLift feature parity (power analysis, market selection, augmented SC) in CausalPy. This feature is complementary — it adds the upstream decision layer that GeoLift doesn't have (which channel? what spend level?) while partially overlapping with #759's G3 (power analysis). The `ExperimentDesigner` can later consume #759's geo-level power analysis and market selection as they're built.

---

## The problem

A practitioner has a fitted MMM with multiple channels whose parameters are uncertain — often severely so, due to correlated spend patterns. They know a lift test would help, but face design questions:

- **Which channel** should they test?
- **At what spend level** — where on the saturation curve should the test probe?
- **What magnitude** of spend change is needed for a detectable effect?
- **For how long** — what experiment duration balances power against cost?
- **What is the cost** of each candidate experiment?

[...]

PR #2356, opened 2026-03-02
Add ExperimentDesigner: posterior-aware experiment design for lift tests

drbenvincent:
## Summary

- Adds `ExperimentDesigner` class that recommends which marketing experiment to run (channel, spend level, duration) based on a fitted MMM's posterior uncertainty about channel response functions
- Implements adstock-aware lift prediction, Bayesian assurance (posterior-predictive power), weighted composite scoring across 5 dimensions, and 5 plotting methods
- Includes fixture generator for creating realistic test posteriors, with both fast synthetic mode and full MCMC fitting

## Details

**Core computation:** For each candidate experiment design (channel × spend change × duration), evaluates the posterior-predicted lift accounting for geometric adstock ramp-up, computes measurement noise σ = σ_residual · √T, and derives Bayesian assurance (expected power over the posterior distribution of the true effect).

**Scoring dimensions:** Channels are ranked by a configurable weighted sum of: posterior uncertainty, spend correlation, saturation gradient, assurance, and cost efficiency — all min-max normalised.

**v1 scope:** National-level experiments with `LogisticSaturation` + `GeometricAdstock` (`adstock_first=True`). Geo-level designs, pulse/switchback experiments, and Fisher Information are deferred to v2.

### New files
- `pymc_marketing/mmm/experiment_design/designer.py` — `ExperimentDesigner` class
- `pymc_marketing/mmm/experiment_design/recommendation.py` — `ExperimentRecommendation` dataclass
- `pymc_marketing/mmm/experiment_design/functions.py` — Numpy `logistic_saturation`
- `pymc_marketing/mmm/experiment_design/fixture.py` — `generate_experiment_fixture()`
- 3 test files with 65 tests covering all components

## Test plan

- [x] 65 new tests pass (functions, designer, fixture, plotting)
- [x] Pre-commit hooks pass (ruff, mypy, formatting)
- [x] Adstock ramp verified against analytic geometric series
- [x] Assurance calibrated (α for zero effects, ~1.0 for large effects)
- [x] NetCDF round-trip for fixtures verified
- [ ] Integration test with fitted MMM (requires MCMC, deferred to CI)

Towards #2355

----
📚 Documentation preview 📚: https://pymc-marketing--2356.org.readthedocs.build/en/2356/


pymc_marketing/mmm/experiment_design/designer.py
```diff
@@ -0,0 +1,1191 @@
...
+    --------
+    .. code-block:: python
+
+        from pymc_marketing.mmm import MMM
+        from pymc_marketing.mmm.experiment_design import ExperimentDesigner
+
+        mmm = MMM(...)
+        mmm.fit(X, y)
+
+        designer = ExperimentDesigner(mmm)
```
williambdean, line 155: Is it possible to just take actuals and posterior predictives? Or just the residuals ?

pymc_marketing/mmm/experiment_design/designer.py
```diff
@@ -0,0 +1,1191 @@
...
+                "lam": posterior[sat_var_map["lam"]]
+                .sel(**sel)
+                .values.astype(np.float64),
+                "beta": posterior[sat_var_map["beta"]]
+                .sel(**sel)
+                .values.astype(np.float64),
+                "alpha": posterior[ads_var_map["alpha"]]
+                .sel(**sel)
+                .values.astype(np.float64),
+            }
```
williambdean: What is actually needed from the media transformation? Maybe there is a way to leverage
the graph directly.

pymc_marketing/mmm/experiment_design/designer.py
```diff
@@ -0,0 +1,1191 @@
...
+        """Evaluate the logistic saturation response in pure numpy.
+
+        Uses the same formula as
+        :func:`pymc_marketing.mmm.transformers.logistic_saturation`
+        but avoids a PyTensor round-trip so callers can pass raw arrays.
+        """
+        x = np.asarray(x)
+        lam = np.asarray(lam)
+        sat = (1.0 - np.exp(-lam * x)) / (1.0 + np.exp(-lam * x))
+        return np.asarray(beta) * sat
```
williambdean: We shouldn't have to do this.

williambdean (commented):
Initial thoughts

pymc_marketing/mmm/experiment_design/designer.py
drbenvincent: Addressed in b36929cb. The entire parameter-extraction approach is gone. We now call `extract_response_distribution` to trace the model's `channel_contribution` graph, substitute posterior samples, and compile it into a reusable PyTensor function via `_build_eval_fn_from_model`. All evaluation (lift prediction, ramp, diagnostics, plotting) goes through this compiled graph — no manual access to `lam`, `beta`, or `alpha` needed.

pymc_marketing/mmm/experiment_design/designer.py
drbenvincent: Addressed in b36929cb. This method (`_eval_saturation`) and the companion `_compute_steady_state_spend` / `_compute_adstock_ramp` are all deleted. Everything now evaluates through the compiled PyTensor graph — no reimplementation of transformation formulas.

pymc_marketing/mmm/experiment_design/designer.py
drbenvincent: Addressed in b36929cb. We kept the `y - mmm.predict(mmm.X)` approach and added a docstring explaining why: `predict()` is the public API for point predictions, it avoids duplicating training data the model already holds, and we only need a point estimate of residual noise — not the full posterior predictive distribution. Falls back to `residual_std=1.0` with a warning if `predict()` fails.

juanitorduz:
@bugbot review

pymc_marketing/mmm/experiment_design/fixture.py
```diff
@@ -0,0 +1,380 @@
...
+    out = np.zeros(n)
+    for t in range(n):
+        for lag in range(min(l_max, t + 1)):
+            out[t] += weights[lag] * x[t - lag]
+    return out
+
+
+def _logistic_saturation_np(x: np.ndarray, lam: float) -> np.ndarray:
+    """Numpy logistic saturation (without beta scaling)."""
+    return (1.0 - np.exp(-lam * x)) / (1.0 + np.exp(-lam * x))
```
juanitorduz, line 95: This seems redundant, cant't we use https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/transformers.py ?

pymc_marketing/mmm/experiment_design/fixture.py
```diff
@@ -0,0 +1,380 @@
...
+
+
+def generate_experiment_fixture(
```
juanitorduz, line 98: Do we need this for the code or just for tests?

juanitorduz: Could we re=use some of https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_data_generator.html ?

pymc_marketing/mmm/experiment_design/recommendation.py
```diff
@@ -0,0 +1,331 @@
...
+        parts = [f"ExperimentRecommendations({len(self)} recommendations"]
+        if self.null_confirmation_candidates:
+            parts.append(
+                f", {len(self.null_confirmation_candidates)} null-confirmation "
+                "candidates"
+            )
+        parts.append(")")
+        return "".join(parts)
+
+    # -- Jupyter rendering ---------------------------------------------------
```
juanitorduz, line 149: remove this comment

pymc_marketing/mmm/experiment_design/recommendation.py
```diff
@@ -0,0 +1,331 @@
...
+                "Ramp": rec.adstock_ramp_fraction,
+                "Net Cost": rec.net_cost,
+                "Score": rec.score,
+            }
+            for i, rec in enumerate(self._recommendations)
+        ]
+        return pd.DataFrame(records)
+
+
+_HIGH_CORRELATION_THRESHOLD = 0.7
```
juanitorduz, line 258: at the top of the file?

juanitorduz (changes requested):
Some initial comments :)

daimon-pymclabs:
# Review: pymc-marketing #2356, `ExperimentDesigner`

**Head reviewed:** `ba0640a` (branch `fix/2355-experiment-designer`, after the `main` merge at 08:53 UTC 2026-09-07)
**Verification:** everything below was executed against that head on a session with the branch's own declared dependency set (pymc 6.0.1, pytensor 3.0.7, arviz 1.2.0, pymc-marketing 1.1.0). Numbers quoted are measured, not inferred.

## Overview

Adds `pymc_marketing/mmm/experiment_design/` with an `ExperimentDesigner` that consumes a fitted MMM and ranks candidate lift tests over a grid of channel x spend-change x duration. For each candidate it predicts cumulative lift by pushing baseline and treatment spend series through the model's own compiled graph (`extract_response_distribution` on `channel_contribution`), computes a cumulative measurement noise `sigma_d`, derives Bayesian assurance (expected two-sided power over the posterior), and scores candidates on a weighted blend of assurance and cost efficiency. Ships a recommendation dataclass with a `Sequence`-protocol container and `_repr_html_`, a fixture generator, five plotting methods, and 4 test modules.

Using the model's own graph rather than reimplementing adstock/saturation is the right architectural choice, and it is what makes the adstock ramp handling credible. The identification caveat in the class docstring is a good instinct. The problems below are in the plumbing around that core, not in the idea.

## Blockers

### 1. Unit mismatch makes every recommendation wrong, and it fails silently

`predicted_lift` comes from `channel_contribution`, which in `multidimensional.py` lives in **scaled** target space (line 1918 divides the target by `target_scale`; line 2049 sets `observed=target_data_scaled`). The repo's own convention for anything compared against real-world quantities is the `*_original_scale` variable, cf. the `extract_response_distribution` default at `multidimensional.py:3387`.

`_compute_residual_std` meanwhile does `mmm.y - mmm.predict(mmm.X)`. `mmm.y` is the raw user series (`multidimensional.py:1591`), while `predict` returns posterior-predictive draws of the **scaled** observed variable with no rescaling. So the two operands are on different scales and the subtraction is not a residual at all: it collapses to the total variation of `y`.

Measured on a 2-channel fit (`target_scale` = 53275):

```
residual_std : 590.3879      <- computed by the designer
std(y)       : 590.3915      <- identical to 4 s.f.; the model fit is effectively ignored
true noise sd: 500

predicted_lift mean (tv, +20%, 8w): 0.0341      (scaled-contribution units)
sigma_d(T=8)                      : 1724.10     (original y units)
=> SNR                            : 1.98e-05    (should be ~1.3)

recommend() -> 0 recommendations
               null_confirmation_candidates = ['tv', 'search']
```

So on a real fitted MMM `recommend()` returns an empty list and asserts that *every* channel has a near-zero effect. No exception, no warning. Dividing `residual_std` by `target_scale` as a one-line manual correction changes the same run to **13 recommendations, nulls `['search']`**. That single factor is the whole difference between "nothing is testable" and a usable answer.

Two distinct fixes are needed: put lift and noise in the same space (either use `channel_contribution_original_scale`, or scale `residual_std` by `target_scale`), and compute the residual from two operands that are actually on the same scale.

`_identify_null_candidates` has the same defect: it compares scaled contributions (~0.05) against `residual_std` (590), so it flags every channel unconditionally.

### 2. `ExperimentDesigner(mmm)` cannot be constructed on the branch's own dependency set

`designer.py:84` does `from pytensor.compile.function import function`. That module does not exist in pytensor 3.x, which the branch requires (`pymc>=6.2.0,<6.3.0`):

```
ModuleNotFoundError: No module named 'pytensor.compile.function'
  at designer.py:84, reached from __init__ -> _build_eval_fn_from_model
```

`from pytensor import function` works.

### 3. `az.hdi(..., hdi_prob=...)` raises on arviz 1.x

The branch pins `arviz>=1.2.0,<2.0`. On arviz 1.2.0:

```
TypeError: hdi got an unexpected keyword argument: 'hdi_prob'
```

Three call sites: `designer.py:647` (`_compute_response_hdi_width`), `:853` (`_evaluate_candidates`, i.e. the main `recommend()` path), `:1086` (`plot_lift_distributions`).

### 4. `az.InferenceData(posterior=...)` raises on arviz 1.x, taking the test suite with it

`fixture.py:330`:

```
TypeError: DataTree.__init__() got an unexpected keyword argument 'posterior'
```

arviz 1.x routes `InferenceData` to `xarray.DataTree` (it emits a `MigrationWarning` on import). Since almost every test builds its designer from a fixture, the suite result on this head is:

```
14 failed, 9 passed, 2 skipped, 92 errors in 6.99s
```

The PR body's "65 new tests pass" was true pre-merge and is not true now. Items 2 to 4 are all straightforward mechanical updates, but they mean nothing in this PR currently executes.

## Correctness and API

### 5. The constructor mutates the caller's InferenceData

`_compute_residual_std` calls `mmm.predict(mmm.X)`, which defaults to `extend_idata=True`. Verified: `posterior_predictive` is absent before `ExperimentDesigner(mmm)` and present after. Constructing a read-only analysis helper should not rewrite the user's idata, and it should not run a full posterior-predictive sample just to obtain a point estimate. Pass `extend_idata=False`, and consider deriving `residual_std` from the existing `y_sigma` posterior instead.

### 6. Two bare `except Exception` blocks hide exactly the failures that matter

`_compute_residual_std` catches everything and falls back to `residual_std=1.0`, an arbitrary number in an unknown scale that silently poisons every SNR and assurance downstream. The spend-correlation block does the same. Combined with item 13, a completely broken designer looks like a working one that simply found nothing. Narrow these to the specific expected exception, or let them propagate.

Related, in the same method: `y_pred_flat[: len(y_actual)]` silently truncates on a shape mismatch. For a panel model that mismatch means the residuals are misaligned rather than merely trimmed. Raise instead.

### 7. `plot_lift_distributions` crashes on a 1x1 grid

```python
d.plot_lift_distributions("tv", spend_changes=[0.2], durations=[8])
# TypeError: 'Axes' object is not subscriptable   (designer.py:1075)
```

`plt.subplots(1, 1)` returns a bare `Axes`, so `axes[np.newaxis, :]` fails. Pass `squeeze=False` and drop both reshape branches.

### 8. `score_weights` merges into the defaults instead of replacing them

`weights.update(score_weights)` keeps whatever the caller did not mention, and unknown keys are dropped silently. Measured:

```
default                : [0.500 0.750 0.500]
{"assurance": 1.0}     : [0.667 0.667 0.333]   <- cost_efficiency is still 0.5, not 0
{"uncertainty": 1.0}   : [0.500 0.750 0.500]   <- typo ignored, silently returns the default ranking
```

A caller asking to rank purely on assurance gets a 2:1 blend. A caller who typos a key gets the default weighting with no signal. Validate keys against `known_dims` and raise on unknown ones; decide explicitly whether partial dicts merge or replace, and document it.

### 9. Low correlations are reported as "high spend correlation"

`_get_correlation_info` hardcodes the word "high" and returns the max regardless of magnitude; `_HIGH_CORRELATION_THRESHOLD` only gates the extra caution paragraph. Measured with r = 0.02:

```
('high spend correlation with search (r = 0.02)', 0.02)
```

Every rationale string therefore carries a scary-sounding clause even for near-orthogonal channels. Only emit the clause above the threshold.

### 10. The AR(1) correction is the large-T limit, applied at small T

`_effective_sigma` inflates the cumulative variance by `(1 + rho) / (1 - rho)`. Against the exact `Var(sum) = sigma^2 [T + 2 * sum_{k=1}^{T-1} (T-k) rho^k]` at rho = 0.5:

```
T=  4  exact=   8.25  used=  12.00  ->  1.45x variance (1.20x sigma)
T=  8  exact=  20.02  used=  24.00  ->  1.20x
T= 12  exact=  32.00  used=  36.00  ->  1.12x
T= 52  exact= 152.00  used= 156.00  ->  1.03x
```

The default durations are 4 to 12 weeks, precisely where the approximation is worst. The bias is conservative (assurance understated), but the exact sum is a two-line closed form, so there is no reason to approximate.

### 11. Ratio estimators with a sign-agnostic epsilon

`_compute_graph_ramp_fraction` and the ramp block in `_channel_metrics` use `ss_per_week + 1e-12`. For spend *decreases* the denominator is negative, so the epsilon pushes it further from zero in one direction and toward a sign flip in the other. Use `np.where(np.abs(ss) > eps, ...)`. Separately, `mean(lift / ss)` is a ratio-of-draws estimator that is heavy-tailed when `ss` is near zero for some draws; the `clip(0, 1)` contains the blow-up but biases the result. A ratio of posterior means would be better behaved, or report the clipped fraction.

### 12. Zero-cost candidates score worst on cost efficiency

`np.where(abs_cost > 1e-10, assurances / abs_cost, 0.0)`. A candidate with meaningful assurance at no net spend change is the best possible efficiency and is assigned the floor. Reachable for a channel with zero current spend.

### 13. An empty result set is returned with no explanation

When the `min_snr` filter drops every candidate, `recommend()` returns an empty `ExperimentRecommendations` and only the notebook `_repr_html_` says anything. This is what masked item 1 end to end. Warn with the best achieved SNR and the threshold when nothing survives.

## Fixture module

### 14. Four documented sampler arguments are silently ignored

`n_chains`, `n_draws`, `n_tune` and `target_accept` are documented as "only used if `fit_model=True`". `_fit_mmm_for_fixture` calls `mmm.fit(X, y_series, random_seed=seed)` and uses none of them. They are in fact only honoured on the `fit_model=False` path. Either forward them or drop them.

### 15. The fixture's `residual_std` is not a residual standard deviation

The DGP builds `y` by applying `lam` to **unscaled** adstocked spend (line 181), while `y_pred_true` applies the same `lam` to **scaled** spend (line 203). The difference between the two is a scale artefact, not model error, so `residual_std` is arbitrary. Independently, `_create_synthetic_posterior` stores `beta` unscaled while `residual_std` is on the `y / max_y` scale, reproducing the same class of mismatch as item 1 inside the fixture used by the golden tests. Whatever those tests are pinning, it is not calibrated assurance.

### 16. Unvalidated type arguments, one unused parameter

`saturation` and `adstock` are accepted, never validated, and written into `attrs`, so `saturation="hill"` yields a logistic fixture labelled `hill`. Validate against the same maps `from_idata` uses. `_simulate_spend`'s `channel_names` parameter is unused.

### 17. The two constructors use different spend conventions

`model["channel_data"]` holds **raw** spend (scaling happens inside the graph at `multidimensional.py:1912`), so `__init__` works in raw units. `from_idata` builds a toy model with no scaler and is fed `spend_scaled`. Each path is internally consistent, but `spend_change_abs`, `net_cost` and `expected_lift` mean different things across them, and a fixture cannot be swapped for a real model. Worth stating explicitly in both docstrings.

## Repo hygiene

- `pymc_marketing/mmm/data/simulated_3channel.nc`, 318 KB, has **zero references** anywhere in the repo (searched `.py`, `.toml`, `.ipynb`, `.md`). hatchling packages everything under `pymc_marketing/`, so it ships in the wheel as dead weight. Delete it, or move it under `tests/` or `docs/` if something is meant to load it.
- `.github/pr-summaries/2355-experiment-designer.md` is an internal planning document committed to the repo.
- The PR body is stale: it describes `functions.py` with a numpy `logistic_saturation` (not in the diff), five scoring dimensions (the code has two, `assurance` and `cost_efficiency`), and three test files (there are four). The unchecked "integration test with fitted MMM" box is exactly the gap that let item 1 through.

## Performance

- No memoization anywhere. `_predict_lift` recomputes the identical baseline contribution for every `(channel, frac)` pair at a given `T`; `_channel_metrics()` re-runs the full per-channel graph sweep on every call, and `recommend()` plus `plot_channel_diagnostics` each call it again. Caching baselines by `T_total` is a cheap large win.
- `plot_saturation_curve` runs 200 separate full-graph evaluations, one per grid point, each over all draws and `l_max + 1` timesteps. Put the grid on the batch axis instead.
- The same method computes `y_all` for **all** draws and only then subsamples to `n_samples=500`, so the subsample costs the full compute and merely adds noise to the band. Subsample the draws first, or drop the argument. The `default_rng(42)` seed is also hardcoded rather than exposed.

## Smaller notes

- `mmm: Any` on the constructor discards type information; a `TYPE_CHECKING` import of `MMM` would keep it without the circular import.
- `from_idata` bypasses `__init__` via `cls.__new__(cls)`. It works because `_init_common` is the single funnel, but a small private dataclass of extracted inputs would be more robust to future attribute additions.
- `n_recent = min(8, ...)` in `__init__` and `spend_scaled[-8:]` in the fixture both hardcode an 8-period lookback for "current spend". Make it a parameter and document it, since it directly sets the operating point of every recommendation.
- `significance_level` is never validated to be in (0, 1).
- `_format_rationale` opens every rationale with "is a high-priority test target", including the lowest-ranked candidate.

## Suggested order of work

1. Items 2, 3, 4: mechanical dependency updates. Nothing runs until these land.
2. Item 1: the scale fix, plus item 5 while touching the same method.
3. Add the deferred integration test: fit a small MMM, assert `recommend()` returns a non-empty set and that SNR is order-1 for a channel with a known effect. That single test catches item 1 and would have caught it originally.
4. Items 7, 8, 9, 13: small, user-visible, each a few lines.
5. Items 10, 11, 14, 15: statistical tightening, including whatever the golden tests are actually pinning.
6. Hygiene and performance.

PR #2363, opened 2026-03-03
Demo notebook for cost per unit

isofer:
Demo notebook for cost per unit

----
📚 Documentation preview 📚: https://pymc-marketing--2363.org.readthedocs.build/en/2363/

### DIFF
--- a/.github/workflows/duplicate-issues.yml
+++ b/.github/workflows/duplicate-issues.yml
@@ -58,6 +58,7 @@ jobs:
       - name: Run duplicate finder agent
         env:
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          GH_TOKEN: ${{ github.token }}
         run: |
           set -o pipefail
           claude -p \
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -51,6 +51,7 @@ jobs:
       - name: Run triage
         env:
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          GH_TOKEN: ${{ github.token }}
         id: triage
         run: |
           set -o pipefail
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
           - --exclude=scripts/
           - --start-year=2022
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.4
+    rev: v0.15.5
     hooks:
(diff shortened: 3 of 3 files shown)


docs/source/notebooks/mmm/mmm_cost_per_unit.ipynb
```diff
@@ -0,0 +1,1268 @@
+{
```
juanitorduz, line 1: <p>missing space 0.10or0.30&nbsp;?</p><p> </p><p>Regarding the comment "I<strong style="letter-spacing: 0.01em;">mpressions are what actually drives revenue". </strong>There are so many strong opinions on this topic that I would like to leave it open for the modeler to decide based on the context. For example, spend data is an apples-to-apples comparison (EUR), but impressions in Google (as an absolute number) probably do not mean the same as impressions in Twitch. I have also seen really bad impressions data XD. </p><p> </p><p>Hence, I will add a comment that the modeler has to be responsable to think about which variable to use :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2363/discussion/#comment-2901663899'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #10.                "chains": 2,
</pre><p>can we add 4 chains instead of 2?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2363/discussion/#comment-2901663901'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

juanitorduz:
Thanks @isofer ! This looks very nice! I just left one comment :)

docs/source/notebooks/mmm/mmm_cost_per_unit.ipynb
isofer: <p>I think the intro was not clear enough. I didn't mean say that you can aggregate multiple social channels together (like Google and twitch). The dataset describe the case that you have only media channels - one social channel measured in impressions and one is TV. I'll clarify this. </p><p>Also if one channel is in Euro and one in USD, then you still should probably model each one in its original currency and use the cost-per-unit to convert to reduce the noise of currency conversion. It's a good example, and I should add a paragraph about it.</p>

PR #2364, opened 2026-03-04
Extends the cost-per-unit support to plotting 

### LINKED ISSUE #2309: Support `cost_per_unit` for channels with non-monetary data (impressions, clicks, etc.) - part 1

isofer:
## Description

Extends the cost-per-unit support introduced in Part 1 to the plotting and summary layers. When a model has `cost_per_unit` set, saturation curve plots and sensitivity analysis plots now display spend on the x-axis instead of raw channel data, and the summary factory adapts its column naming accordingly.

This PR is stacked on top of `isofer/2309-cost-per-unit-part1`.

### Key changes

- **`MMMPlotSuite` accepts `MMMIDataWrapper`**: The constructor now takes an optional `data: MMMIDataWrapper` parameter (mutually exclusive with `idata`), giving plots access to spend-aware helper methods.
- **Spend-scaled x-axes in saturation plots**: `saturation_scatterplot`, `saturation_curves`, and `sensitivity_analysis` accept an `apply_cost_per_unit` parameter. When `True` (default) and cost-per-unit data is available, x-axes show spend values with the label "Spend" instead of raw channel data.
- **Weighted-average cost-per-unit**: `MMMIDataWrapper.get_avg_cost_per_unit()` computes `total_spend / total_data` per channel, used by `saturation_curves` and `plot_interactive` to scale the fitted curve x-axis.
- **Summary column naming**: `MMMSummaryFactory.channel_spend()` returns a column named `channel_spend` when cost-per-unit data is present, or `channel_data` otherwise.

## Related Issue

- Closes #2309

### Core (`pymc_marketing/`)

- **`data/idata/mmm_wrapper.py`**:
  - Added `get_avg_cost_per_unit()` — returns weighted-average cost per unit per channel (`total_spend / total_data`), falling back to `ones_like(channel_scale)` when no spend data exists.

- **`mmm/multidimensional.py`**:
  - `plot` property now passes `MMMIDataWrapper` (`self.data`) to `MMMPlotSuite` instead of raw `self.idata`.

- **`mmm/plot.py`**:
  - `MMMPlotSuite.__init__` accepts `data: MMMIDataWrapper` or `idata` (mutually exclusive), wrapping raw idata automatically.
  - `saturation_scatterplot()`: Added `apply_cost_per_unit` parameter; uses `get_channel_spend()` for x-data and labels x-axis "Spend" when cost-per-unit is available.
  - `saturation_curves()`: Added `apply_cost_per_unit` parameter; scales fitted curve x-coordinates by `channel_scale * avg_cost_per_unit` in original scale mode.
  - `sensitivity_analysis()`: Added `apply_cost_per_unit` parameter; uses spend totals for absolute x-sweep axis scaling.

- **`mmm/plot_interactive.py`**:
  - Saturation curves in original scale now multiply channel scale by `get_avg_cost_per_unit()` for x-axis conversion.

- **`mmm/summary.py`**:
  - `channel_spend()` output column name adapts: `channel_spend` when `channel_spend` exists in `constant_data`, otherwise `channel_data`.


pymc_marketing/mmm/plot_interactive.py
```diff
@@ -1431,9 +1431,10 @@ def saturation_curves(
 
         # Convert x-axis to original scale if requested
         if original_scale:
-            # Get channel scale factors and convert to DataFrame
             channel_scale = self.summary.data.get_channel_scale()
-            scale_df = channel_scale.to_dataframe(name="channel_scale").reset_index()
+            avg_cpu = self.summary.data.get_avg_cost_per_unit()
+            spend_scale = channel_scale * avg_cpu
```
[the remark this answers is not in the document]
isofer: I think it's fine to leave it that way

pymc_marketing/mmm/plot.py
```diff
@@ -214,9 +215,22 @@ class MMMPlotSuite:
...
+
+        if data is not None:
+            self.idata = data.idata
+            self.data = data
+        elif idata is not None:
+            self.idata = idata
+            self.data = MMMIDataWrapper(idata)
+        else:
+            self.idata = cast(az.InferenceData, idata)
+            self.data = cast(MMMIDataWrapper, data)
```
cetagostini, line 233: This sounds like the constructor could accept no args. Meaning, everything downstream will crash? Is it intentional?

pymc_marketing/mmm/plot_interactive.py
cetagostini, line 1436: Are we doing something about the inconsistency? The matplotlib side has `apply_cost_per_unit` to opt out, but the interactive one always applies it. Bugbot mentioned it as well.

pymc_marketing/mmm/plot.py
```diff
@@ -2074,7 +2089,12 @@ def saturation_scatterplot(
                 fallback_title="Channel Saturation Curve",
             )
             ax.set_title(title)
-            ax.set_xlabel("Channel Data (X)")
+            xlabel = (
+                "Spend"
+                if (apply_cost_per_unit and self.data.cost_per_unit is not None)
+                else "Channel Data (X)"
+            )
```
cetagostini: This exact same block shows up in 3 places (here, `saturation_curves`, and `plot_interactive`). Could we pull it into a tiny helper or a constant? Feels like it'll drift over time.

pymc_marketing/mmm/plot.py
```diff
@@ -2046,9 +2061,10 @@ def saturation_scatterplot(
                         indexers[k] = v
             indexers["channel"] = channel
 
-            # Select X data (constant_data)
-            x_data = self.idata.constant_data.channel_data.sel(**indexers)
-            # Select Y data (posterior contributions) and scale if needed
+            if apply_cost_per_unit:
+                x_data = self.data.get_channel_spend().sel(**indexers)
+            else:
+                x_data = self.data.get_channel_data().sel(**indexers)
```
cetagostini: Same if/else pattern appears again in `saturation_curves` below. Maybe a small `_get_x_data(apply_cost_per_unit, **indexers)` helper so we don't repeat ourselves?

tests/mmm/test_cost_per_unit.py
```diff
@@ -53,6 +53,13 @@ def simple_idata(dates, channels):
     """InferenceData without custom dims."""
     n_dates = len(dates)
     n_channels = len(channels)
+    contrib = rng.normal(0.5, 0.1, size=(2, 50, n_dates, n_channels))
```
cetagostini, line 56: Nit — `rng` here is the module-level one, so every fixture call advances its state. If pytest reorders tests, each fixture gets different random values. Could we just do `rng = np.random.default_rng(SEED)` inside each fixture instead?

pymc_marketing/data/idata/schema.py
```diff
@@ -376,13 +376,13 @@ def from_model_config(
             )
 
         groups["posterior"] = InferenceDataGroupSchema(
-            name="posterior", required=True, variables=posterior_vars
+            name="posterior", required=False, variables=posterior_vars
```
cetagostini, line 379: Making `posterior` optional for all consumers feels a bit broad — won't this weaken validation for things that actually need it (like summary, incrementality)? Would a separate schema or skipping validation on the plot side be safer?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1146,7 +1146,10 @@ def plot(self) -> MMMPlotSuite:
         """Use the MMMPlotSuite to plot the results."""
         self._validate_model_was_built()
         self._validate_idata_exists()
-        return MMMPlotSuite(idata=self.idata)
+        data = self.data
+        if data.schema is not None:
+            data.validate_or_raise()
```
cetagostini: This validation already happens inside `MMMIDataWrapper.__init__` and `MMMSummaryFactory.__post_init__`. Running it again on every `.plot` access (which is a property, so no caching) might be redundant work. Is there a reason for a third validation path?

pymc_marketing/mmm/plot.py
```diff
@@ -1936,6 +1950,7 @@ def saturation_scatterplot(
         self,
         original_scale: bool = False,
         dims: dict[str, str | int | list] | None = None,
+        apply_cost_per_unit: bool = True,
```
cetagostini, line 1969: The new `apply_cost_per_unit` param isn't documented in the docstring — there's no Parameters section here at all. Same thing for `saturation_curves` below. Could we add a quick note so users know what it does?

pymc_marketing/data/idata/mmm_wrapper.py
```diff
@@ -373,6 +373,26 @@ def cost_per_unit(self) -> xr.DataArray | None:
...
+        -------
+        xr.DataArray
+            Average cost per unit with dims (channel,) [+ custom dims].
+        """
+        if self.cost_per_unit is None:
+            return xr.ones_like(self.get_channel_scale())
+
+        total_spend = self.get_channel_spend().sum(dim="date")
+        total_data = self.get_channel_data().sum(dim="date")
+        total_data_safe = xr.where(total_data == 0, np.nan, total_data)
```
cetagostini, line 393: When total_data is zero for a channel, this returns NaN, which will silently propagate through the x-axis scaling in saturation plots. Do we have a test for that edge case? Might be worth adding one to make sure the plot doesn't just go invisible.

tests/mmm/test_plot.py
```diff
@@ -5101,3 +5102,193 @@ def test_inference_data_without_posterior_predictive_error(self, mock_suite_basi
...
+        wrapper = MMMIDataWrapper(idata)
+        suite = MMMPlotSuite(data=wrapper)
+        result = suite.sensitivity_analysis(
+            hue_dim="channel",
+            x_sweep_axis="absolute",
+            apply_cost_per_unit=apply_cpu,
+        )
+        fig = result[0] if isinstance(result, tuple) else result.figure
+        plt.close(fig)
+        assert fig is not None
```
cetagostini, line 5293: This only checks `fig is not None` — it'd pass even if `apply_cost_per_unit` were completely ignored. Could we compare the x-axis ranges between `apply_cpu=True` and `apply_cpu=False` to make sure the param actually does something?

pymc_marketing/mmm/summary.py
```diff
@@ -656,7 +656,13 @@ def channel_spend(
         )
 
         spend = self.data.get_channel_spend()
-        df = spend.to_dataframe(name="channel_data").reset_index()
+        col_name = (
+            "channel_spend"
+            if hasattr(self.data.idata, "constant_data")
+            and "channel_spend" in self.data.idata.constant_data
```
cetagostini, line 662: This reaches through the wrapper into `idata.constant_data` directly. The wrapper already has `cost_per_unit` property for this — something like `"channel_spend" if self.data.cost_per_unit is not None else "channel_data"` would keep the abstraction cleaner.

pymc_marketing/mmm/plot_interactive.py
```diff
@@ -1431,9 +1431,10 @@ def saturation_curves(
...
+            scale_df = spend_scale.to_dataframe(name="channel_scale").reset_index()
```
cetagostini, line 1437: Small thing: a few lines below, `xaxis_title` is always `"Spend"` when `original_scale=True`, even if there's no cost-per-unit data. The matplotlib side checks `cost_per_unit is not None` to pick the right label — shouldn't this match?

pymc_marketing/mmm/plot.py
isofer: I myself don't like but this is the current state. There are actually unittests that use MMMPlotSuite(idata=None) at the moment. I don't know why. Maybe because some plots accepts idata??

pymc_marketing/mmm/plot_interactive.py
isofer: I actually left that on purpose. The Matplotlib is aimed at DSs and have more flexibility for deubgging. I reduced the control in the plot_interactive.

pymc_marketing/mmm/multidimensional.py
isofer: The validation doesn't happen in `MMMIDataWrapper.__init__`. and MMMSummaryFactory is not called for `.plot`

pymc_marketing/data/idata/schema.py
isofer: It would, but one of the workflow (in our notebooks as well) is to run prior predictive and then to plot it, and that cause this to fail. So I had to remove it.
I think the solution is clean up the whole MMMPlotSuite as I suggested in a recent discord thread.

pymc_marketing/mmm/plot.py
isofer: fixed.

pymc_marketing/mmm/plot.py
isofer: fixed.

tests/mmm/test_cost_per_unit.py
isofer: fixed

pymc_marketing/data/idata/mmm_wrapper.py
isofer: fixed

pymc_marketing/mmm/plot.py
isofer: fixed

tests/mmm/test_plot.py
isofer: fixed

pymc_marketing/mmm/summary.py
isofer: fixed

pymc_marketing/mmm/plot_interactive.py
isofer: maybe the right thing is to remove `original_scale` support from these functions. There are intended to stakeholders anyway and original_scale should be always true for them?

cetagostini: Like that idea!

pymc_marketing/mmm/plot.py
cetagostini: Okay lets make an issue and tackle this later on!

cetagostini (approved):

PR #2393, opened 2026-03-11
Some fixes TimeSlice Cross Validation and Extend Case Study

juanitorduz:
Some edge cases and support adding a variable in the original scale, and also lift tests. The problem is that `mmm.build_model(X_train, y_train) was called for every fold, so we could not add the lift tests.

Moreover, we show, through time-slice cross-validation, how adding the lift test gets better and better results. :)

image

This extended example also show that out-of-sample accuracy does not imply a good causal model :)

image

- I also compute ROAS with the new `incrementality` module (and fixed an edge case for time-varying parameters).

----
📚 Documentation preview 📚: https://pymc-marketing--2393.org.readthedocs.build/en/2393/


timbo112711 (commented):
Overall this is looking awesome! I think it's a really good example notebook that uses two big methods within MMMs (calibration and stability checks). 

Also love the updates with `_prepare_fold_model()`, much better than we had before 😄

Left some comments in the notebook as well!

timbo112711 (approved):
Really like how this shaped out! Left a brief comment in the notebook, but approving.

juanitorduz:
<p>Good catch! I added an explanation.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2393/discussion/#comment-4071728463'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

PR #2395, opened 2026-03-13
Fix wrong constraints docstrings

### LINKED ISSUE #1707: MMM.optimize_budget Docstring mismatch for custom_constraints
The docstring for the optimize_budget method in the MMM class in mmm.py doesn't match the actual argument name or type for constraints. Specifically:
* Mismatched Argument Name: The docstring describes a parameter named custom_constraints, but the method signature actually uses constraints.
* Incorrect Type Hint and Description: The docstring states that custom_constraints should be list[dict[str, Any]] with a format of [{"key":...,"constraint_fun":...,"constraint_type":...}]. It looks like this gets passed directly to `BudgetOptimizer` via `custom_constraints` which  expects `Sequence[Constraint]`

ricardoV94:
Closes #1707 

----
📚 Documentation preview 📚: https://pymc-marketing--2395.org.readthedocs.build/en/2395/


juanitorduz (approved):
Thanks!

PR #2396, opened 2026-03-13
Fix optimization with time varying media

### LINKED ISSUE #2104: Dimension mismatch in MultidimensionalBudgetOptimizerWrapper with time_varying_media=True
## Problem

When using `mmm.multidimensional.MMM` with `time_varying_media=True` and `MultidimensionalBudgetOptimizerWrapper` for budget allocations, a dimension error occurs.

## Root Cause

The issue is traced to `_replace_channel_data_by_optimization_variable` in `pymc_marketing/mmm/budget_optimizer.py`. The function uses the do-operator to replace `channel_data` with the optimization variable, but it includes the adstock carryover in the date dimension of the replacement tensor (lines 926-946). Specifically:

1. The function pads `repeated_budgets` with zeros to account for carry-over effects, creating a tensor with shape `(num_periods + max_lag, ...)` in the date dimension
2. This extended `channel_data` causes problems for any model components that interact with `channel_data` and require matching length in the date dimension

## Specific Issue with time_varying_media

In `pymc_marketing/mmm/multidimensional.py`:

- `baseline_channel_contribution` (line 1314-1320) depends on `channel_data` via `forward_pass`, so it gets the extended shape `(num_periods + max_lag, ...)`
- `media_latent_process` (line 1322-1326) is created from `time_index` which only has `num_periods` length
- When these are multiplied elementwise (line 1330), there's a dimension mismatch causing the error

## Additional Consideration: MuEffects

If the model has any `MuEffect`s, the budget optimizer should update those too when the date dimension changes. This could potentially be done through `set_data`, since `channel_data` is a data container.

## Proposed Solution

Consider using `pm.set_data` to update `channel_data` instead of the do-operator approach, which would ensure all dependent components (including `media_latent_process` and `MuEffect`s) are properly updated with the correct date dimension.

## Code References

- `_replace_channel_data_by_optimization_variable`: `pymc_marketing/mmm/budget_optimizer.py:896-952`
- `baseline_channel_contribution` and `media_latent_process` multiplication: `pymc_marketing/mmm/multidimensional.py:1314-1332`
- `MuEffect.set_data` protocol: `pymc_marketing/mmm/additive_effect.py:143`

ricardoV94:
Closes #2104

----
📚 Documentation preview 📚: https://pymc-marketing--2396.org.readthedocs.build/en/2396/


juanitorduz (approved):
🙌

PR #2407, opened 2026-03-16
Replace potential by likelihood in calibration by target

cetagostini:
## Description

After talking with @ricardoV94 and see issues with potentials and the new pymc.dims, decide to replace by likelihood.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2407.org.readthedocs.build/en/2407/


pymc_marketing/mmm/lift_test.py
```diff
@@ -796,13 +800,44 @@ def add_cost_per_target_potentials(
     )
 
     with current_model:
-        # Compute mean over the date dimension once
         cpt_mean = cpt_value.mean(dim="date")
```
TeemuSailynoja: Should we do the reversing of the order of operations so that this would be `mean(cost) / mean(target)` instead of `mean(cost/target)`?

Now, if the spend is  very "spiky", and the channel has a slow (or even delayed!) adstock decay, the current order takes the average over an array of mostly `0`s and `spend/1e-12` values.

Or should we add a boolean switch to select the order of operations?

TeemuSailynoja: ^ is not a blocker, I can review with either decision :)

pymc_marketing/mmm/lift_test.py
```diff
@@ -711,13 +712,20 @@ def add_cost_per_target_potentials(
...
+    For each row, we compute the mean of the cost-per-target variable across
+    the date dimension for the specified (dims, channel) slice and add an
+    observed ``Normal`` likelihood term:
+
+    ``Normal(mu=cpt_mean, sigma=sigma, observed=target)``
 
-    For each row, we compute the mean of ``cpt_variable_name`` across the date
-    dimension for the specified (dims, channel) slice and add a soft quadratic
-    penalty:
+    This is equivalent to the former ``Potential``-based quadratic penalty
```
ricardoV94: These LLM replies to prompts shouldn't become line comments. Who cares what was before?

pymc_marketing/mmm/lift_test.py
```diff
@@ -796,13 +800,44 @@ def add_cost_per_target_potentials(
...
+        Use :func:`add_cost_per_target_observations` instead, which uses an
+        observed ``Normal`` likelihood rather than a manual ``Potential``.
+    """
+    warnings.warn(
+        "add_cost_per_target_potentials is deprecated and will be removed in a "
+        "future version. Use add_cost_per_target_observations instead.",
+        FutureWarning,
+        stacklevel=2,
+    )
+    add_cost_per_target_observations(
```
ricardoV94, line 852: are people really using this directly?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -2326,8 +2326,13 @@ def sample_posterior_predictive(
         for mu_effect in self.mu_effects:
             mu_effect.set_data(self, model, dataset_xarray)
 
+        cpt_names: set[str] = getattr(self, "_cpt_calibration_names", set())
+        if cpt_names and "var_names" not in sample_posterior_predictive_kwargs:
+            all_observed = [rv.name for rv in model.observed_RVs]
+            var_names = [n for n in all_observed if n not in cpt_names]
+            sample_posterior_predictive_kwargs["var_names"] = var_names
```
ricardoV94: Don't like this. Can't you hardcode the variable names you expect in the trace and check directly?

Adaptive testing code tends to hide unwanted regressions

pymc_marketing/mmm/multidimensional.py
```diff
@@ -3031,13 +3040,17 @@ def add_cost_per_target_calibration(
...
+        add_cost_per_target_observations(
             calibration_df=calibration_data,
             model=self.model,
             cpt_value=cpt_tensor,
             name_prefix=name_prefix,
         )
 
+        if not hasattr(self, "_cpt_calibration_names"):
+            self._cpt_calibration_names: set[str] = set()
+        self._cpt_calibration_names.add(name_prefix)
```
ricardoV94: what is this variable used for?

ricardoV94 (commented):
Looks alright, left some comments/ questions

pymc_marketing/mmm/multidimensional.py
cetagostini: Done, now I added a simple bool instead.

pymc_marketing/mmm/multidimensional.py
cetagostini: You are right, done!

pymc_marketing/mmm/lift_test.py
cetagostini: What do you mean?

pymc_marketing/mmm/lift_test.py
cetagostini: Done, you are righ!

pymc_marketing/mmm/lift_test.py
cetagostini: Apply both recommendations, I'm agree!

pymc_marketing/mmm/lift_test.py
ricardoV94: what's the risk of just removing the old function?

pymc_marketing/mmm/multidimensional.py
ricardoV94: I still don't know what it is used for. The reason I ask is to know if it is a good idea or not (not knowing anything, I would say is bad)

pymc_marketing/mmm/lift_test.py
cetagostini: Ah, I think just avoid to break users. I know some people use in prod, including some financial companies 😅 So giving a release to catch on this, before break things!

pymc_marketing/mmm/multidimensional.py
cetagostini: My bad, didn't fully follow the comment, but Its only a flag, so that posterior predictive knows to explicitly request only `self.output_var` when nothing is passed and cpt calibration exist, instead of sampling all observed RVs (which would include CPT calibration variables that don't make sense to predict out-of-sample). 

Without the flag, `sample_posterior_predictive` would default to sampling all observed RVs, including "cpt_calibration". If users share the own var names then fine. All we want to avoid is users adding that in sample posterior predictive.

ricardoV94: why isn't that the default (to only ask for output_var)?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -3018,26 +3022,23 @@ def add_cost_per_target_calibration(
...
+        add_cost_per_target_observations(
             calibration_df=calibration_data,
             model=self.model,
-            cpt_value=cpt_tensor,
+            cost_value=as_xtensor(spend_xarray),
+            target_value=self.model["channel_contribution_original_scale"],
             name_prefix=name_prefix,
         )
 
+        self._has_cpt_calibration = True
```
williambdean: Not huge fan of this for bookkeeping. What happens when called a second time? For
example, the lift tests "support" that.

EDIT: See the ValueError that is in the tests (for second call)

williambdean (approved):
Good on the change. Just wondering if this hidden attribute is a larger book keeping issue.

pymc_marketing/mmm/multidimensional.py
ricardoV94: Same, that's why I started the discussion above. I'm also removing a bunch of these in #2413

juanitorduz: I also agree we should avoid these mutable states (they bit me in the past)

juanitorduz: @cetagostini is there is a way to avoid this it would be great! we are very close to the finish line ;)

cetagostini: I this as better option, I think @ricardoV94 you critic at first, but are we good to get back to it?
```py
if "var_names" not in sample_posterior_predictive_kwargs:
            observed_names = {rv.name for rv in model.observed_RVs}
            if observed_names - {self.output_var}:
                sample_posterior_predictive_kwargs["var_names"] = [self.output_var]
```

ricardoV94: why isn't that just `sample_posterior_predictive_kwargs.setdefault("var_names", [self.output_var])` ?

cetagostini: This would unconditionally set `var_names` to `[self.output_var]` whenever the caller didn't explicitly pass it — even for a plain model with no calibration variables at all. The current logic is intentionally conditional.

The inner check (`observed_names - {self.output_var}`) only fires when there are "extra" observed RVs beyond the target — i.e., calibration likelihoods like `cpt_calibration` or `lift_measurements`. Without that guard, a vanilla model (no calibration) would also get var_names forced, which would suppress any other variables that `pm.sample_posterior_predictive` would normally return by default.

ricardoV94: sounds like over complicating to me. it's a method defined for this model, the model is defined as having a single likelihood and that's the one you want to sample. Just say you want to sample it. 

If someone extends the model to have more observed variables they will have to change many other things, of which overriding the default var_names of sample posterior predictive is the least of their problems.

cetagostini: Okay will follow you here.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -2413,8 +2412,9 @@ def sample_posterior_predictive(
         for mu_effect in self.mu_effects:
             mu_effect.set_data(self, model, dataset_xarray)
 
+        sample_posterior_predictive_kwargs.setdefault("var_names", [self.output_var])
```
juanitorduz, line 2415: @cetagostini I think this is breaking the notebooks

PR #2413, opened 2026-03-17
Make MMM workflow idempotent

### LINKED ISSUE #2226: Multidimensional MMM `sample_posterior_predictive(extend_idata=True)` fails on repeated call
If you call it twice, it will fail because `idata.add_groups` fails if a group already exists.

Issue may be present in other methods

### LINKED ISSUE #2239: Don't mutate Prior objects
Several places we assign `.dims` to Prior objects. I don't think we should this. Could lead to very subtle bugs with Prior objects / transforms reused in multiple models / methods

An alternative is to make a copy with the new dims and use that instead. I did something like that in https://github.com/pymc-labs/pymc-marketing/pull/2227 that we could backport

### LINKED ISSUE #2367: We shouldn't mutate MMM.model inplace with methods like `post_sample_model_transformation`
This is a recipe for very subtle / costly bugs.

When calling `.fit()` again or another predictive method weird things will happen. Can we think of a more functional approach?

### LINKED ISSUE #2385: Make use of xtensor.where once implemented
There are a couple of places in the codebase after #2204 that can be simplified once we have xtensor.where (see https://github.com/pymc-devs/pytensor/issues/1501)

Right now we have to convert to tensor to perform boolean advanced indexing and convert back to xtensor

ricardoV94:
Closes #2367 
Closes #2239
Closes #2226

Also closes #2385 in a separate commit. `xtensor.where` actually already exists but doesn't fit the mold. The current strategy is the best I can think of. So I just removed the TODO comments

### DIFF
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -1072,6 +1072,9 @@ class BudgetOptimizer(BaseModel):
             pymc_model=self._pymc_model,
             idata=self.mmm_model.idata,
             response_variable=response_variable,
+            frozen_deterministics=getattr(
+                self.mmm_model, "frozen_deterministics", None
+            ),
         )
 
     def _compile_objective_and_grad(self):
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -764,7 +764,6 @@ class BudgetOptimizer(BaseModel):
         )
 
         # Fill a zero array, then set only the True positions
-        # TODO: We should be able to implement this with xtensor, once we have `.where`
         budgets_zeros = pt.zeros(self._budget_shape)
         budgets_zeros.name = "budgets_zeros"
         bool_mask = np.asarray(self.budgets_to_optimize).astype(bool)
@@ -996,7 +995,6 @@ class BudgetOptimizer(BaseModel):
 
         # Reconstruct the full shape for each time period
         budgets = ptx.zeros_like(budgets).expand_dims(date=num_periods, axis=0)
-        # Need to go to tensor, as we don't have `.where` yet and xtensor don't support >1D boolean indices
         repeated_budgets = budgets.values[:, bool_mask].set(
             repeated_budgets_flat.values
         )
--- a/pymc_marketing/mmm/incrementality.py
+++ b/pymc_marketing/mmm/incrementality.py
@@ -357,10 +357,12 @@ class Incrementality:
         n_draws = posterior_sub.sizes["draw"]
 
         # Extract response distribution (batched over samples)
+        posterior_predictive_model = self.model.model
         response_graph = extract_response_distribution(
-            pymc_model=self.model.model,
+            pymc_model=posterior_predictive_model,
             idata=az.InferenceData(posterior=posterior_sub),
             response_variable="channel_contribution",
+            frozen_deterministics=self.model.frozen_deterministics,
         )
         # Shape: (sample, date, channel, *custom_dims) where sample = chain x draw
 
@@ -382,7 +384,7 @@ class Incrementality:
         # Compile vectorized evaluator (once, reused for both)
         # Use float64 for evaluation to avoid integer truncation when
         # counterfactual_spend_factor produces fractional values (e.g. 1.01).
-        data_shared = self.model.model["channel_data"]
+        data_shared = posterior_predictive_model["channel_data"]
         data_dtype = data_shared.dtype
         if np.dtype(data_dtype).kind != "f":
             raise ValueError(
@@ -403,12 +405,11 @@ class Incrementality:
         # ``time_index``; otherwise it is unused (e.g. time_varying_intercept
         # without time_varying_media) and would trigger an UnusedInputError.
         has_time_index = (
-            "time_index" in self.model.model.named_vars
-            and self.model.model["time_index"]
-            in ancestors([response_graph], blockers=list())
+            "time_index" in posterior_predictive_model.named_vars
+            and posterior_predictive_model["time_index"] in ancestors([response_graph])
         )
         if has_time_index:
-            time_shared = self.model.model["time_index"]
+            time_shared = posterior_predictive_model["time_index"]
             batched_time = ptx.xtensor(
                 name="time_index_batched",
                 dtype=time_shared.dtype,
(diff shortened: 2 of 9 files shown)


ricardoV94:
I checked the bot about other attributes that are stored in the MMM class, but potentially never used and showed these: 
image

Do we keep/remove/mix?

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -1074,6 +1072,9 @@ def extract_response_distribution(self, response_variable: str) -> XTensorVariab
             pymc_model=self._pymc_model,
             idata=self.mmm_model.idata,
             response_variable=response_variable,
+            freeze_deterministics=getattr(
+                self.mmm_model, "frozen_deterministics", None
+            ),
```
williambdean: This makes #2425 a bit harder

ricardoV94: Would need to pass `frozen_deterministics`:  `BudgetOptimizer(model=mmm.model, frozen_deterministics=..., idata=idata)`

Otherwise the proposal in that PR is equally unsafe, because what if the model passed was not "frozen" already with the flat_deterministics thing?

The fundamental issue is that the posterior conditional model is not the same as the observed model/ the one defined in `build_model`. In our case it has an extra input, that is not an input in the original model, and we can obtain it through different strategies, of which frozen_deterministics is one. 

A different approach (the one I actually started this PR with), is to have the MMM be able to define a `post_sample` method that creates such model (not inplace though). In our case we could freeze that HSPG deterministic with `pm.do` and be done with (hopefully `sample_posterior_predictive` wouldn't freak out). Which means your suggestion to decouple would then require the user to do `BudgetOptimizer(model=mmm.post_sample_model(), idata=idata)`. 

(I would still probably allow passing an MMM object, and just teach BudegetOptimizer what to extract in `__init__`) 

There could be other models whose posterior conditional is even more distinct (maybe uncensored variable for a censored model), or a conditional multivariate from the observed one. 

Either case I don't think this PR is incompatible with trying to decouple. It's just trying to remove an implicit state change that happens after fit. The tests will still be relevant even if we go a different route later?

williambdean: There would be a separate / translation of the model that is used for fitting /
optimization. The idea with that issue is that the `BudgetOptimizer` wouldn't have to
know about the operations on the model in order to prep for the optimization.

Passing list[str] for frozen_deterministics into the `BudegetOptimizer`
as an alternative could be way to go.

williambdean: As for the `__init__` . That's bad practice with pydantic. That is ideally removed
anyway

pymc_marketing/pytensor_utils.py
```diff
@@ -267,6 +267,7 @@ def extract_response_distribution(
     pymc_model: Model,
     idata: InferenceData,
     response_variable: str,
+    freeze_deterministics: list[str] | None = None,
```
williambdean: This sounds like a boolean flag. Would think: 

- frozen_deterministics
- deterministics_to_freeze

pymc_marketing/mmm/multidimensional.py
```diff
@@ -2295,6 +2303,8 @@ def _posterior_predictive_data_transformation(
     def _set_xarray_data(
         self,
         dataset_xarray: xr.Dataset,
+        *,
+        model: pm.Model,
         clone_model: bool = True,
```
williambdean, line 2308: Do we need the clone_model flag? 

```python
# Copy case
mmm._set_xarray_data(dataset_xarray, model=model.copy())
# Non copy case
mmm._set_xarray_data(dataset_xarray, model=model)
```

Then no need for the if and would standardize the call.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -3435,7 +3468,8 @@ def _set_predictors_for_optimization(self, num_periods: int) -> pm.Model:
         # Use the model's method to set data
         pymc_model = self._set_xarray_data(
             dataset_xarray=dataset_xarray,
-            clone_model=True,  # Ensure we work on a clone
+            model=self.model.copy(),
+            clone_model=False,
```
williambdean: This is confusing and reason to drop the clone_model flag .

williambdean: This is confusing and reason to drop the clone_model flag.

pymc_marketing/mmm/incrementality.py
```diff
@@ -357,10 +357,12 @@ def compute_incremental_contribution(
         n_draws = posterior_sub.sizes["draw"]
 
         # Extract response distribution (batched over samples)
+        posterior_predictive_model = self.model.model
         response_graph = extract_response_distribution(
-            pymc_model=self.model.model,
+            pymc_model=posterior_predictive_model,
             idata=az.InferenceData(posterior=posterior_sub),
             response_variable="channel_contribution",
+            freeze_deterministics=self.model.frozen_deterministics,
```
williambdean: In favor of the `frozen_deterministics` name due to common naming with the property.

williambdean (commented):
How much time does these graph operations add to the making predictions?

pymc_marketing/mmm/budget_optimizer.py
ricardoV94: The thing that is called BudgetOptimizer can be a thin object (even a function) that creates the actual object if you don't want init to be involved

pymc_marketing/mmm/multidimensional.py
ricardoV94: I have to check again where else is this called from

pymc_marketing/mmm/incrementality.py
ricardoV94: changed

pymc_marketing/mmm/multidimensional.py
ricardoV94: dropped it. We do need to watch out in the future to not accidentally change stuff inplace when we don't want to though.

pymc_marketing/pytensor_utils.py
ricardoV94: changed. Don't love frozen_deterministics much, because it sounds like they are frozen already, but the other is too verbose. I still slightly prefer the original but not by any large margin

juanitorduz (approved):
Thnks @ricardoV94 . LGTM :)

williambdean (approved):
Thanks for the edits! Looks good

PR #2437, opened 2026-03-23
Initial PR for MMMPlotSuite

isofer:
## Description

Following the attack plan in [here](https://github.com/pymc-labs/pymc-marketing/blob/isofer/mmmplotsuite-overall/docs/plans/2026-03-11-mmmplotsuite-v2-attack-plan-design.md#behavioral-rules) this PR implement the first few steps: Introduces a new `pymc_marketing/mmm/plotting/` package with foundation helper utilities and a `TransformationPlots` namespace class. 

Please review [this notebook](https://gist.github.com/isofer/19e1c3d99a5bda35323941a4d113e5f8) to see these plots and provide feedback about the new API

**Note: This PR should be merged after 0.19 is out, since it removes that support for python 3.11!!!!**

**What this adds:**
- `pymc_marketing/mmm/plotting/_helpers.py` — shared utilities used across all plotting namespace classes:
  - `_validate_dims` — validates that dimension filters exist in an xarray object
  - `_dims_to_sel_kwargs` — wraps scalar dim values in lists so `.sel()` preserves dimensions for faceting
  - `_select_dims` — combines validation and `.sel()` into one step
  - `_process_plot_params` — normalizes `figsize`, `backend`, and `return_as_pc` parameters
  - `_extract_matplotlib_result` — converts a `PlotCollection` to `(Figure, NDArray[Axes])`
- `pymc_marketing/mmm/plotting/transformations.py` — `TransformationPlots` class with:
  - `saturation_scatterplot`: scatter of channel spend/data vs. mean posterior contributions, one panel per channel (and custom dims)
  - `saturation_curves`: posterior sample lines + HDI band overlay on scatter, with support for both `original_scale` and model-internal scaled space

**Fixes bundled in this PR:**
- `multidimensional.py`: `sample_saturation_curve` and related methods now drop the `(chain, draw)` MultiIndex after stacking, replacing it with a plain integer `sample` coordinate. This prevents downstream xarray operations (e.g., `unstack`) from failing when the MultiIndex is serialized or passed between contexts.
- `mmm_wrapper.py`: `get_channel_scale` and `get_target_scale` now return copies of the xarray variables to avoid unintentional mutation of `idata.constant_data`.

**Dependency:**
- Adds `arviz-plots[matplotlib]>=1.0.0,<1.1` to `environment.yml` and `pyproject.toml`.

## Changes Made
- New `pymc_marketing/mmm/plotting/` package (`__init__.py`, `_helpers.py`, `transformations.py`)
- New test module `tests/mmm/plotting/` with `test_helpers.py` (321 lines) and `test_transformations.py` (707 lines)
- Fix: plain integer `sample` coordinate after stacking chain/draw in `multidimensional.py`
- Fix: return `.copy()` from `MMMIDataWrapper.get_channel_scale` and `get_target_scale`
- Dependency: `arviz-plots[matplotlib]>=1.0.0,<1.1`


isofer:
@cursoragent review

pymc_marketing/mmm/plotting/transformations.py
```diff
@@ -0,0 +1,387 @@
...
+            original_scale=original_scale,
+            apply_cost_per_unit=apply_cost_per_unit,
+            idata=idata,
+            dims=dims,
+            figsize=figsize,
+            backend=backend,
+            return_as_pc=True,
+            scatter_kwargs=scatter_kwargs,
+            **pc_kwargs,
+        )
```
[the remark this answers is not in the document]
isofer: non issue. This is validated by the call to saturation_scatter.

pymc_marketing/mmm/plotting/_helpers.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+    if not dims:
+        return {}
+    return {
+        k: v if isinstance(v, (list, tuple, np.ndarray)) else [v]
+        for k, v in dims.items()
+    }
+
+
+def _select_dims[XarrayT: (xr.Dataset, xr.DataArray)](
```
juanitorduz, line 56: Shall we define `XarrayT: (xr.Dataset, xr.DataArray)` as a type hint at the top of the file?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -2579,8 +2579,13 @@ def sample_saturation_curve(
...
 
-        # Flatten chain and draw dimensions to sample dimension
-        curve = curve.stack(sample=("chain", "draw"))
+        # Flatten chain and draw dimensions to a plain integer sample dimension
+        n_samples_total = curve.sizes["chain"] * curve.sizes["draw"]
+        curve = (
+            curve.stack(sample=("chain", "draw"))
+            .drop_vars(["chain", "draw"])
+            .assign_coords(sample=np.arange(n_samples_total))
+        )
```
juanitorduz: Can we do this somewhere else? This function should just generate samples with the expected (chain, draw) dims for custom plots and metrics. These post-hoc transforms can be handled by a helper function and called when needed imo.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -2713,8 +2718,13 @@ def sample_adstock_curve(
...
```
juanitorduz: same here (this hintsd a helper function)

juanitorduz (changes requested):
I really like this API. I left some minor comments. 

Once we are happy with it let's remove the plans :)

pymc_marketing/mmm/plotting/_helpers.py
isofer: I actually prefer it here (and the linter as well). This syntax was introduce in [PEP 695](https://peps.python.org/pep-0695/). One of the big benefit is that since it is used only in this function the reader doesn't have to scroll to the op of the file to understand what types are valid. It's also less boilerplate.

pymc_marketing/mmm/multidimensional.py
isofer: done.

pymc_marketing/mmm/multidimensional.py
isofer: done.

isofer:
@juanitorduz any other comments or should we merge?

juanitorduz (approved):
This one LGTM! Thanks! 

I passed it through Code,X and here is some additional additions (some optional) which would be beneficial to have:

## Priority Fixes
- In [`pymc_marketing/mmm/plotting/transformations.py`](pymc_marketing/mmm/plotting/transformations.py), validate `backend`/`return_as_pc` at the start of `saturation_curves` via `_process_plot_params(...)` (or equivalent explicit guard) so non-matplotlib backends fail with a clear `ValueError` instead of downstream extraction errors.
- In [`pymc_marketing/mmm/plotting/transformations.py`](pymc_marketing/mmm/plotting/transformations.py), remove unnecessary `sampled_curves["x"] = spend_data` assignment to avoid potential coordinate misalignment side effects.
- Add regression coverage in [`tests/mmm/plotting/test_transformations.py`](tests/mmm/plotting/test_transformations.py) for:
  - `saturation_curves(backend="plotly", return_as_pc=False)` raising the expected validation error.
  - sampled-curves rendering path preserving expected y-values/line counts without coordinate corruption.

## Consistency Improvements
- Align `arviz-plots` bounds between [`pyproject.toml`](pyproject.toml) and [`environment.yml`](environment.yml) to avoid divergent resolution across install paths.
- Document the Python baseline change (`>=3.12`) explicitly in release/migration notes and confirm no stale CI/docs references remain in changed files.

## API/Docs Polish
- Add short `Examples` blocks to public methods in [`pymc_marketing/mmm/plotting/transformations.py`](pymc_marketing/mmm/plotting/transformations.py) to match project NumPy-doc expectations.
- Optionally tighten helper return annotations in [`pymc_marketing/mmm/plotting/_helpers.py`](pymc_marketing/mmm/plotting/_helpers.py) (e.g., `dict[str, Any]`).

PR #2450, opened 2026-03-27
Halo Effect Modeling Notebook

juanitorduz:
Halo Effect modeling notebook with PyMC-Marketing and Mu (additive effects)

----
📚 Documentation preview 📚: https://pymc-marketing--2450.org.readthedocs.build/en/2450/


Hendrik-86:
Hi Juan,
I went through code. I can follow your reasoning and your approaches (some way of coding it is new to me and I have to get used to it). Before I went through the code, I thought about how I would try to solve this. But in general, the problem of P^2 parameters that would need estimation (in a full cross-product halo effect analysis) is pretty tough to tackle. Here are my thoughts as they came up while reading your code:

a) **Cannibalization**: If the marketing spending on one product reduces sales of another product, this would not be covered forcing coefficients to be positive. This is in general something I do not like (because you could theorize this also across channels), forcing coefficients to be zero or positive with half-normal distribution priors. But I think this argument would be even more pronounced in the product dimension.

b) **Product similarity**: Maybe the (perceived) distance of products could be an explaining factor for either i) halo effects or ii) cannibalization effects. Identifying these could help reducing to "sustainable" amount of cross-product effects. So, something in between your "flagship-approach" and the full "spillover-matrix" approach. So a variable could also be fruitful for what I write in c).

c) **Non-parametric estimation of heterogeneous effects**: I recently explored the possibiltiy to use panel local projections (PLP) for marketing mix modeling (but in a different context than the halo use case, still working on a small conceptual white paper). Panel local projections have a long tradition in macroeconomics to estimate the effect of a macro "shock" (say, changes in government spending) on cross-sectional units, e.g. firms, households. It shares some features with vector autoregressions (VARs) but have some nicer properties. There is also work on Bayesian PLP, see: https://www.marcoschwarzbach.de/uploads/BayesianPanelLocalProjections.pdf.
Recently, there is more work on how to use PLP for heterogeneous treatment effects (See: https://www.bankofengland.co.uk/-/media/boe/files/working-paper/2024/how-do-firms-financial-conditions-influence-the-transmission-of-monetary-policy.pdf. If you compare formula (1) and (2) you understand what I'm thinking of. I do not claim this is the solution, I never estimated heterogeneous effects this way, but this could be an avenue.

d) **Projections (related to c) )**: This is a rather general comment and not directly connected to the halo-Notebook. I always felt uncomfortable with the "chaining" of the two transformations (adstock => saturation) as it is done in the Meridian and PyMC MMM models. For that reason, I was looking for an alternative. Local projections, i.e. estimating effects of time horizons t+h (h=0,1,2,3 etc.) seems to me like a natural way of disentagling the saturation estimation from the adstock estimations (this is not a panacea and has its own difficulties!, especially for longer time horizons h). The resulting impulse-responses are also a natural way in communicating adstock effects to stakeholders as this is quite intuitive to understand.

Happy to hear your thoughts or also discuss some details in a call!

juanitorduz:
Thanks for the input @Hendrik-86 ! I will read the references :)

Point (a) is a fair point that is a clear limitation of this notebook (thanks for bringing this up). The current notebook assumes "the naive?" idea that the halo effects are overall positive.

In the meantime, note that you can turn on and off the adstock and saturation, see https://www.pymc-marketing.io/en/stable/api/generated/pymc_marketing.mmm.components.saturation.NoSaturation.html#pymc_marketing.mmm.components.saturation.NoSaturation :)

Hendrik-86:
> Thanks for the input @Hendrik-86 ! I will read the references :)
> 
> Point (a) is a fair point that is a clear limitation of this notebook (thanks fro brining this up)
> 
> In the meantime, note that you can turn on and off the adstock and saturation, see https://www.pymc-marketing.io/en/stable/api/generated/pymc_marketing.mmm.components.saturation.NoSaturation.html#pymc_marketing.mmm.components.saturation.NoSaturation :)

Than it would be even easier to apply together with PLP :)

juanitorduz:
## Update: mechanism DAGs, forest plots, CRPS

Pushed in 22c85479. The notebook was re-executed end to end with real MCMC (52s, no error outputs) and every numeric and directional claim in the prose was re-checked against the executed outputs.

**Framing and DAGs.** The introduction now presents halo as a family of mechanisms that `MuEffect` and a mask can express, rather than one canonical model. Three graphviz figures carry the argument: the three mechanisms side by side (shared response, common latent driver, directed transfer), each mask drawn as the set of arrows it asserts, and the confounded planning calendar in the identifiability section.

**In-sample CRPS, and a correction.** Each model now reports in-sample CRPS next to RMSE. Writing this up surfaced a claim that the numbers do not support: with `|bias|` on the y axis the six models line up monotonically, so "in-sample score cannot rank halo structures" was too strong. The same table contains the sharper point, so the figure now plots *signed* bias:

| model | in-sample CRPS | bias vs. truth |
| --- | --- | --- |
| baseline (no cross term) | 0.1731 | −30.3% |
| other products' spend as channel | 0.1450 | −30.0% |
| flagship mask (sign-free) | 0.1308 | −5.2% |
| full spillover mask | 0.1295 | −3.0% |
| flagship mask (HalfNormal) | 0.1433 | **+21.3%** |
| brand factor | 0.1297 | −3.7% |

The aggregate-spend workaround and the `HalfNormal` variant tie to within one percent on CRPS while their portfolio values sit fifty points apart on opposite sides of the truth, so the score does not even carry the sign of the error. And the sixteen percent CRPS gain from the baseline to the workaround buys no reduction in bias at all. The prose and the conclusion bullet say that now, and note that the monotone ordering is a property of this DGP rather than something a reader can rely on.

**Plots and typing.** The two coefficient comparisons use `az.plot_forest` with per-receiver truth diamonds instead of overlaid histograms. Every function and `MuEffect` method has input and output type hints.

**Legibility.** Legends are at the default font size and clear of the data: bold axes titles on the forest plots (the bold suptitle collided with the legend title), one legend on the portfolio figure instead of six identical ones, light text on saturated cross-effect cells so `+0.77` is readable, and no gridlines across the mask matrices.

One thing worth flagging for future notebook edits in this repo: editing via jupytext and formatting the intermediate `.py` silently deletes the output-suppressing trailing semicolons, because `ruff format` strips them in a `.py` file but preserves them in an `.ipynb`. That had re-broken the twelve cells fixed in 12cff0ff. They are restored, the new figure cells have them, and formatting now runs on the notebook rather than the Python file.

cetagostini:
I added a few points, not blockers!

juanitorduz:
<p>Added some comment <a href="https://github.com/pymc-labs/pymc-marketing/pull/2450/commits/aa9621354e07242bfece7ec76d760ae21a81e107" target="_blank">https://github.com/pymc-labs/pymc-marketing/pull/2450/commits/aa9621354e07242bfece7ec76d760ae21a81e107</a></p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2450/discussion/#comment-5263583629'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

cetagostini (approved):

PR #2472, opened 2026-04-01
Fix: Latex rendering in new Adstock guide 

timbo112711:
## Description
The latex in our new Adstock guide is not rendering correctly. This may be due to using `\textrm{transf}`  compared to `\text{transf}`. 

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2472.org.readthedocs.build/en/2472/


docs/source/conf.py
```diff
@@ -96,6 +98,43 @@
...
+    try:
+        pio.renderers[renderer].include_mathjax = False
+    except AttributeError:
+        continue
+
+# Force the default to be a renderer we just disabled
+pio.renderers.default = "notebook"
+
+# Force Sphinx to use the specific MathJax 4 version
+mathjax_path = "https://cdn.jsdelivr.net/npm/mathjax@4.1.1/es5/tex-mml-chtml.js"
```
OriolAbril: If this isn't strictly necessary for the workaround to work I would advise against adding it. It means part of the maintenance tasks of pymc-marketing will include keeping this up to date. Sphinx is not the most cutting edge library but sphinx 9.0 already includes mathjax 4 out of the box and will keep it updated for us. Once we check the external extensions and the theme for 9.0 compatibility

docs/source/conf.py
```diff
@@ -268,3 +307,22 @@ def fallback_source():
         "Miscellaneous",
     )
 ]
+
+
+def scrub_plotly_mathjax(app, pagename, templatename, context, doctree):
```
OriolAbril, line 303: Isn't the need for this a sign that the `pio` configuration above isn't actually working?

OriolAbril (commented):
Sorry about the slow response. I was aware this happens but I don't have a good fix. This happened in pymc-examples too but the couple plotly plots there weren't really using interactive features so we switched the renderer to `png` to circumvent the issue which is not a valid workaround here.

I can confirm the plotly scripts no longer seem to include their own mathjax but I still don't see the equations as rendered latex. I would be very interested in getting a proper workaround so we can use it in all the other documentation pages too. So I am also trying to understand everything that is going on.

In general it is a bit slower and a bit uglier (and might make it harder to build docs locally on windows) but a potential alternative workaround could be switching to imgmath instead of mathjax: https://www.sphinx-doc.org/en/master/usage/extensions/math.html. Sphinx has two native ways of rendering equations.

docs/source/conf.py
timbo112711: I agree, can remove. Was testing out if this would work.

docs/source/conf.py
timbo112711: From what I gather, `include_mathjax = False` often fails because Plotly’s rendering logic is split into two separate worlds: the Python side and the JavaScript side. This is basically saying when you generate the HTML snippet for this figure, don't include the `<script>` tag for MathJax.

The problem is that Plotly figures are powered by `plotly.js`. If `plotly.js` sees a math symbol but doesn't detect MathJax already loaded on the page, it has a fail-safe built into its source code. It then programmatically creates a new `<script>` tag pointing to its favorite old version (v2.7.5) and attaches it to the document.

So the goal here would be to figure out a way to let `plotly.js` know that MathJax is loaded. That way it doesn't inject its own version.

juanitorduz (approved):

PR #2479, opened 2026-04-02
Support fixed/user-defined scaling factors for stable model refreshes

### LINKED ISSUE #2478: Support fixed/user-defined scaling factors for stable model refreshes
## Problem

When running MMMs in production with periodic model refreshes (e.g. weekly or monthly refit as new data arrives), the current scaling behavior can destabilize the parameter space between runs.

Today, `VariableScaling` supports `"max"` and `"mean"` methods, both of which derive the scaling factor from the data at fit time. This means that when new data shifts the distribution of a channel — for example, a campaign ramps up spend beyond the previous historical maximum — the scaling factor changes. Because the model's priors were calibrated relative to the previous scale, a shift in scaling effectively moves the goalposts, requiring prior recalibration to maintain comparable parameter interpretations across runs.

This creates a tension in production MMM workflows: the modeler must choose between re-deriving priors every refresh cycle (costly and error-prone) or accepting that parameter estimates may drift in meaning even when the underlying relationship hasn't changed.

## Desired outcome

A way for users to specify **fixed scaling constants per variable** that remain stable across model refreshes, regardless of changes in the incoming data distribution. The scaling factor for a channel would be set once (informed by domain knowledge or an initial data exploration) and then held constant for the lifetime of that model configuration.

This would allow teams to:

- **Set priors once** and have them remain meaningful across refreshes
- **Compare parameters across time** without worrying about scale-induced drift
- **Control scaling with domain knowledge** — e.g. setting a channel's scale to a round business-meaningful number rather than a data-derived statistic

## Context

This need arises specifically in recurring/production MMM use cases where models are periodically refit on expanding or rolling data windows. For one-off exploratory analyses, data-derived scaling works fine. The gap becomes apparent when the same model specification is reused across multiple data snapshots over time.

## Related issues

No directly related issues. Issues below are loosely related:

* #299
* #2436
* #2004

drbenvincent:
## Summary

Closes #2478.

Adds **fixed (user-defined) scaling** so scale factors can stay constant across production refreshes, avoiding scale-induced prior drift when data distributions change.

Scaling is modeled as a **small class hierarchy** instead of a single `VariableScaling` with `method="fixed"`:

- **`DataDerivedScaling`** — `method` is `"max"` or `"mean"`; scales are computed from data at fit time (existing behaviour).
- **`FixedScaling`** — `value` is supplied by the user and does not depend on the current dataset (beyond alignment for multidimensional grids).

### Public API

- **`VariableScaling`** — abstract base; **`DataDerivedScaling`**, **`FixedScaling`** — concrete types.
- **`FixedScaling.value`** — `float | dict[str, float] | xarray.DataArray` (all strictly positive; bool rejected on the scalar path). A **`method`** property returns `"fixed"` for compatibility with code that inspects method names.
- **`Scaling.target` / `Scaling.channel`** — each is a **`VariableScaling`** (either derived or fixed).
- **`FixedScaling.from_long_dataframe`** — build a grid-valued fixed scale from a long table.
- **`Scaling`**, **`VariableScaling`**, **`DataDerivedScaling`**, and **`FixedScaling`** are exported from **`pymc_marketing.mmm`**.

### Legacy `MMM` (1D)

- Fixed scales use **`FixedScaling`**; `_compute_scales()` skips data statistics when the channel (or target) config is **`FixedScaling`**.
- **Scalar** `value` broadcasts to all channels; **dict** maps **`channel_columns`**; target fixed scaling allows **scalar only** (dict raises).
- **Serialization** persists full configs (including **`FixedScaling.value`**); legacy dicts with `method: "fixed"` still deserialize.

### `MultidimensionalMMM`

- **`_compute_scale_for_variable`** routes **`FixedScaling`** through **`_build_fixed_scale`**, which supports:
  - **scalar** → dimensionless `DataArray`;
  - **dict** → only when **exactly one non-date dimension remains** after reduction over `date` and scaling `dims` (labels must match that dimension’s coordinates); clear error if multiple dimensions remain (use **`DataArray`** instead);
  - **`xr.DataArray`** → align / broadcast to the media grid (e.g. country × channel).
- **Init**: dict key checks (**missing / extra keys**) run only when the channel fixed scale is a **dict** and the remaining panel dims after reduction are exactly **`("channel",)`** — otherwise validation is deferred to build/align so multi-dimensional dicts are not mis-validated at construction time.
- Helper **`panel_channel_fixed_scaling_remaining_dims`** documents which non-date dims remain for channel data after reduction.
- Module docstring includes a **fixed scaling** example for multidimensional configs.

### Serialization

- **`FixedScaling` / `Scaling`**: `to_dict` / `from_dict` with JSON-safe encoding for **`DataArray`** values (and legacy deserialization for older encoded payloads).

## Test plan

- **`tests/mmm/test_scaling.py`** — construction, validation, **`DataArray`** / **`from_long_dataframe`**, serialization round-trips.
- **`tests/mmm/test_mmm.py`** — legacy MMM fixed scaling, stability across data changes, dict behaviour, serialization.
- **`tests/mmm/test_multidimensional.py`** — scalar, dict, **`DataArray`**, grids, misaligned coords, init vs build validation, **`idata` round-trip**, and related edge cases.

Pre-commit (ruff, mypy, format) run on touched files.


drbenvincent (commented):
Review from `gpt-5.4-xhigh`.

I like the direction here and think this is close, but I found two merge-relevant issues in the legacy `MMM` path and one smaller robustness gap in the multidimensional path.

1. Legacy `MMM` cannot round-trip fixed scaling through model serialization/loading.

In the legacy `MMM`, `create_idata_attrs()` still serializes scaling as only `method` + `dims`, while `_deserialize_scaling()` reconstructs `VariableScaling(...)` from that partial payload. That worked before, but with `method="fixed"` it now drops the required `value`.

I reproduced this locally with a small legacy `MMM` using fixed scaling:
- Build a model with `scaling=Scaling(... method="fixed", value=...)`
- Call `attrs = mmm.create_idata_attrs()`
- Call `MMM.attrs_to_init_kwargs(attrs)`
- Result: `ValidationError: value is required when method='fixed'`

So a saved legacy MMM with fixed scaling currently cannot be reconstructed correctly. Since this feature is motivated by stable production refreshes, I think this needs to be fixed before merge.

Suggested fix:
- Serialize/deserialize the full `Scaling` object here too, similar to the multidimensional MMM path.
- Add a regression test around `create_idata_attrs()` / `attrs_to_init_kwargs()` or a full `load_from_idata()` round-trip with fixed scaling.

2. Legacy `MMM` does not actually support the dict-valued fixed scales this PR advertises.

The PR description and `VariableScaling.value` type both say fixed scaling accepts `float | dict[str, float]`, but the legacy `MMM` implementation only really handles a scalar.

What I observed locally:
- For channel scaling, `_compute_scales()` does `np.full(n_channels, channel_scaling.value)`. If `value` is a dict, that becomes an object array and `build_model()` crashes with `TypeError: Unsupported dtype for TensorType: object`.
- For target scaling, a dict falls through to `_compute_scale_for_data(..., method="fixed")`, which raises `ValueError: Unknown scaling method: fixed`.

This is more than just an edge case. In the legacy single-dimension MMM, a very natural user need is per-channel fixed scales so the same channel-specific normalization can be preserved across refreshes. Right now the implementation only supports one scalar copied to every channel.

Suggested fix:
- Either support dicts keyed by `channel_columns` in the legacy `MMM`, or explicitly reject dict-valued fixed scales there with a clear validation error.
- If the intended scope is scalar-only for legacy `MMM`, narrow the public API/docs/PR summary to match that.
- Add at least one regression test for the chosen behavior.

3. Nice to tighten: multidimensional dict-valued fixed scales should validate keys more explicitly.

In the multidimensional path, dict-valued fixed scaling works for the happy path, but key mismatches are not handled very cleanly:
- Missing labels fail late with a raw `KeyError`
- Extra labels are silently ignored

I do not see this as a merge blocker if we want to move quickly, but I would prefer either exact-key validation against the remaining coordinate labels or at least a focused follow-up issue/test so this does not stay implicit.

For context, the new happy-path tests do pass locally:
- `tests/mmm/test_scaling.py`
- `tests/mmm/test_mmm.py -k fixed_scaling`
- `tests/mmm/test_multidimensional.py -k fixed_scaling`

So the main issue is not that the feature never works; it is that a couple of important user-facing paths are still uncovered and currently broken/incomplete.

pymc_marketing/mmm/scaling.py
```diff
@@ -23,15 +23,68 @@
...
         ...,
         description="The dimensions to perform operation through.",
     )
+    value: float | dict[str, float] | None = Field(
+        default=None,
+        description=(
+            "Fixed scaling constant(s). Required when method='fixed', "
+            "forbidden otherwise."
+        ),
+    )
```
williambdean: Having this None seems odd the max and mean. Can we break into two classes?

The serialization PR shouldve made this easier

pymc_marketing/mmm/scaling.py
```diff
@@ -46,28 +99,59 @@ def _validate_dims(self) -> Self:
...
+                        )
+            elif self.value <= 0:
+                raise ValueError(
+                    f"Fixed scaling value must be positive, got {self.value}."
+                )
+        else:
+            if self.value is not None:
+                raise ValueError(
+                    f"value must be None when method='{self.method}', got {self.value}."
+                )
```
williambdean: Simplifies with class split

tests/mmm/test_multidimensional.py
```diff
@@ -1817,6 +1817,211 @@ def test_scaling_dict_doesnt_mutate() -> None:
...
+    mmm = MMM(
+        adstock=GeometricAdstock(l_max=2),
+        saturation=LogisticSaturation(),
+        scaling=scaling,
+        date_column="date",
+        target_column="target",
+        channel_columns=["channel_1", "channel_2", "channel_3"],
+        dims=("country",),
+    )
+    with pytest.raises(ValueError, match=r"missing keys.*channel_3"):
```
williambdean, line 2216: Whats stopping this from being raised at initilization?

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1688,19 +1688,73 @@ def _compute_scales(self) -> None:
...
+                values = np.array([scaling.value[str(c)] for c in coords])
+                return xr.DataArray(
+                    values,
+                    dims=(dim_name,),
+                    coords={dim_name: coords},
+                )
+            return xr.DataArray(scaling.value)
+
+        method_fn = getattr(data, scaling.method)
+        return method_fn(dim=reduce_dims)
```
TeemuSailynoja: Coding style question: do we use multiple return statements? Is this important?

drbenvincent:
@williambdean Thanks for the review — great call on the class split. I've refactored \`VariableScaling\` into a two-class hierarchy:

- **\`VariableScaling\`** is now an abstract base (keeps \`dims\` + validation)
- **\`DataDerivedScaling\`** has \`method: Literal["max", "mean"]\` — no \`value\` field at all
- **\`FixedScaling\`** has \`value: float | dict[str, float]\` (required, must be positive) — no ambiguous \`None\`

This eliminates the cross-field \`_validate_fixed_value\` validator entirely, and the \`__type__\` registry handles polymorphic serialization/deserialization (with a legacy fallback for models saved before this change). All existing tests pass.

drbenvincent:
@TeemuSailynoja Good point on the multiple return statements! The original `_compute_scale_for_variable` had three interleaved returns at different nesting levels, which made the control flow harder to follow.

I've refactored it into two methods:
- **`_compute_scale_for_variable`** now has a single return — it dispatches to `_build_fixed_scale` for `FixedScaling` instances, or computes the data-derived statistic otherwise.
- **`_build_fixed_scale`** handles the fixed-value logic (scalar vs dict) in its own focused method.

This keeps each method clear and avoids the scattered returns.

pymc_marketing/mmm/scaling.py
```diff
@@ -23,15 +23,68 @@
...
+    method : ``"max"`` | ``"mean"`` | ``"fixed"``
+        The scaling method.
+    dims : str or tuple of str
+        The dimensions to perform the operation through (``"date"`` is always
+        included implicitly).
+    value : float or dict[str, float] or None
+        Fixed scaling constant(s).  Required when ``method="fixed"``, forbidden
+        otherwise.  A single ``float`` applies uniformly; a ``dict`` maps
+        dimension-level labels to per-level constants (useful for the
+        multidimensional MMM).  All values must be positive.
```
TeemuSailynoja: How to specify multidimensional scales?

For example, suppose I have an MMMM using impressions, and I want to scale these to be relative to the size of the user base of a given (country, product) pair.

pymc_marketing/mmm/scaling.py
drbenvincent: Split into `DataDerivedScaling` (method: max|mean) and `FixedScaling` (required value, no None). Resolved in f7ba39b6.

pymc_marketing/mmm/scaling.py
drbenvincent: Cross-field validator removed — each subclass validates only its own fields now. Resolved in f7ba39b6.

tests/mmm/test_multidimensional.py
drbenvincent: Moved to `__init__` — dict keys are now validated against `channel_columns` at construction time. Resolved in e9f657db.

pymc_marketing/mmm/multidimensional.py
drbenvincent: Extracted `_build_fixed_scale` so `_compute_scale_for_variable` now has a single return. Resolved in 9820d2f3.

tests/mmm/test_multidimensional.py
```diff
@@ -1817,6 +1817,211 @@ def test_scaling_dict_doesnt_mutate() -> None:
...
+        adstock=GeometricAdstock(l_max=2),
+        saturation=LogisticSaturation(),
+        scaling=scaling,
+        date_column="date",
+        target_column="target",
+        channel_columns=["channel_1", "channel_2", "channel_3"],
+        dims=("country",),
+    )
+    with pytest.raises(ValueError, match=r"exactly one remaining dimension"):
+        mmm.build_model(X, y)
```
TeemuSailynoja, line 1968: If the decision is to add support for multidimensional fixed scales, then this will need a rework.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -455,6 +466,8 @@ def __init__(
                 f"Channel scaling dims {self.scaling.channel.dims} must contain {self.dims}, 'channel', and 'date'"
             )
 
+        _validate_fixed_scaling_keys(self.scaling.channel, channel_columns, "channel")
```
TeemuSailynoja: this is not testing for the case where the user sets a fixed scale per country for the media.
Say I scale impressions to be relative to the population of each region.

TeemuSailynoja: Shouldn't we also validate the keys of `self.scaling.target` ?

tests/mmm/test_multidimensional.py
```diff
@@ -1812,9 +1816,203 @@ def test_scaling_dict_doesnt_mutate() -> None:
...
+        target_column="target",
+        channel_columns=["channel_1", "channel_2", "channel_3"],
+        dims=("country",),
+    )
+    mmm.build_model(X, y)
+
+    assert float(mmm.scalers._target) == fixed_target
+    assert float(mmm.scalers._channel) == fixed_channel
+
+
```
TeemuSailynoja, line 1849: Missing a test where there is fixed scaling for media per some other dimension than "channel".
Also, as mentioned above, this case will fail with the current implementation of `_validate_fixed_scaling_keys`.

TeemuSailynoja (changes requested):
The current implementation works for the base MMM class, but I have left a few comments for the multi-dimensional class.

In short, I think the `multidimensional.MMM` calss should support multidimensional fixed scaling. If we decide against that, then we should also state that in the docstrings.

drbenvincent:
## Update (tip: \`5c2bb965\`)

### What changed
- **Multidimensional init:** Channel dict keys are validated against \`channel_columns\` only when the sole remaining non-date dim after reduction is \`channel\` (\`panel_channel_fixed_scaling_remaining_dims\`). Panel-keyed fixed scales are no longer rejected at \`__init__\` by the wrong check.
- **API:** \`FixedScaling.value\` may be a scalar, a **1D dict** (exactly one remaining dim after reduction), or an **\`xr.DataArray\`** that aligns/broadcasts to the reduced grid (e.g. country-only → broadcast over country×channel; full country×channel; country×product×channel on a 2D panel). **\`FixedScaling.from_long_dataframe\`** builds a \`DataArray\` from a long table.
- **Serialization:** \`DataArray\` fixed values round-trip through \`FixedScaling.to_dict\` / \`from_dict\` and idata attrs (see \`test_multidimensional_fixed_scaling_dataarray_idata_roundtrip\`).
- **Tests/docs:** New coverage for broadcast, full grids, 3D panel, misaligned coords, target dict missing a country, dict-vs-\`DataArray\` error text; module + \`FixedScaling\` docstrings.

### Question for @TeemuSailynoja (API shape)
We still support **dict** when there is exactly one remaining dimension (ergonomic / JSON-friendly). Alternatively we could **drop dict for non-scalar** and require a 1D \`DataArray\` everywhere (single code path, slightly more verbose). **Do you prefer keeping dict, deprecating it, or normalizing dict→DataArray internally only?** Happy to follow your preference in a follow-up PR.

I am also replying **once per inline review thread** below with pointers to the relevant commit(s).

pymc_marketing/mmm/scaling.py
drbenvincent: Still addressed by the `DataDerivedScaling` / `FixedScaling` split from **f7ba39b6** (see earlier reply in this thread). That design is unchanged on tip **5c2bb965**.

pymc_marketing/mmm/scaling.py
drbenvincent: Same resolution as the sibling thread: **f7ba39b6** introduced `DataDerivedScaling` / `FixedScaling`; still the same on **5c2bb965**.

tests/mmm/test_multidimensional.py
drbenvincent: Original change: dict key errors for the **channel-keyed** case fire at `MMM.__init__` (**e9f657db**). **5c2bb965** refines multidimensional init: `_validate_fixed_scaling_keys` runs only when the sole remaining dim after reduction is **`channel`**; otherwise panel-keyed fixed scales are validated when building from xarray coords.

pymc_marketing/mmm/multidimensional.py
drbenvincent: Refactor with `_build_fixed_scale` so `_compute_scale_for_variable` has a single return (**9820d2f3**). **5c2bb965** keeps that structure and extends `_build_fixed_scale` for `xr.DataArray` alignment without reintroducing multiple returns in `_compute_scale_for_variable`.

pymc_marketing/mmm/scaling.py
drbenvincent: **5c2bb965**: use `FixedScaling.value` as an `xr.DataArray` that broadcasts to the reduced media grid (e.g. country-only vector, full country×channel, or country×product×channel on a matching panel). `FixedScaling.from_long_dataframe` covers long-format tables. Docs/examples updated in `FixedScaling` and the multidimensional MMM module docstring.

tests/mmm/test_multidimensional.py
drbenvincent: **5c2bb965**: Reworked this test so a dict in that geometry must use a `DataArray` (error text mentions `DataArray`). Added `test_fixed_scaling_dataarray_country_channel_grid` for the successful `DataArray` path with the same country×channel remainder.

pymc_marketing/mmm/multidimensional.py
drbenvincent: **5c2bb965**: (1) Per-country media scaling (e.g. population): `test_fixed_scaling_dataarray_broadcasts_country_only` — country-only `xr.DataArray` broadcasts over country×channel; init no longer applies `_validate_fixed_scaling_keys` when the remaining axis is not `channel`. (2) **Target dict keys:** panel labels are not available at `__init__`, so we validate target dict keys when building scales from `_target` coords (`_build_fixed_scale_from_dict`), same style as channel — see `test_fixed_scaling_target_dict_missing_country`.

tests/mmm/test_multidimensional.py
drbenvincent: **5c2bb965**: Added `panel_channel_fixed_scaling_remaining_dims` and only call `_validate_fixed_scaling_keys` at init when the sole remaining dim is `channel` (fixes the false failure for panel-keyed channel scales). New tests: `test_fixed_scaling_dataarray_broadcasts_country_only`, `test_fixed_scaling_dataarray_three_dims_country_product_channel`, misaligned-coords test, plus `DataArray` grid test.

TeemuSailynoja (approved):
Looks good now. I think keeping the dict in as an input option and making the helper for long format dataframes should keep this approachable for most users.

The tests that suddenly start failing is strange. I see that `fastprocess`  is a dependency for docs, but not for tests. I think you are right that we can add this as a separate issue though.

juanitorduz (changes requested):
thanks @drbenvincent ! Can we add the points from @bugbot?

I also did dome review with Cursor:

## PR #2479 Review — Potential Improvements

---

### 1. Bug / Correctness

**a) Legacy MMM silently breaks on `xr.DataArray`-valued `FixedScaling` for channels** [WE WILL REMOVE THE LEGACY MMM SOON SO THIS ONE IS OK]

The legacy `_compute_scales` dispatches on `isinstance(channel_scaling.value, dict)` vs else, where the else branch does `np.full(n_channels, channel_scaling.value)`. If someone passes an `xr.DataArray` as the channel fixed scale in the legacy MMM, this will produce an object array and crash downstream — but there's no init-time guard and no test covering it.

The multidimensional path handles all three value types, but the legacy `mmm.py` path only handles `dict` and scalar. Either reject `DataArray` at init for the legacy MMM (similar to the dict-target rejection) or support it properly.

**b) `FixedScaling` dict validation misses NaN values** [SIMILAR TO BUGBOT]

The DataArray path validates NaN, and the scalar path validates positivity, but the dict path only checks `val <= 0`. Since `float('nan') <= 0` evaluates to `False`, a dict like `{"US": float('nan')}` passes validation silently. Add `math.isnan(val)` to the dict branch.

**c) `_build_fixed_scale` uses bare `DataArray` instead of `xr.DataArray`**

In `multidimensional.py`, line 1794 uses the bare `DataArray` import from `from xarray import DataArray`, while the rest of the codebase (and `scaling.py` itself) uses `xr.DataArray`. This works but is inconsistent. More importantly, `scaling.py` is the canonical module for this logic but uses `xr.DataArray` everywhere — the multidimensional module diverges.

---

### 2. Design / API Improvements

**a) `_abstract_guard` is a non-standard and confusing pattern**

The abstract base class uses a dummy `_abstract_guard` method solely to prevent direct instantiation. This is an unusual pattern. Consider either:
- Making a semantically meaningful method abstract (e.g. `compute_scale` or `to_scale_value`), or
- Using `__init_subclass__` to enforce the contract, or
- Simply documenting that `VariableScaling` shouldn't be instantiated directly — the `ABC` + `abstractmethod` mechanism is correct, it's just that the abstract method should carry meaning.

pymc_marketing/mmm/scaling.py
```diff
@@ -47,27 +131,288 @@ def _validate_dims(self) -> Self:
...
+
+    @model_validator(mode="after")
+    def _validate_value(self) -> Self:
+        if isinstance(self.value, dict):
+            for key, val in self.value.items():
+                if val <= 0:
+                    raise ValueError(
+                        f"All fixed scaling values must be positive, "
+                        f"got {val} for key '{key}'."
+                    )
```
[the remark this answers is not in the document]
drbenvincent: Resolved in efb4ce53 — added `math.isnan(val)` check in the dict branch of `_validate_value`, consistent with the DataArray path.

pymc_marketing/mmm/mmm.py
```diff
@@ -450,24 +470,42 @@ def _compute_scale_for_data(
...
+        channel_scale: np.ndarray | float
+        if isinstance(channel_scaling, FixedScaling):
+            if isinstance(channel_scaling.value, dict):
+                channel_scale = np.array(
+                    [channel_scaling.value[c] for c in self.channel_columns],
+                    dtype=float,
+                )
+            else:
+                n_channels = len(self.channel_columns)
+                channel_scale = np.full(n_channels, channel_scaling.value)
```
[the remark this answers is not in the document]
drbenvincent: Resolved in 0ef4d083 — added explicit `isinstance(..., DataArray)` guards in both the channel and target branches of `_compute_scales`, raising `ValueError` directing users to `MultidimensionalMMM`.

pymc_marketing/mmm/multidimensional.py
```diff
@@ -1688,19 +1744,127 @@ def _compute_scales(self) -> None:
...
+        if np.isnan(np.asarray(aligned.values)).any():
+            raise ValueError(
+                "Fixed scaling DataArray produced NaNs after alignment — coordinates "
+                "likely do not match the data grid on a shared dimension."
+            )
+        if tuple(aligned.sizes.items()) != tuple(template.sizes.items()):
+            raise ValueError(
+                f"Fixed scaling DataArray has shape {dict(aligned.sizes)} after "
+                f"broadcast; expected {dict(template.sizes)} matching reduced data."
+            )
```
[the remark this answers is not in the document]
drbenvincent: Resolved in 6b8c9105 — changed `tuple(aligned.sizes.items()) != tuple(template.sizes.items())` to `dict(aligned.sizes) != dict(template.sizes)` for order-insensitive comparison.

pymc_marketing/mmm/scaling.py
```diff
@@ -47,27 +131,288 @@ def _validate_dims(self) -> Self:
...
+            if np.isnan(arr).any():
+                raise ValueError("Fixed scaling DataArray must not contain NaN values.")
+            if (arr <= 0).any():
+                raise ValueError(
+                    "All values in a fixed scaling DataArray must be positive."
+                )
+        elif isinstance(self.value, bool):
+            raise TypeError(
+                "FixedScaling.value does not accept bool; use a numeric scalar."
+            )
```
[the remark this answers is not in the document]
drbenvincent: Resolved in ff9a9654 — moved the bool rejection to a `@field_validator("value", mode="before")` that fires before pydantic coercion, and removed the dead `isinstance(self.value, bool)` branch from `_validate_value`.

drbenvincent:
Thanks for the thorough review @juanitorduz! Here's a summary of what was addressed:

**Already resolved** by cursor-bot fix commits:
- 1a) Legacy MMM DataArray → `0ef4d083` (also marked OK by you)
- 1b) Dict NaN validation → `efb4ce53`
- 4a/4c/4d) Tests for legacy DataArray, bool coercion, dict NaN

**New commits addressing your remaining items:**

| Commit | Items addressed |
|--------|----------------|
| `0cd6f511` | **1c)** Replace bare `DataArray` import with `xr.DataArray` in multidimensional.py |
| `3d540143` | **2a)** Replace `_abstract_guard` with meaningful `scaling_description()` abstract method |
| `7eac6588` | **2d)** `from_long_dataframe` duplicate-row guard; **3a)** remove dead `isinstance(val, VariableScaling)` check; **3b)** replace fragile `cast` with runtime `isinstance` |
| `b987ac35` | **2b)** Rename `_validate_fixed_scaling_keys` / `_deserialize_variable_scaling` to public names; **5)** fix docstring typo |

**Intentionally deferred:**
- **2c)** Making `method` an abstract property on `VariableScaling` — Pydantic v2 emits shadow warnings when a field overrides a parent property. The `scaling_description()` approach in `3d540143` gives the ABC a meaningful contract without the Pydantic friction.
- **3c)** `_align_fixed_scale_dataarray` using `data.max()` for shape — noted as not a blocker, deferred to a performance PR.
- **5)** Standalone `to_dict()` not emitting `__type__` — by design since `Scaling.to_dict()` wraps via `serialization.serialize()`.
- **5)** Consolidating `validate_fixed_scaling_keys` and `_build_fixed_scale_from_dict` — would change init-vs-build-time contract.

juanitorduz:
@bugbot Review

juanitorduz:
Thanks @drbenvincent I think we have two small ones and then its good to go :)

pymc_marketing/mmm/scaling.py
```diff
@@ -11,63 +11,423 @@
...
+                raise ValueError("Fixed scaling DataArray must not contain NaN values.")
+            if (arr <= 0).any():
+                raise ValueError(
+                    "All values in a fixed scaling DataArray must be positive."
+                )
+        elif isinstance(self.value, (int, float, np.floating, np.integer)):
+            if float(self.value) <= 0:
+                raise ValueError(
+                    f"Fixed scaling value must be positive, got {self.value}."
+                )
```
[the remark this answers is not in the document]
drbenvincent: Resolved in ac4eacf7 — added `math.isnan(fval)` check to the scalar branch of `_validate_value`, consistent with the dict and DataArray validation paths.

pymc_marketing/mmm/mmm.py
```diff
@@ -205,22 +211,36 @@ def __init__(
...
             if "channel" not in scaling:
-                scaling["channel"] = VariableScaling(method="max", dims=())
+                scaling["channel"] = DataDerivedScaling(method="max", dims=())
             if "target" not in scaling:
-                scaling["target"] = VariableScaling(method="max", dims=())
+                scaling["target"] = DataDerivedScaling(method="max", dims=())
+
+            for key in ("channel", "target"):
+                if isinstance(scaling[key], dict):
+                    scaling[key] = deserialize_variable_scaling(scaling[key])
```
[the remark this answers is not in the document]
drbenvincent: Resolved in 10ddc8dc — removed the redundant `for key in ("channel", "target")` deserialization loops from both `MMM.__init__` and `MultidimensionalMMM.__init__`. The `Scaling._coerce_dict_values` model validator already handles this when `Scaling(**scaling)` is called.

juanitorduz (approved):
thank you @drbenvincent

PR #2483, opened 2026-04-07
Fix `model.named_vars` changes after saving and loading a model

### LINKED ISSUE #2467: model.named_vars changes after saving and loading a model
It seems that after saving and loading a model some variables in original scale are lost. The following code

```
namem = "00_test_202604_v05.nc"
named_vars1 = mmm.model.named_vars.keys()
with tempfile.NamedTemporaryFile(suffix=".nc", delete=False) as tmpfile:
    tmp_path = tmpfile.name

mmm.save(tmp_path, engine="h5netcdf")

volume_path = f"{namem}"
copyfile(tmp_path, volume_path)

mmmX = MMM.load(volume_path)
named_vars2 = mmmX.model.named_vars.keys()
print(set(named_vars1).difference(named_vars2))
```

outputs

{'channel_contribution_original_scale', 'intercept_contribution_original_scale', 'y_original_scale'}

The variables are present in idata but code like

y_out_of_sample = mmmX.sample_posterior_predictive(
    x_test,
    var_names=[
        "channel_contribution_original_scale",
    ],
    extend_idata=False,
    include_last_observations=True,
    random_seed=123,
)

throws a

KeyError: 'channel_contribution_original_scale'

try:
  return self.named_vars[self.name_for(key)]
 except KeyError:

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/2467

## Summary

- After saving and loading a Multidimensional MMM, `_original_scale` variables (e.g. `channel_contribution_original_scale`, `intercept_contribution_original_scale`, `y_original_scale`) added via `add_original_scale_contribution_variable()` were lost from `model.named_vars`, causing a `KeyError` when passed as `var_names` to `sample_posterior_predictive`.
- Added `_restore_original_scale_variables()` to `build_from_idata()` which detects `_original_scale` variables present in the posterior but missing from the rebuilt model graph, and recreates them.

## Root Cause

In the Multidimensional MMM, `_original_scale` variables are created by the user calling `add_original_scale_contribution_variable()` **after** `build_model()`. During `load`, `build_from_idata()` only calls `build_model()`, which does not recreate these user-added variables. The posterior samples for them were persisted correctly in `idata`, but the model graph was missing the corresponding entries in `named_vars`.

The 1D `BaseMMM` is not affected because it creates `_original_scale` variables inside `build_model()` via `_add_original_scale_deterministics()`.

## Fix

After `build_model()` completes in `build_from_idata()`, a new `_restore_original_scale_variables()` method scans `idata.posterior` for variables ending in `_original_scale` that are:
1. Not already present in `model.named_vars` (e.g. `total_media_contribution_original_scale` is created inside `build_model` and is skipped)
2. Have a valid base variable in the model (validated via `add_original_scale_contribution_variable`)

This is backward-compatible: it works for models saved before the fix since it requires no changes to `save`.

## Test plan

- [x] Regression test `test_save_load_preserves_original_scale_variables` that creates an MMM with three `_original_scale` variables, saves, loads, and asserts they are all present in the loaded model's `named_vars`
- [x] Test fails without the fix, passes with it
- [x] All 191 existing tests in `test_multidimensional.py` continue to pass
- [x] Pre-commit hooks pass

----
📚 Documentation preview 📚: https://pymc-marketing--2483.org.readthedocs.build/en/2483/


juanitorduz:
@carlosagostini @williambdean friendly ping :)

juanitorduz:
This was solved in https://github.com/pymc-labs/pymc-marketing/pull/2508 

fyi @PabloRoque @carlosagostini

PR #2487, opened 2026-04-09
Remove old MMM Class

juanitorduz:
Remove `mmm.MMM` in favor of `multidimensional.MMM`. 

In a later PR we will rename `multidimensional` -> `mmm`.

----
📚 Documentation preview 📚: https://pymc-marketing--2487.org.readthedocs.build/en/2487/


juanitorduz:
@bugbot review

williambdean (commented):
Isn't `pymc_marketing.mmm.MMM` now just the `multidimensional.MMM`? The notebooks are explicit about this import.

I'm in favor of*: 

```python
```

\* For public facing API docs

.claude/skills/code-best-practice/SKILL.md
williambdean, line 1: Can we consllidate these or sym link? For another PR

juanitorduz: I can do this here.

juanitorduz:
> ```python
> from pymc_marketing.mmm import MMM
> ```

I do not think we have this yet, but we will do it for sure. I wanna do it in a next PR as this one is already big enough :)

.claude/skills/code-best-practice/SKILL.md
juanitorduz: Actually, let me do it in the next PR as I want to have this as minimal as possible 🙏

juanitorduz:
@carlosagostini @williambdean friendly ping :)

cetagostini:
> Write by cetagostini-clawbot

## Deep Review — Hardening & Test-Coverage Assessment

I went through the whole diff (~9.2k LOC of source/test changes, plus ~21k LOC of `.po` locale files), inspected every removed method against the new `multidimensional.MMM`, ran the modified budget-optimizer tests, and probed a few behaviors live to confirm. Below is a prioritized list of what could be hardened, plus concrete tests to add.

---

### 1. Real bugs / footguns left behind

#### 1.1 `BudgetOptimizer` default `response_variable="total_contribution"` is now broken
**Severity: high** (silent footgun for external users)

The new multidim `MMM.build_model` produces `total_media_contribution` and `total_media_contribution_original_scale`, **not** `total_contribution`. Confirmed live:

```
False False True   # total_contribution / total_media_contribution / total_media_contribution_original_scale
```

But `BudgetOptimizer` still defaults to `response_variable=\"total_contribution\"`:

```python
# pymc_marketing/mmm/budget_optimizer.py L648-649
response_variable: str = Field(
    default=\"total_contribution\",
```

The module docstring also contradicts itself — line 90 uses `total_media_contribution_original_scale` in the example, line 201 says the default is `total_contribution`.

I reproduced the failure: `BudgetOptimizer(model=wrapper, num_periods=4).allocate_budget(10.0)` → `KeyError: 'total_contribution'`.

`MultiDimensionalBudgetOptimizerWrapper.optimize_budget` masks this because it overrides the default to `total_media_contribution_original_scale`, but anyone instantiating `BudgetOptimizer` directly (which is the documented \"custom pymc model\" workflow in the docstring) hits the wall.

**Fix options, pick one:**
1. Change the default to `\"total_media_contribution_original_scale\"` and update the docstring.
2. Keep the default, but raise a clear, actionable error early in `BudgetOptimizer.__init__` if the response variable is not in `idata.posterior`, listing what is available.

**Test to add** in `tests/mmm/test_budget_optimizer.py`:

```python
def test_budget_optimizer_clear_error_on_missing_response_variable(mmm_wrapper):
    \"\"\"Default response_variable must either work out-of-the-box on the new MMM,
    or raise an actionable error naming the missing key and listing alternatives.\"\"\"
    with pytest.raises((KeyError, ValueError)) as exc_info:
        BudgetOptimizer(model=mmm_wrapper, num_periods=4).allocate_budget(10.0)
    msg = str(exc_info.value)
    assert \"total_media_contribution_original_scale\" in msg or \\
           \"response_variable\" in msg
```

#### 1.2 Broken Sphinx cross-references in published notebooks
**Severity: medium** (visible in the docs build)

`pymc_marketing.mmm.mmm` and `pymc_marketing.mmm.base` no longer exist as modules, but several published notebooks still cite them:

- `docs/source/notebooks/general/prior_predictive.ipynb` — 3 refs
- `docs/source/notebooks/mmm/mmm_case_study.ipynb` — 1 ref (and references the removed method `allocate_budget_to_maximize_response`)
- `docs/source/notebooks/mmm/mmm_example.ipynb` — 8 refs
- `docs/source/notebooks/mmm/dev/mmm_example_new.ipynb` — 8 refs

These are `` {class}\`MMM <pymc_marketing.mmm.mmm.MMM>\` ``-style and produce broken cross-refs in the rendered docs. **Minimum fix**: replace `pymc_marketing.mmm.mmm.MMM` → `pymc_marketing.mmm.MMM` (or `pymc_marketing.mmm.multidimensional.MMM`) and remove the `allocate_budget_to_maximize_response` ref.

#### 1.3 `MultiDimensionalBudgetOptimizerWrapper` is not exported from `pymc_marketing.mmm`
It's the only documented way to do budget optimization on the new MMM, but `pymc_marketing/mmm/__init__.py` doesn't re-export it. Users must do `from pymc_marketing.mmm.multidimensional import MultiDimensionalBudgetOptimizerWrapper`. Add it to `__all__`.

#### 1.4 Pre-commit hook versions silently downgraded

```diff
# .pre-commit-config.yaml L12-25
  - repo: https://github.com/astral-sh/ruff-pre-commit
-   rev: v0.15.12
+   rev: v0.15.11
  - repo: https://github.com/pre-commit/mirrors-mypy
-   rev: v1.20.2
+   rev: v1.20.1
```

This looks like an accidental backwards merge resolution (`v0.15.12` / `v1.20.2` are on `main`). Should be re-bumped before merge.

---

### 2. Test coverage that disappeared and should be ported

The deleted `tests/mmm/test_mmm.py` had a `TestMMMEdgeCases` block (~30 tests) and a `TestMMMHelperMethods` block. Some of these check behaviors that the multidim MMM also has but where I could not find an equivalent test in `tests/mmm/test_multidimensional.py`. These are concrete suggestions:

| Old test in `test_mmm.py` | Status on multidim | Suggested action |
|---|---|---|
| `test_single_row_dataframe` | not covered | Add — the model should fail clearly, not silently |
| `test_all_zeros_in_channel` | not covered | Add — scaling by `max=0` is the classic divide-by-zero |
| `test_negative_values_in_channels` | not covered | Add — old MMM warned, what does new do? |
| `test_nan_values_in_input` | not covered | Add |
| `test_infinite_values_in_input` | not covered | Add |
| `test_duplicate_channel_names` | not covered | Add |
| `test_unsorted_dates` / `test_duplicate_dates` | partially via `_validate_dims_in_dataframe` | Add explicit tests |
| `test_mismatched_X_y_lengths` | not covered | Add |
| `test_target_with_zeros` / `test_target_all_zeros` | not covered | Add — `method=\"max\"` with zero target |
| `test_data_setter_with_different_length` | not covered | Add — high value, common save/load case |
| `test_invalid_scaling_method` | not covered | Add |
| `test_transformer_fitting_failure` | not covered | Add |
| `test_get_errors_bad_y_length` | partially in `TestResidualsOverTime` | Verify and reuse in new style |
| `test_get_group_predictive_data` | not directly covered | Add — pre/post fit error paths |

These edge cases were the strongest part of the old test suite (real engineering hardening), and dropping them all in one PR is risky. Even if behaviors slightly differ between old and new, it's safer to port them than rely on coverage by accident.

**Suggested first batch (highest ROI)**, all using `simple_mmm_data` from `tests/mmm/conftest.py`:

```python
class TestMultidimMMMEdgeCases:
    def test_single_row_dataframe(self):
        ...

    @pytest.mark.parametrize(\"zero_columns\", [[\"channel_1\"], [\"channel_1\", \"channel_2\"]])
    def test_all_zeros_in_channel_does_not_silently_divide_by_zero(self, simple_mmm_data, zero_columns):
        X = simple_mmm_data[\"X\"].copy()
        for c in zero_columns:
            X[c] = 0.0
        # default scaling=\"max\" should either succeed with finite scales or raise a clear error
        ...

    def test_target_all_zeros_with_max_scaling(self, simple_mmm_data):
        # Document and lock the behavior. Old MMM had MaxAbs which handled this; new MMM uses max.
        ...

    def test_nan_in_channel_raises_clear_error(self, simple_mmm_data):
        ...

    def test_inf_in_target_raises_clear_error(self, simple_mmm_data):
        ...

    def test_mismatched_X_y_lengths_raises_clear_error(self, simple_mmm_data):
        ...

    def test_duplicate_channel_columns_raises(self):
        ...

    def test_data_setter_handles_different_length_X(self, simple_fitted_mmm):
        # Critical: predict on shorter/longer windows
        ...

    def test_unsorted_dates_raises_or_sorts(self, simple_mmm_data):
        ...

    def test_duplicate_dates_raises(self, simple_mmm_data):
        ...
```

#### `test_plotting.py` parity (deleted plot tests)

The deleted `TestBasePlotting.test_plots` was a parametric matrix over `plot_*` methods × kwargs. The new `MMMPlotSuite` is well-tested for `posterior_predictive`, `residuals_*`, `waterfall_*`, `channel_parameter`, `prior_vs_posterior`, `posterior_distribution`, and `channel_contribution_share_hdi`. What I could **not** find in `tests/mmm/test_plot.py`:

- `prior_predictive` with combinations of `add_mean`, `add_gradient`, `original_scale`, `hdi_list`. The old test had ~20 combos; the new test seems to cover the basic call only.
- Error paths on `plot.prior_predictive()` / `plot.posterior_predictive()` when `prior_predictive` / `posterior_predictive` group is missing in `idata`. Old MMM had `test_plot_prior_predictive_no_fitted` and `test_plot_posterior_predictive_no_fitted`.

**Suggested test:**

```python
class TestPriorPredictivePlot:
    @pytest.mark.parametrize(\"kwargs\", [
        {},
        {\"original_scale\": True},
        {\"add_mean\": True, \"add_gradient\": True},
        {\"hdi_list\": []},
        {\"hdi_list\": [0.94, 0.5]},
    ])
    def test_prior_predictive_kwargs_matrix(self, mock_suite_with_prior_predictive, kwargs):
        fig = mock_suite_with_prior_predictive.prior_predictive(**kwargs)
        assert isinstance(fig, plt.Figure)

    def test_prior_predictive_raises_when_not_sampled(self, mock_suite):
        with pytest.raises(RuntimeError, match=r\"prior_predictive\"):
            mock_suite.prior_predictive()
```

---

### 3. Behavioral changes worth pinning with a regression test

#### 3.1 Budget allocation values changed in `test_allocate_budget`

```
{\"channel_1\": 54.78357587906867, \"channel_2\": 45.21642412093133}, 48.8     # old
{\"channel_1\": 58.97600120944057, \"channel_2\": 41.02399879055943}, 44.94    # new
```

This is **expected** — the old test optimized `total_contribution` (scaled, includes intercept), the new one optimizes `total_media_contribution_original_scale` (original scale, media-only) — but the change is non-obvious. A short docstring on `test_allocate_budget` explaining why the expected values differ from the old MMM would prevent confusion in future debugging. (When I dug in, I had to verify by inspection that this isn't a bug in the optimizer.)

#### 3.2 Default scaling: `MaxAbsScaler` → `method=\"max\"`

Old MMM used `MaxAbsScaler` (divides by `max(|x|)`), new MMM uses `DataDerivedScaling(method=\"max\")` (divides by `max(x)`). These are equivalent for non-negative data (the typical MMM case) but **differ if any channel/target ever goes negative**. Worth a regression test:

```python
def test_default_scaling_with_mixed_sign_target_is_documented(simple_mmm_data):
    \"\"\"Lock in the behavior of method='max' when target has negative values.
    Old MMM used MaxAbsScaler; new MMM uses max. Same for non-negative data,
    differs when negative values exist.\"\"\"
    ...
```

---

### 4. Dead code / API surface to clean up later (not blocking)

`pymc_marketing/mmm/__init__.py` still re-exports `validating` and `preprocessing` (and their decorator helpers), but the new multidim MMM does not use `MMMModelBuilder.validate` / `.preprocess` / the `_tags` system at all — it has its own `_validate_dims_in_*`, `_validate_metrics`, and `Scaling`. These modules are dead code from this PR's perspective:

- `pymc_marketing/mmm/preprocessing.py` (`MaxAbsScaleChannels`, `MaxAbsScaleTarget`, `StandardizeControls`, decorators)
- `pymc_marketing/mmm/validating.py` (`Validate*` mixins, decorators)

Removing them is a breaking change (anyone importing them externally), so it's reasonable to keep them for the `0.20.0` cycle. But it's worth either:
- Adding a `DeprecationWarning` on import, **OR**
- Filing a follow-up issue to remove them in `0.21`.

---

### 5. Quick summary of \"are we taking something out which will bring bugs?\"

| Removed thing | Replacement | Risk |
|---|---|---|
| `MMM.optimize_budget()` | `MultiDimensionalBudgetOptimizerWrapper.optimize_budget()` | **Low** — covered by deprecation in old class, wrapper tested |
| `plot_*` methods directly on `MMM` | `mmm.plot.*` (`MMMPlotSuite`) | **Low** — well covered |
| `compute_channel_contribution_original_scale` | `add_original_scale_contribution_variable` + `idata` access | **Medium** — no equivalent helper, requires user code |
| `get_channel_contribution_share_samples` (data) | `mmm.plot.channel_contribution_share_hdi()` returns plot, not data | **Medium** — no public data accessor; users who pulled the array out lose that |
| `new_spend_contributions` / `plot_new_spend_contributions` | `mmm.sample_response_distribution()` (different API) | **Medium** — not 1-to-1, migration guide doesn't mention |
| `format_recovered_transformation_parameters` | nothing | **Low** — niche |
| `MMM.fit_result`, `MMM.prior_predictive` access patterns | unchanged | None |
| `BaseValidateMMM`, `MMMModelBuilder` (used as base classes by external code) | nothing | **Breaking** — should be called out in the changelog/migration guide |

The 3 medium-risk items above are not blockers, but the migration guide should explicitly call them out so external users don't silently lose functionality.

---

### Minimum recommended action list before merging

1. **Fix the `BudgetOptimizer` default** (sec 1.1) — either change the default or raise a clear error. Add the test in 1.1.
2. **Fix the broken Sphinx refs** (sec 1.2) — find/replace `pymc_marketing.mmm.mmm.MMM` and `pymc_marketing.mmm.base.*` across the 4 affected notebooks.
3. **Re-bump pre-commit versions** (sec 1.4).
4. **Export `MultiDimensionalBudgetOptimizerWrapper`** from `pymc_marketing.mmm.__init__`.
5. Port at least the **top-5 edge case tests** from sec 2 (zero-channel, NaN, mismatched lengths, single row, target-all-zeros). These are quick wins on robustness and prevent silent regressions.
6. Add a one-line docstring on the changed `test_allocate_budget` expected values (sec 3.1) explaining the response-variable semantic shift.
7. Extend the migration guide with the methods called out in section 5 (medium-risk removals) so external users see them.

Items 1–4 are tightly scoped fixes. Items 5–7 can be done in a follow-up PR if this one is already big, but I'd put #1, #2, #3, #4 inside this PR.

Happy to start with #1+#2+#4 (the smallest, most clearly correct fixes) plus the edge-case tests in #5 if that's useful.

juanitorduz:
@bugbot review . Especially in view of https://github.com/pymc-labs/pymc-marketing/pull/2487#issuecomment-4345954701

juanitorduz:
@carlosagostini would you mind giving another review 🙏

cetagostini:
## Review notes — refined after independent verification

> _AI-assisted review (Claude Opus). Findings were dispatched to independent critic agents for verification before posting; refuted/overstated items were dropped. Treat each item as a starting point for human review, not a final verdict._

Eight items worth addressing before merge, ordered by severity:

---

### 1. Silent NaN propagation when target is all-zero (High)

`pymc_marketing/mmm/multidimensional.py:2132`

The channel path guards against `0/0` via `pmd.math.switch(pmd.math.isnan(...), 0.0, ...)` (lines 2127–2129). The target path doesn't:

```python
target_data_scaled = _target / _target_scale  # no guard
```

When the target is all-zero, `_target_scale` is `0` and the likelihood receives `NaN`. The new test `test_target_all_zeros_with_max_scaling` only asserts `scaler == 0` and never exercises the model graph, so this passes silently.

**Fix:** mirror the channel-side switch, or fall back to `_target_scale = 1.0` when the reduction is zero. Then update the test to assert `target_data_scaled.eval()` is finite.

---

### 2. `mlflow.py` docstrings reference a removed method (Medium)

`pymc_marketing/mlflow.py:114`, `:867`, `:1161`

All three sit inside `.. code-block:: python` blocks within the module docstring, `log_mmm`, and `autolog`. Sphinx renders these as user-facing examples; `plot_components_contributions` no longer exists on `MMM`, so copy-pasting yields `AttributeError`.

**Fix:** replace each `mmm.plot_components_contributions()` with `mmm.plot.contributions_over_time()`.

---

### 3. Migration guide misroutes `sample_response_distribution` (Medium)

`docs/source/notebooks/mmm/mmm_migration_guide.ipynb` — "Methods Removed Without 1-to-1 Replacement" section.

> `new_spend_contributions` / `plot_new_spend_contributions` -> use `mmm.sample_response_distribution(...)`

`def sample_response_distribution` is defined on `MultiDimensionalBudgetOptimizerWrapper` (`multidimensional.py:3886`), not on `MMM`. Calling it on a bare MMM `AttributeError`s.

**Fix:** rewrite the bullet to show the wrapper construction:
```python
optimizable_model = MultiDimensionalBudgetOptimizerWrapper(model=mmm, start_date=..., end_date=...)
optimizable_model.sample_response_distribution(...)
```

---

### 4. Migration guide silently drops two old methods (Medium)

The "Methods Removed" section doesn't mention:
- `get_ts_contribution_posterior` — no method exists, but `mmm.idata.posterior[\"channel_contribution\"]` (or `channel_contribution_original_scale` after `add_original_scale_contribution_variable`) exposes the same `(chain, draw, date, channel, *dims)` array.
- `plot_grouped_contribution_breakdown_over_time` — no built-in stacked-area + group-merging plot exists. `plot.contributions_over_time` is a line plot; `plot.waterfall_components_decomposition` is a single-snapshot bar.

**Fix:** add both to the "Methods Removed Without 1-to-1 Replacement" list. For `plot_grouped_…`, document the manual stacking workaround using `summary.contributions()` or direct posterior indexing.

---

### 5. New test has a dead error-matcher (Medium)

`tests/mmm/test_multidimensional.py:226-235`

```python
with pytest.raises(
    (TypeError, ValueError),
    match=r\"y length must match X|cannot concatenate object\",
):
    mmm.build_model(X, y)
```

The implementation deterministically raises `ValueError(\"y length must match X when passed as ndarray\")` at `multidimensional.py:3309`. The `TypeError` slot and the `|cannot concatenate object` alternative are unreachable — the test would silently pass even if the clear-error path regressed to a leaky pandas error.

**Fix:** tighten to a single exception/regex:
```python
with pytest.raises(ValueError, match=r\"y length must match X\"):
    mmm.build_model(X, y)
```

---

### 6. No `prior_predictive` plot test (Medium)

`tests/mmm/test_plot.py` has a single smoke test for `posterior_predictive` (lines 237–246) and **none** for `prior_predictive`. The old `test_plotting.py` covered ~50 parametrizations across `add_mean`, `add_gradient`, `original_scale`, and varying `hdi_list`. Most of that combinatorial coverage didn't migrate.

**Fix:** at minimum, add a parametrized smoke test for `mmm.plot.prior_predictive(...)` mirroring the `posterior_predictive` smoke test, with one or two `hdi_prob` values and an `original_scale` toggle. Full restoration of the old matrix is optional but recommended.

---

### 7. Misleading test name on the all-zero-channel test (Low)

`tests/mmm/test_multidimensional.py:187-202` — `test_all_zeros_in_channel_does_not_silently_divide_by_zero`

The name promises no silent divide-by-zero; the body only asserts `scale == 0` and `np.isfinite(channel_scales).all()`. It never builds beyond `build_model` and never checks the scaled tensor. The `pmd.math.switch(isnan, 0.0, …)` only guards `NaN`, not `Inf` — so a non-zero numerator over a zero scale (which can happen under multidim `scaling.dims` with one slice all-zero and another not) would propagate `Inf`.

**Fix:** either rename to `test_all_zeros_channel_yields_zero_scale` (matching what's actually asserted) or extend the test to evaluate `channel_data_scaled` and assert finite values, plus add a parametrized case with `scaling.dims=(\"country\",)` to exercise the heterogeneous-slice path.

---

### 8. No coverage for infinite values in input (Low)

The new edge-case class covers `NaN` in channels but not `+inf` / `-inf`. The old `test_mmm.py::test_infinite_values_in_input` is gone.

**Fix:** add a test asserting either a clear `ValueError` or finite scalers when `X` contains `np.inf`.

---

### Items I dropped after verification

For transparency: the first pass also flagged a `BudgetOptimizer.response_variable` default change as a breaking change (it's purely naming alignment — the new MMM never registered `total_contribution`), claimed `compute_channel_contribution_original_scale` had no replacement (`add_original_scale_contribution_variable` covers it), and called out a list of \"missing\" tests that turned out to be migrated to `test_validating.py` or `test_plot.py`. Those were not real issues and aren't included above.

juanitorduz:
@bugbot review

juanitorduz:
@carlosagostini , the final changes have been addressed 🙏

cetagostini:
Hey, I just finished another round of review and I think we are ready for merge!

cetagostini (approved):
Let's roll baby!

PR #2512, opened 2026-04-24
Add Zarr save/load support to ModelIO (zarr>=3 shim)

### LINKED ISSUE #2511: `zarr` support is blocked
Zarr support comes handy on cloud environments when dealing with moderate/large inference objects. 

Right now, Zarr support is blocked:

* `pymc-marketing` pins `arviz<1.0.0`:
* `arviz<1.0.0` does not support `zarr>=3.0`: https://github.com/arviz-devs/arviz/blob/v0.23.4/arviz/data/inference_data.py#L817.
* current version of nutpie requires `zarr>=3.0`

As a consequence:
```
def test_save_load_zarr_roundtrip(fit_mmm: MMM, tmp_path):
   
    store = tmp_path / "model.zarr"
    fit_mmm.idata.to_zarr(str(store))
    idata_loaded = az.from_zarr(str(store))
    loaded = MMM.load_from_idata(idata_loaded)
    assert isinstance(loaded, MMM)
    assert loaded == fit_mmm
```

> test_save_load_zarr_roundtrip - ImportError: Found zarr>=3, which is not supported by ArviZ. Instead, you can use 'dt = InferenceData.to_datatree' followed by 'dt.to_zarr()' (needs xarray>=2024.11.0)

I think we can unblock it by moving to `arviz>=1.0`, so linked to #2165

PabloRoque:
## Summary

- `arviz<1.0` `InferenceData.to_zarr()` / `from_zarr()` hard-error on `zarr>=3`, which breaks save/load when `nutpie` pulls in `zarr>=3` as a dependency.
- This PR shims the gap using the same pattern `arviz>=1.0` adopted internally:
  - `idata_to_zarr` → `idata.to_datatree().to_zarr(store)`
  - `idata_from_zarr` → `az.from_datatree(xr.open_datatree(store, engine="zarr"))`
- `ModelIO.save()` / `load()` dispatch on `.zarr` extension; all existing `.nc` paths are unchanged.
- I/O helpers live in `pymc_marketing/data/idata/utils.py`; `pymc_marketing/utils.py` re-exports them for backward compatibility.
- When the `arviz>=1.0` migration happens, this file is the only one to touch.

## Testing

Added `test_save_load_zarr_roundtrip` to `tests/mmm/test_multidimensional.py` which exercises the full `MMM.save("model.zarr")` / `MMM.load("model.zarr")` roundtrip.

Closes #2511 

----
📚 Documentation preview 📚: https://pymc-marketing--2512.org.readthedocs.build/en/2512/


ColtAllen (commented):
Thanks! I've been wanting to add Zarr support for model persistence.

Can we add some TODO and/or FIXME comments highlighting that these changes are temporary until `arviz>=1.0`

PabloRoque:
@ColtAllen, I added a couple of TODOs to replace the functionality once Arviz>=1.0.0 lands.

pymc_marketing/data/idata/utils.py
```diff
@@ -26,6 +29,70 @@
...
+    Returns
+    -------
+    az.InferenceData
+    """
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            category=UserWarning,
+            message=r"fit_data group is not defined in the InferenceData scheme",
+        )
```
williambdean: I think this is completely stale now. There was change in arviz to allow other group I thought

PabloRoque: It is stale indeed. 

I would argue that we don't need this definition at all, and we can just do `az.from_netcdf` where this function is called.

williambdean: For it!

PabloRoque: Removed. 

Thanks for the nudge: I also found out we were not supporting a `groups` parameter on `idata_to_zarr` the way `to_netcdf` had. I added it to keep the two backends consistent, preserving `idata.attrs` on the filtered payload so model metadata survives a partial save.

juanitorduz (approved):

Issue #2513, opened 2026-04-24
Max-diff modelling


NathanielF:
The Louviere Sequential Best-Worst (SBW) model is the industry standard for analyzing MaxDiff (Best-Worst Scaling) data. Developed by Jordan Louviere, it treats a single MaxDiff task as a sequence of two discrete choices.

When a respondent looks at a set of items and picks the "Best" and "Worst," the SBW model assumes they follow a specific cognitive path. First choose Best of a set of options, then choose the worst. It's a type of discrete choice model often applied in UX research where agents are asked to choose between goods displayed in sets of 3+ and rank most preferred and least preferred. I think it's pretty straightforward to implement for pymc marketing.

PR #2514, opened 2026-04-25
Add Maxdiff methods to Consumer Choice

NathanielF:
## Description
Add MaxDiff (Best-Worst Scaling) hierarchical Bayesian model

Overview
Implements the Louviere sequential best-worst model for MaxDiff surveys, supporting both item-intercept and part-worth (conjoint) formulations with optional respondent heterogeneity.

New Features
MaxDiffMixedLogit model class with two modes:

- Item-intercept mode: Free utility per item with optional per-respondent random intercepts (HB-MaxDiff)

- Part-worths mode: Utilities as linear combinations of item attributes via patsy formulas, with optional random coefficients on specified features

- prepare_maxdiff_data: Reshapes long-format survey data into padded arrays for the likelihood, with input validation

- generate_maxdiff_data / generate_maxdiff_conjoint_data: Synthetic data generators for testing and parameter recovery

Counterfactual API: apply_intervention for assortment changes and score_new_items for extrapolating to new SKUs (part-worths mode)

- Batched prediction: predict_choices with optional draw_batch_size for memory-efficient counterfactual simulation

Documentation
Comprehensive notebook covering:

- Model specification and identification strategy

- Synthetic data generation and parameter recovery

- Respondent heterogeneity visualization

- Share-of-preference calculations

- Counterfactual assortment analysis

Part-worths mode with attribute importance and new-item extrapolation

## Related Issue
https://github.com/pymc-labs/pymc-marketing/issues/2513
- [X] Closes # https://github.com/pymc-labs/pymc-marketing/issues/2513
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2514.org.readthedocs.build/en/2514/


NathanielF:
## Design Note: Prediction vs Posterior Predictive

predict_choices (used by apply_intervention) bypasses PyMC's posterior predictive sampler. It manually recomputes per-task utilities from the posterior, then samples best → worst sequentially using pre-drawn uniforms for reproducibility. This ensures the worst pick is conditional on the freshly sampled best pick — the correct generative joint distribution for counterfactual simulation.

This split exists because PyMC's Categorical likelihood only observes a single outcome per task, but the MaxDiff generative process requires sampling from two linked categorical distributions where the second depends on the first's realized value.

juanitorduz:
Hey @NathanielF ! thank you for this one! I first ran a Claude review :)

I will look into the notebook next week :) 

## Detailed review notes

Claude checked out the PR locally, ran `ruff check` (clean), and ran the non-slow test suite (`tests/customer_choice/test_maxdiff.py -k "not Recovery"`) → **84 passed in ~93s** on Python 3.13.12 in `pymc-marketing-dev`. Comments are organized by importance.

---

### Blocking

#### B1. `prepare_maxdiff_data` silently corrupts on duplicated `items`

`pymc_marketing/customer_choice/maxdiff.py:173`

```python
item_to_idx = {item: i for i, item in enumerate(items)}
```

If a user passes `items=["a", "b", "a"]` (an easy mistake when merging from multiple sources) the dict silently collapses both `"a"` entries into one index, producing wrong `n_items`, mis-padded arrays, and a model whose coords don't match the data. There is no validation. Add up front:

```python
if len(set(items)) != len(items):
    seen, dupes = set(), []
    for x in items:
        if x in seen:
            dupes.append(x)
        seen.add(x)
    raise ValueError(f"items must be unique; duplicates: {sorted(set(dupes))}")
```

Mirror this in `MaxDiffMixedLogit.__init__`.

#### B2. `sample_posterior_predictive` ignores coord changes when `task_df` shape changes

`pymc_marketing/customer_choice/maxdiff.py:1183-1231`

When a different `task_df` is passed:

```python
arrays = self.preprocess_model_data(task_df)
with self.model:
    pm.set_data({
        "item_idx": arrays["item_idx"],
        "mask": arrays["mask"],
        ...
    })
```

If `n_tasks` or `k_max` differ from the trained shapes, the `tasks` and `positions` coords on the existing model are stale and the posterior predictive is silently mis-coordinated (and may shape-error). At minimum, raise if `arrays["k_max"] != self.arrays["k_max"]` (same for `n_tasks`) or pass `coords={...}` to `pm.set_data`. Document the constraint either way. The `apply_intervention` path goes through `predict_choices` (no `set_data`), so that path is fine.

#### B3. `_ensure_best_worst_flags` warning text is misleading after upstream sorts

`pymc_marketing/customer_choice/maxdiff.py:541-596`

The auto-dummy-flag logic is the right call, but the warning says

> "first row = best, last row = worst"

`groupby(... sort=False).cumcount()` returns `0..k-1` *in row order as it appears in `df`*. If the user did `df.sort_values("item_id")` upstream, "first row" depends on the sort key, so the message is misleading. Either drop the "first row = best" detail (flags are unused anyway) or clarify that the assignment is arbitrary and only present to satisfy the prepare-data validator.

---

### Should fix before merge

#### S1. `pop_shift` not applied to new-respondent draws in part-worths mode

`pymc_marketing/customer_choice/maxdiff.py:992-1027` and `1478-1498`

In `_build_partworths_utility` the model centers utilities by subtracting `pop_shift = pt.mean(U_item_pop_raw)`:

```python
U_item_pop = pm.Deterministic("U_item_pop", U_item_pop_raw - pop_shift, ...)
U_item_r   = pm.Deterministic("U_item_r",  pt.dot(beta_full_r, X_data.T) - pop_shift, ...)
```

But `_draw_new_respondent_utilities` for part-worths returns `beta_full @ self.X.T` — *not* shifted. Each respondent's softmax is shift-invariant within a single task, so this is operationally fine for `predict_choices` (each row is self-consistent), but:

- The mixed tensor `pred_beta_r` then has known-respondent rows centered and unknown-respondent rows uncentered. Anyone inspecting `pred_beta_r` directly, or computing cross-row statistics (e.g. differences between known and population draws), gets silently wrong numbers.
- The class docstring says new-respondent draws come from "the fitted population distribution" — readers expect them on the same scale as `U_item_pop`.

Apply the same shift, e.g. store `pop_shift` as a Deterministic and pull it from `posterior`. Add a regression test asserting `mean(_draw_new_respondent_utilities(...))` matches `mean(U_item_r)` per draw.

#### S2. `intervention_idata` is `xr.Dataset` here but `az.InferenceData` everywhere else

In `mnl_logit`, `nested_logit`, `mixed_logit`, `consideration_set_logit`, `self.intervention_idata` is an `az.InferenceData`. In MaxDiff (`maxdiff.py:656,738`) it's an `xr.Dataset`. Downstream tooling that type-checks (e.g. `idata.posterior_predictive`) breaks unexpectedly when switching modules.

Either (a) wrap the result as `az.InferenceData(posterior_predictive=result)` in `apply_intervention` / `score_new_items`, or (b) rename the attribute to `intervention_dataset` and document the deviation. (a) is preferable.

#### S3. `predict_choices` upfront `pred_beta_r` allocation undermines `draw_batch_size`

`pymc_marketing/customer_choice/maxdiff.py:1437-1460`

Docstring claim: "Lowers peak memory linearly". But `pred_beta_r` of shape `(C, D, n_pred, I)` is allocated *before* batching. For a mid-sized study (4 chains × 2000 draws × 1000 respondents × 50 items, float64) that's already ~3 GB. Batching only reduces the per-task tensor.

Two reasonable fixes:

- Move `pred_beta_r` construction inside the batch loop, driven by a slice of the posterior + a slice of new uniforms.
- Or document the footprint clearly: "`draw_batch_size` chunks the per-task softmax `(C, D_b, T, K_max)` but the per-respondent utility tensor is allocated in full".

#### S4. `__init__` validation order

`pymc_marketing/customer_choice/maxdiff.py:418-471`

`self.reference_item = reference_item if reference_item is not None else items[-1]` runs before the "in pool" check. In part-worths mode you set `self.random_intercepts = False`, which is then re-asserted by `full_covariance and not self.random_intercepts → raise`. The error chain reads "use random_intercepts=True" even though the user *did* and only `_is_partworths` flipped it off. Reorder:

1. validate `(item_attributes, utility_formula)` consistency,
2. set `self._is_partworths`,
3. validate `full_covariance` against the *user-supplied* `random_intercepts` (not the internal-rewritten one) and against `_is_partworths`,
4. validate `reference_item`.

Also, `reference_item` is checked twice (here + `prepare_maxdiff_data`) — fine, but in part-worths mode it's not actually used, so the check is a footgun. Either skip it in part-worths mode or document.

#### S5. `_create_fit_data` round-trips integer `task_id` as objects

`pymc_marketing/customer_choice/maxdiff.py:1096-1100`

`self.task_df.reset_index(drop=True).to_xarray()` preserves dtypes inside the Dataset, but reload via `idata["fit_data"].to_dataframe().reset_index(drop=True)` can return mixed-type columns as `object`. `prepare_maxdiff_data` is robust to this through `groupby`, but `respondent_to_idx` keys on whatever dtype was restored. Extend `TestSaveLoadRoundtrip` to assert e.g. `loaded.task_df["task_id"].dtype == task_df["task_id"].dtype`.

#### S6. No parameter-recovery test for `full_covariance=True`

`TestParametricRecovery` covers diagonal random intercepts; `TestPartWorthsRecovery` covers part-worths. There's no slow recovery test that fits with `full_covariance=True` and asserts the posterior `corr_matrix` mean is close to `ground_truth["item_correlation"]`. `generate_maxdiff_data` now supports `item_correlation` and the docstring there explicitly invites this validation. Please add a `@pytest.mark.slow` test under `TestFullCovariance`:

```python
def test_correlation_matrix_recovered():
    corr_true = np.array([[1, .8, 0], [.8, 1, 0], [0, 0, 1]])
    task_df, gt = generate_maxdiff_data(
        n_respondents=300, n_items=3, subset_size=2,
        sigma_respondent=1.0, item_correlation=corr_true, random_seed=7,
    )
    model = MaxDiffMixedLogit(task_df, items=gt["items"], full_covariance=True)
    model.fit(...)
    corr_post = model.idata["posterior"]["corr_matrix"].mean(("chain", "draw")).values
    assert np.linalg.norm(corr_post - corr_true) < 0.2
```

---

### Suggestions

#### G1. Note that `_softmax_stable` is intentionally numpy

In `_predict_choices_batch` the helper `_softmax_stable` is reimplemented (not `pm.math.softmax`) because `pred_beta_r` is already realized as numpy values. A one-line comment would save the next reader 30 seconds.

#### G2. `_sample_categorical_from_u` precondition is implicit

`pymc_marketing/customer_choice/maxdiff.py:56-66`

Relies on `cumulative[..., -1] == 1.0` to guarantee a valid index. Defensive code is cheap:

```python
return np.clip(np.sum(cumulative < u[..., None], axis=-1), 0, p.shape[-1] - 1)
```

Avoids out-of-bounds indices in numerical-precision corner cases.

#### G3. `MaxDiffArrays` TypedDict vs `dataclass`

`MaxDiffArrays` is a `TypedDict` of 11 fields used as `arrays["item_idx"]` everywhere. A `@dataclass(frozen=True)` would be more pythonic, type-checks better, and lets you `arrays.item_idx` (consistent with the rest of pymc-marketing). The TypedDict gives no real benefit since callers never construct it from a dict literal.

#### G4. Patsy is now a hard import

`pymc_marketing/customer_choice/maxdiff.py:32` and `synthetic_data.py:18`

`import patsy` at module top means anyone importing `MaxDiffMixedLogit` for the *intercept* path also pays the patsy import cost. Confirm `pyproject.toml` lists patsy as a non-optional dependency (transitive via pymc is likely, but worth a sanity check).

#### G5. `_is_partworths` could be a property

`pymc_marketing/customer_choice/maxdiff.py:425-446`

`self._is_partworths = has_attrs` is a snapshotted boolean. After `__init__`, if a user reassigns `item_attributes`, `_is_partworths` won't track. Make it a `@property` returning `self.item_attributes is not None and self.utility_formula is not None`, or document that mutation is unsupported.

#### G6. `prepare_maxdiff_data` Python-side loop

`pymc_marketing/customer_choice/maxdiff.py:211-247`

The NOTE acknowledges the O(T·K_max) cost. A vectorized rewrite via `groupby().agg()` + `np.add.reduceat` would help users with industrial-scale studies (>1M tasks). Not blocking — file an issue if this isn't tackled here.

#### G7. Notebook — section ordering

The notebook is 7723 lines. A small markdown TOC at the top (most other customer-choice notebooks have one) would help navigation.

---

### Minor / nits

- `maxdiff.py:46`: prefer `NEG_INF = -np.inf` over `float("-inf")` for symmetry with the rest of the codebase.
- `maxdiff.py:909`: `self.items.index(self.reference_item)` is O(I). You computed `item_to_idx` already — cache `self._ref_idx` on the model.
- `maxdiff.py:933`: the `chol_L` Deterministic duplicates `chol_cov` from the LKJ unpacking. You can drop one if downstream readers don't depend on the specific dim names.
- `maxdiff.py:1029-1094`: JSON-encoding all init kwargs is consistent with the rest of customer_choice, but a single helper (`_dump_attrs(self, mapping)`) would cut repetition.
- `maxdiff.py:1031-1033`: `attrs["task_df"] = json.dumps("Placeholder for DataFrame")` matches `mixed_logit.py`'s `"Placeholder for DF"` — promote a shared sentinel constant.
- `synthetic_data.py:286-290`: `_softmax_1d` duplicates `_softmax_stable` from `maxdiff.py`. Promote one to a shared util.
- `synthetic_data.py:218`: `true_utilities = true_utilities - true_utilities[-1]` silently mutates the supplied array's apparent meaning. `warnings.warn` if `true_utilities[-1] != 0`.
- `tests/customer_choice/test_maxdiff.py:209-216`: `import pymc as pm` inline — move to top of file (workspace `no-inline-imports` rule).
- `tests/customer_choice/test_maxdiff.py:878,927`: `import xarray as xr` inline — same.
- `tests/customer_choice/test_maxdiff.py:81-87`: `FAST_FIT_KWARGS` uses `nuts_sampler="pymc"` while `default_sampler_config` uses `numpyro`. Add a one-liner comment so the next contributor doesn't "fix" the inconsistency.
- `maxdiff.py:902,956`: `(v0.1 behaviour)` / `(v0.1/v0.2) path` are insider references. Replace with concrete behavior descriptions ("diagonal-only respondent covariance") so they survive future refactors.
- `score_new_items` einsum `"cdp,ip->cdi"` (`maxdiff.py:722`): a tiny inline assertion that `posterior["beta_feat"].shape[-1] == X_all.shape[-1]` would yield clearer failure messages when feature_names drift.
- `predict_choices` doesn't honor `idata_kwargs`. Probably out of scope — document.

---

### Checks I could not run

I did not exercise the `@pytest.mark.slow` recovery tests, so I'm trusting the author and CI on the recovery numbers (`rho >= 0.85`). The deterministic slice (84 tests) passes locally and ruff is clean.

I also didn't validate the notebook end-to-end — ReviewNB is the right tool there.

NathanielF:
I took a pass at addressing these @juanitorduz . One note i think the model finds the parameter recovery on the covariance cases quite difficult especially with mock sampling. I've added a loose test to ensure it's directionally correct, but I'm open to revisiting that test and stress testing it a bit. But let me know what you think .

juanitorduz (approved):
Thank you @NathanielF !

PR #2531, opened 2026-05-01
migrate budget allocation plots 

### LINKED ISSUE #2525: migrate the budget allocation plots to MMMPlotSuite v2
migrate the budget allocation plots to MMMPlotSuite v2 

Methods:
budget_allocation → mmm.plot.budget.allocation()
allocated_contribution_by_channel_over_time → mmm.plot.budget.contribution_over_time()

isofer:
## Description

Adds a `BudgetPlots` namespace class to the MMM plotting package with two methods for visualising budget allocation optimisation results. The class follows the same namespace pattern already used by `DecompositionPlots`, `DiagnosticsPlots`, `SensitivityPlots`, and `TransformationPlots`.

**`allocation_roas`** replaces `MMMPlotSuite.budget_allocation`. The old method was a side-by-side bar chart showing allocated spend vs channel contribution per channel. Two problems: (1) it plotted the posterior mean of a scaled contribution, discarding all uncertainty; (2) the reason users called it was to judge whether each channel's return justified its spend — which is exactly ROAS. The new method makes that intent explicit by computing ROAS directly (`sum(channel_contribution_original_scale) / allocation`) and showing the full posterior distribution as a forest plot with HDI bars, so under- and over-performing channels are immediately visible with their uncertainty.

**`contribution_over_time`** replaces `MMMPlotSuite.allocated_contribution_by_channel_over_time`. 

The time-series rendering logic previously inlined in `DecompositionPlots.components_contributions` has been extracted into a shared `_plot_timeseries_channel` helper in `_helpers.py`; both `DecompositionPlots` and `BudgetPlots` now delegate to it.

`channel_contribution_original_scale` has been added to the default `var_names` in `MultiDimensionalBudgetOptimizerWrapper` so it is available in the returned dataset without callers having to request it explicitly.

see demo [here](https://gist.github.com/isofer/bf5aed78653eeacd8e51733cabb52176)

## Related Issue

- Closes #2525

## Changes Made

- `pymc_marketing/mmm/plotting/budget.py` — new file: `BudgetPlots` with `allocation_roas` and `contribution_over_time`
- `pymc_marketing/mmm/plotting/_helpers.py` — extracted `_plot_timeseries_channel` shared helper
- `pymc_marketing/mmm/plotting/decomposition.py` — replaced inline time-series rendering with `_plot_timeseries_channel`
- `pymc_marketing/mmm/plotting/__init__.py` — export `BudgetPlots`
- `pymc_marketing/mmm/multidimensional.py` — add `channel_contribution_original_scale` to default sampled var names
- `tests/mmm/plotting/test_budget.py` — new: 343-line test file covering both methods (happy path, panel/multi-geo, dim subsetting, error cases, visual structure assertions)

----
📚 Documentation preview 📚: https://pymc-marketing--2531.org.readthedocs.build/en/2531/


isofer:
@juanitorduz ready for review

isofer:
@cursor push 04b5e3901d

isofer:
@juanitorduz @carlosagostini ready for review

pymc_marketing/mmm/plotting/budget.py
```diff
@@ -0,0 +1,222 @@
...
+        pc_kwargs = _process_plot_params(
+            figsize=figsize,
+            backend=backend,
+            return_as_pc=return_as_pc,
+            **pc_kwargs,
+        )
+
+        roas_da = (
+            samples["channel_contribution_original_scale"].sum("date")
+            / samples["total_allocation"]
```
juanitorduz, line 116: Can we use the incrementality module to compute roas?

pymc_marketing/mmm/plotting/budget.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+        azp.add_lines(pc, 1.0, orientation="vertical")
+
+        return _extract_matplotlib_result(pc, return_as_pc)
+
+    def contribution_over_time(
+        self,
+        samples: xr.Dataset,
+        dims: dict[str, Any] | None = None,
+        hdi_prob: float = 0.85,
```
juanitorduz: can we keep the default to 0.94?

juanitorduz (commented):
Looks very nice! I left two comments :)

pymc_marketing/mmm/plotting/budget.py
isofer: @juanitorduz @carlosagostini 

It's not clear to me if it's the right thing to do. It depends on what type of plots or analysis we want to provide for the user.

The current workflow and plots supports answering the following questions:
1. if using a given allocation, what would be the future **total** ROAs for each channel?
2. if using a given allocation, what would be the contribution of each channel at each point in time?

with the current plan, after we finish the work, the user will use the following workflow to answer the questions:

```
allocation_xarray, res_scipy = optimizable_model.optimize_budget(...)

sample_response_given_allocation = optimizable_model.sample_response_distribution(allocation_xarray, ...)

optimizable_model.plot..allocation_roas(sample_response_given_allocation, ...)
```

This is working well now.
If we want to use Incrementality here, we would need to update the Incrementatly module to support arbitraty allocation. Then we can replace the call of  `optimizable_model.sample_response_distribution`, but we won't gain anything since it would give the same answer. 
So there's no point in doing that. 

I believe, that the only reason to bring in Incrementality, is if we want to answer the question:  if using a given allocation, what would be the **ROAs** of each channel at different points in time in the future (like each month)?

Then we would need to use incrementally, since it can correctly compute the effect of carryover on each point in time.
The workflow then will look like:
```
allocation_xarray, res_scipy = optimizable_model.optimize_budget(...)

roas = mmm.incrementality.contribution_over_spend(
(allocation_xarray, ...)

optimizable_model.plot.roas_over_time(roas)
```

This would require to change incrementality to support arbitrary allocation, and adding plot.roas_over_time. This is not a lot of work, but the question is if this is useful for our users.


So thee main questions that need to be answered is will this be useful for our users.  i.e. When they allocate budgets, which is usually at the quarter level, do they want to see what is the predicted ROAs for each month? 

Another related question is that the budget allocation

2. AFAIK, The ROAs computation now and the budget optimizer overall doesn't take into account the carryover affect of the budgets that were used before the allocation period. So if you optimize for Q2, it ignores any effect from Q1. Do we want to address that (both at the budget optimization code, and the ROAs calculation code)?

juanitorduz: Thanks @isofer !

1. Regarding this point. I think it would be nice to have the incrementality module so that we are consistent. I understand the extra work, so we can tackle this in another PR (please create an issue 🙏 )

2. I am not sure if it does it. It would be important to do it, but again, we can tackle this in another PR then :) (please create and issue 🙏 )

isofer: done.
https://github.com/pymc-labs/pymc-marketing/issues/2544
https://github.com/pymc-labs/pymc-marketing/issues/2545

pymc_marketing/mmm/plotting/budget.py
isofer: done

juanitorduz (approved):
Thanks @isofer !

PR #2547, opened 2026-05-10
docs: cleanup sphinx warnings + add audit-only docs CI

anevolbap:
Closes #1210.

Sphinx warning log on `main`: **1295 stderr lines (964 distinct sphinx warnings) → 0**. CI is green: 54 checks pass (including the new enforcing `Docs` job and the RTD build).

**Drafted with heavy assistance from Claude Opus 4.7 and reviewed by me before opening.**

## What "CI gate" means here

This PR adds a docs-build job that runs `sphinx-build -W --keep-going` on every PR (`.github/workflows/docs.yml`) and sets `sphinx.fail_on_warning: true` on Read the Docs (`.readthedocs.yml`). With `-W`, any sphinx/MyST/numpydoc warning becomes an error and fails the build. `--keep-going` makes sphinx collect every warning before failing instead of stopping at the first, so the job summary lists all of them.

`make check_docs` mirrors the same locally so contributors can reproduce CI without remembering the flags.

Net effect: from this PR on, any PR that introduces a new docs warning fails CI and cannot merge until fixed. Before this PR there was no docs gate at all (RTD built on push but RTD failures didn't block PRs).

## TL;DR — how the count came down

All counts below are distinct sphinx WARNING:/ERROR: lines from successive `make html` runs.

| Build | Distinct warnings | What changed since previous |
|---|---|---|
| `main` (pre-PR) | 964 | Baseline from a clean `make html` (1295 stderr lines) |
| After autosummary template extension | 85 | The first big win — see below |
| After gallery toctree fix | 12 | The second big win |
| After plot example + autosummary stub fixes | **0** | Final |

## The two big wins

**1. Excluding pydantic / mlflow inherited methods from autosummary** (commit `0d804b94`).

About 770 of the original ~900 warnings traced to a single root cause: `pydantic.BaseModel` and `mlflow.pyfunc.model.PythonModel` methods (`model_dump`, `model_dump_json`, `model_validate`, `predict_stream`, etc.) being inherited into our Field-based config classes (HSGPBase, BassPriors, etc.) and PyFuncModel wrappers. autosummary enumerated them, autodoc re-parsed their (malformed) upstream docstrings, and the warnings fanned out ~16x via inheritance.

[...]

PR #2550, opened 2026-05-11
fix(docs): give InferenceData accessor properties a real docstring

### LINKED ISSUE #1992: Docs not rendering for ModelBuilder accessors
This might have to be attached to the property below

https://github.com/pymc-labs/pymc-marketing/blob/9af7f676dadf6ef29455b6fababb53bba2246e4c/pymc_marketing/model_builder.py?plain=1#L100

anevolbap:
Closes #1992.

## Bug

`create_idata_accessor` in `pymc_marketing/model_builder.py` returns a `property` used as a class attribute on `RegressionModelBuilder` and `CLVModel` to expose `.prior`, `.posterior`, `.predictions`, etc. The accessor's docstring was meant to be assigned via:

```python
def accessor(self) -> xr.Dataset:
    __doc__ = f"Access the '{value}' attribute of the InferenceData object."  # noqa: F841
    ...
```

This assigns to a local variable named `__doc__`, not the function's `__doc__` attribute (hence the lingering `noqa: F841` for an unused local that no one removed). The resulting `property.__doc__` was therefore `None`, so Sphinx rendered nothing for the 7 accessor properties built with this factory:

- `prior`, `prior_predictive`, `posterior`, `posterior_predictive`, `predictions`, `fit_data` (in `RegressionModelBuilder` / `CLVModel` / MV-ITS)
- plus one in the deprecated MMM path

## Fix

Pass the docstring via `property(..., doc=...)` so the resulting property exposes `__doc__` correctly. The `noqa` is no longer needed and goes away with the unused local.

Verified at the REPL:

```pycon
>>> from pymc_marketing.model_builder import create_idata_accessor
>>> p = create_idata_accessor('posterior', 'msg')
>>> p.__doc__
"Access the 'posterior' attribute of the InferenceData object."
```

Before this PR, the same call returned `None`.

## Test plan

- [x] REPL verification above.
- [ ] CI passes.
- [ ] Spot-check rendered docs: the `MMM.posterior`, `MMM.prior`, etc. property entries should now have a one-line description instead of being blank (verified once #2547 merges and a fresh RTD preview is available).

----
📚 Documentation preview 📚: https://pymc-marketing--2550.org.readthedocs.build/en/2550/


juanitorduz (approved):
Thanks @anevolbap

Issue #2108, opened 2025-12-01
v1.0 Release Checklist


ColtAllen:
**Repo Maintainers: Please update this checklist and link to appropriate issues per internal discussions**
**Contributors: Feedback welcome!**

# MMM
- [ ] Deprecate legacy model class in favor of `MultiDimensionalMMM`

# CLV
- [ ] API change: `CLVModel(data,model_config).fit() -> CLVModel(model_config).fit(data)`
- [ ] sBG Notebook
- [ ] BG/BB Notebook
- [ ] Consolidate fit methods into `ModelBuilder` for cross-module support
- [ ] Expand `mlflow` Support

# Customer Choice
- [ ] Model Builder inheritance for save/load and expanded fit functionality? (requires API change)

# Bass Diffusion
- [ ] Model Builder inheritance for save/load and expanded fit functionality

ColtAllen:
> Isn't the customer choice model already supporting save / load.

It isn't being tested in the Logit models, and the internal changes I have in mind for the CLV models may break that functionality.

@NathanielF how large are the datasets typically for the Logit models? The reason I'm redoing the CLV API is because those models sometimes rival open-sourced LLMs in size due to the massive datasets, creating I/O issues when logging to MLFlow. The only way to fix this is to have the option to exclude fit data when save/loading, which isn't possible when the model is initialized with data.

NathanielF:
@williambdean  on the mvits model i believe @drbenvincent originally implemented that class, but i think i substantially copied the class design for the MnLogit and Nested logit classes. 

@ColtAllen on the question of size it's likely worth distinguishing between the mvits model and the nestedl logit and mnl logit. For the latter two, the data sets i'd expect are quite small ussually survey response data. The data itself is typically expensive to gather so it's hard to accumulate huge amounts of data. Think of the effort required to accurately describe your own company's product +  your major competitors in a market. 

For the mvits that takes timeseries data of product sales and this can be easier to gather, but you still need accurate sales data for your competitors. Not trivial. But i'd expect medium to large data sets for MVITs models. @drbenvincent  might be able to say more....

twiecki:
Should we rename `MultiDimensionalMMM` to just be `MMM` or will we break too much code?

williambdean:
Is still the checklist that we are going by?

I've checked off another item in the existing checklist. If we use checklist of github
issues, I think that they close the checkmark automatically ...


```
- [ ] #<number>
```

[...]

PR #2558, opened 2026-05-11
fix(utils): drop obsolete fit_data UserWarning workaround in from_netcdf

### LINKED ISSUE #1716: Remove custom `from_netcdf` function
This warning doesnt appear in the new arviz versions

https://github.com/pymc-labs/pymc-marketing/blob/ea4498bfabba3cae68581e90d2a80bbaedafdbea/pymc_marketing/utils.py?plain=1#L22-L41

anevolbap:
Closes #1716.

## Summary

`pymc_marketing.utils.from_netcdf` wrapped `arviz.from_netcdf` in a `warnings.catch_warnings()` block that silenced this UserWarning:

```
{group} group is not defined in the InferenceData scheme
```

Modern arviz gates that warning behind `warn_on_custom_groups=False` (default), so it no longer fires when loading an InferenceData with a custom group like `fit_data`. The filter block is dead code.

## Changes

- `pymc_marketing/utils.py`: drop the `warnings.catch_warnings` block, return `az.from_netcdf(filepath)` directly. Update the docstring summary.

The function itself stays in its deprecation cycle from #2529 and will be removed in the 1.0 release per #2108. This PR only removes the obsolete workaround.

## Verification

Locally on `arviz==0.23.4`:

```python
import warnings, arviz as az, xarray as xr, numpy as np, tempfile

idata = az.InferenceData(
    posterior=xr.Dataset({"a": xr.DataArray(np.zeros((1, 10)), dims=["chain", "draw"])}),
    fit_data=xr.Dataset({"x": xr.DataArray([1, 2, 3], dims=["idx"])}),
)
with tempfile.NamedTemporaryFile(suffix=".nc", delete=False) as tf:
    idata.to_netcdf(tf.name)
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    reloaded = az.from_netcdf(tf.name)
print([str(x.message) for x in w if issubclass(x.category, UserWarning)])  # []
print(reloaded.groups())  # ['posterior', 'fit_data']
```

No UserWarning is emitted and the `fit_data` group survives the round-trip.

## Test plan

- [x] `pytest tests/test_utils.py -x --no-cov` passes (both deprecation-warning and load tests).
- [x] `pre-commit run --files pymc_marketing/utils.py` clean.

----
📚 Documentation preview 📚: https://pymc-marketing--2558.org.readthedocs.build/en/2558/


juanitorduz (approved):
thanks @anevolbap !

PR #2559, opened 2026-05-11
fix(plot): drop plt.show from MMMPlotSuite.cv_predictions

### LINKED ISSUE #2373: MMMPlotSuite: plt.show() inside library methods (cv_predictions, param_stability)
## Problem

`cv_predictions` (line 4615) and `param_stability` (multiple times) call `plt.show()`. No other methods do. This is an anti-pattern in library code because:
- It forces immediate rendering, preventing further customization
- It doesn't work in non-interactive backends
- It's inconsistent with 18 other methods that correctly return without showing

anevolbap:
Closes #2373.

Removes the single `plt.show()` call in `MMMPlotSuite.cv_predictions`. Library methods should return the figure and let the caller render.

Adds a regression test that monkeypatches `plt.show` and asserts it is not called.

Note: the issue also mentions `param_stability`, but it has no `plt.show()` calls. Happy to extend the PR if a different location was intended.


williambdean (approved):
LGTM!

juanitorduz (changes requested):
thanks @anevolbap I do not think we need the test, just remove the plt.show()(s) wherever they are present :)

anevolbap:
Removed the test, thanks for the review!

anevolbap:
No other `plt.show()` calls exist in real code paths. The 63 remaining occurrences are all inside docstring examples (`>>>` blocks), left alone since they are illustrative and out of scope for #2373.

juanitorduz (approved):

PR #2560, opened 2026-05-12
fix(ci): stop duplicate-issues cancellations; harden triage

### LINKED ISSUE #2474: Proposal: Fix duplicate-finder cancellations and triage output format failures
## Background

Two GitHub Actions workflows that power our issue automation are currently broken in different ways.

---

### What's happening

The workflow is configured with \`cancel-in-progress: true\`. When an issue is edited (which triggers a new run for the same issue), the in-progress run is cancelled and a new one starts. Because the workflow begins with a 10-minute \`sleep\` to allow for rapid edits, most runs spend the entire sleep period and then get cancelled when the user makes another edit.

### Evidence

Out of the last 60 issue-triggered runs:
- **30 cancelled** (50%)
- **23 success** (38%)
- **7 failure** (12%)

Representative cancelled runs:
- [Run 23852795768](https://github.com/pymc-labs/pymc-marketing/actions/runs/23852795768) — WindowedYearlyFourier issue
- [Run 23850579633](https://github.com/pymc-labs/pymc-marketing/actions/runs/23850579633) — MuEffect protocol issue
- [Run 23846754104](https://github.com/pymc-labs/pymc-marketing/actions/runs/23846754104) — multidimensional MMM issue
- [Run 23836547637](https://github.com/pymc-labs/pymc-marketing/actions/runs/23836547637) — model.named_vars issue (4 separate cancellations for same issue)
- [Run 23836401223](https://github.com/pymc-labs/pymc-marketing/actions/runs/23836401223) — Add a component for the marketing funnel
- [Run 23792875775](https://github.com/pymc-labs/pymc-marketing/actions/runs/23792875775) — MaskedPrior issue (2 cancellations)
- [Run 23776224166](https://github.com/pymc-labs/pymc-marketing/actions/runs/23776224166) — Optimize total response (2 cancellations)
- [Run 23449912645](https://github.com/pymc-labs/pymc-marketing/actions/runs/23449912645) — target_scale issue (3 cancellations)
- [Run 23405774476](https://github.com/pymc-labs/pymc-marketing/actions/runs/23405774476) — Temporal lift calibration (3 cancellations)

### Root cause

\`\`\`yaml
concurrency:
  group: issue-\${{ github.event.issue.number }}-duplicate
  cancel-in-progress: true   # ← cancels mid-run when next edit arrives
\`\`\`

Combined with:
\`\`\`yaml
- name: Wait 10 minutes before checking
  run: sleep 600             # ← 10 minutes of wasted runner time before cancellation
\`\`\`

### Note on data freshness

The issue title and body are captured from \`\${{ github.event.issue.* }}\` at the time the workflow is **triggered**. So:
- Each edit fires a new workflow run with the latest body at that moment
- With \`cancel-in-progress: false\`, multiple runs may queue — but the final edit's run will always have the up-to-date content
- The sleep window naturally batches rapid edits (only the last edit's run survives the queue)

---

### What's happening

The triage agent is actually running and producing high-quality analysis, but the workflow reports failure every time. The step that extracts labels uses \`grep\`, which exits with code 1 when the pattern is not found — causing the entire job to fail.

### Evidence

All 30 of the last 30 triage runs have failed:
- [Run 23852795787](https://github.com/pymc-labs/pymc-marketing/actions/runs/23852795787) — WindowedYearlyFourier issue
- [Run 23850579642](https://github.com/pymc-labs/pymc-marketing/actions/runs/23850579642) — MuEffect protocol issue
- [Run 23848020420](https://github.com/pymc-labs/pymc-marketing/actions/runs/23848020420) — \`dim\` missing in adstock issue
- [Run 23846754065](https://github.com/pymc-labs/pymc-marketing/actions/runs/23846754065) — multidimensional MMM issue
- [Run 23836401261](https://github.com/pymc-labs/pymc-marketing/actions/runs/23836401261) — Add a component for the marketing funnel
- [Run 23836352076](https://github.com/pymc-labs/pymc-marketing/actions/runs/23836352076) — model.named_vars issue
- [Run 23792861022](https://github.com/pymc-labs/pymc-marketing/actions/runs/23792861022) — MaskedPrior issue
- [Run 23776206931](https://github.com/pymc-labs/pymc-marketing/actions/runs/23776206931) — Optimize total response issue
- [Run 23754294100](https://github.com/pymc-labs/pymc-marketing/actions/runs/23754294100) — MMMPlotSuite v2 issue

### Root cause

The agent does not reliably emit the expected \`TRIAGE: Added labels: ...\` line. The workflow step does:

\`\`\`bash
LABELS="\$(grep '^TRIAGE: Added labels:' /tmp/claude_output.log | sed '...')"
\`\`\`

When the agent outputs free-form analysis without that exact line, \`grep\` exits with code 1, and because \`set -o pipefail\` is active in the previous step, the whole job fails. Labels are never applied and \`Needs Triage\` is never removed.

---

### Fix 1: \`duplicate-issues.yml\` — remove cancellations, smarter sleep

**Change \`cancel-in-progress\` to \`false\`**
Queued runs will still execute, but no run is killed mid-flight. The final edit's run always has the latest data.

**Replace the flat 10-minute sleep with a shorter event-aware wait**
- \`opened\` event: sleep 2 minutes (catches quick "oops, typo" fixes)
- \`edited\` event: sleep 5 minutes (batches a typical editing session)

This means a user can edit an issue repeatedly; each edit queues a run, but only the last edit's run (and any not yet past the sleep) will be doing meaningful work — and nothing is wasted.

\`\`\`yaml
concurrency:
  group: issue-\${{ github.event.issue.number }}-duplicate
  cancel-in-progress: false

steps:
  - name: Wait for edits to settle
    run: |
      if [ "\${{ github.event.action }}" == "opened" ]; then
        sleep 120   # 2 min for new issues
      else
        sleep 300   # 5 min for edits
      fi
\`\`\`

### Fix 2: \`triage.yml\` — robust label extraction + post analysis as comment

**Make the label extraction step non-fatal**
Use \`grep ... || true\` so a missing format line doesn't fail the job.

**Strengthen the agent's output format instructions**
Update \`.claude/agents/triage.md\` to make the \`TRIAGE: Added labels: ...\` line a hard requirement, with examples, placed at the very end of the response.

**Post the triage analysis as a comment**
Currently the analysis is only visible in the workflow logs (behind an artifact download). Instead, post it directly on the issue in a collapsed \`<details>\` block so it is accessible without leaving GitHub, but doesn't take up much space:

\`\`\`markdown
<details>
<summary>🤖 Triage analysis</summary>

{full agent output here}

</details>
\`\`\`

### Fix 3: \`triage.yml\` — add \`edited\` trigger (optional, discuss)

Currently triage only fires on \`opened\`. If the issue body is substantially changed, re-triaging might be useful. This mirrors the behaviour already in \`duplicate-issues.yml\`. This is lower priority and can be a follow-up.

---

## Files to change

| File | Change |
|---|---|
| \`.github/workflows/duplicate-issues.yml\` | \`cancel-in-progress: false\`, smarter sleep, increase timeout |
| \`.github/workflows/triage.yml\` | Robust label extraction, post comment with analysis |
| \`.claude/agents/triage.md\` | Strengthen output format requirements |

---

## Out of scope

- Rate limiting (not needed given the 5-minute sleep batching)
- Changes to \`.claude/agents/duplicate-finder.md\` (works correctly)
- Triage on \`edited\` events (separate issue/follow-up)

anevolbap:
Closes #2474.

## duplicate-issues.yml

- `cancel-in-progress: true` -> `false`. Queued runs finish instead of being killed mid-sleep when the issue is edited again.
- Flat 10-minute sleep replaced with event-aware wait: 2 min on `opened`, 5 min on `edited`. Job timeout bumped to 20 min.
- Post-or-update instead of always-post. Each duplicate-finder comment is wrapped in a hidden HTML sentinel (`<!-- duplicate-finder-comment -->`); later queued runs find the existing comment and update it via the `updateIssueComment` GraphQL mutation instead of stacking one comment per edit. Without this, `cancel-in-progress: false` would let an editing session post N duplicate-finder comments on the same issue.

## triage.yml

- Label extraction moved into its own step with `|| true`, so a missing `TRIAGE: Added labels:` line no longer fails the job. Pipeline ends with `tail -n 1` so a stray duplicated prefix can't inject a newline into `$GITHUB_OUTPUT`.
- `Remove Needs Triage` and the new analysis-comment step run only when label extraction completed (`steps.extract_labels.conclusion == 'success'`), not on a full agent crash; in that case the issue stays in Needs Triage for a human.
- New step posts the agent's analysis as a collapsed `<details>` comment on the issue, with the `TRIAGE: Added labels: ...` plumbing line stripped out.

## .claude/agents/triage.md

- The `TRIAGE: Added labels:` line is required as the very last line of every response, with an explicit empty-list form when no labels apply.

Per the issue's "out of scope": no changes to `duplicate-finder.md`, and triage on `edited` is left for a follow-up.

## Known limitation

If a later queued run decides the issue has no duplicates after an earlier run posted `FOUND`, the existing comment is not cleared. This is the inverse of the dedup case the PR fixes; addressing it would add a delete path on `has_duplicates == 'false'`. Out of scope here, can be a follow-up if it bites.

## Test plan

- [x] `pre-commit run --files .github/workflows/duplicate-issues.yml .github/workflows/triage.yml .claude/agents/triage.md` clean.
- [x] Local simulation of the extract pipeline against canned `claude_output.log` payloads (labels present, empty list, trailing space, no marker, padded labels, duplicate prefix). All pass; not committed (sandbox-only).
- [ ] Workflows can only be exercised on real issue events; happy to validate on the next opened/edited issue.

----
:books: Documentation preview :books: https://pymc-marketing--2560.org.readthedocs.build/en/2560/


.github/workflows/duplicate-issues.yml
```diff
@@ -6,21 +6,26 @@ on:
...
   group: issue-${{ github.event.issue.number }}-duplicate
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 jobs:
   check-duplicates:
     if: github.event.action == 'opened' || github.event.changes.body
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 20
```
anevolbap, line 15: Bumped from 15 to 20 minutes. The issue asked to 'increase timeout' without a specific number. Worst case is a 5 min sleep plus a 50-turn agent run plus the comment step, so 20 leaves headroom. Open to a smaller value if preferred.

.github/workflows/duplicate-issues.yml
```diff
@@ -6,21 +6,26 @@ on:
...
 
     steps:
-      - name: Wait 10 minutes before checking
-        run: sleep 600
+      - name: Wait for edits to settle
+        run: |
+          if [ "${{ github.event.action }}" = "opened" ]; then
+            sleep 120
+          else
+            sleep 300
```
anevolbap, line 27: POSIX `=` is used here in the `[ ... ]` test (the issue's snippet had bash-only `==`). Functionally identical under bash, just matches what shellcheck prefers. Easy to revert if the snippet form is preferred.

.github/workflows/triage.yml
```diff
@@ -62,26 +62,48 @@ jobs:
             --print \
             < /tmp/issue_content.txt 2>&1 | tee /tmp/claude_output.log
 
-          # Extract labels from triage output
-          LABELS="$(grep '^TRIAGE: Added labels:' /tmp/claude_output.log | sed 's/^TRIAGE: Added labels: //')"
-          echo "labels=${LABELS}" >> $GITHUB_OUTPUT
+      - name: Extract labels
+        id: extract_labels
+        run: |
+          LABELS="$(grep '^TRIAGE: Added labels:' /tmp/claude_output.log | sed 's/^TRIAGE: Added labels: //' || true)"
```
anevolbap: Extraction is split into its own step so the `|| true` is the only thing protecting against a missing line. The previous structure ran extraction inside the agent step, which is what made the whole job fail (grep returning 1 under `set -e`).

.github/workflows/triage.yml
```diff
@@ -62,26 +62,48 @@ jobs:
...
+          echo "labels=${LABELS}" >> "$GITHUB_OUTPUT"
 
       - name: Apply labels
-        if: success() && steps.triage.outputs.labels
+        if: steps.extract_labels.outputs.labels
         run: |
```
anevolbap, line 75: The `success()` guard is dropped. The extract step now never fails (it returns an empty string on no match), so the only useful gate is whether labels were actually captured. If `labels` is the empty string, this evaluates to false and the step is skipped.

.github/workflows/triage.yml
```diff
@@ -62,26 +62,48 @@ jobs:
...
+          LABELS="${{ steps.extract_labels.outputs.labels }}"
           if [ -n "$LABELS" ]; then
             gh issue edit "${{ github.event.issue.number }}" --add-label "$LABELS"
           fi
         env:
           GH_TOKEN: ${{ github.token }}
 
       - name: Remove Needs Triage label
-        if: success()
+        if: always()
```
anevolbap: Gated on `steps.extract_labels.conclusion == 'success'` in the follow-up commit. An earlier draft used `always()`, which cleared the label even when the agent crashed outright; in that case the issue should stay in Needs Triage for a human. This condition matches "the agent ran and extraction completed" without firing on a hard crash.

.github/workflows/triage.yml
```diff
@@ -62,26 +62,48 @@ jobs:
...
         run: gh issue edit "${{ github.event.issue.number }}" --remove-label "Needs Triage" || true
         env:
           GH_TOKEN: ${{ github.token }}
 
+      - name: Post triage analysis as comment
```
anevolbap, line 89: Gated on `steps.extract_labels.conclusion == 'success'` in the follow-up commit. Posts the analysis when the agent ran (including the no-labels case). On a hard agent crash the comment is skipped; the artifact upload still happens for debugging. The `TRIAGE: Added labels: ...` plumbing line is stripped before posting since it isn't user content.

.github/workflows/triage.yml
```diff
@@ -62,26 +62,48 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+        run: |
+          if [ ! -s /tmp/claude_output.log ]; then
+            echo "No agent output to post"
+            exit 0
+          fi
+          {
+            printf '%s\n' '<details>'
+            printf '%s\n' '<summary>:robot: Triage analysis</summary>'
```
anevolbap: Dropped from the summary in the follow-up commit. `<summary>Triage analysis</summary>` is plain text now; the emoji wasn't carrying information.

.github/workflows/triage.yml
```diff
@@ -62,26 +62,48 @@ jobs:
...
+            printf '%s\n' ''
+            cat /tmp/claude_output.log
+            printf '%s\n' ''
+            printf '%s\n' '</details>'
+          } > /tmp/triage_comment.md
+          gh issue comment "$ISSUE_NUMBER" --body-file /tmp/triage_comment.md
```
anevolbap, line 113: Using `printf` + `cat` instead of a heredoc with `$RESPONSE` substitution. The agent output is arbitrary text and would need careful quoting otherwise. `gh issue comment --body-file` avoids any further shell escaping.

.claude/agents/triage.md
```diff
@@ -75,18 +75,28 @@ Add labels based on issue content (title + body):
...
+format (no trailing text after it):
 
 ```
 TRIAGE: Added labels: label1, label2
 ```
 
-For example:
+This line is required on every response, even when no labels apply. In that
+case emit it with an empty label list:
+
```
anevolbap, line 87: Added an explicit empty-list form: `TRIAGE: Added labels:`. This softens the existing rule below ('add enhancement as default'). The intent is to avoid forcing the agent to invent a label just to satisfy the format. It does deviate from the prior contract though, so flagging for explicit consent. Easy to drop if the old default-to-enhancement behavior is preferred.

anevolbap (commented):
Self-review against the issue. Items I want explicit eyes on before flipping out of draft.

williambdean (approved):
Looks fine by me! Thanks for the edits. Let's try it out.

PR #2561, opened 2026-05-12
fix(plot): use ceil division so plot_curve grid fits all axes

### LINKED ISSUE #1834: plot_curve ncols silently cuts off axes
Example with 9 axes but with ncols=4

2x4 grid cuts off the last plot

anevolbap:
Closes #1834.

`set_subplot_kwargs_defaults` derived `nrows` from `ncols` (and vice versa) with floor division, so a 9-axis call with `ncols=4` produced a 2x4 grid and silently dropped the 9th plot. Switch to ceil division.

Side effect: when `total_size` is not a multiple of `ncols`/`nrows`, the grid now has empty trailing axes (e.g. 9 plots with `ncols=4` -> 3 empty cells). That is visible empty rather than missing data, which is the behaviour the issue asks for.

Added a parametrized test covering both `ncols`-given and `nrows`-given paths, including the divides-evenly case to guard against regressions.

----
📚 Documentation preview 📚: https://pymc-marketing--2561.org.readthedocs.build/en/2561/


pymc_marketing/plot.py
```diff
@@ -176,9 +176,9 @@ def set_subplot_kwargs_defaults(
         subplot_kwargs["ncols"] = total_size
 
     if "ncols" in subplot_kwargs:
-        subplot_kwargs["nrows"] = total_size // subplot_kwargs["ncols"]
+        subplot_kwargs["nrows"] = -(-total_size // subplot_kwargs["ncols"])
```
anevolbap, line 179: `-(-total_size // k)` is the floor-to-ceiling trick: floor-divide the negation, then negate back. Same as `math.ceil(total_size / k)` but stays in integer math and avoids the `math` import.

Happy to swap to `math.ceil` if you prefer the readability.

juanitorduz (approved):

PR #2562, opened 2026-05-12
chore(model-builder): remove X_pred/y_pred deprecated aliases

### LINKED ISSUE #1768: Deprecate X_pred warning
This has been warned since Feb 2025

https://github.com/pymc-labs/pymc-marketing/blob/c24b97be3ca322e8ff1d31500b700631a0975065/pymc_marketing/model_builder.py?plain=1#L50-L74

anevolbap:
Closes #1768.

The `_handle_deprecate_pred_argument` helper has emitted a `DeprecationWarning` for `X_pred` / `y_pred` since #1450 (Feb 2025, ~15 months). Drop the helper and its call sites.

Behavior for callers that still pass `X_pred=`:

- `sample_prior_predictive`, `sample_posterior_predictive`, `predict_posterior`: `X` stays `None`, so the existing `Please provide X.` `ValueError` fires before `**kwargs` is forwarded downstream. The error points at the new argument name.
- The "both `X` and `X_pred` provided" `ValueError` is gone; passing both now silently ignores `X_pred`. Unlikely in practice.

Tests:

- Removed `test_deprecation_warnings` and `test_handle_deprecate_pred_argument` (alias and helper are gone).
- Added `test_pred_alias_no_longer_accepted` covering all three methods to lock in the new behavior.


anevolbap:
Out-of-scope nit, surfacing for triage: while editing this file I noticed [`mmm/mmm.py:2375`](https://github.com/pymc-labs/pymc-marketing/blob/9ceebc57e77bc586b011be20c176cf115ee9bffe/pymc_marketing/mmm/mmm.py#L2375) has the comment `# Transform X and y_pred to xarray`, but `y_pred` is not referenced in that function (predates this PR). Happy to fix here as a drive-by, or file a separate issue, whichever you prefer.

pymc_marketing/model_builder.py
```diff
@@ -1281,8 +1254,8 @@ def sample_prior_predictive(
             Prior predictive samples for each input X
 
         """
-        X = _handle_deprecate_pred_argument(X, "X", kwargs)
-        y = _handle_deprecate_pred_argument(y, "y", kwargs, none_allowed=True)
+        if X is None:
+            raise ValueError("Please provide X.")
```
williambdean: Do we need this? Can't we just leverage typing and remove | None

anevolbap: Good point. One thing to confirm before making the change: the old `_handle_deprecate_pred_argument` raised `ValueError("Please provide X.")` when `X` was missing, and the current inline check keeps that exact error. Dropping the default and making `X` required turns it into a `TypeError` from Python. OK with that small behavior change?

williambdean: Yeah, I think that is fine

anevolbap: Two follow-up commits on this branch (still need to push):

- `2a8e816d`: `X` is now a required positional in `predict`, `predict_proba`, `predict_posterior`, `sample_prior_predictive`, `sample_posterior_predictive`. Inline `if X is None` checks removed. Missing `X` now raises `TypeError` from Python instead of `ValueError`. The deprecation-removal test was updated accordingly.
- `63c453e7`: while at it, the two `sample_*` methods had no annotation on `X` at all. Added `X: np.ndarray | pd.DataFrame | pd.Series` so all five methods share the same input contract. This surfaced a pre-existing LSP divergence in `MMM.sample_posterior_predictive` (extra `include_last_observations` and `clone_model` kwargs, narrower `DataFrame` type). Replaced the per-parameter `# type: ignore` with a single `# type: ignore[override]` on the def line.

Happy to drop the second commit if it feels out of scope.

juanitorduz (approved):
Thanks :)

PR #2563, opened 2026-05-12
refactor(mmm): rename optimize_budget **minimize_kwargs to **allocate_budget_kwargs

### LINKED ISSUE #1660: Be explicit about the kwargs going to `allocate_budget`
Within the `optimize_budget` method, the `allocate_budget` method is called with 
a `minimize_kwargs` kwargs that is unpacked:

https://github.com/pymc-labs/pymc-marketing/blob/982c97ae36b2c5c31bf245376536cb7b3634f4df/pymc_marketing/mmm/mmm.py?plain=1#L2389-L2393

However, there is also a `minimize_kwargs` that is passed to the underlying method: 

https://github.com/pymc-labs/pymc-marketing/blob/982c97ae36b2c5c31bf245376536cb7b3634f4df/pymc_marketing/mmm/budget_optimizer.py?plain=1#L372-L379

It ends up being called like this: 

```python
minimize_kwargs = {
    "x0": ..., 
    "return_if_fail": ...,
    "minimize_kwargs": { ... },
}

mmm.optimize_budget(
    ..., 
    **minimize_kwargs, 
)
```

I think using `allocate_budget_kwargs` would be a bit clear

Thoughts?

anevolbap:
Closes #1660.

`BudgetOptimizerWrapper.optimize_budget`'s variadic relays every keyword to `BudgetOptimizer.allocate_budget` (`x0`, `minimize_kwargs`, `return_if_fail`, ...), not only the scipy minimize options, so the previous name was misleading. Renaming to `**allocate_budget_kwargs` matches what the parameter actually does and avoids the awkward `minimize_kwargs={"minimize_kwargs": {...}}` reading.

User-facing calls are unaffected: callers pass `minimize_kwargs=`, `x0=`, `return_if_fail=` as named kwargs that the variadic absorbs and re-forwards by name.

----
📚 Documentation preview 📚: https://pymc-marketing--2563.org.readthedocs.build/en/2563/


juanitorduz (approved):
Thanks!

WDYT @cetagostini ?

williambdean (approved):

anevolbap:
This also applies cleanly to v1.0.0: `optimize_budget` there has the same `**minimize_kwargs` in the three spots this renames, they are the only occurrences in `mmm.py`, and the patch applies with a line-offset only. Should it target v1.0.0 instead of main?

PR #2564, opened 2026-05-12
docs: fix code comments and typos in mmm_fivetran_connectors notebook

### LINKED ISSUE #1970: code comments need in cell
https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_fivetran_connectors.html#loading-data-from-fivetran-dbt-schema

anevolbap:
Closes #1970.

The python example under "Loading Data from Fivetran dbt Schema" had two prose lines without a `#` prefix, so the snippet was not valid Python:

```python
Create your database connection
engine = create_engine('your_database_connection_string')

Query the standardized ad reporting table
query = "SELECT * FROM <schema>.ad_reporting__ad_report"
```

Prepended `#` to both. The sibling block lower in the notebook (the `shopify__orders` example) already does it this way.

Also fixed two unrelated typos I noticed in the same notebook:
- "lternative way to load data" -> "Alternative"
- "take a look to the followinc documents" -> "following"

----
📚 Documentation preview 📚: https://pymc-marketing--2564.org.readthedocs.build/en/2564/


juanitorduz (approved):
Thanks!

PR #2567, opened 2026-05-13
refactor(budget-optimizer): use PrivateAttr and model_post_init

### LINKED ISSUE #1662: Use `PrivateAttr` and `model_post_init` with pydantic v2
We can use `PrivateAttr` for all of these attrs to have them defined along the other Fields. This seems to be closer to best practices with pydantic v2

https://github.com/pymc-labs/pymc-marketing/blob/982c97ae36b2c5c31bf245376536cb7b3634f4df/pymc_marketing/mmm/budget_optimizer.py?plain=1#L165-L217

anevolbap:
Closes #1662.

`BudgetOptimizer` used a custom `__init__` to assign 12 underscore-prefixed instance attributes after `super().__init__`. Declared each as `PrivateAttr` next to the `Field` declarations and renamed the body to `model_post_init`, matching the pattern already used by `CustomModelWrapper` in the same file.

No behavior change. Private attribute names are unchanged, so the tests that reach into them (`_budgets_flat`, `_budget_dims`, `_budget_coords`, `_budget_distribution_over_period_tensor`) keep working.

Verified locally with `pytest tests/mmm/test_budget_optimizer.py tests/mmm/test_budget_optimizer_mmm.py tests/mmm/test_cost_per_unit.py tests/test_pytensor_utils.py` (105 passed).


juanitorduz (approved):
Thanks!

PR #2571, opened 2026-05-14
fix(hsgp): stop forwarding deprecated drop_first to pm.gp.HSGP

### LINKED ISSUE #1460: Handle the deprecation of drop_first in HSGP
ckages/pymc/gp/hsgp_approx.py:297: DeprecationWarning: The drop_first a
rgument will be deprecated in future versions. See https://github.com/p
ymc-devs/pymc/pull/6877
    warnings.warn(

anevolbap:
Closes #1460.

`pm.gp.HSGP(drop_first=True)` emits a `DeprecationWarning` and pymc plans to remove the argument (pymc PR #6877). The wrapper forwarded `self.drop_first`, so the warning fired and the call site would break on removal.

Fix: drop the kwarg and slice `phi[:, 1:]` / `sqrt_psd[1:]` after `prior_linearized` returns. Same numerical result.

Added a regression test that asserts no `DeprecationWarning` when `drop_first=True`.


juanitorduz (approved):

juanitorduz:
Thanks @anevolbap !

PR #2575, opened 2026-05-15
docs(clv): align MAP vs MCMC comparison in quickstart

### LINKED ISSUE #2338: Misleading model comparison (MAP vs MCMC) in CLV Quickstart due to uncontrolled variables (priors & data split)
**Describe the issue**
In the "CLV Quickstart" notebook, there is a section comparing the performance of the BG/NBD model when fitted with [MAP](https://www.pymc-marketing.io/en/stable/notebooks/clv/clv_quickstart.html#model-fitting-with-map) vs. [MCMC](https://www.pymc-marketing.io/en/stable/notebooks/clv/clv_quickstart.html#model-fitting-with-mcmc). The notebook concludes that *"Looking at the time plots again, we can see results track much more closely when fit with MCMC."*

However, this comparison is currently an "apples-to-oranges" evaluation because two major variables are entirely different between the MAP and MCMC fitting steps:

1. **In-sample vs. Out-of-sample Data:** 
   - The **MAP** model is evaluated on out-of-sample data. It explicitly uses `rfm_train_test_split` (52 weeks for training, 26 weeks for testing). 
   - The **MCMC** model (`bgm`) is evaluated in-sample. It uses the entire 78-week `rfm_data` dataset for both fitting and visualization.
2. **Prior Configurations:** 
   - The **MAP** model (`bgm_map`) uses the default, uninformative priors (e.g., `HalfFlat` and pooled hierarchical priors).
   - The **MCMC** model (`bgm`) is initialized with a custom `model_config` using informative `HalfNormal(0, 10)` priors.

**Why this is misleading**
The text implies that the MCMC algorithm itself is responsible for the vastly improved tracking of expected purchases. However, my independent testing reveals that the visible improvement in the expected mean is primarily driven by the optimized prior structure and the in-sample data leakage, rather than the choice between MAP and MCMC. 

While MCMC is absolutely necessary for providing uncertainty estimates and credible intervals (which the notebook correctly points out), demonstrating its superiority via an unaligned expected mean comparison is confusing for users trying to understand the mathematical differences.

**My Testing / Evidence**
I have run controlled tests where both MAP and MCMC use the exact same data split and prior configurations. As shown in the screenshots below, when the variables are properly controlled, the difference in the expected mean tracking between MAP and MCMC is negligible, and the true optimization comes from the structural changes (priors).

MAP with optimized structure & data split:

Image

MCMC with optimized structure & data split:

Image

MAP with data split and without optimized structure:

Image

MAP with optimized structure & without data split:

Image

**Proposed Direction**
To make the tutorial more robust and pedagogically sound, I suggest:
1. Re-running both MAP and MCMC using the **exact same model configuration** (preferably the custom `HalfNormal` priors to show good results).
2. Applying the **same train/test split** for both methods so both evaluate out-of-sample prediction accuracy.

anevolbap:
Closes #2338.

The CLV Quickstart compared `bgm_map` (default `HalfFlat` priors, fit on the train/test split, plotted with `t_start_eval=52`) against `bgm` (`HalfNormal` priors, fit on the full `rfm_data`, plotted in-sample) and concluded that results "track much more closely when fit with MCMC". Two variables were confounded, so the conclusion misattributed the improvement to the inference method.

Changes:
- Apply `model_config` (`HalfNormal` priors) to `bgm_map`.
- Add a `bgm_mcmc` fit on `rfm_train_test_data` with the same `model_config` and plot it with `t_start_eval=52` for an apples-to-apples comparison.
- Refit `bgm` on the full `rfm_data` afterwards so downstream sections (matrices, per-customer predictions, CLV) keep using all available data.
- Reword the surrounding narrative. `plot_expected_purchases_over_time` only renders the posterior mean, so the value of MCMC over MAP shows up later in parameter posteriors and credibility intervals on per-customer predictions and CLV, not in this plot.

----
📚 Documentation preview 📚: https://pymc-marketing--2575.org.readthedocs.build/en/2575/


juanitorduz (commented):
Thank you @anevolbap

juanitorduz (approved):

Issue #2158, opened 2025-12-26
Deprecate Unused CLV Distributions


ColtAllen:
The following distributions in [`clv/distributions.py`](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/distributions.py) and their respective RVs are circa-2022 relics and have never been used anywhere in the library other than to generate synthetic data for testing:

- [ ] [ContNonContract](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/distributions.py#L80)
- [ ]  [ContContract](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/distributions.py#L197)

Removing these will speed up the CI checks when creating new PRs:

- [ ] Add deprecation warnings to `__init__` for next release.
- [ ] Delete all classes and respective unit tests for [v1.0 release](https://github.com/pymc-labs/pymc-marketing/issues/2108).

naveen28204280:
Hey, can this issue be assigned to me ? I'd like to work on it as it's my first contribution to pymc

ColtAllen:
> Hey, can this issue be assigned to me ? I'd like to work on it as it's my first contribution to pymc

Thanks @naveen28204280, you've been assigned!!

naveen28204280:
@ColtAllen So, I need to just make a PR for the current version with a deprecation warning and another for v1.0 with the classes removed ?

Issue #2157, opened 2025-12-26
Deprecate `ShiftedBetaGeoModelIndividual`


ColtAllen:
The frequentist implementation of Shifted Beta-Geo Model assumes all customers start in the same time period. `ShiftedBetaGeoModelIndividual` overcame this by estimating an individual dropout posterior for each individual customer, but it's a rather "academic" solution that doesn't hold up well in practice. Many users have complained to me about the slow fit times of this model, difficulties in convergence, and the fact the individual dropout rate is fixed over time. 

The new `ShiftedBetaGeoModel` groups customers by cohort (i.e., the first time period they started their contract) and uses hierarchical parameter pooling estimate a full dropout distribution for each cohort and control indirectly for time-varying factors. It also supports covariats and has additional predictive methods like expected remaining lifetime and retention elasticity rate.

I propose we deprecate `ShiftedBetaGeoModelIndividual` in two PRs:

- [ ] Add a deprecation warning to `ShiftedBetaGeoModelIndividual.__init__`
- [ ] Remove `ShiftedBetaGeoModelIndividual` and its respective unit tests altogether in the [1.0 release](https://github.com/pymc-labs/pymc-marketing/issues/2108).

**I'm open to feedback on this.**

Issue #1516, opened 2025-02-22
CLV Deprecations


arthurmello:
- For a future release (2025-Q4 or early 2026)
- Related to #1498

Remove:
- `_rename_posterior_variables`  and `deprecated_keys = [key for key in model_config if key.endswith("_prior")]`, in basic CLV model
- the related tests (such as `test_backward_compatibility_with_old_config` and `test_deprecation_warning_on_old_config`)

ColtAllen:
Adding additional deprecations to this; will update as PRs are merged.

- [ ] `clv.utils.clv_summary`

ColtAllen:
https://github.com/pymc-labs/pymc-marketing/issues/2158 (deprecate unused distribution blocks)

ColtAllen:
Remaining CLV deprecations before 1.0. These three already raise warnings, and the first item on this list is described in the original message of this issue:

- [old prior dict keys](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/models/basic.py#L49) in CLV model configs
- [Plotting parameter deprecation](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/plotting.py#L460)
- [`clv_model.fit(fit_method)`](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/clv/models/basic.py#L133) parameter

**In [version 1.0](https://github.com/pymc-labs/pymc-marketing/issues/2108), all five remaining deprecations described here and in previous two messages will neither raise warnings nor support backwards compatibility.**

ColtAllen:
`CLVModel.build_with_data` (and the methods it calls) should also be removed when the old API is fully deprecated because it has some redundancies with `ModelIO` for model loading.

ColtAllen:
> CLVModel.build_with_data (and the methods it calls) should also be removed when the old API is fully deprecated because it has some redundancies with ModelIO for model loading.

Addressed in https://github.com/pymc-labs/pymc-marketing/pull/2723

> https://github.com/pymc-labs/pymc-marketing/issues/2157 (deprecate ShiftedBetaGeoModelIndividual)

Done

PR #2576, opened 2026-05-15
refactor(clv): drop _prior suffix backwards-compat shim

anevolbap:
## Summary

Removes the `_prior`-suffix backwards-compat shim from `CLVModel`. It was added in #1498 as a v1.0 deprecation. #1516 is the CLV-deprecations umbrella; this PR covers the first item there.

Retargeted from `main` to `v1.0.0` and rebased onto it, since this is a hard breaking change meant for the v1.0 cut. On `v1.0.0` the pymc-6 migration (#2616) already renamed the covariate convergence-test keys, so the remaining changes are just the shim removal, the two deprecation tests, and one dev notebook.

Three sites removed in `pymc_marketing/clv/models/basic.py`:
- the `deprecated_keys` rename loop in `__init__`
- the `_rename_posterior_variables` call in `build_from_idata`
- the `_rename_posterior_variables` method itself

Two deprecation tests dropped in `tests/clv/models/test_basic.py` (`test_backwards_compatibility_with_old_config`, `test_deprecation_warning_on_old_config`).

`docs/source/notebooks/clv/dev/bg_nbg_covariates_test_issues.ipynb` still passed `a_prior`/`b_prior`, relying on the shim's silent rename. Renamed those keys to drop the suffix.

## Breaking change

Users still passing `_prior`-suffixed `model_config` keys now get a silent fallback to defaults instead of a rename, and saved netCDFs with `_prior`-suffixed posterior vars will not load cleanly. This is intended for v1.0.

## Remaining CLV deprecations

Tracked in #1516 and out of scope here: `clv.utils.clv_summary`, `ShiftedBetaGeoModelIndividual` (#2157), unused distribution blocks (#2158), plotting parameter, `fit(fit_method=...)`, `CLVModel.build_with_data`. Each lands as a separate PR.

## Test plan

- `ruff check` on touched files: pass
- `pytest tests/clv/` on pymc 6.0.1 / arviz 1.2.0: 317 passed, 17 skipped


ColtAllen (approved):
Thanks for opening this @anevolbap ! I'm about to remove the remaining deprecations, so let's get this merged.

PR #2580, opened 2026-05-16
fix(customer_choice): distinguish products in MVITS HDI legend labels

### LINKED ISSUE #1810: improve legend labels for `plot_causal_impact_sales`
We can see in the notebook here: 

https://www.pymc-marketing.io/en/latest/notebooks/customer_choice/mv_its_saturated.html#scenario-1-own-versus-competitor-sales

that the legend labels are all the same for each color

Hard to distinguish with many like with `model3.plot_causal_impact_sales()` call

anevolbap:
Closes #1810

In `MVITS.plot_fit`, `plot_counterfactual`, and `plot_causal_impact_sales`, every product's posterior predictive HDI band was labeled with the same string `"Posterior predictive (HDI)"`, so the legend could not tell the products apart (visible in the notebook output linked from #1810).

The issue only named `plot_causal_impact_sales`, but the same bug existed in `plot_fit` and `plot_counterfactual`. Fixing all three keeps them in line with `plot_causal_impact_market_share`, which already prefixes the product name.


juanitorduz (approved):
Thanks @anevolbap

PR #2583, opened 2026-05-17
feat(mmm): generalize supplementary data persistence via MuEffect.idata_groups()

williambdean:
Replace the hardcoded `isinstance(effect, EventAdditiveEffect)` check in `MMM.save()` with a generic `idata_groups()` hook on `MuEffect`.

Any custom `MuEffect` subclass that needs to persist DataFrames (or other non-JSON-serializable data) alongside the model can now override `idata_groups()` to declare what goes into InferenceData — no need to modify `MMM.save()`.

```python
class MyCustomEffect(MuEffect):
    df: pd.DataFrame

    def idata_groups(self) -> dict[str, xr.Dataset]:
        return {"supplementary_data_my_effect": xr.Dataset.from_dataframe(self.df)}
```

The load side was already extensible via custom deserializer + `DeserializationContext` — this closes the gap on the save side.

### Backwards compatibility

Fully compatible — old saved models load fine, new saved models load in old code, and existing custom MuEffects that don't override `idata_groups()` silently inherit the default empty dict.

### Testing

Added a module-level `_CustomEffectWithSuppData` that exercises the generic `idata_groups()` path through a full save/load round-trip, proving the pattern works for custom effects without any `isinstance` checks. A pytest fixture with yield handles registration cleanup to avoid test pollution.


juanitorduz (approved):
super cool!

PR #2586, opened 2026-05-17
docs(mmm): document parameters on Saturation/Adstock wrapper classes

### LINKED ISSUE #1188: `LogisticSaturation` class does not document all of it's parameters
The logistic saturation function seems to come in 2 varieties. There is the `LogisticSaturation` class defined [here](https://github.com/pymc-labs/pymc-marketing/blob/e80ea88f4f8f4ba3f33641f13f4f350f679656d1/pymc_marketing/mmm/components/saturation.py#L178C7-L178C25) and then there is the `logistic_saturation` function defined [here](https://github.com/pymc-labs/pymc-marketing/blob/e80ea88f4f8f4ba3f33641f13f4f350f679656d1/pymc_marketing/mmm/transformers.py#L451).  The former has no documentation of parameters (though there is a brief vignette?) and instead says it is a "Wrapper around logistic saturation function" and see the documentation on the latter.  The problem is that there the former takes a parameter (`beta`) that is not found anywhere in the latter.  And the vignette doesn't help because it passes in no parameter values.  At best, this seems very confusing.

Are there other functions and wrappers that behave like this?

anevolbap:
Closes #1188.

Each Saturation/Adstock wrapper in `pymc_marketing.mmm.components.{saturation,adstock}` only pointed at the underlying transformer for parameter docs, but most wrappers add a `beta` scaling parameter that the underlying transformer does not take. The result: e.g. `LogisticSaturation` accepts `beta` with no documentation anywhere on the page.

Per the plan agreed in the issue thread:

- Added a Notes section in `saturation.py`'s module docstring listing which wrappers introduce an extra scaling parameter (`beta`) and which ones expose the asymptote through a parameter of the underlying function (`alpha`, `b`, `sigma`).
- Added a numpydoc `Parameters` section to each wrapper class (9 in `saturation.py`, 6 in `adstock.py`) listing the function arguments and their default priors.

No behavior change. Pure docstrings. `pre-commit` and `sphinx-build -W` clean locally.

### DIFF
--- a/pymc_marketing/mmm/components/adstock.py
+++ b/pymc_marketing/mmm/components/adstock.py
@@ -196,7 +196,14 @@ class AdstockTransformation(Transformation):
 class BinomialAdstock(AdstockTransformation):
     """Wrapper around the binomial adstock function.
 
-    For more information, see :func:`pymc_marketing.mmm.transformers.binomial_adstock`.
+    Calls :func:`pymc_marketing.mmm.transformers.binomial_adstock` with the wrapper's
+    ``l_max``, ``normalize`` and ``mode`` settings.
+
+    Parameters
+    ----------
+    alpha : tensor
+        Retention rate of the ad effect; must be between 0 and 1. Default prior:
+        ``Prior("Beta", alpha=1, beta=3)``.
 
     .. plot::
         :context: close-figs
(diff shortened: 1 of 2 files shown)


williambdean:
Thanks @anevolbap

williambdean (approved):

PR #2591, opened 2026-05-17
fix(mmm): correct return type on sample_posterior_predictive and sample_response_distribution

### LINKED ISSUE #2582: update sample_response_distribution return type
Out-of-scope nit: this method is annotated `-> az.InferenceData` (line 3969), but the runtime return is `xr.Dataset` (the `az.extract(...).merge(constant_data).merge(_dataset)` chain produces a Dataset, confirmed at runtime). Same story for `MMM.sample_posterior_predictive` (`-> xr.DataArray`, actually returns a Dataset since `var_names` requests multiple vars). Happy to fix in place here, or punt to a separate PR. Preference?

_Originally posted by @anevolbap in https://github.com/pymc-labs/pymc-marketing/pull/2578#discussion_r3253029303_

anevolbap:
Closes #2582.

Two methods in `pymc_marketing/mmm/mmm.py` advertised the wrong return type:

- `MMM.sample_posterior_predictive` was annotated `-> xr.DataArray`. It returns `az.extract(post_pred, "posterior_predictive", combined=combined)` with no `var_names` filter, which is an `xr.Dataset`.
- `BudgetOptimizerWrapper.sample_response_distribution` was annotated `-> az.InferenceData`. Its return chain is `sample_posterior_predictive(...).merge(constant_data).merge(_dataset)`, which is also an `xr.Dataset`.

Updates both annotations and the matching "Returns" sections in the docstrings. No behavior change.

----
📚 Documentation preview 📚: https://pymc-marketing--2591.org.readthedocs.build/en/2591/

### DIFF
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -2531,7 +2531,7 @@ class MMM(RegressionModelBuilder):
         include_last_observations: bool = False,  # type: ignore
         clone_model: bool = True,  # type: ignore
         **sample_posterior_predictive_kwargs,  # type: ignore
-    ) -> xr.DataArray:
+    ) -> xr.Dataset:
         """Sample from the model's posterior predictive distribution.
 
         Parameters
@@ -2552,7 +2552,7 @@ class MMM(RegressionModelBuilder):
 
         Returns
         -------
-        xr.DataArray
+        xr.Dataset
             Posterior predictive samples.
         """
         # Update model data with xarray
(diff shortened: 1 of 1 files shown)


williambdean (approved):
Thanks @anevolbap

anevolbap:
Note on why CI did not catch this:

The mypy config in `pyproject.toml`:

```toml
[tool.mypy]
files = "pymc_marketing/*.py"
```

uses a non-recursive glob, so `pymc_marketing/mmm/mmm.py` is excluded from type checking. The pre-commit `mypy` hook passes vacuously on this file.

Even if the file were in scope, `arviz.extract` has no return annotation, so mypy treats its return as `Any`. `Any` satisfies any annotation, including the incorrect `xr.DataArray` / `az.InferenceData` ones.

Widening to `pymc_marketing/**/*.py` is not a side fix: running mypy across the full package today surfaces around 1,264 errors, of which only 40 are `[return-value]` (the category this PR belongs to). The rest is PyTensor `[operator]` noise (531), missing third-party stubs (143), and `[union-attr]` (94). Tightening the gap is a separate infra discussion, not in scope here.

juanitorduz (approved):
Thanks!

PR #2594, opened 2026-05-18
feat(mlflow): resolve `_log__` parameters in create_log_callback

### LINKED ISSUE #1615: Automatically handle `<variable>_log__` and common transformations
This would be nice. Good first start is `<variable>_log__` only

https://github.com/pymc-labs/pymc-marketing/blob/06439db8d2dcf7c8a9a2fa74794b030671e20cfa/pymc_marketing/mlflow.py?plain=1#L296-L301

At the moment, the user needs to know how the sampler will handle the variable

anevolbap:
## Summary

Auto-resolve `_log__` parameter names in `create_log_callback`. Pass `sigma` for a `HalfNormal` and the callback maps it to `sigma_log__` from `draw.point` on the first draw, then caches the mapping for the rest of sampling. Existing callers passing `sigma_log__` keep working because the resolver prefers exact matches.

Scope is `_log__` only per the issue's "good first start" note. The comment block on `_TRANSFORM_SUFFIXES` documents the two caveats for follow-up PRs: append only scalar-valued transforms (`mlflow.log_metric` is scalar-only), and custom user transforms need a model handle to resolve (out of scope here). An inline comment on the `log_metric` call flags the scalar-only constraint for future maintainers.

Closes #1615.

## Test plan

- [x] `pytest tests/test_mlflow.py -k "resolve or logging_callback or callback_raises"` (6 tests) passes locally.
- [x] `test_logging_callback_resolves_log_transform`: model with `HalfNormal("sigma")`, asks for `parameters=["mu", "sigma"]`, asserts metrics under `chain_0/mu` and `chain_0/sigma`.
- [x] Unit tests on the resolver: exact-match wins, `_log__` fallback, KeyError otherwise.
- [x] `pre-commit` clean.


anevolbap:
This targets `main`, but it merges cleanly onto `v1.0.0` and `tests/test_mlflow.py` passes there. The changes are all in `create_log_callback` and a new `_resolve_parameter` helper, clear of the `InferenceData` to `DataTree` migration on `v1.0.0`, so there is no conflict. Happy to retarget to `v1.0.0` if you want it in the release, otherwise fine to leave it on `main`.

williambdean:
Lets do all new features into v1.0.0

anevolbap:
Done, retargeted to `v1.0.0`. CI is green on the new head.

williambdean (approved):
Thanks for the adjustments! @anevolbap

PR #2606, opened 2026-05-27
Freeze MMM model before sampling

ricardoV94:
Together with https://github.com/pymc-devs/pytensor/pull/2175  I get a 1.5x speedup MCMC sampling in the demo MMM with default nutpie+numba. On CPU it's better than JAX

The PyTensor improvement part will only be seen when we update the dependency ofc.

### DIFF
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -193,6 +193,7 @@ import pymc as pm
 import pymc.dims as pmd
 import xarray as xr
 from pydantic import ConfigDict, Field, InstanceOf, StrictBool, validate_call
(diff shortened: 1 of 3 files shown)


juanitorduz:
> Together with [pymc-devs/pytensor#2175](https://github.com/pymc-devs/pytensor/pull/2175) I get a 1.5x speedup MCMC sampling in the demo MMM with default nutpie+numba. On CPU it's better than JAX
> 
> The PyTensor improvement part will only be seen when we update the dependency ofc.

This looks great! There is a test failing (and thats it :D)

ricardoV94:
The test was the one WRONG hahaha

juanitorduz (approved):
Amazing! Thanks @ricardoV94

PR #2616, opened 2026-06-04
chore: migrate to pymc>=6.0.0, arviz>=1.1.0, pytensor>=3.0.0

williambdean:
Migrate from `arviz.InferenceData` to `xr.DataTree`, upgrade to
`pymc>=6.0.0`, `arviz>=1.1.0`, `pytensor>=3.0.0`.

### Dependencies
- `pymc>=6.0.0`, `arviz>=1.1.0`, `pytensor>=3.0.0`
- `pymc-extras>=0.11.0`

### Source code (~44 files)
- All `az.InferenceData` → `xr.DataTree` type annotations
- `idata.extend()` → `idata.update()`, `add_groups()` → bracket assignment
- `az.plot_*` → `azp.plot_*` (arviz_plots), `pm.plot_trace` → `azp.plot_trace`
- DataTree child variable access: `.posterior.VAR` → `.posterior["VAR"]` (25 fixes)
- `az.hdi()`: `hdi_prob=` → `prob=`, `ci_bounds` → `ci_bound`, `"higher"` → `"upper"`
- `shifted_beta_geo.py`: restore `pm.sample_posterior_predictive()` with `sample_vars`
- `plot.py`: DataTree→Dataset guard for `uplift_curve`/`marginal_curve`, `.dataset` for reads
- `pytensor`: import paths restructured, `samples` → `draws`
- All docstrings updated for new API

### Notebooks (54/54 pass locally)
- arviz_plots 2.x API: `ax=`, `model_names=`, `figure=fig` removed
- `az.plot_posterior` → `azp.plot_dist`, `az.plot_hdi` → manual HDI + `fill_between`
- `az.rcParams`/`azp.rcParams` → `arviz_base.rcParams`
- DataTree access in notebook cells
- `az.style.use("arviz-darkgrid")` → `"arviz-vibrant"


williambdean:
Maybe the arviz plots isnt installed
https://github.com/pymc-labs/pymc-marketing/actions/runs/26949882488/job/79512149716#step:5:1

pymc_marketing/bass/model.py
```diff
@@ -548,7 +549,7 @@ def sample_posterior_predictive(
         extend_idata: bool = True,
         combined: bool = True,
         **sample_posterior_predictive_kwargs: Any,
-    ) -> xr.Dataset:
+    ) -> xr.DataArray:
```
williambdean: Seems wrong

pymc_marketing/clv/models/basic.py
```diff
@@ -251,7 +246,6 @@ def _fit_approx(
         with self.model:
             approx = pm.fit(
                 method=method,
-                callbacks=[pm.callbacks.CheckParametersConvergence(diff="absolute")],
```
williambdean, line 254: Need to double check what happened with this

pymc_marketing/clv/models/basic.py
```diff
@@ -289,17 +283,17 @@ def _fit_approx(
 
     # TODO: Remove for v1.0
     @classmethod
-    def idata_to_init_kwargs(cls, idata: az.InferenceData) -> dict:
+    def idata_to_init_kwargs(cls, idata: xr.DataTree) -> dict:
         """Create the initialization kwargs from an InferenceData object."""
         kwargs = cls.attrs_to_init_kwargs(idata.attrs)
         if "fit_data" in idata:
-            kwargs["data"] = idata.fit_data.to_dataframe()
+            kwargs["data"] = idata.fit_data.to_dataset().to_dataframe()
```
williambdean: Need to look into this

williambdean:
I will just start touching some of the notebooks  to try to get the run through. Follow up PR can be for closer look

pymc_marketing/bass/model.py
OriolAbril: It is correct, the docstring was right and the type hint was wrong. `pymc.sample_posterior_predictive` returns a Dataset but this is then passed to `arviz.extract` with a string as `var_names` which will return a DataArray.

williambdean: Thanks for confirming @OriolAbril !

pymc_marketing/bass/model.py
```diff
@@ -728,20 +728,20 @@ def fit(  # type: ignore[override]
...
-            self.idata.extend(idata, join="right")
+            self.idata.update(idata)
         else:
             self.idata = idata
 
         self.idata["posterior"].attrs["pymc_marketing_version"] = __version__
 
-        if "fit_data" in self.idata:
-            del self.idata.fit_data
+        if "/fit_data" in self.idata.groups:
```
OriolAbril: mostly fyi, unless the datatree has nested groups this is equivalent to `"fit_data" in self.idata.children`. If there are nested groups `.children` will only show the level immediately below whereas `.groups` will show everyithing with `/` indicators mid element for deeper levels.

pymc_marketing/bass/model.py
```diff
@@ -728,20 +728,20 @@ def fit(  # type: ignore[override]
...
+            self.idata = self.idata.drop_nodes("fit_data")
 
-        self.idata.add_groups(fit_data=ds)
+        self.idata["/fit_data"] = ds
```
OriolAbril: also fyi, the leading `/` is optional, `dt["fil_data"]` and `dt["/fit_data"]` will create the same group.

pymc_marketing/customer_choice/mv_its.py
```diff
@@ -491,17 +492,19 @@ def plot_fit(
         x = self.X.index.values  # type: ignore
         existing_products = self.coords["existing_product"]
         for i, existing_product in enumerate(existing_products):
-            az.plot_hdi(
-                x,
+            hdi = az.hdi(
```
OriolAbril: I would recommend computing the hdi outside the loop, then subset it for plotting.

pymc_marketing/clv/models/basic.py
OriolAbril: Here it would be better to use only `.dataset` to get a view of the `fit_data` group as a dataset instead of a datatree. `to_dataset()` will create a shallow copy of the group as a dataset which allows for example to edit its contents.

pymc_marketing/data/idata/utils.py
```diff
@@ -29,48 +28,38 @@
...
-        )
-        filtered_idata.attrs = idata.attrs.copy()
-        idata = filtered_idata
+        filtered = {
+            f"/{g}": idata[f"/{g}"].to_dataset()
+            for g in groups
+            if f"/{g}" in idata.groups
+        }
+        idata = xr.DataTree.from_dict(filtered)
+        idata.attrs = idata.attrs.copy() if hasattr(idata, "attrs") else {}
```
OriolAbril: If there is only one level of nesting like there used to be in InferenceData, all this should be equivalent to `dt.filter(lambda g: g.name in groups)`

pymc_marketing/data/idata/utils.py
```diff
@@ -79,62 +68,66 @@ def idata_from_zarr(store: str | Path) -> az.InferenceData:
...
-    for group_name in idata.groups():
-        group = getattr(idata, group_name)
-        if "date" in group.dims:
-            filtered_groups[group_name] = group.sel(**date_slice)
-        else:
-            filtered_groups[group_name] = group
+    for path in idata.groups:
+        if path == "/":
+            continue
+        group_name = path.lstrip("/")
```
OriolAbril: Here it would be much more convenient to loop over `.children`

OriolAbril (commented):
I started taking a quick look while traveling, left some datatree usage comments mostly. There might be some extra interesting nuggets in https://python.arviz.org/en/stable/user_guide/migration_guide.html#datatree

williambdean:
Filed upstream bug report for the remaining Numba JIT failures: https://github.com/pymc-devs/pytensor/issues/2201

The 4 failing `test_logp_finite` tests are now skipped with a reference to that issue until the fix lands.

juanitorduz:
> Filed upstream bug report for the remaining Numba JIT failures: [pymc-devs/pytensor#2201](https://github.com/pymc-devs/pytensor/issues/2201)
> 
> The 4 failing `test_logp_finite` tests are now skipped with a reference to that issue until the fix lands.

It seems it's fixed! :) Thanks @ricardoV94 👯

juanitorduz:
wow! @williambdean, this looks great! :) Please let me know if you need a review round :)

williambdean:
Aiming for later today or tomorrow. Will ping you!

Notebooks were ☠️

ricardoV94:
Should open issue to update/remove notebook for distributions that no longer exist? ImportError: cannot import name 'ContNonContract' from 'pymc_marketing.clv.distributions' (/Users/will/github/pymc-eco/pymc-marketing/worktrees/pymc6-migrate/pymc_marketing/clv/distributions.py)

pymc_marketing/mmm/dims.py
```diff
@@ -30,6 +30,8 @@ def _vectorize_shape_i(op: Shape_i, node, batched_x):
     [old_x] = node.inputs
     core_ndims = old_x.type.ndim
     batch_ndims = batched_x.type.ndim - core_ndims
+    if isinstance(batched_x, XTensorVariable):
```
ricardoV94: This doesn't sound right, if shape_i was already operating on an XTensorVariable, there's no reason to convert in this vectorize step

pymc_marketing/pytensor_utils.py
```diff
@@ -88,8 +87,10 @@ def _prefix_model(f2, prefix: str, exclude_vars: set | None = None):
             base_to_prefixed[name] = new_name
 
     # Don't rename dimensions that belong to excluded variables
+    from pymc.pytensorf import StringConstant
```
ricardoV94: don't import locally. Also make a note that https://github.com/pymc-devs/pymc/pull/8340 will change the IR of these objects. One of these is getting rid of this StringConstant. I'd suggest upper pinning pymc to the last release and opening an issue to track this

pyproject.toml
```diff
@@ -25,31 +25,30 @@ classifiers = [
...
     "matplotlib>=3.5.1,<3.11",
     "narwhals",
     "numpy>=2.0",
     "pandas",
     "patsy",
     "pydantic>=2.1.0",
-    # NOTE: Used as minimum pymc version with test.yml `OLDEST_PYMC_VERSION`
-    "pymc>=5.28.5,<6.0.0",
-    "pytensor>=2.38.2,<3.0.0",
+    "pymc>=6.0.0",
```
ricardoV94: I suggest alway upper pinning pymc on the intermediate version, that's the one we bump when we do (or notice) breaking changes.

ricardoV94: Can perhaps adapt this script to automatically check if new pymc version was released and try to increase the upper pin: https://github.com/pymc-devs/pytensor/blob/main/.github/workflows/numba-upper-bound.yml

(It hasn't yet fired since it was merged, so it's untested on the pytensor side)

pyproject.toml
```diff
@@ -25,31 +25,30 @@ classifiers = [
 ]
 
 dependencies = [
-    "arviz>=0.13.0,<1.0.0",
+    "arviz>=1.1.0",
```
ricardoV94: take this as an opportunity to major version pin already, so <2.0, and for all libraries. Will be a good think to have done by the time it happens

pyproject.toml
```diff
@@ -208,3 +210,18 @@ ignore_missing_imports = true
 
 [tool.mypy-scipy]
 ignore_missing_imports = true
+
+# Temporary: pin pymc-extras to a specific commit until a release containing
+# the pymc>=6 / DataTree changes is published on PyPI. Pinning the commit keeps
```
ricardoV94, line 215: That's already the case?

ricardoV94:
Can be a follow-up but the two utilities in this module shouldn't be needed anymore: https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/model_graph.py

pytensorf has [get_symbolic_rv_shapes](https://github.com/pymc-devs/pymc/blob/fffac800a9a8b7c125833b8efa4d66af79089cbd/pymc/pytensorf.py#L1027) and `sample_posterior_predictive` has `freeze_vars` that takes care of what the other helper was about

pymc_marketing/data/idata/utils.py
```diff
@@ -29,48 +28,33 @@
...
         Groups to save. If None, all groups are saved.
     """
     if groups is not None:
-        filtered_idata = az.InferenceData(
-            **{group: getattr(idata, group) for group in groups if group in idata}
-        )
-        filtered_idata.attrs = idata.attrs.copy()
-        idata = filtered_idata
+        idata = idata.filter(lambda g: g.name in groups)
+        idata.attrs = idata.attrs.copy() if hasattr(idata, "attrs") else {}
```
cetagostini: `idata_to_zarr(..., groups=[...])` silently drops the **root attrs** (model `id`, version, config).

After `idata = idata.filter(...)` reassigns `idata`, the next line `idata.attrs = idata.attrs.copy()` is a self-copy of the *already-filtered* tree — and `DataTree.filter()` does not carry over the root node's attrs. The sibling helpers in this file (`filter_idata_by_dates`, `aggregate_idata_time`, ...) avoid this by referencing the original `idata` through a separate `result` variable.

Reproducible (on this PR's stack: xarray 2026.4.0, arviz 1.1.0, zarr 3.2):

```python
import numpy as np, xarray as xr, tempfile, os
from pymc_marketing.data.idata.utils import idata_to_zarr, idata_from_zarr

ds = xr.Dataset({"a": ("x", np.arange(3))})
dt = xr.DataTree.from_dict({"/posterior": ds, "/fit_data": ds})
dt.attrs.update(id="MODEL123", version="1.0")

tmp = tempfile.mkdtemp()
idata_to_zarr(dt, os.path.join(tmp, "all.zarr"))                        # all groups
idata_to_zarr(dt, os.path.join(tmp, "sub.zarr"), groups=["posterior"])  # subset

print(idata_from_zarr(os.path.join(tmp, "all.zarr")).attrs)  # {'id': 'MODEL123', 'version': '1.0'}
print(idata_from_zarr(os.path.join(tmp, "sub.zarr")).attrs)  # {}   <-- root attrs lost
```

Because `ModelBuilder.load()` relies on `idata.attrs["id"]`, a model saved to a zarr store with a group subset can't be reloaded. Suggested fix:

```python
if groups is not None:
    attrs = dict(idata.attrs)
    idata = idata.filter(lambda g: g.name in groups)
    idata.attrs = attrs
```

(Also `hasattr(idata, "attrs")` is always `True` for a `DataTree`, so that guard is dead.)

pymc_marketing/mmm/mmm.py
```diff
@@ -1385,7 +1385,7 @@ def compute_counterfactual_contributions_dataset(
         """
         self._validate_idata_exists()
 
-        idata: az.InferenceData = cast(az.InferenceData, self.idata)
+        idata: xr.DataTree = cast(xr.DataTree, self.idata)
         posterior: xr.Dataset = idata.posterior
```
cetagostini: `idata.posterior` returns an `xr.DataTree` node, not an `xr.Dataset`, so this annotation is inaccurate — the Dataset-only operations used afterwards work only because `DataTree` incidentally mirrors part of the `Dataset` API.

```python
import numpy as np, xarray as xr
dt = xr.DataTree.from_dict({"/posterior": xr.Dataset({"a": ("x", np.arange(3))})})
print(type(dt.posterior).__name__)           # DataTree
print(isinstance(dt.posterior, xr.Dataset))  # False
```

Prefer `idata["/posterior"].dataset` to obtain an actual `Dataset` view that matches the annotation.

cetagostini (commented):
Two verified issues:

pyproject.toml
```diff
@@ -25,31 +25,30 @@ classifiers = [
...
+    "pytensor>=3.0.0,<4.0",
     "scikit-learn>=1.1.1",
     "seaborn>=0.12.2",
     "tqdm",
     "xarray>=2024.1.0",
     "xarray-einstats>=0.5.1",
     "pyprojroot",
-    "pymc-extras>=0.9.2,<0.11.0",
-    "preliz>=0.20.0",
+    "pymc-extras>=0.11.0,<0.12.0",
```
cetagostini: The `dependencies` cap here (`pymc-extras>=0.11.0,<0.12.0`) is inconsistent with the `[tool.uv.sources]` git pin below (line ~218), which resolves to **`0.12.2.dev1`** — outside the declared `<0.12.0` range. Verified on a fresh install of this branch:

```
$ uv pip install -e ".[test]"
 - pymc-extras==0.10.0
 + pymc-extras==0.12.2.dev1+gee8cc37df (from git+https://github.com/pymc-devs/pymc-extras.git@ee8cc37)

$ python -c "import importlib.metadata as m; print(m.version('pymc-extras'))"
0.12.2.dev1+gee8cc37df          # <-- violates the declared <0.12.0 cap
```

Two consequences:
1. `uv` overrides the version range via `[tool.uv.sources]`, so the declared `<0.12.0` is misleading.
2. `[tool.uv.sources]` is **uv-only** — a plain `pip install pymc-marketing` ignores it and resolves `pymc-extras<0.12.0` from PyPI (0.11.x), which does not contain the pymc>=6 / DataTree changes this PR depends on. That gives pip users an import/runtime-broken install.

This is the loose end from the earlier thread: if the required changes are already in a published `pymc-extras` release, drop the git source and pin to that release; otherwise the cap should be widened to admit the pinned dev version and the pip divergence called out. Either way the temporary git source needs resolving before release.

cetagostini (commented):
One more verified packaging issue (pip vs uv divergence for pymc-extras), with the install output that demonstrates it.

docs/source/notebooks/clv/dev/continuous_contractual.ipynb
```diff
@@ -12,51 +12,82 @@
    "cell_type": "code",
```
juanitorduz, line 12: <p>This has to be fixed:</p><p> </p><p>ImportError                               Traceback (most recent call last)Cell In[1], line 8</p><p>      4 import arviz_plots as azp</p><p>      5 import numpy as np</p><p>      6 import pymc as pm</p><p>      7 </p><p>----&gt; 8 from pymc_marketing.clv.distributions import ContContract </p><p>ImportError: cannot import name 'ContContract' from 'pymc_marketing.clv.distributions' (/Users/will/github/pymc-eco/pymc-marketing/worktrees/pymc6-migrate/pymc_marketing/clv/distributions.py)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2616/discussion/#comment-3480877078'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

docs/source/notebooks/clv/dev/continuous_non_contractual.ipynb
```diff
@@ -12,18 +12,35 @@
    "cell_type": "code",
```
juanitorduz, line 12: <p>same here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2616/discussion/#comment-3480877086'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

docs/source/notebooks/clv/dev/pareto_nbd_advi.ipynb
```diff
@@ -6,12 +6,12 @@
    "metadata": {},
```
juanitorduz, line 6: <pre class="ql-syntax" spellcheck="false">Line #4.    idata_nutpie = pnbd_nutpie.fit(fit_method="mcmc", nuts_sampler="nutpie")
</pre><p>can we also fix this ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2616/discussion/#comment-3480877091'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

juanitorduz:
@williambdean this looks great! I suggest: you verify/address Carlos ' comments and fix the remaining tests, and merge to the branch `v1.0.0` and then we branch from this one and re-run one notebook at a time and work on small pull requests to find edge cases (and help you parallelize) . WDYT?

williambdean:
Yes, thatd be great!
Let me wrap this up this morning

docs/source/notebooks/clv/dev/continuous_contractual.ipynb
williambdean: I just reverted the dev notebooks — will leave those for future iterations.

docs/source/notebooks/clv/dev/continuous_non_contractual.ipynb
williambdean: I just reverted the dev notebooks — will leave those for future iterations.

docs/source/notebooks/clv/dev/pareto_nbd_advi.ipynb
williambdean: I just reverted the dev notebooks — will leave those for future iterations.

pymc_marketing/data/idata/utils.py
williambdean: Good catch, fixed.

pymc_marketing/mmm/mmm.py
williambdean: Good catch, fixed.

pyproject.toml
williambdean: Good catch, fixed.

pymc_marketing/mmm/dims.py
williambdean: PyTensor 3.x in CI still blocks implicit XTensorVariable → Tensor. The .values conversion is needed for incrementality vectorization. Will revisit when upstream handles it. Please make an upstream issue if you see that is needed, Ricardo.

https://github.com/pymc-labs/pymc-marketing/actions/runs/28235216418/job/83648445118

williambdean:
All review comments addressed. @daimon-pymclabs take another look please.

pymc_marketing/mmm/dims.py
ricardoV94: PyTensor will keep blocking implicit XTensorVariable. But if Shape_i doesn't support XTensorVariables, how comes it shows up here?

juanitorduz (approved):
AMAZING! Let's merge to `v1.0.0` and iterate :) 

(I just re-triggered daimon)

pymc_marketing/model_builder.py
```diff
@@ -1033,30 +1052,22 @@ def fit(  # type: ignore[override]
...
-            idata.posterior = pm.compute_deterministics(
-                idata.posterior, merge_dataset=True
+            idata["/posterior"] = pm.compute_deterministics(
+                idata["/posterior"], merge_dataset=True
             )
 
         if self.idata:
-            self.idata = self.idata.copy()
-            self.idata.extend(idata, join="right")
+            self.idata.update(idata)
```
daimon-pymclabs, line 1060: The `self.idata = self.idata.copy()` that preceded the old `extend(...)` was dropped here, so `self.idata.update(idata)` now mutates `self.idata` in place on a re-fit (same change at L1214 in `approximate_fit`).

The `customer_choice` models (`mixed_logit`, `mnl_logit`, `nested_logit`, `maxdiff`) and `BassModel.fit` all kept the `self.idata = self.idata.copy()` line before `update`. Worth aligning the two so the no-side-effects-on-refit semantics are consistent — either re-add the copy here or drop it from the choice models.

pymc_marketing/customer_choice/mixed_logit.py
```diff
@@ -1144,34 +1143,34 @@ def fit(
                 category=UserWarning,
                 message="The group fit_data is not defined in the InferenceData scheme",
             )
-            self.idata.add_groups(fit_data=fit_data)
+            self.idata["/fit_data"] = fit_data
```
daimon-pymclabs, line 1146: This `catch_warnings` / `filterwarnings("...not defined in the InferenceData scheme")` wrapper is now dead code: that warning was emitted by arviz's `add_groups`, which has been replaced by the plain `self.idata["/fit_data"] = fit_data` assignment below — a DataTree assignment doesn't raise it. `model_builder.py` correctly removed these blocks; the same now-inert wrapper still remains here and in `mnl_logit.py` (L616), `nested_logit.py` (L927), and `maxdiff.py` (L1197). Safe to simplify to the bare assignment.

pymc_marketing/clv/models/shifted_beta_geo.py
```diff
@@ -960,11 +960,14 @@ def _distribution_new_customer(
...
+            post_pred = pm.sample_posterior_predictive(
                 self.idata,
-                var_names=var_names,
+                var_names=list(var_names),
+                sample_vars=["theta", "churn"],
                 random_seed=random_seed,
-            ).posterior_predictive
+            )
+            result = post_pred["posterior_predictive"].dataset
+            return result.stack(sample=("chain", "draw")).reset_index("sample")
```
daimon-pymclabs, line 970: Heads-up on a public output-shape change: `distribution_new_customer_churn_time` and `distribution_new_customer_theta` previously returned a `DataArray` with `(chain, draw, new_customer_id)` dims (straight from `.posterior_predictive`). After `.stack(sample=("chain", "draw")).reset_index("sample")` they now return `(sample, new_customer_id)`.

If intended (it does match the "draws" wording in the docstrings) that's fine — just flagging that any downstream code / tests / notebooks relying on the `chain`/`draw` dims here will need updating.

daimon-pymclabs (commented):
Another review round, focused on the source `.py` changes (skipped the notebooks). The DataTree migration reads cleanly and consistently across the board. A few non-blocking observations below — mostly consistency cleanups plus one output-shape heads-up on `shifted_beta_geo`. None of these should hold up merging to `v1.0.0` and iterating.

Issue #1706, opened 2025-05-23
Dual y-axes for the "Innovators vs Imitators" plots in bass notebook


williambdean:
Looks better.  At some point (not necessarily now), I might suggest using 2 different y-axes for the "Innovators vs Imitators" plots.  Because the counts and the cumulative counts are on such different scales, it might make sense to put them on 2 different axes so that you can actually see the innovator/imitator counts.  Right now only the cumulative plot is effectively visible.

_Originally posted by @cluhmann in https://github.com/pymc-labs/pymc-marketing/pull/1702#issuecomment-2904630387_

williambdean:
Connected: #2589 — Plotting methods for BassModel (this should be addressed there)

PR #2628, opened 2026-06-12
feat(bass): plotting methods for BassModel

### LINKED ISSUE #2589: Plotting methods for BassModel
Add dedicated plotting methods to `BassModel` for visualizing adoption curves and diagnostics.

### Motivation

The existing Bass notebook shows several useful plots, but users must build them manually. Adding `model.plot_*` methods (consistent with MMM and CLV patterns) makes analysis easier and enables MLflow integration via `mlflow.log_figure`.

### Proposed plots

| Method | Description |
|--------|-------------|
| `plot_adoption_curve` | Adoption curve: observed vs. predicted adopters over time with HDI |
| `plot_cumulative` | S-curve: cumulative adoption with HDI |
| `plot_decomposition` | Breakdown into innovators vs. imitators (see #1706 for dual y-axes) |
| `plot_peak` | Posterior distribution of peak adoption time |

### Design notes

- Follow the pattern in `pymc_marketing/clv/plotting.py` (class methods that accept `ax` and `figsize`)
- Support both single-product and multi-product data (with `product` parameter to select)
- Use ArviZ for HDI computation where applicable
- Reuse `F` and `f` functions for theoretical curves

### Related

- #1706 - Dual y-axes for innovator vs. imitator plots
- #2588 - MLflow integration depends on this

anevolbap:
Closes #2589
Related to #1706

Retargeted to `v1.0.0`.

Adds four plotting methods to `BassModel`, built on `pymc_marketing.plot.plot_curve`:

- `plot_adoption_curve`: posterior new adopters per period with HDI, observed counts overlaid.
- `plot_cumulative`: cumulative adoption S-curve with HDI, observed cumulative overlaid.
- `plot_decomposition`: innovators and imitators per period on the left y-axis, cumulative adoption on a twin right y-axis (the dual-axes design from #1706).
- `plot_peak`: posterior of the peak adoption time.

All live in `pymc_marketing.bass.plotting` and are exposed as `plot_*` methods on `BassModel`. Each returns `(fig, axes)`, so `mlflow.log_figure` works with the autolog workflow in #2624.

v1 adaptations:

- `plot_peak` uses `arviz_plots.plot_dist` (arviz 1.2 removed `arviz.plot_posterior`), faceting by product for multi-product models.
- Fixes a v1 regression in `BassModel.fit`: under arviz 1.2 `InferenceData` subclasses `xarray.DataTree`, so `idata.posterior = compute_deterministics(...)` set a shadowing attribute and dropped the deterministics (`adopters`, `innovators`, `imitators`, `peak`) from the fitted posterior. The plots need those, so `fit` now writes to the `idata["posterior"]` node.

### DIFF
--- a/pymc_marketing/bass/plotting.py
+++ b/pymc_marketing/bass/plotting.py
@@ -1,0 +1,350 @@
+#   Copyright 2022 - 2026 The PyMC Labs Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""Plotting functions for the Bass diffusion model.
+
+Each function takes a fitted :class:`~pymc_marketing.bass.model.BassModel`
+and returns a ``(Figure, ndarray of Axes)`` tuple, following the convention
+of :func:`pymc_marketing.plot.plot_curve`. Multi-product models plot one
+subplot per product; pass ``product`` to select a single one.
+
+The functions are also exposed as ``plot_*`` methods on
+:class:`~pymc_marketing.bass.model.BassModel`.
+"""
+
+from typing import TYPE_CHECKING, Any
+
+import arviz as az
+import matplotlib.pyplot as plt
+import numpy as np
+import numpy.typing as npt
+import xarray as xr
+from matplotlib.axes import Axes
+from matplotlib.lines import Line2D
+
+from pymc_marketing.plot import plot_curve
+
+if TYPE_CHECKING:
+    from pymc_marketing.bass.model import BassModel
+
+__all__ = [
+    "plot_adoption_curve",
+    "plot_cumulative",
+    "plot_decomposition",
+    "plot_peak",
+]
+
+
+def _select_product(da: xr.DataArray, product: str | None) -> xr.DataArray:
+    if product is None:
+        return da
+    if "product" not in da.dims:
+        raise ValueError(
+            "The model has no 'product' dimension. "
+            "Remove the product argument for single-product models."
+        )
+    return da.sel(product=product)
+
+
+def _get_observed(model: "BassModel", product: str | None) -> xr.DataArray | None:
+    idata = model.idata
+    if idata is None or "fit_data" not in idata or "observed" not in idata.fit_data:
+        return None
+    return _select_product(idata.fit_data["observed"], product)
+
+
+def _overlay_observed(
+    observed: xr.DataArray,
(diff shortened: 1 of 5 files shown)


anevolbap:
The three Bass PRs are now retargeted to `v1.0.0` and adapted to the new stack (pymc 6, pytensor 3, arviz 1.2, arviz-plots 1.2):

- #2624 (mlflow): ports as-is; the autolog test forces `nuts_sampler="pymc"` since the pymc6 default (nutpie) does not log `sampling_time` / `time_per_draw`.
- #2628 (plots): `plot_peak` moves to `arviz_plots.plot_dist` (arviz 1.2 dropped `arviz.plot_posterior`); the other three plots are unchanged since `plot.py` stayed on matplotlib.
- #2630 (notebook): rebuilt on the `BassModel` workflow and re-executed on v1 with real sampling.

While adapting the plots I found a v1.0.0 regression in `BassModel.fit`: it drops the deterministics (`adopters`, `innovators`, `imitators`, `peak`) from the fitted posterior. Under arviz 1.2 `InferenceData` subclasses `xarray.DataTree`, so `idata.posterior = pm.compute_deterministics(...)` sets a shadowing attribute and leaves the actual node untouched. Any `BassModel.fit` followed by reading `model.posterior["adopters"]` (or the new plots) hits this. v1 CI does not catch it because the only deterministics test uses `create_bass_model` + `sample_prior_predictive`, never `BassModel.fit`. Fixed here by writing to `idata["posterior"]`. Happy to split that into a standalone fix if you prefer it lands independently.

Suggested merge order: #2624 and #2628 first (independent), then rebase and merge #2630.

williambdean:
Had some comments on #2630. If those are supported in the API already, then this is good to go

pymc_marketing/bass/plotting.py
```diff
@@ -0,0 +1,361 @@
...
+
+__all__ = [
+    "plot_adoption_curve",
+    "plot_cumulative",
+    "plot_decomposition",
+    "plot_peak",
+]
+
+
+def _select_product(da: xr.DataArray, product: str | None) -> xr.DataArray:
```
williambdean: This might be a bit too specific. What are your thoughts?

Maybe fine for a first iteration...

We assume some coordinates like "date" or "T". However, "product" might be pushing it.

williambdean (commented):
Comment on the naming of "product'

pymc_marketing/bass/plotting.py
anevolbap: I gave the generalization a try (e86ad570). The plots now infer the faceting dim from the data instead of assuming `product`, so a model with e.g. a `region` dimension works with no extra args. Added an optional `dim` to override it and a `coord` to select one series, plus tests.

anevolbap:
> Had some comments on #2630. If those are supported in the API already, then this is good to go

They are, nothing new was needed here: `subplot_kwargs` (including `ncols`) flows through to `plot_curve` for `plot_adoption_curve`/`plot_cumulative`/`plot_decomposition`, and `plot_peak` forwards `col_wrap` to `plot_dist`. The legend placement was an internal tweak to `plot_decomposition`, and the ipywidgets warning was just the run env (it is a declared dep). So the plotting API already covers all of it.

williambdean (approved):
Thanks for the edits @anevolbap

PR #2623, opened 2026-06-11
[HOT FIX] Dependencies: Upper bound pymc-extras

juanitorduz:
Upper bound pymc-extras

----
📚 Documentation preview 📚: https://pymc-marketing--2623.org.readthedocs.build/en/2623/

PR #2570, opened 2026-05-14
refactor(budget-optimizer): collapse default_constraints into constraints

anevolbap:
Closes #1416. See the issue thread for design discussion. Notebook updates land in a follow-up PR.

## Summary
- Rename `custom_constraints` to `constraints` on `BudgetOptimizer`.
- Remove `default_constraints`. Empty `constraints` auto-adds the default sum constraint; non-empty means the caller is in charge. Pass `build_default_sum_constraint()` explicitly to keep the default alongside customs.
- Remove `default_constraints` from `MMM.optimize_budget`.
- Simplify `set_constraints` to `set_constraints(constraints)`.

Hard removal, no deprecation cycle: v1.0 is the next release and carries breaking changes anyway (per @juanitorduz's review).

**BREAKING CHANGE**: `custom_constraints` and `default_constraints` are removed. Use `constraints` instead.

## Test plan
- [x] `tests/mmm/test_budget_optimizer.py` (22 pass)
- [x] `tests/mmm/test_budget_optimizer_mmm.py` + `tests/mmm/test_cost_per_unit.py` (77 pass)
- [x] pre-commit clean
- [ ] Notebook updates (follow-up PR)

PR #2629, opened 2026-06-12
fix: pin pymc-extras<0.11.0 and matplotlib<3.11 on v1.0.0 branch

cetagostini:
## Summary
Backports two dependency pins to the `v1.0.0` branch so the "build source distribution" job (and any fresh install) works again. Both mirror fixes on `main`; lockfile metadata updated to match (resolved versions were already compatible, so only the requirement specifiers change).

- `pymc-extras>=0.9.2,<0.11.0` (from #2623, merged on `main`): pymc-extras 0.11.0+ requires `arviz>=1.1`, conflicting with our `arviz<1.0.0` pin — pip fails with `ResolutionImpossible`.
- `matplotlib>=3.5.1,<3.11` (from #2626, approved on `main`): matplotlib 3.11.0 removed the deprecated `matplotlib.style.core` shim that arviz-plots 1.1.0 accesses at import time, so importing `pymc_marketing` fails with `AttributeError`. Droppable once an arviz-plots release includes arviz-devs/arviz-plots#466.

These currently break every PR targeting `v1.0.0` (e.g. #2570).

## Test plan
- [x] `uv lock` resolves cleanly (only requirement specifiers change in the lockfile)
- [x] pre-commit clean
- [ ] "build source distribution" job green on this PR

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 dependencies = [
     "arviz>=0.13.0,<1.0.0",
     "arviz-plots[matplotlib]>=1.0.0,<2.0",
-    "matplotlib>=3.5.1",
+    "matplotlib>=3.5.1,<3.11",
     "narwhals",
     "numpy>=2.0",
     "pandas",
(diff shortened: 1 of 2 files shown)


williambdean (approved):
LGTM

PR #2632, opened 2026-06-14
Warn on unused model_config keys

### LINKED ISSUE #1256: Raise or warn when model_config keys are not used
I want all of the items in the `model_config` parameter at initialization to be used and warn / raise if something was typed wrong / not used

MECoban:
Closes #1256.

## Problem

`ModelBuilder.__init__` merges user config with `self.default_model_config | model_config`. Any key the user passes that is **not** part of `default_model_config` is silently merged in and then ignored by the model. A typo like `model_config={"alphaa": ...}` (instead of `"alpha"`) leaves the user wondering why their model didn't change.

## Change

After the merge, warn about keys that the model does not use:

```python
unused_model_config_keys = set(model_config) - set(default_model_config)
if unused_model_config_keys:
    warnings.warn(
        "The following model_config keys are not used by the model "
        f"and will be ignored: {sorted(unused_model_config_keys)}. "
        f"Valid keys are: {sorted(default_model_config)}.",
        UserWarning,
        stacklevel=2,
    )
```

- Chose a **`UserWarning`** rather than raising, to stay backward compatible (existing code passing extra keys keeps working). Happy to switch to raising a `ModelConfigError` if you'd prefer stricter behavior.
- `default_model_config` is cached in a local so the property isn't evaluated twice.
- Updated the `model_config` docstring to mention the warning.

## Tests

Added to `tests/test_model_builder.py`:
- `test_model_config_warns_on_unused_keys` — unknown key triggers the warning.
- `test_model_config_no_warning_for_valid_keys` — valid-only config stays silent.

This assumes `default_model_config` enumerates every key a model uses (per its docstring). If some model intentionally accepts keys outside it, the warning is harmless (no break) and we can refine the check.

----
📚 Documentation preview 📚: https://pymc-marketing--2632.org.readthedocs.build/en/2632/

### DIFF
--- a/pymc_marketing/model_builder.py
+++ b/pymc_marketing/model_builder.py
@@ -603,6 +603,8 @@ class ModelBuilder(ABC, ModelIO):
         model_config : Dictionary, optional
             dictionary of parameters that initialise model configuration.
             Class-default defined by the user default_model_config method.
+            A ``UserWarning`` is raised for any key not present in
+            ``default_model_config``, since such keys are ignored by the model.
         sampler_config : Dictionary, optional
             dictionary of parameters that initialise sampler configuration.
             Class-default defined by the user default_sampler_config method.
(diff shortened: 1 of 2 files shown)


juanitorduz (approved):
Thnks @MECoban !

PR #2638, opened 2026-06-22
fix(mmm): make WeibullPDFAdstock differentiable for NUTS

### LINKED ISSUE #2637: WeibullPDFAdstock cannot be sampled with NUTS: `Min` Op has no gradient
## Description

`WeibullPDFAdstock` (and `weibull_adstock(..., type=WeibullType.PDF)`) produces a
model that **cannot be sampled with NUTS**. Any model that puts priors on the
Weibull adstock parameters fails when PyMC tries to build the log-probability
gradient:

```
NotImplementedError: grad not implemented for Op Min{axis=0}
```

Prior predictive sampling works (it needs no gradients), so the problem only
shows up at fit time — and because PyMC silently falls back from NUTS, the
failure can be confusing (e.g. `nutpie` aborts with "Model can not be sampled
with NUTS alone").

## Root cause

In PDF mode the kernel is min-max normalized
([`pymc_marketing/mmm/transformers.py`](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/transformers.py)):

```python
w = (w - w.min(dim=kernel_dim)) / (
    w.max(dim=kernel_dim) - w.min(dim=kernel_dim)
)
```

The xtensor `.min()` reduction lowers to PyTensor's `Min` Op, which has **no
gradient implemented** (only `Max` does). PyMC therefore cannot build
`d(logp)/d(adstock_lam)` and `d(logp)/d(adstock_k)`, so the model is not
NUTS-samplable. `WeibullCDFAdstock` is unaffected (it uses `cumprod`, not `min`).

## Reproducible example

```python
import numpy as np
import pymc as pm
from pytensor.xtensor import as_xtensor
from pymc_marketing.mmm.components.adstock import WeibullPDFAdstock

x = np.ones(20)
y_obs = np.random.default_rng(0).normal(size=20)
with pm.Model(coords={"time": range(20)}) as model:
    mu = WeibullPDFAdstock(l_max=10).apply(as_xtensor(x, dims=("time",)), core_dim="time")
    sigma = pm.HalfNormal("sigma", 1)
    pm.Normal("obs", mu=mu.values, sigma=sigma, observed=y_obs, dims=("time",))

model.compile_dlogp()   # raises NotImplementedError: grad not implemented for Op Min{axis=0}
```

Swapping `WeibullPDFAdstock` for `WeibullCDFAdstock` above compiles fine.

## Expected behavior

`WeibullPDFAdstock` models should be samplable with NUTS, like every other
adstock transformation.

## Proposed fix

Express the minimum via the grad-supported `Max` Op — `-((-w).max(...))` is
mathematically identical and differentiable. PR incoming.

## Environment

- pymc-marketing 0.19.4
- pymc 5.28.5
- pytensor 2.38.3
- Python 3.12

hermesdt:
## Description

`WeibullPDFAdstock` produced models that could not be sampled with NUTS. In PDF
mode the kernel is min-max normalized:

```python
w = (w - w.min(dim=kernel_dim)) / (w.max(dim=kernel_dim) - w.min(dim=kernel_dim))
```

The xtensor `.min()` reduction lowers to PyTensor's `Min` Op, which has no
gradient implemented (only `Max` does). PyMC therefore could not build
`d(logp)/d(adstock_lam)` / `d(logp)/d(adstock_k)`, and `model.compile_dlogp()`
raised `NotImplementedError: grad not implemented for Op Min{axis=0}`. Prior
predictive worked (no gradients needed), so the failure only surfaced at fit
time. `WeibullCDFAdstock` was unaffected.

This PR expresses the minimum as `-((-w).max(...))`, which is mathematically
identical but uses the grad-supported `Max` Op. The numerical output of the
transformation is unchanged (covered by the existing `test_weibull_pdf_adstock`).

A regression test (`test_adstock_logp_is_differentiable`) builds a small model
for every adstock transformation and asserts `model.compile_dlogp()` works and
returns a finite gradient — so NUTS-samplability is now guarded for all adstocks.

## Related Issue

- [x] Closes #2637

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2638.org.readthedocs.build/en/2638/


juanitorduz (approved):
Thanks @hermesdt

juanitorduz:
Actually, we might wanna fix this upstream in pytensor

hermesdt:
Make sense @juanitorduz , do you want me to open a PR there?

juanitorduz:
> Make sense @juanitorduz , do you want me to open a PR there?

That would be great, if yout have the time :)

hermesdt:
Implemented here @juanitorduz

juanitorduz:
👀  https://github.com/pymc-devs/pytensor/pull/2253

williambdean:
Agreed on moving to pytensor. Like the test though

hermesdt:
Done in https://github.com/pymc-devs/pytensor/pull/2253

PR #2640, opened 2026-06-22
fix(mmm): keep RootSaturation gradient finite at zero spend

### LINKED ISSUE #2639: RootSaturation NaN gradient at zero spend: `alpha * x**(alpha-1)` is infinite at x=0
## Description

`RootSaturation` (and `root_saturation`) produces a model that **cannot be
sampled with NUTS** when a channel has zero-spend periods. `nutpie` aborts with
"All initialization points failed" / `Logp function returned error:
ErrorCode(3)` because `d(logp)` is `NaN`.

## Root cause

`root_saturation(x, alpha) = x ** alpha` with `alpha` in `(0, 1)`. The
derivative with respect to the **base** is

```
d/dx (x ** alpha) = alpha * x ** (alpha - 1)
```

which is `0 ** (negative) = inf` at `x == 0` for `alpha < 1`. In an MMM the
saturation input is a function of random variables (adstocked spend, or any
scaled channel), and channels routinely have **exact** zero-spend weeks, so this
`inf` flows into `d(logp)/d(theta)` for every upstream parameter and becomes
`NaN`.

Note this is **not** the `x ** alpha * log(x)` term (the derivative w.r.t. the
*exponent*) — PyTensor's `Pow` gradient already guards that with
`switch(eq(x, 0), 0, ...)`. A model where the saturation input is *constant*
data therefore samples fine, which is why this is easy to miss; the bug only
appears once the input depends on a free RV.

## Reproducible example

```python
import numpy as np
import pymc as pm
from pytensor.xtensor import as_xtensor
from pymc_marketing.mmm.components.saturation import RootSaturation

x = np.linspace(0.0, 1.0, 40)
x[:5] = 0.0  # exact zero-spend weeks
y_obs = np.random.default_rng(0).normal(size=40)
with pm.Model(coords={"time": range(40)}) as model:
    scale = pm.HalfNormal("scale", 1)              # an RV the input depends on
    xx = as_xtensor(x, dims=("time",)) * scale     # == 0 where x == 0
    mu = RootSaturation().apply(xx)
    sigma = pm.HalfNormal("sigma", 1)
    pm.Normal("obs", mu=mu.values, sigma=sigma, observed=y_obs, dims=("time",))

grad = model.compile_dlogp()(model.initial_point())
print(grad)   # -> [nan, ...]  ; pm.sample(nuts_sampler="nutpie") then fails to initialize
```

With constant data (no `scale` RV) the gradient is finite — the singularity only
bites when the input is differentiable.

## Expected behavior

`RootSaturation` models with zero-spend periods should be samplable with NUTS.

## Proposed fix

Shift the input by a tiny `eps` before exponentiation:
`root_saturation(x, alpha) = (x + eps) ** alpha`. `eps = 1e-6` is negligible
against inputs scaled to `[0, 1]` and keeps the gradient finite. Trade-off:
`f(0) = eps ** alpha` is no longer exactly `0` (and is non-trivial for very
small `alpha`, e.g. `1e-6 ** 0.1 ≈ 0.25`), so the exact `eps` value / approach
is worth a maintainer's eye. PR incoming.

## Aside (separate issue, not addressed here)

While writing the regression test I noticed `TanhSaturationBaselined` also
yields a `NaN` gradient at its default initial point — but for an unrelated
reason (`arctanh(r)` with `r == 1` at the initial point), independent of zero
spend. Happy to open a separate issue if useful.

## Environment

- pymc-marketing 0.19.4
- pymc 5.28.5
- pytensor 2.38.3
- Python 3.12

hermesdt:
## Description

`root_saturation(x, alpha) = x ** alpha` (with `alpha` in `(0, 1)`) has a
derivative with respect to the **base**

```
d/dx (x ** alpha) = alpha * x ** (alpha - 1)
```

that is `inf` at `x == 0` for `alpha < 1`. In an MMM the saturation input is a
function of random variables (adstocked spend, or any scaled channel) and
channels routinely have **exact** zero-spend periods, so this `inf` propagates
into `d(logp)/d(theta)` for every upstream parameter and becomes `NaN` —
`nutpie` then fails with "All initialization points failed".

(For the record, this is *not* the `x ** alpha * log(x)` term w.r.t. the
exponent — PyTensor's `Pow` gradient already guards that with
`switch(eq(x, 0), 0, ...)`. The bug only surfaces once the input depends on a
free RV, which is why constant-data models sample fine.)

This PR shifts the input by a tiny `eps` (default `1e-6`):
`root_saturation(x, alpha) = (x + eps) ** alpha`. `eps` is negligible against
inputs scaled to `[0, 1]` and keeps the gradient finite everywhere.

**Trade-off worth a review:** `f(0) = eps ** alpha` is no longer exactly `0`,
and for very small `alpha` it is not negligible (`1e-6 ** 0.1 ≈ 0.25`). Happy to
adjust the `eps` magnitude, switch to `maximum(x, eps) ** alpha`, or use a
no-bias piecewise variant if maintainers prefer.

## Related Issue

- [x] Closes #2639

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2640.org.readthedocs.build/en/2640/


daimon-pymclabs (commented):
Thanks for the clear diagnosis and minimal repro — the `inf`-at-`x=0` gradient propagating into `d(logp)/d(theta)` is exactly right, and the fix does unblock `nutpie`. I'd push back on the additive shift itself, though: `eps` is a knob with no good setting, and a small change makes the question disappear.

## Why there's no optimal `eps`

With `(x + eps)**alpha`, two quantities move in opposite directions:

- **Bias at zero spend:** `f(0) = eps**alpha`, which should be ~0 (no spend → no contribution).
- **Gradient at zero spend:** `f'(0) = alpha * eps**(alpha-1)` — finite now, but it grows as `eps` shrinks, re-approaching the original singularity numerically.

For `alpha < 1` you can't make both small. The bias `eps**alpha = exp(alpha·ln eps)` decays only *linearly in ln eps*, so dropping `eps` by four orders of magnitude (1e-4 → 1e-8) only takes the `alpha=0.1` bias from 0.40 to 0.16:

| alpha | f(0) @ eps=1e-4 | f(0) @ eps=1e-6 | f(0) @ eps=1e-8 | grad(0) @ eps=1e-6 |
| --- | --- | --- | --- | --- |
| 0.05 | 0.63 | 0.50 | 0.40 | 2.5e4 |
| 0.10 | 0.40 | 0.25 | 0.16 | 2.5e4 |
| 0.20 | 0.16 | 0.063 | 0.025 | 1.3e4 |
| 0.30 | 0.063 | 0.016 | 0.004 | 4.8e3 |
| 0.50 | 0.010 | 0.001 | 0.0001 | 5.0e2 |

To get the bias under 0.01 at `alpha=0.1` you'd need `eps ≈ 1e-20` (below float precision, and that pushes `f'(0)` back above 1e6).

This isn't a tail concern given the default `RootSaturation` prior `Beta(1, 2)`, which puts **~19% of its mass below `alpha=0.1`** (median ≈ 0.29). With `eps=1e-6`, a meaningful slice of the prior produces a quarter-to-half of a channel's max contribution *out of exactly zero spend*, which biases `beta`, the contribution decomposition, and ROAS — and breaks the "paused channel ⇒ no effect" property.

## Suggested fix: guard the gradient instead of shifting the input

The canonical PyTensor pattern for a singular gradient at zero is the "safe input" double-`where`. It keeps the offending value out of the differentiated branch:

```python
import pytensor.tensor as pt

x = as_xtensor(x)
alpha = as_xtensor(alpha)
x_safe = pt.where(x > 0, x, 1.0)          # keep 0 out of the branch that gets differentiated
return pt.where(x > 0, x_safe**alpha, 0.0)
```

This gives:
- `f(0) = 0` **exactly** — zero bias, `f(x) = x**alpha` untouched for all real spend;
- a finite gradient everywhere — `0` at `x=0`, exact `alpha·x**(alpha-1)` for `x>0`;
- no `eps` to tune.

The only behavioral change is that the gradient w.r.t. upstream params is 0 at exact-zero-spend points — which is arguably correct, since a zero-spend period shouldn't inform `alpha`.

`maximum(x, eps)**alpha` is a reasonable runner-up (real spend stays unbiased, only the degenerate region is floored), but it retains `f(0)=eps**alpha` bias on exact zeros and adds a derivative kink at `x=eps`, so the `where` form is cleaner.

The existing `test_root_saturation_logp_is_differentiable_at_zero_input` test should pass unchanged with the `where` version (and you could add an assertion that `root_saturation(0, alpha) == 0` to lock in the no-bias property).

*(Review on behalf of @juanitorduz.)*

hermesdt:
Change implemented @daimon-pymclabs , thank you for the feedback!

daimon-pymclabs:
Looks great — this is exactly the safe-input pattern I had in mind. `x_safe = pt.where(x > 0, x, 1.0)` keeps the zero out of the differentiated branch, so you get `f(0) = 0` exactly, a finite gradient everywhere, and no `eps` knob to tune. The `compile_dlogp` test that runs the input through a free RV (mirroring post-adstock spend) is a nice touch — that's the path that was actually NaN-ing out NUTS. 👍

daimon-pymclabs (approved):
Approving — gradient guard resolves the zero-spend NaN without the bias/eps tradeoff. Thanks for the quick turnaround.

juanitorduz (approved):

hermesdt:
@juanitorduz I pushed another commit as many tests were failing. Pipeline requires manual approval to trigger.

juanitorduz:
Thanks @hermesdt

pymc_marketing/mmm/transformers.py
```diff
@@ -1265,4 +1265,9 @@ def root_saturation(
     """
     x = as_xtensor(x)
     alpha = as_xtensor(alpha)
-    return x**alpha
+    x_tensor = x.values
+    alpha_tensor = alpha.values
+    x_safe = pt.where(x_tensor > 0, x_tensor, 1.0)
+    result = pt.where(x_tensor > 0, x_safe**alpha_tensor, 0.0)
```
williambdean, line 1271: Does pytensor.xtensor not have this?

williambdean: Fine for follow up

PR #2642, opened 2026-06-23
Use pathmc as the backend for the causal tooling

cetagostini:
## Description

Re-points the causal tooling at the standalone **[pathmc](https://github.com/pymc-labs/pathmc)** library, which now owns PyMC-Labs' path-analysis / structural-causal-modeling stack. The module becomes a thin front end instead of a second implementation of the same algorithms, and the extra shrinks to a single dependency.

### `mmm/causal.py`

- **`TBFPC`, `BuildModelFromDAG`, `TestResult`** are re-exported from pathmc through a module-level `__getattr__` (PEP 562). Importing the module — and with it the core MMM — does not require pathmc; only *reaching for* a causal algorithm does, and the error then names the extra to install.
- **`CausalGraphModel`** keeps the surface `MMM` consumes (`build_graphical_model`, `get_unique_adjustment_nodes`, `compute_adjustment_sets`) but swaps dowhy for pathmc's native identification. `PathModel.adjustment_sets` returns *minimal* valid backdoor sets: descendants of the treatment are excluded, a set only qualifies if it d-separates treatment from outcome once the treatment's incoming edges are removed, and colliders therefore never enter the set.

  This is a correctness fix, not just a re-plumb. dowhy's `get_unique_adjustment_nodes` took the **union of every node lying on any backdoor path**, so for `Z1 -> X; Z1 -> C; Z2 -> C; Z2 -> Y; X -> Y` it returned `{Z1, C, Z2}` — adjusting for the collider `C` and thereby *opening* the path it was meant to block. pathmc returns `[]`: the path is already blocked. (The `mmm_funnel_mueffect_advanced` notebook called this coarseness out explicitly; that note is now updated.)
- **`get_backdoor_paths` is removed.** It wrapped a dowhy private API (`CausalModel._graph.get_backdoor_paths`) and has no pathmc equivalent — the adjustment sets are computed directly. The pathmc model stays reachable as `CausalGraphModel.path_model` for anyone who wants to inspect alternatives via `adjustment_sets` / `is_identifiable` / `collider_warnings`.
- Adjustment sets and the control columns derived from them are now **sorted**, so the MMM's `control` coord no longer depends on `set` iteration order (i.e. on `PYTHONHASHSEED`).

### `causal_utils.py`

`same_markov_equivalence_class_CPdag` stays a thin, lazily imported wrapper over `pathmc.same_markov_equivalence_class` (skeleton + unshielded-collider equality, dependency-free DOT reader — no `pydot`).

### `pyproject.toml`

`dag = ["pathmc>=0.2.0"]`. **`dowhy`, `osqp` and `pygraphviz` are gone**, as is the `ignore::SyntaxWarning:dowhy.*` filter. `0.2.0` is the first pathmc release carrying the discovery front end re-exported here (`0.1.x` has `PathModel` but no `TBFPC` / `BuildModelFromDAG` / `dag_to_spec` / `same_markov_equivalence_class`); the floor is the minimum API, so the resolver installs the newest pathmc the pinned PyMC/PyTensor stack allows. Net effect on the lockfile: 45 insertions, 352 deletions (`dowhy` pulled in `osqp`, `pydot`, `pygraphviz`, `qdldl`, `scs`, `statsmodels`, `sympy`).

> **On versions:** pathmc `0.2.1+` requires `pytensor>=3.1.1`, while `pymc>=6.0.0,<6.1.0` requires `pytensor<3.1`, so today the extra resolves to pathmc `0.2.0`. Once #2883 (`pymc>=6.2`) lands, the same declaration resolves to `0.3.x` with no change here. Both were exercised — see below.

### Fallout from dropping dowhy

Two things only worked because dowhy was in the environment; both are fixed here.

- **DOT comment headers.** pathmc's DOT readers require the `graph`/`digraph` keyword to come first (`re.match(r"\s*(?:strict\s+)?(graph|digraph)\b", ...)` in `pathmc.cpdag._parse_dot`, `s.lower().startswith("digraph")` in `pathmc.dag._parse_dag`), while `graphviz.Digraph(comment=...)` emits `// <comment>` above it. dowhy tolerated that, so `same_markov_equivalence_class_CPdag(true_dag.source, ...)` in `mmm_causal_reasoning_and_discovery.ipynb` broke on the switch — and `MMM(dag=...)` would have broken the same way. Both entry points now strip the DOT preamble (whitespace, `//`, `#`, `/* */`) before delegating; regression tests cover a real `graphviz.Digraph(comment=...)` and each comment style. Worth fixing upstream in pathmc too, since it affects any caller passing a commented DOT.
- **statsmodels.** `mmm_brand_metrics_long_term.ipynb` needs statsmodels through impulso and only had it as a dowhy transitive. The `docs` extra now asks for what the notebook uses: `impulso[diagnostics]`.

## Verification

Run in a worktree venv synced from the updated lockfile (`pymc 6.0.1`, `pytensor 3.0.7`, `pathmc 0.2.0`):

- `tests/mmm/test_causal.py` + `tests/test_causal_utils.py` — **26 passed**, **100% coverage** of `pymc_marketing/mmm/causal.py` and `pymc_marketing/causal_utils.py`. Re-run against pathmc `0.3.0`: **26 passed**.
- `tests/mmm/test_mmm.py` + `tests/pie` — **275 passed, 1 skipped**.
- `pytest tests/ -k "causal or dag or adjustment"` — **41 passed**.
- Core MMM imports with `pathmc` unavailable; `causal.TBFPC` and `CausalGraphModel.build_graphical_model` then raise `ImportError` naming `pip install pymc-marketing[dag]`.
- End-to-end smoke test: an `MMM` given a DAG plus `control_columns=["control_1", "control_2"]` derives `adjustment_set == ["control_1"]` from pathmc, drops the irrelevant control with a warning, and builds a model whose `control` coord is `["control_1"]`.
- Notebook smoke runs (mock runner): `mmm_causal_reasoning_and_discovery.ipynb` (TBFPC / BuildModelFromDAG / Markov equivalence, all now from pathmc) and `mmm_causal_identification.ipynb` (`MMM(dag=...)` → pathmc identification) both pass. The adjustment sets the latter prints are unchanged as *sets*; one displayed tuple was reordered to the new sorted output.

- **CI on this branch is green**: 56 checks pass (full test matrix on 3.12 and 3.14 x numba/cvm x 6 splits, every notebook batch, Read the Docs build, `uv lock --check`, pre-commit); 2 skipped, 0 failures.

## Related Issue

- Related to pymc-labs/pathmc#307 (discovery port) and the structure-uncertainty epic pymc-labs/pathmc#306
- Sequencing note: #2883 (`pymc>=6.2`) lets the extra pick up pathmc `0.3.x`, but is not required to merge this.

## Checklist

- [x] `pre-commit run --all-files` passes
- [x] Tests included: `CausalGraphModel` adjustment-set behaviour (incl. the collider case dowhy got wrong and a latent-confounder non-identifiability warning), lazy-import wiring, and the actionable errors raised without the extra
- [x] Documentation updated: module docstrings explain the pathmc backend and the `dag` extra; the three notebooks that referenced dowhy are updated


juanitorduz:
As this is a breaking change, I just added it to the v1.0 milestone

juanitorduz:
@carlosagostini , I think now that we have merged `v1.0.0`, you can continue working on this one

cetagostini:
@juanitorduz @drbenvincent migration to pathmc done!!

juanitorduz:
@drbenvincent  @cetagostini ...pathmc also lazily depends on pymc-marketing ... is this circular dependency a problem? can we add unit-test about it?

drbenvincent (changes requested):
## Strong direction, with a few safeguards needed before merge

This is a very positive architectural move. The PR gives PyMC-Marketing a cleaner responsibility boundary: MMM keeps the user-facing workflow of turning a DAG into model controls, while `pathmc` becomes the focused home for causal discovery, path analysis, and structural causal-modeling algorithms. Replacing a large in-repo implementation and the DoWhy stack with a thin `pathmc` front end should reduce duplicated maintenance, simplify the dependency graph, and let causal capabilities evolve in the package that owns them.

The lazy-import design is particularly well judged. Core MMM users do not need to install the causal stack, while users who do reach causal functionality receive an actionable `pymc-marketing[dag]` installation message. The PR also does a good job of preserving the familiar `CausalGraphModel` surface and of addressing migration fallout such as Graphviz DOT comment preambles and explicitly declaring the diagnostics dependency used by the documentation.

Most importantly, this is a correctness improvement rather than a cosmetic backend substitution. Delegating adjustment-set identification to `PathModel.adjustment_sets` replaces the previous broad union-of-backdoor-path-nodes approach with minimal, d-separation-valid sets. In particular, avoiding collider adjustment is a substantial improvement for DAG-guided MMM workflows. The focused tests, notebook runs, and coverage are encouraging.

Before merging, I would ask for a few guardrails that make the new causal contract as robust as the implementation itself.

### 1. Make the multi-treatment estimand explicit and test the final set

`get_unique_adjustment_nodes` computes a minimal valid set per treatment and then unions those sets. That is a reasonable first policy, but a union of sets that are valid one treatment at a time is not automatically valid for the joint treatment intervention an MMM is often interpreted as estimating: adding a variable can, in some DAGs, open a collider path. Please either validate the combined set against the joint estimand in `pathmc`, or document that the contract is explicitly per-channel identification and add a regression test demonstrating the intended multi-treatment semantics.

### 2. Do not make non-identification too easy to overlook

When no backdoor set exists, the current behavior warns and continues with no adjustment for that treatment. That may be useful for exploratory modeling, but it is a risky default for an API presenting causal guidance: a warning can be lost while the resulting fit still appears causal. Please consider a `strict_identification` mode (or another explicit opt-in to continue) that raises with the treatment/outcome pair and suggested next steps. At minimum, document prominently that a successful model build is not evidence that every requested causal effect was identified.

### 3. Lock down the cross-package and version contract

The review discussion correctly raises the possible lazy, bidirectional relationship between `pathmc` and PyMC-Marketing. Please add an import-order test covering core PyMC-Marketing without `pathmc`, causal PyMC-Marketing with the extra, and `pathmc` when PyMC-Marketing is present. Also test the oldest compatible `pathmc` release and the newest one admitted after the PyMC/PyTensor pins relax. The dependency is currently `pathmc>=0.2.0`, so this is the place to make behavioral compatibility an intentional contract rather than something the resolver happens to preserve.

Finally, please add a concise v1 migration note for the removed `get_backdoor_paths` API and for cases where existing DAGs now return different, safer control sets. With those points addressed, I would be enthusiastic about merging this migration.

PR #2645, opened 2026-06-24
Address review feedback on the pymc6 / DataTree migration

daimon-pymclabs:
Follow-up to the review of #2616, addressing the concrete issues raised. Scoped to library source + tests; no notebook changes.

## Bug fixes (with tests)

**1. `mlflow._force_load_idata_groups` was a silent no-op.**
```python
for group in idata.groups:        # DataTree.groups -> "/posterior", "/sample_stats", ...
    if hasattr(idata, group):     # hasattr(idata, "/posterior") -> always False
        ...
```
`DataTree.groups` yields `/`-prefixed paths, so the guard never passed and no group was ever materialized — lazy (netCDF/Zarr-backed) idata was not force-loaded before MLflow logging. Replaced with iteration over `idata.subtree`, loading each group's dataset in place. Verified the data is genuinely materialized (the backing file can be deleted afterwards and values still read), and added `test_force_load_idata_groups_visits_every_group`.

**2. `pytensor_utils._as_symbolic` was dead and broken.** Never called (the old `as_symbolic` usage was replaced inline by `StringConstant(dim.type, ...)`), and `Variable(name)` mis-constructs a `Variable` (first positional arg is `type`). Removed.

## Reproducibility

**3. `pymc-extras` was pinned to a moving git branch.** `[tool.uv.sources]` pointed at `pymc-extras.git` with no rev, so it resolved to whatever HEAD happened to be. Pinned to the exact commit already in `uv.lock` (`ee8cc37`) so the build is reproducible. The lock change is just the `?rev=` annotation of the same commit — no resolution change. Left a comment noting this is temporary until a pymc-extras release with the pymc>=6/DataTree changes lands on PyPI.

## Docstring / message cleanup

**4. Remaining `InferenceData` references** where the object is now concretely a `DataTree`: the two user-facing `KeyError` messages in `log_sample_diagnostics`, `log_inference_data` / `autolog` docstrings, `create_idata_accessor` docstrings, and the `DeserializationContext.idata` docstring. (Deliberately left alone: the `InferenceDataGroupSchema` class name, and prose referring to the arviz *concept* — renaming those is out of scope for this migration.) Updated the corresponding test assertion for the new message text.

**5. `model_builder.fit_result` setter** — added a comment explaining the flatten-into-posterior branch (when `res` has no explicit posterior group).

## Deliberately NOT changed (need author input)

- **ADVI convergence callback** removed from `clv/models/basic.py::_fit_approx` (`CheckParametersConvergence`). This is a behavior change, not part of the API migration, and @williambdean's own inline "Need to look into this" is on this file — flagging rather than guessing. Was it intentional (e.g. callback API changed under pymc 6) or an accidental drop?
- **`fit()` idiom drift** between `bass/model.py` (keeps `.copy()` before `.update()`) and `model_builder.py` (mutates in place); and `fit_data` assignment via `xr.DataTree(ds)` vs raw Dataset. All functionally correct — left for a consistency pass if desired rather than risk altering `fit()` semantics on a green branch.
- **`residuals_posterior_distribution` rendering** (`az.plot_dist` → `sns.kdeplot` + quantile `axvline`s) — a necessary consequence of the arviz API change; visual semantics differ slightly, worth an eyeball but not a bug.
- **`@OriolAbril`'s `.dataset` (view) / `.children` style suggestions** — left as-is to avoid introducing view-aliasing subtleties / behavior changes for nested trees; happy to apply if you'd prefer.

## Testing

Ran `tests/test_mlflow.py` for the touched paths — the new force-load test and the updated `test_log_sample_diagnostics_missing_group` pass. `ruff check` / `ruff format --check` clean on all changed files. (Couldn't run the full suite locally — optional samplers and graphviz aren't installed in the sandbox — but confirmed the unrelated failures are identical with and without these changes.)

----
📚 Documentation preview 📚: https://pymc-marketing--2645.org.readthedocs.build/en/2645/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -211,8 +211,11 @@ ignore_missing_imports = true
 [tool.mypy-scipy]
 ignore_missing_imports = true
 
+# Temporary: pin pymc-extras to a specific commit until a release containing
+# the pymc>=6 / DataTree changes is published on PyPI. Pinning the commit keeps
+# the build reproducible (an unpinned branch would resolve to a moving target).
 [tool.uv.sources]
-pymc-extras = { git = "https://github.com/pymc-devs/pymc-extras.git" }
+pymc-extras = { git = "https://github.com/pymc-devs/pymc-extras.git", rev = "ee8cc37df25f177ae7aea8a76587fb26da261a80" }
 
 [dependency-groups]
 dev = [
--- a/pymc_marketing/mlflow.py
+++ b/pymc_marketing/mlflow.py
@@ -332,19 +332,21 @@ def _log_and_remove_artifact(path: str | Path) -> None:
 
 
 def _force_load_idata_groups(idata: xr.DataTree) -> None:
-    """Force load all groups into memory since ArviZ does lazy loading.
+    """Force load all groups into memory since the store is read lazily.
+
+    netCDF/Zarr-backed ``DataTree`` objects load their groups lazily, so we
+    materialize every group in the tree before logging to MLflow.
 
     Parameters
     ----------
     idata : xr.DataTree
         The DataTree object to force load.
     """
-    for group in idata.groups:
-        # Convert each group to an in-memory dataset
-        if hasattr(idata, group):
-            group_data = getattr(idata, group)
-            if hasattr(group_data, "load"):
-                group_data.load()
+    # ``DataTree.groups`` yields ``/``-prefixed paths (e.g. ``"/posterior"``),
+    # which are not attribute names, so iterate the nodes directly and load
+    # each group's dataset in place.
+    for node in idata.subtree:
+        node.dataset.load()
 
 
 def _attach_run_id(idata: xr.DataTree) -> None:
@@ -561,10 +563,10 @@ def log_sample_diagnostics(
 
     """
     if "posterior" not in idata:
-        raise KeyError("InferenceData object does not contain the group posterior.")
+        raise KeyError("DataTree object does not contain the group posterior.")
 
     if "sample_stats" not in idata:
-        raise KeyError("InferenceData object does not contain the group sample_stats.")
+        raise KeyError("DataTree object does not contain the group sample_stats.")
 
     posterior = idata["posterior"]
     sample_stats = idata["sample_stats"]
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -211,8 +211,11 @@ ignore_missing_imports = true
 [tool.mypy-scipy]
 ignore_missing_imports = true
 
+# Temporary: pin pymc-extras to a specific commit until a release containing
+# the pymc>=6 / DataTree changes is published on PyPI. Pinning the commit keeps
+# the build reproducible (an unpinned branch would resolve to a moving target).
 [tool.uv.sources]
-pymc-extras = { git = "https://github.com/pymc-devs/pymc-extras.git" }
+pymc-extras = { git = "https://github.com/pymc-devs/pymc-extras.git", rev = "ee8cc37df25f177ae7aea8a76587fb26da261a80" }
```
juanitorduz, line 218: @williambdean can we have a lower/upper bound?

juanitorduz (commented):
@williambdean here are some initial suggestions from daimon :)

daimon-pymclabs:
@juanitorduz good question — there actually already is a PEP 508 bound. It lives in `[project.dependencies]`:

```toml
"pymc-extras>=0.11.0,<0.12.0",
```

The `[tool.uv.sources]` entry doesn't replace that bound — it only overrides *where* uv fetches pymc-extras from (the git repo at a fixed commit instead of PyPI). The version constraint still applies on top of it.

Two things worth flagging now though, both pointing the same way:

1. **The pin and the bound have drifted out of sync.** The pinned commit `ee8cc37` is one past the `v0.12.1` tag (it's the "Implement basic INLA interface" merge), so its version metadata reports ≥0.12.1 — i.e. *outside* the declared `<0.12.0` bound. So the git rev and the version range no longer agree.

2. **`pymc-extras 0.12.1` is now on PyPI** (released after this PR was opened — it's the current latest). The whole point of the temporary git pin was "until a release with the pymc>=6 / DataTree changes lands on PyPI." If 0.12.1 already carries those changes (worth a 2-min check against what we actually needed), we can drop the `[tool.uv.sources]` block entirely and just point at PyPI with a clean bound, e.g.:

```toml
"pymc-extras>=0.12.1,<0.13.0",
```

That removes the git dependency, gives you the lower/upper bound you're asking for, and resolves the inconsistency. Happy to make that change here — @williambdean, do you want me to swap the git pin for the PyPI `>=0.12.1,<0.13.0` bound (assuming 0.12.1 has the DataTree changes), or keep the git pin for now?

PR #2652, opened 2026-06-26
chore: remove dead catch_warnings blocks from add_groups migration

williambdean:
The 'fit_data not defined in InferenceData scheme' warning was emitted by arviz's `add_groups`, which was replaced by `idata["/fit_data"] = fit_data` in the PyMC 6 migration (#2616). DataTree assignment never raises this warning, making the `catch_warnings` / `filterwarnings` blocks dead code.

Removed in: `mixed_logit.py`, `mnl_logit.py`, `nested_logit.py`, `maxdiff.py`.

Flagged by @daimon-pymclabs in #2616 review.

----
📚 Documentation preview 📚: https://pymc-marketing--2652.org.readthedocs.build/en/2652/


williambdean:
Please take a look @daimon-pymclabs

daimon-pymclabs:
Took a look — this is good to merge. Checked the mechanical side carefully:

- The four `customer_choice` `fit()` methods and the two conftest fixtures all suppressed the same `UserWarning` ("The group fit_data is not defined in the InferenceData scheme"), which no longer fires now that `fit_data` is a recognized group — so the blocks are genuinely dead.
- The `import warnings` removals are correct: dropped from `mixed_logit.py`, `mnl_logit.py`, `nested_logit.py` and both conftests where it's now unused, and **retained** in `maxdiff.py` (still used by `warnings.warn(...)` at ~L601). No dangling references left in any changed file.
- Full test matrix + customer_choice notebooks are green, so the suppressed path is exercised.

One optional belt-and-suspenders: a quick run of the customer_choice fit tests under `-W error::UserWarning` would prove the warning is gone rather than merely silent, but I don't think it's necessary to block on. LGTM 👍

williambdean:
@daimon-pymclabs any other missing similar spots for this change?

daimon-pymclabs:
Good catch to check. I grepped the repo for the `"The group fit_data is not defined in the InferenceData scheme"` suppression. Two files outside this PR still contain it:

- `pymc_marketing/model_builder.py` (~L1053 and ~L1216)
- `pymc_marketing/clv/models/basic.py` (~L109)

But these are **not** dead in the same way the ones in this PR are. The blocks here became dead because #2616 migrated the write from `self.idata.add_groups(fit_data=...)` to the DataTree assignment `self.idata["/fit_data"] = fit_data`, and DataTree assignment never raises that `UserWarning`. The three spots above **still call `self.idata.add_groups(fit_data=...)`**, which does still emit the warning — so their `catch_warnings` blocks are live, not dead, and removing them now would let the warning leak.

So for this PR's scope (removing suppressions left dead by the add_groups→DataTree migration), the 6 files you've already changed cover everything. The model_builder/clv-basic spots are a separate, slightly larger follow-up: migrate them to `idata["/fit_data"] = ...` as well, and *then* drop the suppression. Happy to open that as a follow-up PR if you'd like to keep this one a clean no-behavior-change cleanup.

(Note: the unrelated `catch_warnings` at `model_builder.py:522` and `clv/models/basic.py:305`/`387` suppress different messages, so they're out of scope here.)

daimon-pymclabs:
Closing in favour of #2869, which scopes the removal to the four source files without the 59 re-executed notebooks. Reopening the work there keeps the history rewrite in #2846 unblocked.

PR #2621, opened 2026-06-09
feat(mmm): OptimizableMuEffect protocol + DiscountedEventEffect (discount-depth lever)

PabloRoque:
## Summary

Two additions on top of #2654, kept in strictly separate layers:

1. **`BudgetOptimizer.optimizable_vars`** — a graph-only generalization of the optimizer: co-optimize any named `pm.Data` variable (single non-date dim) alongside the media budgets, in its native units, with optional native bounds. The optimizer imports nothing from the marketing layer (an AST test enforces this).
2. **`DiscountedEventEffect`** — a concrete `OptimizableMuEffect` whose lever is promotional discount depth `d_k ∈ [discount_min, discount_max]`, with link-correct price × volume economics.

### `BudgetOptimizer.optimizable_vars: dict[str, list[tuple[low, high]] | None]`

- Each entry's dim/coords/size are read off the model graph by name; its segment joins the **same flat decision vector and the same `do()`** as the media budgets, so gradients flow through one compiled objective.
- Extra variables **never enter the default budget-sum constraint** (`constraints.py` is untouched; `_budgets` is sliced to the media head). A cash-denominated lever can opt into the pot via a custom `constraints=`.
- Optimal values come back as `result.optimized_vars[name]` — a DataArray over the variable's own dim (`{}` when unused; documented in `allocate_budget`).
- **Inert-lever guards** (structural + numeric): construction raises if `response_variable` doesn't depend on a declared variable (graph ancestry on the compiled model), and warns when levers are scored against the media-contribution default (reachable under the log link, but then tuned to maximize incremental media contribution rather than revenue); `allocate_budget` additionally warns per lever coordinate whose gradient is identically zero (verified at x0 **and** at a perturbed point, so a warm start sitting exactly at an optimum is not mislabeled) — e.g. an event window outside the optimization window; those return at their current values.
- **Warm start**: levers seed at the variable's current model value (clipped to bounds) — e.g. the historical discount depth — not at a bound. This is what makes the raw objective converge for both blocks; the objective is deliberately passed to SLSQP **unscaled** (an earlier |f(x0)| objective rescale left the compiled constraints untouched, which froze the media block at its uniform seed — measured and removed; a regression test asserts media moves off its seed in the joint problem).
[...]

PR #2654, opened 2026-06-26
refactor(budget-optimizer): BudgetOptimizer accepts pm.Model + DataTree directly

### LINKED ISSUE #2425: BudgetOptimizer to take pm.Model and InferenceData
Take out the middle man of the Wrapper class and make BudgetOptimzer take pm.Model and InferenceData directly. 

There would still be a bridge between common MMM class and this optimizer. But it the Optimzwr it self doesnt need to know about how to create the optimization. 

```python
optimizer = mmm.budget_optimizer(...)

result = optimizer.allocate_budget(...)
```

Will still be compatible with the combined model with processing steps before initializing the OptimizerBudget.

```python
model, idata = combine_model_andidata([geo1_mmm.create_optimization_model(...), ...],...)

optimizer = BudgetOptimizer(model=model, idata=idata)
result = optimizer.allocate_budget(...)
```

williambdean:
Closes #2425.

BudgetOptimizer no longer requires a wrapper object. Pass a pre-built
`pm.Model` and its `DataTree` directly. All existing wrapper paths
continue to work via a transparent backward-compat validator.

```python
# New API
optimizer = mmm.budget_optimizer(start_date="2025-01-01", end_date="2025-03-31")
optimal, result = optimizer.allocate_budget(total_budget=100_000)

# Or with a custom model
optimizer = BudgetOptimizer(
    model=pymc_model,
    idata=idata,
    num_periods=13,
    adstock_periods=4,
)
```

Configurable variable and dimension names (`channel_data_var`,
`date_dim`, etc.) replace the old hardcoded strings. A `mu_effects`
field is forward-compatible with future OptimizableMuEffect PRs (#2621).

New standalone utilities for multi-model merging:

```python
merged_model, merged_idata = merge_models_and_idata(
    models=[m1, m2], idatas=[idata1, idata2], prefixes=["north", "south"],
)
```

`BuildMergedModel`, `BudgetOptimizerWrapper`, and `CustomModelWrapper`
now emit `DeprecationWarning` on instantiation.

----
📚 Documentation preview 📚: https://pymc-marketing--2654.org.readthedocs.build/en/2654/


williambdean:
Please review @daimon-pymclabs 
Looking for more user flexibility and developer benefits

daimon-pymclabs (commented):
## Review

Overall this is a solid, well-structured refactor and the direction is right: decoupling `BudgetOptimizer` from the `MMM` wrapper so it accepts a plain `pm.Model` + `DataTree`, making the variable/dimension names configurable (`channel_data_var`, `date_dim`, `channel_contribution_var`), and keeping every legacy path alive through a transparent `model_validator` + deprecation warnings. The configurable names and the `merge_models_and_idata` convenience are genuine flexibility/DX wins — a non-MMM PyMC model can now be optimized without subclassing, and the multi-model merge workflow is a single call. Docstrings and examples are excellent.

A few things to address before merge, roughly in order of importance.

### 1. `frozen_deterministics` is silently dropped — regression for HSGP / time-varying models
`BudgetOptimizer.extract_response_distribution` previously passed `frozen_deterministics=getattr(self.mmm_model, "frozen_deterministics", None)`. The refactor drops that argument entirely:

```python
return extract_response_distribution(
    pymc_model=self._pymc_model,
    idata=_extract_dataset(self.idata, "posterior"),
    response_variable=response_variable,
)
```

But `extract_response_distribution` still accepts `frozen_deterministics`, and its docstring is explicit: *"Some models (e.g. those containing HSGP) need this to obtain a valid conditional posterior graph."* `MMM.frozen_deterministics` also still exists and is used elsewhere in `mmm.py`. The new `mmm.budget_optimizer()` entry point doesn't thread it through either. So for any MMM with frozen deterministics (HSGP, time-varying intercept/media), the optimizer will now build the response graph without freezing them — risking either a `RuntimeError("RVs found in the extracted graph")` or silently different optimization results. Suggest adding a `frozen_deterministics: list[str] | None` field on `BudgetOptimizer` and having `mmm.budget_optimizer()` pass `self.frozen_deterministics`.

### 2. mu_effects budget slots are half-wired — `allocate_budget` will break once an optimizable effect is present
`model_post_init` enlarges the optimization vector to `size_budgets + size_mu_effect_budgets` and slices out `_effect_budget_xtensors`. Two gaps:

- `_effect_budget_xtensors` is built but **never consumed** anywhere in the file (no `replace_for_optimization` call, not referenced in the objective graph) — currently dead scaffolding.
- `allocate_budget` sizes `x0` and `bounds` from `budgets_size = self.budgets_to_optimize.sum()` (i.e. `size_budgets` only), while the compiled objective expects `_budgets_flat` of length `size_budgets + size_mu_effect_budgets`. So `self._budgets_flat.type.filter(x0)` will raise on a shape mismatch (and `bounds` would be the wrong length) the moment an effect with `replace_for_optimization` is passed.

This is dormant today because stock `mu_effects` lack `replace_for_optimization`, so `_optimizable_mu_effects` is empty — but it's untested, fragile, and will bite #2621. Either fully wire it (consume the xtensors + extend `x0`/`bounds`/per-effect bounds) or gate it behind a guard that raises a clear "not yet supported" error until #2621 lands.

### 3. New merge utilities have no unit tests
Codecov flags 59% patch coverage / ~95 missing lines in `budget_optimizer.py`, and `merge_inference_data` / `merge_models_and_idata` have non-trivial logic (variable/dim prefixing, `shared_dims` derivation, draw thinning, `xr.merge` alignment) but no direct tests in the diff. The `xr.merge` step in particular aligns on shared `chain`/`draw` indexes across independently-sampled models — worth a test asserting the merged shapes/coords are what you expect (and that thinning + prefixing round-trip correctly), since silent outer-join NaNs there would be hard to catch downstream.

### 4. `idata` argument to `extract_response_distribution` changed type
It now receives `_extract_dataset(self.idata, "posterior")` (a bare `xr.Dataset`) instead of the full `DataTree`/`InferenceData` it got before. `extract_response_distribution` runs `az.extract(idata)`. Please confirm `az.extract` behaves identically on a bare posterior `Dataset` vs the previous container across the supported arviz range — `az.extract` historically expects an InferenceData/DataTree with a `posterior` group and a `group=` arg, so passing the inner Dataset may rely on convert-to-dataset fallthrough.

### 5. Minor
- **`merge_models_and_idata` docstring/behavior mismatch:** the docstring (and `Raises`) say it requires ≥2 models, but the body only checks `len(models) == len(idatas)`. Either enforce `len(models) >= 2` or correct the docstring.
- **`"optimization_model" in type(model).__dict__` breaks inheritance** (used in the validator and in `BuildMergedModel.optimization_model`): a subclass that *inherits* `optimization_model` without redefining it won't be found in `type(model).__dict__` and will fall back to the deprecated `_set_predictors_for_optimization`. Consider walking the MRO or using a sentinel that distinguishes "real impl" from the Protocol stub.
- **Redundant work in `mmm.budget_optimizer`:** `create_zero_dataset(...)` runs once inside `create_optimization_model()` and again directly just to compute `num_dates`. You can derive `num_periods` from the built model's `date` coord instead of rebuilding the zero dataset.
- **DRY:** the `idata` → `DataTree.from_dict({g: getattr(idata, g) ...})` conversion appears three times in `_handle_legacy_model_arg`; pull it into a small helper (it already overlaps with `_extract_dataset`).

Happy to dig into any of these further — #1 and #2 are the only two I'd consider blocking.

cetagostini (approved):

daimon-pymclabs (approved):
Re-reviewed after 6b5d263 + b84c46a — thanks @williambdean, this addresses both blocking items and most of the minors. ✅

- **#1 frozen_deterministics** — fully threaded: new `frozen_deterministics` field on `BudgetOptimizer`, inferred from the wrapper in `_handle_legacy_model_arg`, passed into `extract_response_distribution`, and forwarded by `mmm.budget_optimizer()`. HSGP / time-varying models are covered now.
- **#2 mu_effects slots** — good call gating it: `_optimizable_mu_effects` now raises a clear `NotImplementedError` pointing at #2621, and the dead `_effect_budget_xtensors` scaffolding + mismatched `_budgets_flat` sizing are gone. No more dormant shape-mismatch trap.
- **#5** — docstring ≥2-models line removed, `_to_datatree` helper de-dupes the three conversion sites, and `budget_optimizer()` now derives `num_periods` from the built model's `date` coord instead of rebuilding the zero dataset.

Approving. Two non-blocking follow-ups I'd still flag (fine to do separately or wave off):
- **#3** — `merge_inference_data` / `merge_models_and_idata` still have no direct unit tests (codecov patch coverage). The `xr.merge` alignment on shared chain/draw indexes is where a silent outer-join NaN would hide — a small round-trip test (prefixing + thinning + merged shapes/coords) would be worth it.
- **minor** — `"optimization_model" in type(model).__dict__` still misses a subclass that *inherits* (rather than redefines) the impl and would fall back to the deprecated path; walking the MRO would be more robust. And worth a quick confirm that `az.extract` behaves identically on the bare posterior `Dataset` now passed to `extract_response_distribution` vs the previous container.

daimon-pymclabs:
Thanks for the ping, @williambdean — reviewed. Overall this is a solid, well-motivated refactor that delivers exactly what #2425 asked for: the wrapper "middle man" is gone, `BudgetOptimizer(model=..., idata=...)` is the canonical path, and the legacy wrappers keep working through a `model_validator`. Parameterizing the variable/dimension names (`channel_data_var`, `date_dim`, `channel_contribution_var`) is a genuine flexibility win for non-MMM PyMC models, and `date_dim` looks correctly threaded through the `_validate_and_process_*` helpers and `_replace_channel_data_by_optimization_variable`.

I traced the `num_periods`/`adstock_periods` arithmetic end-to-end and it's consistent: `create_zero_dataset(..., include_carryover=True)` yields `W + l_max` dates, `budget_optimizer()` recovers `num_periods = n_dates - l_max = W` with `adstock_periods = l_max`, and `_replace_channel_data_by_optimization_variable` rebuilds `W + l_max` steps — matching the legacy `BudgetOptimizerWrapper`. No double-counting. 👍

A few things worth addressing before merge:

**1. The two new public merge utilities are untested.** `merge_inference_data` and `merge_models_and_idata` (~161 new lines) have no tests — neither is referenced in any of the three budget test files. Since both are now exported from `pymc_marketing.mmm.__init__`, this is most of the ~92 lines Codecov flagged. Worth covering: the auto-prefix path (`prefixes=None`), the length-mismatch `ValueError`s, `merge_on=None`, the `n == 1` early return, and a round-trip (merge two idatas → feed to a `BudgetOptimizer`). Also untested: the `DeprecationWarning`s on the three wrappers, and the legacy-unpacking branch of `_handle_legacy_model_arg`. A `pytest.warns(DeprecationWarning)` per wrapper would lock in the "transparent backward-compat" contract.

**2. `merge_inference_data` — `xr.merge` conflict on shared vars.** The cross-model merge uses `xr.merge([...])`, which defaults to `compat="no_conflicts"` and will **raise** if two models share a `constant_data`/`observed_data` variable (e.g. the deliberately-unprefixed `merge_on` var like `channel_data`) with differing values. That's a plausible real case — two regions, same channel names, different spend — and it'll surface as an opaque `xr.MergeError`. Either document that `merge_on` vars must be identical across models, or handle the conflict explicitly. Relatedly, the docstring says it "keeps every n-th posterior draw" but only the `posterior` group is thinned (other groups pass through) — worth a one-line note.

**3. `adstock_periods` silently defaults to 0 on the direct path.** On the new canonical API, forgetting `adstock_periods` drops the carryover tail — but the optimization still runs and returns a plausible-looking result, so the mistake is invisible. Since `create_optimization_model` already bakes carryover into the model's date length, consider inferring/validating `adstock_periods` from the model's date dim vs `num_periods` and warning on mismatch. This is the ergonomics edge most likely to bite a new user.

**Lower severity:**
- `_handle_legacy_model_arg`: `BudgetOptimizer(model=legacy_wrapper, idata=my_idata)` silently discards `my_idata` (the legacy branch overwrites `data["idata"]` from the wrapper). Nobody should pass both, but a guard or docstring note avoids a confusing silent override.
- `extract_response_distribution` now gets a bare `Dataset` instead of the DataTree. I checked — `az.extract` → `convert_to_dataset` returns a Dataset as-is, so this is **safe**, just an input-type change to a shared helper worth a brief comment.
- `mu_effects` accepts input but always raises `NotImplementedError` if any effect sets `replace_for_optimization` — reasonable forward-compat stub for #2621, but please cover that raise with a test so it doesn't regress into a silent no-op.
- Naming: `OptimizerCompatibleModel` docstrings still say "model wrappers" — minor, but the PR's whole point is that you *don't* need a wrapper, so tightening that language reduces confusion.

None of these block the core value — the direct-model path and the legacy bridge both look correct. Items **1** and **2** are the two I'd want resolved before merge; the rest is polish.

PabloRoque:
@williambdean — I merged `main` into this branch to resolve the conflict state and the inflated diff. Summary of what happened and what I did:

**Why the PR showed ~100 files / CONFLICTING:** this branch tracked `v1.0.0` by merging the actual branch (last on Jul 8), while `main` received v1.0.0 as a single squashed commit (#2677, Jul 28) — and v1.0.0 kept evolving for those three weeks in between (314 files). The merge-base was stuck pre-v1.0.0, so GitHub displayed all of that as this PR's diff.

**Resolution policy for the merge (`67fbc7db`):**
- 58 conflicted files + 6 silently auto-merged notebooks: taken from `main` wholesale — the branch's side was provably the stale Jul-8 v1.0.0 snapshot with no changes of this PR's own.
- `mmm/__init__.py`: union of both import blocks. `mmm.py`: 6 hunks, all InferenceData→DataTree docstring wording from #2782 — main's terminology taken.
- `budget_optimizer.py`: 6 hunks, refactor side kept — three were against pre-refactor code (referencing `self.mmm_model` / `is_wrapper`, which don't exist here), one keeps the `model_validator` import.
- **One judgment call to flag:** in three spots main tightened `idata` to `xr.DataTree` (per #2782) where this branch deliberately uses `Any` + conversion for the legacy-wrapper path. I kept the branch's permissive typing since it's the reviewed back-compat contract — happy to tighten instead if you prefer.

**Result:** the PR now shows its true diff — **7 files, +1088/−154** — and all 105 optimizer tests pass locally on the merged branch (`test_budget_optimizer.py`, `test_budget_optimizer_mmm.py`, `test_cost_per_unit.py`), pre-commit clean.

juanitorduz:
@juanitorduz merge this one

juanitorduz (approved):

Issue #2659, opened 2026-06-29
`budget_allocation` plot missing from v2 `BudgetPlots` namespace (plot_suite="new")


juanitorduz:
## Summary

When a model opts into the new plot suite (`mmm.plot_suite = "new"`), `optimizable_model.plot` returns the v2 `BudgetPlots` namespace, which only implements `allocation_roas` and `contribution_over_time`. The legacy `budget_allocation` bar chart (allocated spend vs. channel contribution per channel) was **not** ported to the v2 namespace, so existing code that calls it raises an `AttributeError`.

## Reproduction

```python
mmm.plot_suite = "new"
# ... fit, optimize, sample_response_distribution -> sample_allocation ...
optimizable_model.plot.budget_allocation(samples=sample_allocation)
```

```
AttributeError: 'BudgetPlots' object has no attribute 'budget_allocation'
```

This previously worked (see the v1.0.0 multidimensional MMM notebook:
https://github.com/pymc-labs/pymc-marketing/blob/v1.0.0/docs/source/notebooks/mmm/mmm_multidimensional_example.ipynb).

## Relevant code

- `optimizable_model.plot` dispatch: `pymc_marketing/mmm/mmm.py` (`BudgetOptimizerWrapper.plot`, returns `BudgetPlots()` when `plot_suite == "new"`).
- v2 namespace: `pymc_marketing/mmm/plotting/budget.py` (`class BudgetPlots` — only `allocation_roas` and `contribution_over_time`).
- Legacy implementation that still exists: `pymc_marketing/mmm/plot.py` (`MMMPlotSuite.budget_allocation`).

## Current workaround

`budget_allocation` is effectively stateless (it only reads `channel_contribution*` and `allocation` from the `samples` dataset), so it can be called on the legacy suite directly without flipping the global `plot_suite`:

```python
from pymc_marketing.mmm.plot import MMMPlotSuite

fig, axes = MMMPlotSuite(idata=mmm.idata).budget_allocation(samples=sample_allocation)
```

## Proposed resolution

Port `budget_allocation` to the v2 `BudgetPlots` namespace (so `optimizable_model.plot.budget_allocation(...)` keeps working under `plot_suite="new"`), or, if it is intentionally deprecated in favor of `allocation_roas`, document the migration path and emit a clear, actionable error message pointing users to the replacement / workaround.

juanitorduz:
wont solve as we will remoev these plots

PR #2674, opened 2026-07-01
[FIX] budget_allocation plot missing from v2 BudgetPlots namespace (plot_suite="new")  

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/2659

----
📚 Documentation preview 📚: https://pymc-marketing--2674.org.readthedocs.build/en/2674/

### DIFF
--- a/docs/source/notebooks/mmm/mmm_plot_suite_migration_guide.ipynb
+++ b/docs/source/notebooks/mmm/mmm_plot_suite_migration_guide.ipynb
@@ -302,7 +302,6 @@
     "\n",
     "fig, axes = optimizer.plot.allocation_roas(samples=samples)\n",
     "fig, axes = optimizer.plot.contribution_over_time(samples=samples)\n",
-    "fig, axes = optimizer.plot.budget_allocation(samples=samples)\n",
     "```"
    ]
   },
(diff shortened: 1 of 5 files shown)


juanitorduz:
@bugbot review

isofer:
budget_allocation was not migrated to the new suite on purpose because:
1. We thought it was not a great plot. I originally discussed that with @carlosagostini, and we decided that ROA graph makes more sense. We didn't think there is actually a use for the current plot. At least that what I recall.
2. Arviz doesn't have bar plots. So it would be good to have an example here of the final output.

cetagostini:
> budget_allocation was not migrated to the new suite on purpose because:
> 1. We thought it was not a great plot. I originally discussed that with @carlosagostini, and we decided that ROA graph makes more sense. We didn't think there is actually a use for the current plot. At least that what I recall.
> 2. Arviz doesn't have bar plots. So it would be good to have an example here of the final output.

Indeed, I'm thinking this migration it's perfect moment to take it out fully from the examples, and totally erase. Wdyt? @juanitorduz

juanitorduz:
perfect! makes sense! thanks for the input

PR #2675, opened 2026-07-01
fix: preserve metadata attrs on sample_prior/sample_posterior_predictive return values

williambdean:
Pass `keep_dataset=True` to `az.extract` in both `sample_prior_predictive` and `sample_posterior_predictive` to prevent xarray's `Dataset.__getitem__` from stripping group-level attrs when converting single-variable groups to DataArrays.

This restores metadata (`created_at`, `inference_library`, `pymc_marketing_version`, etc.) on the returned Dataset, which was lost after the PyMC 6 / DataTree migration.

**Root cause:** `az.extract` converts single-variable datasets to DataArrays via `data[var_name]` (xarray's `__getitem__`), which does not propagate dataset-level attrs to the returned DataArray. `keep_dataset=True` skips this conversion.

**Added tests:** parametrized by `combined=True/False` for both methods, verifying result is a Dataset with metadata attrs.

----
📚 Documentation preview 📚: https://pymc-marketing--2675.org.readthedocs.build/en/2675/

### DIFF
--- a/tests/test_model_builder.py
+++ b/tests/test_model_builder.py
@@ -763,6 +763,28 @@ def test_sample_prior_predictive_has_pymc_marketing_version(
     )
 
 
+@pytest.mark.parametrize("combined", [True, False])
+def test_sample_prior_predictive_returns_dataset_with_attrs(toy_X, combined):
+    model = RegressionModelBuilderTest()
+    result = model.sample_prior_predictive(toy_X, combined=combined)
+    assert isinstance(result, xr.Dataset)
+    assert "pymc_marketing_version" in result.attrs
+    assert "inference_library" in result.attrs
+    assert "inference_library_version" in result.attrs
+
+
+@pytest.mark.parametrize("combined", [True, False])
+def test_sample_posterior_predictive_returns_dataset_with_attrs(
+    fitted_regression_model_instance, toy_X, combined
+):
+    result = fitted_regression_model_instance.sample_posterior_predictive(
+        toy_X, combined=combined, extend_idata=False
+    )
+    assert isinstance(result, xr.Dataset)
+    assert "inference_library" in result.attrs
+    assert "inference_library_version" in result.attrs
+
+
 def test_fit_after_prior_keeps_prior(
     model_with_prior_predictive,
     toy_X,
--- a/pymc_marketing/model_builder.py
+++ b/pymc_marketing/model_builder.py
@@ -1286,10 +1286,9 @@ class RegressionModelBuilder(ModelBuilder):
             else:
                 self.idata = prior_pred
 
-        result = az.extract(prior_pred, "prior_predictive", combined=combined)
-        if isinstance(result, xr.DataArray):
-            result = result.to_dataset()
-
+        result = az.extract(
+            prior_pred, "prior_predictive", combined=combined, keep_dataset=True
+        )
         return result
 
     def sample_posterior_predictive(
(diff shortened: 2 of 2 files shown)


daimon-pymclabs:
The fix looks right — `keep_dataset=True` avoids the `Dataset.__getitem__` path in `az.extract` that was dropping the group-level attrs, and it's cleaner than the manual `.to_dataset()` round-trip it replaces.

One small gap: `test_sample_posterior_predictive_returns_dataset_with_attrs` only asserts `isinstance(result, xr.Dataset)` — it doesn't actually check any attr survived, so it wouldn't have caught the original bug (the pre-fix code also returned a Dataset, just without attrs). The prior-predictive test asserts `"pymc_marketing_version" in result.attrs`; worth mirroring that here (e.g. assert one of `created_at`/`inference_library` is present) so the test name matches what it verifies.

tests/test_model_builder.py
```diff
@@ -763,6 +763,24 @@ def test_sample_prior_predictive_has_pymc_marketing_version(
     )
 
 
+@pytest.mark.parametrize("combined", [True, False])
+def test_sample_prior_predictive_returns_dataset_with_attrs(toy_X, combined):
+    model = RegressionModelBuilderTest()
+    result = model.sample_prior_predictive(toy_X, combined=combined)
+    assert isinstance(result, xr.Dataset)
+    assert "pymc_marketing_version" in result.attrs
```
williambdean, line 771: What is our expectation here? @juanitorduz

tests/test_model_builder.py
```diff
@@ -763,6 +763,24 @@ def test_sample_prior_predictive_has_pymc_marketing_version(
...
+
+@pytest.mark.parametrize("combined", [True, False])
+def test_sample_posterior_predictive_returns_dataset_with_attrs(
+    fitted_regression_model_instance, toy_X, combined
+):
+    result = fitted_regression_model_instance.sample_posterior_predictive(
+        toy_X, combined=combined, extend_idata=False
+    )
+    assert isinstance(result, xr.Dataset)
+
```
williambdean, line 786: @daimon-pymclabs please push directly here with some actual checks of the attrs

daimon-pymclabs:
@williambdean pushed 632dcdc with real attr checks:

- `test_sample_prior_predictive_returns_dataset_with_attrs`: now asserts `result.attrs["pymc_marketing_version"] == __version__` (value, not just presence) plus that the standard ArviZ/PyMC metadata attrs (`created_at`, `inference_library`, `inference_library_version`) survive the `az.extract` conversion.
- `test_sample_posterior_predictive_returns_dataset_with_attrs`: previously only checked `isinstance(..., Dataset)`; now also asserts those same metadata attrs are preserved — i.e. the exact regression this PR fixes. (The posterior-predictive path doesn't stamp `pymc_marketing_version`, so I checked the metadata attrs instead.)

Introspected the returned Datasets locally to confirm which attrs are actually present before writing the assertions; all 6 tests pass.

williambdean:
@daimon-pymclabs please make sure that pre commit is fixed

daimon-pymclabs:
@williambdean pre-commit was failing on an import-ordering issue in my test commit (`from pymc_marketing import __version__` was placed out of order). Fixed it with ruff and pushed (3f099f5). pre-commit passes clean locally now — the pre-commit.ci run should go green on this push.

juanitorduz (approved):

Issue #2614, opened 2026-06-02
Migrate all notebooks to `pymc>6.0` and `arviz>1.0`


juanitorduz:
We want to migrate all the notebooks. Let's do it ONE BY ONE on separate PRs 

- The target branch should be [`v1.0.0`](https://github.com/pymc-labs/pymc-marketing/tree/v1.0.0)
- For MMM notebooks, we should use `from pymc_marketing.mmm import MMM` (not multidimensional!)
- We need to use the new plotting namespace https://github.com/pymc-labs/pymc-marketing/blob/main/docs/source/notebooks/mmm/mmm_plot_suite_migration_guide.ipynb
- We can keep using `az.style.use("arviz-darkgrid")`

#### Fundamentals
- [x] `mmm/mmm_quickstart` — MMM Quickstart Guide, https://github.com/pymc-labs/pymc-marketing/pull/2657
- [x] `mmm/mmm_example` — MMM Example: Extensive Introduction Notebook, https://github.com/pymc-labs/pymc-marketing/pull/2657
- [x] `mmm/mmm_multidimensional_example` — MMM Multidimensional Example Notebook (e.g. Geo-MMM), https://github.com/pymc-labs/pymc-marketing/pull/2657
- [x] `mmm/mmm_media_saturation` — Understanding Media Saturation https://github.com/pymc-labs/pymc-marketing/pull/2672/
- [x] `mmm/adstock_functions_guide` — Guide to Adstock Transformations, https://github.com/pymc-labs/pymc-marketing/pull/2664
- [x] `mmm/mmm_build_from_yml_example` — Build MMM from YAML Configuration https://github.com/pymc-labs/pymc-marketing/pull/2665
- [x] `mmm/mmm_sensitivity_analysis` — Sensitivity Analysis and Marginal Effects https://github.com/pymc-labs/pymc-marketing/pull/2661
- [x] `mmm/mmm_multiplicative` — Additive vs Log vs Log-Log MMMs https://github.com/pymc-labs/pymc-marketing/pull/2672/
- [x] `mmm/plot_interactive` — Interactive MMM Visualizations https://github.com/pymc-labs/pymc-marketing/pull/2672/
- [x] `mmm/mmm_cost_per_unit` — Cost-per-Unit: Model What Drives Revenue, Report What You Spend https://github.com/pymc-labs/pymc-marketing/pull/2665
[...]

PR #2683, opened 2026-07-03
[Notebook Re-Run] TVP Baseline

timbo112711:
## Description

Re-run of the TVP Baseline notebook

Part of https://github.com/pymc-labs/pymc-marketing/issues/2614.

Changes:

- Switch to arviz-darkgrid style
- Update plot annotations
- Adds `sampler_config=copy.deepcopy(sampler_config)`

`sampler_config` is a plain dict defined once at module level and passed into `MMM()` on every call to `create_and_fit_mmm`. Inside `MMM.__init__`, PyMC mutates that dict and it pulls `target_accept` out of the top level and moves it into a nested nuts key. Because it's the same dict object, the second call to `create_and_fit_mmm` receives a `sampler_config` that now has `target_accept` in both the top-level dict (still there from the original definition) and the nuts sub-dict (added by the first call's mutation). PyMC then raises:

```ValueError: `target_accept` was defined twice.```

The fix was to pass copy.deepcopy(sampler_config) instead of sampler_config directly:

```
mmm = MMM(
    ...
    sampler_config=copy.deepcopy(sampler_config),
    ...
)
```

Each call now gets its own independent copy of the dict, so mutations from one call don't bleed into the next.

@juanitorduz @williambdean for the above we can add in a one-liner to the model builder in `create_samlpe_kwargs` that fixes this on the backend. Should we do that here, open up another issue, or even need to worry about it??


juanitorduz (approved):
Thanks @timbo112711

PR #2684, opened 2026-07-04
fix(clv): restore (chain, draw) dims on ShiftedBetaGeo new-customer distributions

### LINKED ISSUE #2651: ShiftedBetaGeoModelIndividual.distribution_new_customer_* dims changed from (chain, draw) to (sample)
PyMC 6 migration (#2616) stacks chain/draw → sample at shifted_beta_geo.py:970. Breaking API change for distribution_new_customer_churn_time and distribution_new_customer_theta. Evaluate whether to keep sample or restore original dims.

Flagged by @daimon-pymclabs in #2616 review. @ColtAllen for CLV context.

anevolbap:
Fixes #2651

The pymc 6 migration (#2616) added `result.stack(sample=("chain","draw")).reset_index("sample")` at the end of `ShiftedBetaGeoModelIndividual._distribution_new_customer`, flattening `distribution_new_customer_theta` and `distribution_new_customer_churn_time` to a single `sample` dim. On pymc 6 the raw `post_pred["posterior_predictive"].dataset` already carries `(chain, draw)`, so the stack was the only thing collapsing it. That is a silent breaking change against the pre-migration public API and inconsistent with `distribution_customer_churn_time` and the other CLV models, which all keep `(chain, draw)`.

## Change
- Return `post_pred["posterior_predictive"].dataset` directly, keeping the real pymc 6 bits (`sample_vars`, `list(var_names)`, `.dataset`).
- Revert the `sample` assertions in the two affected tests back to `(chain, draw)`, and add a `"sample" not in res.dims` check so the next upstream change is caught in CI.

## Verification
Installed this branch on pymc 6.0.1 / arviz 1.2 and ran `tests/clv/models/test_shifted_beta_geo.py`: `39 passed, 4 skipped`. Baseline reproduces green with the old `sample` assertions, confirming the flip.

Discussion and repro in #2651.

----
📚 Documentation preview 📚: https://pymc-marketing--2684.org.readthedocs.build/en/2684/


williambdean (approved):
Thanks @anevolbap

PR #2686, opened 2026-07-06
Update model data before repeated fit

### LINKED ISSUE #2605: mmm.fit(X, y) is ignored on second run or after calling build_model
https://github.com/pymc-labs/pymc-marketing/blob/e2982b61493e27e18f54aecb80e2bdfd567ffe22/pymc_marketing/model_builder.py#L999-L1027

This simply ignores X, y... the API strongly suggests you should be able to call it twice?

ahfoysal:
Fixes #2605.

When a regression model has already been built, `fit(X, y)` now updates the model data before sampling. This keeps a second fit, or a fit after `build_model`, from sampling with stale inputs.

I added a regression test that builds with one dataset, fits with another, and checks both the PyMC data containers and stored `fit_data` use the new inputs.

Checks:
- `uvx ruff check pymc_marketing/model_builder.py tests/test_model_builder.py`
- `python3.12 -m py_compile pymc_marketing/model_builder.py tests/test_model_builder.py`

----
📚 Documentation preview 📚: https://pymc-marketing--2686.org.readthedocs.build/en/2686/


daimon-pymclabs:
Thanks for this, @ahfoysal — and welcome! This is a really nice first contribution: the diagnosis is spot on, the fix is the minimal expression of the intent, the commit is clean, the issue is linked, there's a test, and you'd already merged `main` before anyone asked. `RegressionModelBuilder.fit` genuinely did silently drop `X, y` when a model already existed, and that `else` branch is the right shape for it.

I ran this on a v1 environment (pymc 6.0.1 / pytensor 3.0.7 / arviz 1.2.0) with the branch installed editable. `tests/test_model_builder.py` → 78 passed, and downstream `tests/customer_choice/test_mv_its.py` (24 passed) and `tests/pie` (31 passed, 1 skipped) are clean too.

There are two things I'd like to work through with you before it lands. Neither is a problem with your reasoning — the second one is really a design question about the codebase that we should answer, not you.

### 1. `fit(X)` with `y` omitted now writes zeros into the observed data

A few lines above your new branch, `fit` does:

```python
if y is None:
    y = np.zeros(X.shape[0])
```

So by the time `self._data_setter(X, y)` runs, `y` is a zeros array rather than `None`, and a conforming `_data_setter` writes those zeros in as the observation. Measured on this branch with a subclass following the pattern documented on the abstract method:

```
observed y_data after build_model(X, y):   [3.0  3.0847  3.1695  3.2542]
observed y_data after fit(X) w/ y omitted: [0.  0.  0.  0.]
```

On `main` this path was a no-op, so the real `y` survived. Calling `fit(X)` with the target carried inside `X` is documented usage (`MMM.build_model`: *"If omitted, X must contain a target variable"*), so it's reachable. Tracking whether `y` was actually supplied, and only forwarding it when it was, should close this cleanly.

### 2. The change can't reach MMM, which is what #2605 is about

This one isn't obvious from the outside at all, so please don't take it as a miss. `MMM._data_setter` is a no-op stub:

```python
def _data_setter(self, X, y=None): ...
```

MMM does its data swapping in `_set_xarray_data` via its own `sample_posterior_predictive`, so the stub only exists to satisfy the abstract method. Measured end-to-end on this branch, `build_model(X1, y1)` then `fit(X2, y2)`:

```
channel_data STILL equals the X1 values: True
target_data  STILL equals the y1 values: True
target_scale STILL the y1 scale:         True
fit_data     c1 == X2 c1:                True
```

So the posterior is still fit to X1/y1 while `idata.fit_data` records X2/y2. (That inconsistency predates your PR — `create_fit_data(X, y)` runs unconditionally at the end of `fit`.) `MVITS._data_setter` is an empty stub as well, so of the three in-repo `RegressionModelBuilder` subclasses, `PIEModel` is the only one your fix currently benefits.

Behind this sits a design decision that's on us rather than you: MMM's scalers and coords are computed in `_generate_and_preprocess_model_data` at build time, so `pm.set_data` alone can't be correct for a genuinely different dataset. Either MMM's `_data_setter` delegates to `_generate_and_preprocess_model_data` + `_set_xarray_data`, or the `else` branch rebuilds via `self.build_model(X, y)`. Rebuilding is simpler and matches "fit means fit to this data", but it would discard post-build mutations like `add_original_scale_contribution_variable`. Happy to hash this out here — and equally happy for the PR to be scoped to the base-class fix with a follow-up issue for MMM, if you'd prefer that. Just say which you'd rather do.

### 3. A note on the test

Since `fit` ends with `fit_data = self.create_fit_data(X, y)` unconditionally, the two `idata.fit_data` assertions also pass on `main`. I reverted just the source hunk and re-ran: the only failing assertion is the `model.model["x"].get_value()` check at line 854. Nothing wrong with keeping the extra assertions, it's just worth knowing that the `get_value()` pair is what's actually pinning the fix — a comment marking it as the regression assertion would help future readers.

### Optional follow-ups, all fine as-is

- A second fit with a different row count raises `ValueError: The 'x' variable already had 10 coord values defined for its numbers dimension...` out of `pm.set_data`. Loud and reasonable, but untested, and the base `fit` never forwards coords.
- For `PIEModel`, `_target_scale` and `_encoders` are frozen at build time, and BART's leaf-prior `Y=y_scaled` can't be `set_data` at all, so a second fit reuses the first dataset's scaling and BART prior.
- The abstract `_data_setter` docstring still frames the method as prediction-only, and after this change it sits on the fit path too — a one-line update would keep it honest.

One housekeeping item on our side: the `Test`, `Docs`, and `Test Notebooks` workflows are all sitting in `action_required` (standard for first-time contributors), so CI hasn't actually run on GitHub yet. A maintainer needs to click "Approve and run workflows" — I'll make sure that happens.

Thanks again for digging into this one. It's a subtle bug in a corner of the codebase that's easy to get wrong, and you found the right line.

.vscode/extensions.json
williambdean, line 1: We will not want any .vscode of public/fonts

.gitignore
```diff
@@ -158,3 +157,10 @@ sandbox/
 
 .cursor/plans
 automatic_translation.py
+
+# uv
+uv.lock
+branch_structure.json
```
williambdean, line 163: Not needing this change either

pymc_marketing/mmm/mmm.py
```diff
@@ -4050,15 +4021,19 @@ def optimize_budget(
...
-        response_variable: str | None = None,
+        response_variable: str = "total_media_contribution_original_scale",
         utility_function: UtilityFunctionType = average_response,
         constraints: Sequence[Constraint] = (),
+        default_constraints: bool = True,
         budgets_to_optimize: xr.DataArray | None = None,
         budget_distribution_over_period: xr.DataArray | None = None,
         cost_per_unit: pd.DataFrame | xr.DataArray | None = None,
         callback: bool = False,
         **allocate_budget_kwargs,
```
williambdean, line 4032: Seems unrelated to PR. Can you rebase

williambdean (commented):
Thanks for starting the PR. It seems to not have the changes yet. Feel free to put in draft state.

PR #2630, opened 2026-06-12
docs(bass): update notebook to the BassModel workflow

anevolbap:
Closes #2590
Closes #1706

Retargeted to `v1.0.0`. #2624 has merged; this is now stacked on #2628 only.

Rewrites `bass_example.ipynb` around the `BassModel` class: `fit` with `model_config` / `sampler_config`, the `plot_*` methods from #2628, out-of-sample forecasting, a save/load round-trip, and MLflow autologging from #2624. `create_bass_model` stays in the simulation section as the lower-level API.

Two source fixes with regression tests:

- `BassModel.load` returned plain dicts instead of `Prior` objects; fixed with `parse_model_config` in `__init__` plus a serialization round-trip for `Scaled` priors.
- Multi-product forecasting without observed data failed on a 1-D `y_obs` placeholder.

The notebook was re-executed on the v1 stack (pymc 6, pytensor 3, arviz 1.2, arviz-plots 1.2) with real sampling. With #2624 merged, the mlflow commits have dropped out of this diff. Merging #2628 next and rebasing shrinks it further to just the notebook plus the two fixes.

Note: `bass/bass_example` was re-executed for the pymc6/arviz migration in #2658 (part of #2614). This PR goes further and rewrites the notebook around the `BassModel` class per #2590, building on the #2658 version (re-executed on `v1.0.0`). It supersedes the Bass entry in the #2614 checklist.

Issue #2598, opened 2026-05-18
Migrate Bass Model dimension handling from create_dim_handler to pymc.dims


williambdean:
## Summary

The `create_bass_model` function currently uses `create_dim_handler` from `pymc_extras.prior` to align variables with different dimensions (e.g., `m` has dims `("product",)`, `p` has dims `("region",)`, time has `("T",)`). The rest of the codebase (MMM, special priors, etc.) has migrated to `pymc.dims` (PR #2204), and the Bass model should follow suit.

**Status**: Waiting for upstream `pymc.dims` to support all distributions used by the Bass model. See "Blockers" below.

## Motivation

- Consistency with the rest of the codebase (MMM, special priors, etc.)
- Remove `create_dim_handler` dependency from `pymc_extras.prior` — this is the last major usage outside the Prior class itself
- Leverage named dims in the type system for automatic broadcasting

## Blockers

The ideal migration uses `xdist=True` on all priors:

```python
m = priors["m"].create_variable("m", xdist=True)
p = priors["p"].create_variable("p", xdist=True)
q = priors["q"].create_variable("q", xdist=True)
```

This keeps everything in xtensor space from the start — clean, no back-conversion needed.

However, two-step patterns like creating regular variables and then converting with `pmd.as_xtensor` are not acceptable:

```python
# NOT the desired pattern
m_reg = priors["m"].create_variable("m")
m_xt = pmd.as_xtensor(m_reg, dims=priors["m"].dims or ())
```

For `xdist=True` to work everywhere, `pymc.dims` needs to support all distributions the Bass model uses:

| Distribution | In pymc.dims? | Used for |
|---|---|---|
| Normal | Yes | m prior |
| Beta | Yes | p, q priors |
| Gamma | Yes | m prior |
| HalfNormal | Yes | m prior |
| Poisson | **No** | Likelihood (default) |
| NegativeBinomial | **No** | Likelihood (alternative) |
| DiracDelta | **No** | Fixed m in examples |

Poisson is the critical missing piece. It's the default likelihood and the most commonly used.

### pymc_marketing/bass/model.py

- Import `pymc.dims as pmd`, remove `create_dim_handler` import
- Use `xdist=True` on all Prior.create_variable calls (m, p, q)
- Replace `pm.Deterministic(... dims=...)` with `pmd.Deterministic(...)` for adopters, innovators, imitators, peak
[...]

PR #2724, opened 2026-07-09
refactor(bass): migrate dim handling to pymc.dims

anevolbap:
Addresses #2598

Migrates the Bass model's dimension handling from `create_dim_handler` (`pymc_extras.prior`) to `pymc.dims`, for consistency with the rest of the codebase.

Stacked on #2628 and #2630. The diff is large because it includes those two; once they merge into `v1.0.0` and this rebases, it shrinks to just `pymc_marketing/bass/model.py`.

In `create_bass_model`:

- `m`/`p`/`q` via `create_variable(xdist=True)`; deterministics via `pmd.Deterministic`.
- Inline xtensor `_f_xt`/`_F_xt` for the model graph; the public `F`/`f` stay on PyTensor for the float/NumPy API and tests.
- `t` and `y_obs` stay `pm.Data` wrapped with `pmd.as_xtensor`, so the data-setter and out-of-sample forecasting keep working. `as_xtensor` is applied only to data, not to priors.
- Deterministics transposed back to the historical `(T, ...)` order.
- The xdist likelihood path does not accept `observed=None`, so the no-data (prior predictive) path sets `mu` on a copy and creates the variable directly.

`DiracDelta` is not in `pymc.dims` in released pymc, so for that distribution only it falls back to a regular variable wrapped as an xtensor. It is implemented upstream (pymc-devs/pymc#8350, shipping in 6.1.0); the fallback can be removed once a pymc release includes it and the pin allows it.

Verified: full bass test suite, mlflow autolog, `adopters`/`peak` matching the closed-form formula to machine precision, and the bass notebook mock runner.

Draft until the stacked PRs merge.


pymc_marketing/bass/model.py
```diff
@@ -237,6 +243,36 @@ def f(
...
+    NumPy arrays; this variant uses ``pymc.dims`` ops so it composes with
+    named-dims (xtensor) variables.
+    """
+    return (1 - pmd.math.exp(-(p + q) * t)) / (1 + (q / p) * pmd.math.exp(-(p + q) * t))
+
+
+def _f_xt(p, q, t):
+    """Xtensor version of :func:`f` for the ``pymc.dims`` model graph."""
+    exp_t = pmd.math.exp(t * (p + q))
+    return (p * (p + q) ** 2 * exp_t) / (p * exp_t + q) ** 2
```
williambdean: Let's just consolidate with the F . No need to support both.

pymc_marketing/bass/model.py
```diff
@@ -237,6 +243,36 @@ def f(
...
+
+    Uses ``xdist=True`` so the variable lives in xtensor space from the
+    start. ``pymc.dims`` does not yet support ``DiracDelta``; for that case
+    only, create a regular variable and wrap it with ``pmd.as_xtensor``.
+    This wrap is applied to data, not to a supported prior, so it does not
+    reintroduce the two-step conversion pattern for the general case.
+    """
+    if getattr(prior, "distribution", None) == "DiracDelta":
+        return pmd.as_xtensor(prior.create_variable(name), dims=prior.dims or ())
+    return prior.create_variable(name, xdist=True)
```
williambdean, line 266: Let's add an issue to deprecate this when the pymc PR goes through and next release

pymc_marketing/bass/model.py
```diff
@@ -325,37 +361,42 @@ def create_bass_model(
...
-        )
+        if observed is None:
+            # No data: an unobserved likelihood for prior predictive. The
+            # xdist likelihood path does not accept ``observed=None``, so set
+            # ``mu`` on a copy and create the variable directly.
+            likelihood = priors["likelihood"].deepcopy()
+            likelihood.parameters["mu"] = adopters
+            likelihood.create_variable("y", xdist=True)
+        else:
+            observed_xt = pmd.as_xtensor(observed, dims=combined_dims)
```
williambdean, line 377: Do we still need the combined_dims then. xtensor should be able to handle the dims logic, right?

pymc_marketing/special_priors.py
```diff
@@ -65,6 +66,25 @@ def _is_xarray_dataarray_dict(data: Any) -> bool:
 )
 
 
+def _serialize_scaled(obj: Scaled) -> dict:
+    return {
+        "__type__": f"{Scaled.__module__}.{Scaled.__qualname__}",
+        "dist": obj.dist.to_dict(),
+        "factor": obj.factor,
+    }
```
williambdean, line 74: I think this should already be injected? 

EDIT: Seeing that `to_dict` isn't handled in pymc_extras.prior

williambdean (commented):
Looks pretty solid. Thanks for starting.

pymc_marketing/bass/model.py
anevolbap: Done in a3d14ee4. A single `_exp` helper dispatches `pmd.math.exp` for xtensor and `pt.exp` otherwise, so `F`/`f` work for both the float/NumPy API and the model graph. Dropped the duplicate `_f_xt`/`_F_xt`.

pymc_marketing/bass/model.py
anevolbap: Opened #2726 to track removing the fallback once a pymc release ships `dims.DiracDelta` (merged in pymc-devs/pymc#8350).

pymc_marketing/bass/model.py
anevolbap: Dropped the transpose in a3d14ee4, xtensor orders the dims and consumers select by name. `combined_dims` stays only to wrap the observed data into an xtensor with the right dims and to set the likelihood dims, which cannot be inferred from a raw array.

pymc_marketing/special_priors.py
anevolbap: Right, `pymc_extras.prior` does not serialize `Scaled` (`to_dict` is not handled there), so the save/load round-trip needs this. It comes from #2630 and shows here because #2724 is stacked on it.

anevolbap:
CI's oldest-deps job runs pymc 6.0.0, where `pymc.dims` doesn't have `Poisson` yet (added in 6.0.1). This PR uses `xdist=True` for the Poisson likelihood, so it fails there. It's fine on 6.0.1+.

Simplest path: keep this a draft until v1.0.0's pymc floor is `>=6.1.0`, which has every distribution the Bass model needs (the DiracDelta fallback also goes away then). I'll rebase once the floor moves.

anevolbap:
<p>Done. It is now a figure legend to the right of the grid instead of inside the first subplot. Fixed in plot_decomposition itself, so every caller gets it.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2724/discussion/#comment-4929441061'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

anevolbap:
<p>Right, this was just <code>ipywidgets</code> missing from the run env (it is a declared dep). Re-ran with it installed, so the progress bars render as widgets and the warning is gone, matching the other v1 re-run notebooks. No code change.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2724/discussion/#comment-4929441188'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

anevolbap:
<p>&nbsp;Done. The MLflow section called plot_adoption_curve() / plot_decomposition() without subplot_kwargs, so they fell back to a single row. Added subplot_kwargs={"ncols": 3, "figsize": (15, 12)} to both, matching the other plot cells, and re-ran. They render as a 3×3 grid now.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2724/discussion/#comment-4929441287'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

pymc_marketing/special_priors.py
anevolbap: Opened pymc-devs/pymc-extras#709 to add serialization to `Scaled` upstream. Once it lands, this shim can go away and save/load will round-trip `Scaled` natively.

anevolbap:
Following up on the oldest-deps failure above: it is only the `pymc==6.0.0` shards, where `pymc.dims` lacks `Poisson`/`NegativeBinomial` (both landed in 6.0.1). Raising the v1.0.0 floor to `pymc>=6.0.1` would clear it with no code change here, and it does not widen the test surface, the non-oldest shards already resolve 6.0.1 and pass. This keeps the DiracDelta fallback, so it stays on pymc 6.0.x and does not need the larger `>=6.1.0` bump (that one belongs to the separate follow-up that lifts the pymc-extras pin and drops the fallback). Would it be possible to move the v1.0.0 floor to 6.0.1?

Issue #2722, opened 2026-07-09
Resolve `xarray` FutureWarning for variable combination


ColtAllen:
Future updates to the `xarray` dependency may cause behavior changes, as described in this warning when predictions are ran for `ShiftedBetaGeoModel`:

```shell
pymc-marketing/pymc_marketing/clv/models/shifted_beta_geo.py:536: FutureWarning: In a future version of xarray the default value for compat will change from compat='no_conflicts' to compat='override'. 
This is likely to lead to different results when combining overlapping variables with the same name. 
To opt in to new defaults and get rid of these warnings now use `set_options(use_new_combine_kwarg_defaults=True) or set compat explicitly.
  return xarray.combine_by_coords(
```

ColtAllen:
Resolved in https://github.com/pymc-labs/pymc-marketing/pull/2728

PR #2723, opened 2026-07-09
Remove support for pre-v1.0.0 CLV API

ColtAllen:
## Description

This PR removes all support and deprecation messages for the pre-`v1.0.0` CLV API. For users who have not yet migrated to the new API, this will be a breaking change in `v1.0.0`.

## Related Issue

- [ ] Closes https://github.com/pymc-labs/pymc-marketing/issues/1516, https://github.com/pymc-labs/pymc-marketing/issues/2082

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--2723.org.readthedocs.build/en/2723/

PR #2728, opened 2026-07-12
Fix `xarray` combine_by_coords FutureWarning

ColtAllen:
## Related Issue

- [ ] Closes https://github.com/pymc-labs/pymc-marketing/issues/2722
- [ ] Related to https://github.com/pymc-labs/pymc-marketing/pull/2723

## Context

xarray 2026.4.0 emits a `FutureWarning` on every `combine_by_coords()` call because the default `compat` will change from `'no_conflicts'` to `'override'` in a future release. This warning fires in all four CLV model `_extract_predictive_variables` methods, which combine model parameters (alpha, beta, r, s, etc.) with customer variables (T, frequency, recency) into a single `xarray.Dataset`.

Since all combined variables have **distinct names** (no overlapping variables), both `compat='no_conflicts'` and `compat='override'` produce identical results. The fix is to explicitly pass `compat` to silence the warning and lock in deterministic behavior.

## Approach: Explicit `compat` at each call site

Pass `compat="override"` explicitly to all four `combine_by_coords()` calls. This:
- Opts into the future default, so no behavior change when xarray eventually flips
- Is semantically correct — we're combining distinctly-named variables, so no conflict checking is needed
- Is slightly faster than `'no_conflicts'` since it skips value comparisons

**Why not `xarray.set_options(use_new_combine_kwarg_defaults=True)` in `__init__.py`?**
That flag changes defaults for `concat`, `merge`, and `open_mfdataset` too — not just `combine_by_coords`. It also changes `join` from `'outer'` to `'exact'` and `coords` from `'different'` to `'minimal'`. These side effects could break other xarray usage in the library or downstream user code. Explicit `compat` at each call site is safer and more targeted.

## Files to modify

1. **`pymc_marketing/clv/models/shifted_beta_geo.py`** (line 536) — add `compat="override"`
2. **`pymc_marketing/clv/models/pareto_nbd.py`** (line 502) — add `compat="override"`
3. **`pymc_marketing/clv/models/beta_geo.py`** (line 463) — add `compat="override"`
4. **`pymc_marketing/clv/models/beta_geo_beta_binom.py`** (line 336) — add `compat="override"`

Each change is the same pattern — add a single kwarg:
```python
# Before
xarray.combine_by_coords((var1, var2, ...))

# After
xarray.combine_by_coords((var1, var2, ...), compat="override")
```

## Tests

Add a test to each model's existing test file that asserts `_extract_predictive_variables` does not emit a `FutureWarning`:

- `tests/clv/models/test_shifted_beta_geo.py`
- `tests/clv/models/test_pareto_nbd.py`
- `tests/clv/models/test_beta_geo.py`
- `tests/clv/models/test_beta_geo_beta_binom.py`

Each test will:
1. Use an existing fitted model fixture from the test file
2. Call `_extract_predictive_variables` inside `pytest.warns` / `warnings.catch_warnings` to assert no `FutureWarning` is raised
3. Verify the returned dataset contains expected variables

## Verification

```bash
uv run python -W error::FutureWarning -c "
import warnings; warnings.filterwarnings('error', category=FutureWarning)
# Quick smoke test that none of the combine_by_coords calls emit warnings
"
uv run pytest tests/clv/models/ -x -q --no-header -W error::FutureWarning 2>&1 | tail -20
uv run pre-commit run --files pymc_marketing/clv/models/shifted_beta_geo.py pymc_marketing/clv/models/pareto_nbd.py pymc_marketing/clv/models/beta_geo.py pymc_marketing/clv/models/beta_geo_beta_binom.py
```

----
📚 Documentation preview 📚: https://pymc-marketing--2728.org.readthedocs.build/en/2728/


tests/clv/models/test_beta_geo_beta_binom.py
```diff
@@ -320,6 +320,21 @@ def test_fit_result_without_fit(self, mocker, model_config):
         assert len(idata.posterior.draw) == 10
         assert model.idata is idata
 
+    def test_extract_predictive_variables_no_future_warning(self):
+        import warnings
+
+        pred_data = self.pred_data.assign(future_t=1)
+        with warnings.catch_warnings():
+            warnings.simplefilter("error", FutureWarning)
```
ricardoV94: pytest.filterwarning("error") can be used to do the same. decate the test function with it or assign a pytestmark = ... at the top to cover all tests in the file

ColtAllen: Requested change has been applied.

tests/clv/models/test_beta_geo_beta_binom.py
```diff
@@ -319,6 +319,18 @@ def test_fit_result_without_fit(self, mocker, model_config):
         assert len(idata.posterior.draw) == 10
         assert model.idata is idata
 
+    @pytest.mark.filterwarnings("error::FutureWarning")
```
ricardoV94: Why FutureWarning only? Presumably we don't want any warnings whatsoever?

ColtAllen: Perhaps we just remove these tests altogether? `FutureWarning` no longer appears in `sbg.ipynb` (see watermark at bottom for before/after verification.)

Screenshot 2026-07-13 at 2 19 24 PM
Screenshot 2026-07-13 at 2 25 37 PM

ricardoV94:
didn't they disappear because you fixed the warning?

ricardoV94:
I don't think we need the new tests, I would just make the test module fail on any warning `pytestmark = pytest.filterwarnings("error")` and then it's still a regression test

ColtAllen:
> didn't they disappear because you fixed the warning?

Correct - unit tests checking for warnings seems unnecessary now.

OriolAbril (approved):

ColtAllen:
> I don't think we need the new tests, I would just make the test module fail on any warning `pytestmark = pytest.filterwarnings("error")` and then it's still a regression test

Claude is barking about this being too aggressive. It is causing `DeprecationWarning` tests for the `pymc_extras.Prior` class to fail:

```shell
FAILED tests/clv/models/test_beta_geo.py::TestBetaGeoModel::test_posterior_distributions[map] - DeprecationWarning: alpha is automatically converted to Prior("Weibull", al...
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
==================== 1 failed, 5 p
```

@juanitorduz @williambdean should this `Prior` class deprecation be removed for `v1.0.0` as well?

ricardoV94:
> Claude is barking about this being too aggressive. It is causing `DeprecationWarning` tests for the `pymc_extras.Prior` class to fail:

That's exactly why it's good to fail at warnings. So we have time to fix it before the change. Anyway you can open an issue and exclude this specific warning with "ignore:xxx"

juanitorduz:
@ColtAllen, is this one good to merge? Can you please open the issue Ricardo suggested 🙏 ?

williambdean:
> > Claude is barking about this being too aggressive. It is causing `DeprecationWarning` tests for the `pymc_extras.Prior` class to fail:
> 
> That's exactly why it's good to fail at warnings. So we have time to fix it before the change. Anyway you can open an issue and exclude this specific warning with "ignore:xxx"

🙌

ColtAllen:
> > That's exactly why it's good to fail at warnings. So we have time to fix it before the change.
> 
> 🙌

Damn straight 💯

> Can you please open the issue Ricardo suggested 🙏 ?

https://github.com/pymc-labs/pymc-marketing/issues/2732

PR #2742, opened 2026-07-17
fix(model_builder): update model data on repeated fit (#2605)

### LINKED ISSUE #2605: mmm.fit(X, y) is ignored on second run or after calling build_model

shivamlalakiya:
## Description

Fixes #2605

```python
if not hasattr(self, "model"):
    self.build_model(X, y)

So a second fit, or a fit after an explicit build_model, silently ignored the new X, y and sampled from the data baked in when the model was first built. fit_data, meanwhile, was recorded from the new X, y — so the stored data and the data actually sampled disagreed.

This PR refreshes the existing model's data via the existing _data_setter mechanism (the same one sample_posterior_predictive uses) before sampling, in both fit and approximate_fit:
    self.build_model(X, y)
else:
    self._data_setter(X, y)
```

Added regression tests (test_second_fit_uses_new_data, test_fit_after_build_model_uses_new_data, test_approximate_fit_after_build_model_uses_new_data) that assert both the PyMC data containers and stored fit_data use the new inputs. They fail on main (100% mismatched elements) and pass with this change.

Scope notes: Fixes the base class and subclasses with a real _data_setter (e.g. BassModel). MMM and mv_its implement _data_setter as a deliberate no-op (MMM stores scaled data and re-derives scaling in build_model), so they are not changed here — but they aren't regressed either, and the documented build_model → add_lift_test_measurements → fit workflow is preserved. A real MMM refit path can be a follow-up. Overlaps with #2686 (which changes fit only); this PR also covers approximate_fit and adds broader tests.

Related Issue

- [x] Closes #2605
- [ ] Related to #

Checklist

- [x] Checked that the pre-commit linting/style checks pass. Feel free to comment pre-commit.ci autofix to auto-fix.
- [ ] Added necessary documentation (docstrings and/or example notebooks) using numpydoc format.

----
📚 Documentation preview 📚: https://pymc-marketing--2742.org.readthedocs.build/en/2742/

### DIFF
--- a/pymc_marketing/model_builder.py
+++ b/pymc_marketing/model_builder.py
@@ -1054,6 +1054,11 @@ class RegressionModelBuilder(ModelBuilder):
 
         if not hasattr(self, "model"):
             self.build_model(X, y)
+        else:
+            # Refresh data on the existing model so repeated fits (or a fit
+            # after build_model) sample from the new X, y instead of the
+            # values baked in when the model was first built.
+            self._data_setter(X, y)
 
         sampler_kwargs = create_sample_kwargs(
             self.sampler_config,
@@ -1195,6 +1200,11 @@ class RegressionModelBuilder(ModelBuilder):
 
         if not hasattr(self, "model"):
             self.build_model(X, y)
+        else:
+            # Refresh data on the existing model so repeated fits (or a fit
+            # after build_model) sample from the new X, y instead of the
+            # values baked in when the model was first built.
+            self._data_setter(X, y)
 
         # Prepare kwargs for pymc.fit
         _fit_kwargs: dict[str, Any] = {}
--- a/tests/test_model_builder.py
+++ b/tests/test_model_builder.py
@@ -842,6 +842,61 @@ def test_second_fit(toy_X, toy_y, mock_pymc_sample):
     assert id_before != id_after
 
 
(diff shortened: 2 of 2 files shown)


shivamlalakiya:
@ricardoV94 this should resolve #2605.

`RegressionModelBuilder.fit()` and `approximate_fit()` only ran `build_model(X, y)` when no model existed yet, so a second `fit()` — or a `fit()` after an explicit `build_model()` — silently sampled the data baked in at build time and ignored the new `X, y`. The fix refreshes the existing model via `_data_setter(X, y)` before sampling. `MMM.fit()` delegates to `super().fit()`, so `mmm.fit(X, y)` is covered as well.

The regression tests assert the PyMC data containers (and `idata.fit_data`) actually hold the new `X, y` after a second fit, a fit-after-`build_model`, and `approximate_fit`.

One heads-up: I rebased the branch onto `v1.0.0` and retargeted the PR base there, since that's where the `blackjax<1.6` CI fix lives — **all checks are green** now. Happy to rebase back onto `main` if that's the intended base instead.

Does this look good, or is there anything you'd like changed?

Issue #2471, opened 2026-04-01
Route `yearly_seasonality` and `control_columns` through the `MuEffect` protocol internally


williambdean:
## Summary

`yearly_seasonality` and `control_columns` are currently built inline inside `build_model` and `_set_xarray_data` in `multidimensional.py`. The `MuEffect` protocol (`create_data` / `create_effect` / `set_data`) already exists and handles exactly this pattern — `FourierEffect`, `LinearTrendEffect`, and `EventAdditiveEffect` all use it. This issue tracks refactoring the two remaining inline effects to go through the same protocol, making the model-building pipeline uniform.

**No public API changes.** `yearly_seasonality` and `control_columns` remain as `__init__` parameters. Deterministic variable names (`yearly_seasonality_contribution`, `fourier_contribution`, `control_contribution`) are preserved for backward compatibility.

---

### 1. New `ControlEffect(MuEffect)` in `additive_effect.py`

No `MuEffect` subclass exists for control columns today. A new one is needed:

- `create_data`: registers `pmd.Data("control_data", mmm.xarray_dataset._control)`
- `create_effect`: creates the `gamma_control` prior variable, multiplies by `control_data`, returns `control_contribution.sum(dim="control")`
- `set_data`: calls `pm.set_data({"control_data": ...})` with dtype handling
- `to_dict` / `from_dict`: custom implementations for the nested `Prior` field, following the same pattern as `FourierEffect` — written to be compatible with the `SerializableBaseModel` / `TypeRegistry` system from #2391

The `Prior` for `gamma_control` is snapshotted from `model_config["gamma_control"]` at construction time so the effect is self-contained.

### 2. Add `contribution_name: str | None = None` to `FourierEffect`

`FourierEffect.create_effect` currently names the output deterministic `f"{self.fourier.prefix}_contribution"` (e.g. `fourier_mode_contribution`). Downstream code references `yearly_seasonality_contribution` by name. An optional override field lets the auto-constructed instance use the legacy name without breaking anything.

### 3. `MMM.__init__`: prepend effects into `self.mu_effects`

Replace the current block that constructs `self.yearly_fourier` then sets `self.mu_effects = []` with logic that prepends the two framework effects before any user-added ones:

```python
self.mu_effects: list[MuEffect] = []

if self.yearly_seasonality is not None:
    self.yearly_fourier = YearlyFourier(...)          # kept for backward compat
    self.mu_effects.append(FourierEffect(
[...]

PR #2757, opened 2026-07-27
feat(mmm): DataVarMuEffect, MediaMuEffect, ControlMuEffect for xarray data referencing

williambdean:
Add three MuEffect subclasses that read data directly from named variables in the training `xr.Dataset`, enabling multi-granularity media and controls without NaN from xarray coordinate unions.

```python
mmm = (
    MMM(
        date_column="date",
        channel_columns=["tv", "digital"],
        dims=("product", "geo"),
    )
    .add_mu_effect(MediaMuEffect(
        data_vars=["media_product"],
        media_transformation=MediaTransformation(
            adstock_first=True,
            dims=("product", "product-channel"),
        ),
        channel_dim="product-channel",
        prefix="product_media",
    ))
    .add_mu_effect(MediaMuEffect(
        data_vars=["media_geo"],
        media_transformation=MediaTransformation(
            adstock_first=True,
            dims=("geo", "geo-channel"),
        ),
        channel_dim="geo-channel",
        prefix="geo_media",
    ))
    .add_mu_effect(ControlMuEffect(
        data_vars=["control_national"],
        prefix="national_ctrl",
    ))
    .add_mu_effect(ControlMuEffect(
        data_vars=["control_product"],
        prefix="product_ctrl",
    ))
)

mmm.fit(X, y)  # X is an xr.Dataset with all named variables
```

Also extends the `Model` protocol with `xarray_dataset` so all MuEffects can access the training data. Also adds `xarray_dataset` to `MVITS` to satisfy the protocol.

The `DataVarMuEffect` base class is designed to be compatible with the `OptimizableMuEffect` protocol in #2621.


daimon-pymclabs:
Review of the three `MuEffect` subclasses. Core mechanism (reading model inputs directly from named vars in the training `xr.Dataset` to allow media/controls at different granularities without coordinate-union NaNs) looks sound, and the tests exercise create/set_data, prior-predictive sampling, and serialization roundtrips.

Findings, most-significant first:

**1. `pymc_marketing/mmm/additive_effect.py:405`** — `DataVarMuEffect`'s docstring claims "`create_data` and `set_data` are provided by default," but the base only implements `create_data`; `set_data` stays abstract (declared `@abstractmethod` on `MuEffect` at line 358). A user who follows the documented contract and subclasses `DataVarMuEffect` implementing only `create_effect` gets a `TypeError: Can't instantiate abstract class ... set_data` at construction. The docstring promises an API the class doesn't deliver.

**2. `additive_effect.py:528` and `:610`** — `MediaMuEffect.set_data` and `ControlMuEffect.set_data` are byte-for-byte identical. This is the same duplication the docstring meant to describe: lift a single `set_data` onto `DataVarMuEffect` and both the duplication and finding #1 disappear together. Right now the two copies drift independently.

**3. `additive_effect.py:497`** — `MediaMuEffect.create_effect` reads only `self.data_vars[0]`, but the inherited `create_data` registers every entry in `data_vars` as a `pm.Data`. If a caller passes `data_vars=["media_a", "media_b"]`, `media_b` is registered (and updated by `set_data`) yet never enters the transformation or the contribution: a silent no-op that looks wired up. Either consume all vars or validate `len(data_vars) == 1` in the constructor.

**4. `additive_effect.py:510`** — `MediaMuEffect.to_dict` serializes `"effect_dims": list(self.effect_dims)`, but `effect_dims` is a read-only `@property` derived from `media_transformation.dims`, and `from_dict` never reads the key back. The serialized field is dead weight that can silently disagree with the reconstructed value; drop it from `to_dict`. (Relatedly, the tests pass `effect_dims=...` into the constructor, which pydantic silently ignores since the config leaves `extra="ignore"`, so it's harmless but misleading.)

Nothing here is a crash-on-the-happy-path bug; findings 1 and 3 are the ones a maintainer would most want addressed before merge.

juanitorduz:
@williambdean, from my side, LGTM. I just pushed some small remarks Daimon found :)

williambdean:
Addressed all feedback from the review:

1. **`set_data` lifted to `DataVarMuEffect`** — removed the byte-for-byte duplicate from `MediaMuEffect` and `ControlMuEffect`. The base class now provides `set_data` as documented.

2. **`Field(min_length=1)` on `data_vars`** — catches empty `data_vars=[]` at construction with a clear `ValidationError`. The `data_vars[0]` access in `MediaMuEffect.create_effect` is now safe by construction.

3. **Removed stale `effect_dims` from `to_dict`** — it was a dead key; `from_dict` never read it. The property is still available at runtime, derived from `media_transformation.dims`.

Ready for re-review.

pymc_marketing/mmm/additive_effect.py
```diff
@@ -285,6 +400,223 @@ def idata_groups(self) -> dict[str, xr.Dataset]:
...
+        mmm : Model
+            The MMM model instance.
+        model : pm.Model
+            The PyMC model.
+        X : xr.Dataset
+            The new prediction dataset.
+        """
+        for var_name in self.data_vars:
+            if var_name in X.data_vars:
+                pm.set_data({var_name: X[var_name].values}, model=model)
```
PabloRoque, line 451: Should we use pm.dims ?

PabloRoque (approved):
I really like this, thanks @williambdean 

Nit: how would it look to have an intercept term here as well. I think we would benefit quite a lot to move from the monolithic `build_model` to this more composable approach.

pymc_marketing/mmm/additive_effect.py
williambdean: pm.set_data is still used just like pm.do, pm.sample, pm.Model, etc

```
>>> import pymc.dims as pmd
>>> pmd.set_data
  File "<python-input-1>", line 1, in <module>
    pmd.set_data
AttributeError: module 'pymc.dims' has no attribute 'set_data'
```

juanitorduz (approved):
Great @williambdean !

williambdean:
Thanks for the review @PabloRoque 
As for the intercept comment, maybe good to discuss here: https://github.com/pymc-labs/pymc-marketing/issues/2471

cetagostini:
Let's add this into [here](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_gam_options.html) under the *Arbitrary terms in PyMC-Marketing* section! @williambdean

Issue #3, opened 2022-03-09
Add `tanh` saturation function


juanitorduz:
Could be useful to add the 2-parameter saturation function we've got outlined in this blog post https://www.pymc-labs.io/blog-posts/reducing-customer-acquisition-costs-how-we-helped-optimizing-hellofreshs-marketing-budget/ We find this works well because the parameterisation has independent control of the initial slope and the saturation level.

PR #2765, opened 2026-07-28
Funnel Effect Example via MuEffects

juanitorduz:
Funnel Effect Example via MuEffects: Simple Case

----
📚 Documentation preview 📚: https://pymc-marketing--2765.org.readthedocs.build/en/2765/


daimon-pymclabs (commented):
Reviewed the full notebook (56 cells), the gallery wiring, and cross-checked a few API conventions against the repo at `v1.0.0`. Overall this is a strong, well-written methodological example. The `MuEffect`-with-second-likelihood pattern is used cleanly and the causal story is honest. A few things to fix before merge.

## Should fix

1. **Missing gallery thumbnail.** The card in `gallery.md` points at `../gallery/images/mmm_funnel_aware.png`, but that PNG is not in the PR and does not exist in `docs/source/gallery/images/` on the branch. Every sibling entry (e.g. `mmm_upper_funnel_causal_approach.png`) ships a committed thumbnail, so the card will render broken. Add the image.

2. **The written DGP contradicts the implemented one.** Cell 1 defines the mediator with noise, `M_t = baseline + γC_t + f_{u→l}(U_t) + ε^M_t`, and then writes the target as `Y_t = intercept + f_direct(U_t) + f_{l→y}(M_t) + ε^Y_t`, i.e. `Y` depends on the noisy `M_t`. But the code has the target consume the noise-free latent demand: `funnel_lower_demand` is a `Deterministic` (no noise), observed `lower_spend ~ TruncatedNormal(mu=demand)`, and the returned contribution is `saturation_lf(adstock_lf(demand))`. So observed spend = latent demand + ε, and `Y` responds only to the latent demand, never to ε^M. Cell 5 actually brags about this ("the target consumes the latent demand, not the observed lower spend"), which is the right choice, but it directly contradicts the equations three cells earlier. Introduce a distinct latent `D_t` (demand) vs observed `M_t` (spend), write `M_t = D_t + ε^M` and `Y_t = ... + f_{l→y}(D_t)`, and add one sentence that this encodes the assumption that the target responds to the systematic part of lower-funnel demand while the idiosyncratic execution noise in observed spend has no causal effect on `Y`. That assumption is defensible but currently unstated and mis-stated.

3. **`hdi_bounds` passes the wrong kwarg.** `az.hdi(da, prob=prob)` uses `prob`, but arviz's public parameter, and every other `az.hdi` call in this repo, is `hdi_prob`. If your arviz version silently ignores unknown kwargs, the HDI is pinned at the 0.94 default no matter what `prob` you pass, which is a latent trap the day someone calls `hdi_bounds(da, 0.5)`. Rename to `hdi_prob` and confirm it runs on the notebook's arviz. (The `ci_bound="lower"/"upper"` selection is fine, it matches the repo's own `ci_bound` dimension convention, so no change there.)

## Minor / nits

4. **R² rescale is fragile (cell 53).** Multiplying `posterior_predictive["y"]` by `df["y_obs"].max()` assumes that group is stored in scaled units. It clearly is on your stack (the R² table is sensible), but this silently couples the punchline to an internal scaling convention. Prefer the `*_original_scale` predictive variable or an explicit original-scale flag so it can't break quietly later.

5. **Mediator scale.** The mediator equation runs on raw `lower_spend` / `lower_control` while the base MMM scales its channel to `[0,1]`. Self-consistent here because the data is generated in-model, but anyone adapting this to real data needs to know the second likelihood's `sigma` and the `HalfNormal(1)` priors on `baseline`/`gamma` live on the raw scale. Worth one line in the caveats.

6. **Cosmetic.** Notebook H1 is "Funnel-Aware MMM: Modeling Upper- to Lower-Funnel Mediation" but the gallery card drops "Modeling". And `%autoreload 2` is left in a docs notebook. Both harmless.

7. **Naive B explanation is thin (cell 30/50).** It's thinner than Naive A's: you say the two naive models are "biased in opposite directions" but only give a mechanism for A (mediator conditioning blocks the indirect path). A one-line reason for B's direction of bias would round it out; the chart shows it empirically but the prose leaves it as "misestimates."

## What's good

The counterfactual construction is correct in both places that matter: the ground-truth indirect effect via `f(baseline+control+U) − f(baseline+control)` (upper set to zero and subtracted), and the recovery-side indirect via the same zero-upper counterfactual. The note that the direct path is already a counterfactual because `saturation(0)=0` is right. The identifiability caveats (cells 43, 54) are honest and correct, the direct/indirect split is genuinely weakly identified because both paths scale with the same `U`, and `lower_control` as an exclusion restriction on the mediator is sound. The "bias is invisible in predictive fit" section is the strongest pedagogical beat in the notebook. The caveats section (DAG-blindness, `BudgetOptimizer` not substituting into `MuEffect` subgraphs, effect ordering) is exactly the right set of honest limitations.

Net: fix the missing thumbnail and the DGP/notation contradiction (#1, #2) before merge; #3 is a quick, worthwhile correctness fix. The rest are polish.

juanitorduz:
Thanks for the thorough review @daimon-pymclabs. Pushed 38528e21 addressing the feedback.

**Fixed**

- **DGP notation vs implementation (should-fix #2).** Rewrote the structural equations to distinguish the latent lower-funnel demand `D_t` (deterministic) from the observed spend `M_t ~ TruncatedNormal(D_t, σ_M, lower=0)`, with the target now written as depending on `D_t` (matching the code). Added a sentence stating the assumption that `Y` responds to the systematic demand while the idiosyncratic execution noise `ε^M` in observed spend has no causal effect on `Y`. The intro and the `FunnelEffect` walkthrough are now consistent.
- **R² rescale (nit #1).** Now reads each model's own scale via `idata.constant_data["target_scale"]` instead of `df["y_obs"].max()`, so it no longer couples to an assumed scaling convention. Table is unchanged (0.965 / 0.941 / 0.968).
- **Naive B mechanism (nit #4).** Added a one-line explanation: with the mediator gone, the single upper-funnel curve must also absorb the exogenous, control-driven lower-funnel demand that upper spend never caused, which here biases it upward.
- **Mediator scale (nit #2).** Added a caveat that the mediator equation runs on the raw `lower_spend`/`lower_control` scale, so `σ_M` and the `HalfNormal(1)` priors on `baseline`/`gamma` live on that raw scale when adapting to real data.
- **Title (nit #3-cosmetic).** Dropped "Modeling" from the H1 so it matches the gallery card.

**Respectfully declining two, with evidence**

- **Missing thumbnail (should-fix #1).** `docs/source/gallery/images/` is gitignored (`.gitignore` L152-153) and RTD regenerates thumbnails in `pre_build` via `python scripts/generate_gallery.py` (`.readthedocs.yaml`). No sibling notebook ships a committed thumbnail (`git ls-files docs/source/gallery/images/` is empty), so this matches convention; the committed notebook carries image outputs, so the card renders on the docs build. The RTD preview confirms it. Committing the PNG would add a gitignored build artifact.
- **`prob` → `hdi_prob` (should-fix #3).** On this branch's arviz (1.2.0), the parameter is `prob`, not `hdi_prob`: `az.hdi(da, hdi_prob=0.94)` raises `TypeError: hdi got an unexpected keyword argument: 'hdi_prob'`, and the signature is `hdi(data, prob=None, ...)`. The repo itself uses `prob=` at `pymc_marketing/customer_choice/mv_its.py:669`. So the current `hdi_bounds(..., prob=...)` is correct and does vary with the value; renaming would crash the notebook. (`hdi_prob` was the pre-1.0 arviz name.)

Left `%autoreload` in place to match the other notebooks (e.g. `mmm_data_generator.ipynb`).

daimon-pymclabs:
## Round 2 review

I re-checked my round-1 review against the actual arviz version and build setup, and re-read all 56 cells against Juan's commit. Two of my three round-1 "should fix" items were wrong — correcting the record here.

### Retractions from round 1

**#3 (hdi_bounds `prob` kwarg) — retracted, was wrong.** I checked against the arviz 0.x API from memory. This notebook runs on arviz 1.2.0 (watermark cell), which is the major API rewrite. I installed 1.2.0 and confirmed: the signature is `az.hdi(data, prob=None, ...)`, the kwarg really is `prob`, and it returns a `ci_bound` dimension with `["lower","upper"]`. So `az.hdi(da, prob=prob)` and `.sel(ci_bound=...)` are both correct. No change needed.

**#1 (missing gallery thumbnail) — retracted, was wrong.** `docs/source/gallery/images/` is gitignored (`.gitignore:153`), and `.readthedocs.yaml:19` runs `python scripts/generate_gallery.py` on every build, which regenerates `gallery.md` from `gallery.yaml` and extracts each thumbnail from the notebook's first image cell. None of the sibling pngs (`mmm_quickstart.png`, etc.) are committed either. The png can't and shouldn't be committed, and the notebook has plenty of image cells so it won't fall back to the default logo. Non-issue.

### Fixes verified correct

- **#2 DGP/notation contradiction:** cell 1 now defines latent demand `D_t` (deterministic) separately from observed `M_t = D_t + ε^M`, with the target responding to `D_t`, plus an explicit sentence on the noise assumption. Matches the code exactly now.
- **#4 R² rescale:** cell 53 now reads each model's own `target_scale` from `constant_data` instead of `df["y_obs"].max()`. Removes the hidden coupling.
- **#5/#7 Naive B:** cell 30 now gives the upward-bias mechanism, cell 50 explains the opposite-direction biases, and cell 54 adds the "Mind the mediator scale" caveat about the raw-scale `sigma` and `HalfNormal(1)` priors.
- **#6 Title:** H1 now reads "Funnel-Aware MMM: Upper- to Lower-Funnel Mediation", matching the gallery card in both `gallery.md` and `gallery.yaml`.

### Fresh pass

Re-read all 56 cells looking for anything new or introduced by the edits. The counterfactual construction (cells 22, 45), the two-likelihood forward simulation, and the parameter-recovery / effect-decomposition logic all hold up. Nothing substantive. The only remaining nit is cosmetic and pre-existing: `%autoreload 2` is still in a docs notebook (harmless).

**Verdict: ready to merge.** The one genuinely important round-1 catch was #2, and it's fixed cleanly.

daimon-pymclabs:
Chiming in on two of @TeemuSailynoja's points, since they touch the `MuEffect` contract rather than just the prose.

**Generalizing the transform annotations** — agreed, and it's free. `apply(..., core_dim=)` and `to_dict()` are both defined on `Transformation` in `pymc_marketing/mmm/components/base.py` (lines 620 and 168), so `FunnelEffect` only ever uses base-class API. Widening `InstanceOf[GeometricAdstock] -> InstanceOf[AdstockTransformation]` and `InstanceOf[LogisticSaturation] -> InstanceOf[SaturationTransformation]` changes nothing about this notebook's run but stops the example from teaching a narrower contract than the code actually supports — worth doing precisely *because* it's a docs notebook people copy from.

**The out-of-sample point is sharper than "doesn't support it"** — `lower_spend` and `lower_control` are constructor fields (fixed-length `list[float]` captured at build time), and `set_data` writes those same stored arrays back:

```python
pm.set_data({f"{self.prefix}_lower_spend": np.asarray(self.lower_spend, ...)}, model=model)
```

So there is no channel for new-period mediator values at all. Calling `predict` on an `X` of different length gives a length-T array against a resized `date` coord, i.e. a shape conflict rather than a graceful degradation. Given the notebook does argue about predictive performance, I'd state the limitation explicitly next to `set_data` — something like *"this effect holds its mediator data fixed at the training values, so it is an in-sample decomposition example; supporting OOS prediction means threading the lower-funnel series through `X` instead of the constructor."*

On the budget-optimizer note: agreed the `MaskedPrior` route (`pymc_marketing/special_priors.py`) is worth a forward-pointing sentence so readers don't conclude MuEffect and the optimizer are fundamentally incompatible.

The two "reads like an LLM artifact" observations are fair calls and worth acting on — meta-commentary that only makes sense relative to an earlier draft state should just be cut, not explained.

juanitorduz:
Thanks for the careful read @TeemuSailynoja! Pushed 833d3833 addressing all ten comments; the notebook was re-executed end-to-end with full MCMC. Point by point:

**Prose**

- **Cell 0 (redundant note block):** Removed the "internal, self-contained example" note entirely — it repeated the numbered points above it.
- **Cell 1 (spell out the indirect path):** The `lower_control` bullet now reads "…separate the **direct** path from the **indirect** path -- the part of the upper-funnel effect on the target that is mediated by lower-funnel demand -- an **exclusion restriction** on the mediator equation." Glad the staged DGP intro works!
- **Cell 6 (repeated "structural mediation" sentence):** Cut the closing paragraph — fair call, it repeated "The funnel structure" and only made sense relative to an earlier draft.
- **Cell 44 (scaling meta-commentary):** Rewrote the paragraph to state the scaling convention plainly (amplitude coefficients are estimated on the internally standardized target, so the plot multiplies them by `target_scale`), dropping the "apparent gap / not a modeling failure" framing.
- **Cell 51 (emphasis):** Now opens with "**Both naive models are biased.**" and notes that the opposite directions are incidental; the point is that neither naive specification recovers the truth.
- **Cell 31 (prominence of the "common error / unbiased" point):** Added an `{important}` admonition in the Naive A section: treating funnel layers as independent channels is the default in production MMMs and a common error; the funnel model is not just more descriptive, it is the only one of the three that recovers unbiased upper-funnel estimates.

**Code**

- **`GeometricAdstock` / `LogisticSaturation` annotations:** Widened to `InstanceOf[AdstockTransformation]` / `InstanceOf[SaturationTransformation]` (all four fields). The effect only uses base-class API (`apply(..., core_dim=)`, `to_dict()`), so the example now teaches the contract the code actually supports.
- **Out-of-sample prediction:** Went one step further than documenting the limitation — `FunnelEffect` is now a `DataVarMuEffect` subclass, so `lower_spend` / `lower_control` are read from the `X` dataset at fit time and refreshed at prediction time instead of being frozen in the constructor. OOS prediction now works: pass future `lower_control` (exogenous) and placeholder `lower_spend` (the second likelihood is resampled — the model *forecasts* lower-funnel spend from the upper-funnel plan). One consequence: the funnel model's `X` is an `xr.Dataset` (built by a small `make_mmm_dataset` helper), since the DataFrame path only carries channel/control columns. Verified with an actual 10-week OOS predictive run in a standalone script. The notebook itself stays an in-sample decomposition study; the new **Caveats** bullet spells out the OOS semantics.
- **Budget optimizer caveat:** Now says a `BudgetOptimizer`-compatible `MuEffect` **is** possible via careful use of `MaskedPrior`, just more convoluted and out of scope here.

Results are essentially unchanged (same seed, same model graph): bias table and R² tell the same story.

TeemuSailynoja (approved):

juanitorduz:
Thank you very much @TeemuSailynoja ❤️

TeemuSailynoja:
@juanitorduz Thank you for the edits! No further requests from me, I think it's a clear intro to funnel modeling!

juanitorduz:
> @juanitorduz Thank you for the edits! No further requests from me, I think it's a clear intro to funnel modeling!

Great! I think there are so many ways we can extend this initial example ;) PR's welcome hehe

drbenvincent:
* I'd be tempted to have the DAG of the DGP right at the start of the The Funnel Structure section. In that DAG, lower_spend is confusing. If it's spend then it should drive the target. If it's purely a noisy observation of latent demand then it needs a more descriptive name. `lower_spend` doesn't seem like the right name.
* I would only show one equation for M_t, the Truncated Normal one.
* Collapse all plot-only code blocks - it makes for an easier, less interrupted flow.
* Over-liberal use of bolding words/phases. Reserve it for when it's really needed

drbenvincent:
Already merged, but @juanitorduz wanted more comments here. This is what I put together with the help of an agent, but heavy input and iteration from me

---

This review has two layers, kept deliberately separate:

1. **The notebook** — is what it shows correct, and are its claims supported by what it runs?
2. **The direction** — does `FunnelEffect`-on-`MMM` scale as the route from "everything is a
   direct effect" to a DAG-first MMM?

The short version: the notebook is good work and its core result is real, with a few claims that
overreach what it computes. The direction concern is separate and unaffected by those fixes.

---

# Part 1 — The notebook

## What it establishes

**The estimation result holds.** On the synthetic DGP, Naive A (mediator as channel) and Naive B
(mediator omitted) are both biased; the funnel model recovers direct, indirect, and total
upper-funnel effects.

**The `FunnelEffect` pattern works.** A `MuEffect` that declares its own likelihood inside
`create_effect` turns "add a term to `mu`" into "fit a second structural equation jointly". That is
a genuinely useful demonstration, and it plugs into `MMM` without touching the class.

**It is the first *joint* funnel notebook.** The repo already ships `mmm_intro_upper_funnel` and
`mmm_upper_funnel_causal_approach`, which tell the same funnel story with a **two-stage** approach
(separate outcome model and mediation model). Joint estimation in one model is the new
contribution and should be credited as such — but see Part 2 on what three overlapping funnel
notebooks with three different mechanics implies.

## Claims that overreach

**"The bias is invisible in predictive fit" is only true for Naive A.** In-sample R² is Naive A
0.9649, Naive B 0.9412, funnel 0.9675. Naive A carries the worst bias (−27.2%) at an R²
indistinguishable from the funnel model — that comparison alone makes the point, emphatically.
Naive B has ~80% more residual variance, so fit *does* partially flag it. The notebook then
concludes "held-out error alone cannot catch it" without ever computing held-out error. Narrow the
claim to Naive A, or compute the holdout.

**The funnel model is the exact generative model.** Data is produced by `MMM(...) +
FunnelEffect(...)` clamped with `do()` and forward-sampled, then refit by the same classes with the
same transform families. "Only the funnel model recovers the truth" therefore holds under *zero*
structural misspecification. That is fine for parameter recovery, but it demonstrates
self-consistency, not robustness. The untested question is what the funnel model does when its DAG
is wrong — when the observed lower spend genuinely drives the target, or when `U` and `C` are
correlated. The notebook's own punchline (fit will not tell you your model is wrong) applies to the
funnel model too, and saying so pre-empts the obvious "but the funnel model works" reply.

**There is no lower-funnel lever in this DAG.** The graph is U→Y, U→D, C→D, D→M, D→Y. Nothing
points *out of* `M`: observed lower-funnel spend is a **measurement** of latent demand, not a cause
of the target, so `do(lower_spend)` has exactly zero effect on `Y`. The consequence is that this
model cannot answer "should I move budget from upper to lower funnel?" — arguably the main business
reason to build a funnel MMM. The notebook is honest that `Y` depends on `D` rather than `M`, but
files the allocation gap under "budget optimization is out of scope", which reads as a tooling
limitation. Even a perfect optimizer would find nothing to allocate here. Worth stating plainly.

**Terminology.** `lower_control` is described as "an exclusion restriction on the mediator
equation". The restriction doing the identification work is `C`'s *absence from the outcome
equation*. Minor, but the identification story leans on that sentence.

## Things that silently give the wrong answer today

None of these are debatable and all are reproducible:

| What | What happens |
|------|--------------|
| Waterfall decomposition | The notebook sets `plot_suite = "new"`; `DecompositionPlots.waterfall` goes through `get_contributions()`, which returns fixed buckets (channels / baseline / controls / seasonality) with no `mu_effects` branch. The funnel contribution is **dropped entirely** and the decomposition no longer sums to the fitted mean. (The *legacy* suite auto-detects `*_contribution_original_scale` and would show it as one mixed bar.) |
| `incrementality` ROAS | Hardcoded to `channel_contribution`. `upper_roas(funnel)` returns a plausible-looking number that omits the mediated path, with **no warning**. The notebook catches this by hand; a user would not. |
| `MMM.save()` / `load()` | `serialize()` succeeds, `deserialize()` raises `ValidationError` ("Input should be an instance of `AdstockTransformation`"): `to_dict` is supplied with no matching `from_dict` (contrast `MediaMuEffect`). The registered type key is `__main__.FunnelEffect`, so it would not resolve in a fresh process anyway. |
| Two decomposition paths disagree | `compute_counterfactual_contributions_dataset` **does** enumerate `mu_effects`; `get_contributions` does not. The same library answers "what are the contributions?" two different ways depending on which door you use. |

## Smaller gaps

- **Data plumbing.** `to_mmm_dataset(DataFrame)` reads only channel/control columns, so the funnel
  model needs an `xr.Dataset`. `make_mmm_dataset` is notebook-local; there is no library helper.
- **Scaling.** Channels are scaled to [0, 1]; the mediator equation runs on raw `lower_spend` /
  `lower_control`, so its `HalfNormal(1)` priors live on a different scale. Self-consistent here
  because the data is generated in-model; a trap on real data. The notebook flags this.
- **Library component.** `FunnelEffect` is notebook-only, untested.

---

# Part 2 — The direction

## `MMM` has no causal DAG in its representation

`MMM` holds channels, controls, transforms, and an ordered list of `mu_effects`. What gets compiled
is:

```text
μ = intercept + Σ channels + Σ controls + seasonality + Σ mu_effects
→ one primary target likelihood on y
```

The optional `dag` argument does **not** define structural equations. It is consumed only in
`__init__`, where it computes a backdoor adjustment set and intersects it with `control_columns`
(and silently nulls `yearly_seasonality` if seasonality is not in that set). It requires
`outcome_node` as well — passing `dag` alone does nothing at all. `BuildModelFromDAG` exists in
`causal.py` but builds linear `slope * parent` edges with no adstock or saturation, and is not used
by `MMM`.

So the PyMC graph contains `funnel_lower_demand` and a second likelihood, while to the `MMM` object
`funnel_effect_contribution` is just another summand. There is no first-class notion of treatment,
mediator, outcome, or valid intervention.

## The wiring makes this concrete

This is not only an architectural framing — it shows up in the code the notebook has to write:

- `create_effect` reads its treatment as `mmm.channel_data_scaled.isel(channel=0)`, a **hardcoded
  positional index**. Add a channel or reorder `channel_columns` and it silently transforms the
  wrong one.
- `channel_data_scaled` is a plain Python attribute carrying `## TODO: Find a better way to save it
  or access it in the pytensor graph.` in `mmm.py`. The only wiring point for a treatment is an
  acknowledged hack.
- `DataVarMuEffect.create_data` registers `pmd.Data(var_name, ...)` **unprefixed**, so two mediator
  effects sharing a control name collide with `ValueError: Variable name lower_control already
  exists.`

An effect cannot ask the model "which variable is my treatment?" because the model does not know.
It can only reach in by position and by global string name.

## Arbitrary causal queries will not generalize

Interventions are possible in PyMC (`pm.do`, graph surgery). They are not generic in `MMM` because
the library cannot represent what is being intervened on.

| Query | Productized `FunnelEffect` | Richer funnel graphs |
|-------|--------------------------|----------------------|
| Total / direct / indirect effects | Could be automated | Needs new components or a graph layer |
| Total ROAS (incl. induced spend) | Feasible with optimizer work | Custom counterfactuals per specification |
| Path-specific effects (one mediator only) | Needs graph semantics | Bespoke analysis |
| Valid `do()` vs input substitution | Needs documented assumptions | Analyst's responsibility |

The notebook's indirect counterfactual (set `upper_spend = 0`, subtract contributions) is correct
for this exogenous DGP — I checked that the scalers stay pinned, so the substitution is sound. It
is not a general intervention operator: under confounded plans or richer graphs, substitution and
`do()` diverge, and the framework cannot warn about it because it does not know the graph.

Identification is a separate axis again. Even with a causal query API, direct vs indirect paths can
be weakly identified (four independent media transforms), confounded plans add sensitivity (#2769),
and fit metrics will not catch misspecification.

## The open question

Productizing `FunnelEffect` — direct/indirect/total estimands, funnel ROAS, eventually funnel-aware
optimization — is a sensible next step and scales to **fixed funnel templates**.

The question to settle early is whether that is the endpoint, or whether funnel models need a
structural representation layer that supports arbitrary path queries. The evidence that it is not
the endpoint: the repo now has three funnel notebooks, three different mechanics, no shared
abstraction, and a decomposition API that disagrees with itself about whether `mu_effects` exist.
That destination is not reachable by accumulating more one-off effect subclasses.

I am not proposing an alternative here — that is a separate piece of work. The ask is only that we
agree on what the current approach can and cannot do before we build more on top of it.

## Suggested follow-ups

1. Decide whether `FunnelEffect`-on-`MMM` is the long-term architecture or a bridge.
2. Fix the silent-wrong-answer cases: waterfall dropping `mu_effects`, direct-only `incrementality`
   ROAS reported without warning, broken `save`/`load` round-trip.
3. Narrow the notebook's predictive-fit claim, and state that lower-funnel spend is a measurement,
   not a lever, in this DAG.
4. Promote `FunnelEffect` to a tested library component with consistent scaling and a working
   `from_dict`.

juanitorduz:
Thanks @drbenvincent ! Can you please create an issue (just copy-paste), and I will take care of this valuable feedback!

PR #2770, opened 2026-07-28
docs(bass): update notebook to the BassModel workflow

### LINKED ISSUE #1706: Dual y-axes for the "Innovators vs Imitators" plots in bass notebook

### LINKED ISSUE #2590: Update Bass notebook to use BassModel class
Update `docs/source/notebooks/bass/bass_example.ipynb` to showcase the new `BassModel` class.

### Motivation

The current notebook uses `create_bass_model` directly. Since the refactor (#2584) added the `BassModel(ModelBuilder)` class as the recommended entry point, the notebook should demonstrate the class-based workflow.

### Sections to update

1. **Import** — Use `from pymc_marketing.bass import BassModel` as primary entry point
2. **Model creation** — `BassModel(model_config=...)` instead of separate priors dict
3. **Fitting** — `model.fit(data=...)` with data in various formats (array, DataFrame, xr.Dataset)
4. **Posterior analysis** — Use `model.idata` directly with ArviZ
5. **Posterior predictive** — `model.sample_posterior_predictive(X=future)` for forecasting
6. **Save/Load** — `model.save()` / `BassModel.load()` round-trip
7. **Plotting** — `model.plot_*` methods (once #<plotting-issue-number> is complete)
8. **MLflow** — `pymc_marketing.mlflow.autolog(log_bass=True)` (once #2588 is complete)

### Notes

- Keep the existing simulation + multi-product structure — it's a good demo
- Replace `pm.sample` calls with `model.fit` where possible
- The `create_bass_model` function should still be mentioned as a lower-level alternative

anevolbap:
Closes #2590
Closes #1706

Reopens #2630, which GitHub auto-closed when the `v1.0.0` branch was deleted and merged into `main`. Same work, retargeted to `main`. The plotting methods (#2628) are already in `main`, so this no longer carries those commits, just the notebook and two source fixes.

Rewrites `bass_example.ipynb` around the `BassModel` class: `fit` with `model_config` / `sampler_config`, the `plot_*` methods, out-of-sample forecasting, a save/load round-trip, and MLflow autologging. `create_bass_model` stays in the simulation section as the lower-level API.

Two source fixes with regression tests:

- `BassModel.load` returned plain dicts instead of `Prior` objects; fixed with `parse_model_config` in `__init__` plus a serialization round-trip for `Scaled` priors.
- Multi-product forecasting without observed data failed on a 1-D `y_obs` placeholder.

The notebook was re-executed on the v1 stack (pymc 6, pytensor 3, arviz 1.2, arviz-plots 1.2) with real sampling.

----
📚 Documentation preview 📚: https://pymc-marketing--2770.org.readthedocs.build/en/2770/


daimon-pymclabs:
Really nice PR, @anevolbap — thanks for taking this on. I set up a real v1 environment (pymc 6.0.1 / pytensor 3.0.5 / arviz 1.1.0 / pymc-extras 0.12.2.dev1) with your branch installed editable, and `pytest tests/bass/test_model.py` gives **40 passed**, with the 3 new tests passing in isolation too.

**Both source fixes are correct**, and I verified the save/load one end-to-end. Before it, `model_config` came back as plain dicts; with your change `BassModel.load` returns real `Prior` objects, `loaded.model_config == model.model_config` is `True`, and `loaded.id == model.id` is `True` (which matters, since `load_from_idata` compares ids). The `parse_model_config`-in-`__init__` placement matches the established convention (`clv/models/basic.py:52`, `mmm/mmm.py:547`, all of `customer_choice/`), and `model_builder.py:532` already wraps `cls(**init_kwargs)` in `simplefilter("ignore", DeprecationWarning)` for exactly this pattern, so the "automatically converted to Prior" warning doesn't leak on load — I measured zero warnings. The `y_obs` shape fix is right and multi-product forecasting works.

Below is everything I noticed, roughly in priority order. Most of it is polish; only the first two are things I'd want in before this becomes public docs. Feel free to push back on any of it.

## Notebook

**The `p` prior is unsatisfiable, and the notebook text contradicts it.** `Prior("Beta", mu=0.03).constrain(lower=0.01, upper=0.03)` puts the mean exactly on the upper bound, so the constraint can't be met. Measured achieved mass in [0.01, 0.03] is **0.528**, with **47.2% of the prior above 0.03**. preliz says so out loud and that warning is in cell 10's stored output: `The requested mass is 0.95, but the computed one is 0.528`. Cell 12 then tells the reader "we have chosen the priors within the usual ranges of empirical studies: p typically 0.01–0.03". An interior mean (`mu=0.02`) fixes both the prior and the warning. `q` is fine — it achieves exactly 0.95.

**Cell 10's stored stderr leaks a local path into the published docs:** `/home/pablo/Documents/repos-mios/pymc-marketing/.claude/worktrees/bass-v1-rebase/.venv/...`. Because `nb_execution_mode = "auto"` (`docs/source/conf.py:122`) skips notebooks that already have outputs, this ships verbatim to readthedocs. Re-running the cell after the prior fix should clear it.

**Cell 34 builds `ref_val` and never uses it** — `azp.plot_dist` is called without it, so the figure titled "Estimated Market Cap (m) vs Observed Cumulative at T=51" draws no observed reference. Worth passing it (or dropping the variable).

**Off-by-one in cell 47:** `T_extended = np.arange(int(T.max()) + 26)` with `T = arange(52)` gives indices 0..76, which is **25** points past the last observation rather than the 26 the title and heading claim. `np.arange(len(T) + 26)` does it.

**A recovery-check opportunity in cell 30.** The text says the posteriors are "centered around the true values (from the generative model)", but the dashed lines in cells 28/29 are prior central tendencies rather than true values — the data came from one specific prior draw, and `sample_prior_bass_data` discards that draw's `p`/`q`/`m` by returning only `y`. If you keep the full prior idata and plot the actual per-product true values, this becomes a genuine parameter-recovery check, which feels like the real payoff of a simulated-data notebook. Same applies to `m` in cell 34. (Tiny sub-note: cell 28 uses `parameters["mu"]` and cell 29 uses `preliz.mean()`; for `p` those are numerically identical, so purely style.)

**Cell 49 could show off the fix.** It's the one place that demos the bug this PR fixes, but it only runs `az.summary` on the restored idata. Printing `loaded_model.model_config["m"]` would render `Scaled(Prior("Gamma", ...), factor=50000)` and make the fix visible as living documentation.

Smaller notebook nits, take or leave: cell 49 writes `bass_model.nc` into the docs source dir and cell 51 drops `mlruns/` plus the mlflow sqlite store there, neither cleaned up (harmless while notebooks aren't re-executed, but `tempfile` is cheap insurance); `%load_ext autoreload` / `%autoreload 2` in cell 2 is dev cruft for a docs notebook; cell 3 defines `seed`/`rng` while `setup_simulation_parameters` re-derives the same seed and its own local `rng` internally; and `plot_peak` takes `col_wrap`/`figure_kwargs` (it forwards to `arviz_plots.plot_dist`) while its sibling methods take `subplot_kwargs` (matplotlib) — correct as written, but a one-line note would save readers a trip to the source.

## Source

**Nested-prior serialization won't compose.** `_serialize_scaled` writes the inner dist with a raw `obj.dist.to_dict()`, which carries no `__type__` (I checked: `serialization.serialize(Scaled(Prior(...)))["dist"]` has no `__type__` key), and `_deserialize_scaled` reads it back with `pymc_extras.deserialize`. That's fine for a plain `Prior`. But the `SpecialPrior` classes in the same module are registered with `@serialization.register`, which wraps their `to_dict` to inject `__type__` — so a `Scaled(LogNormalPrior(...))` would emit a `__type__`-tagged inner dict and then hand it to `pymc_extras.deserialize`, which doesn't dispatch on `__type__`. Routing the inner dist through `serialization.serialize` / `serialization.deserialize` would make nesting compose, plus a test with a non-plain-`Prior` inner dist.

**`factor` is assumed to be a JSON scalar.** `Scaled.__init__` types it `XTensorLike` and documents it as "broadcastable to the dimensions of the distribution", so per-dim factors are a supported idea upstream. `json.dumps` on an ndarray factor raises, and through the real save path `_json_default` falls through to `str(obj)` — so it would silently round-trip as a string. Low priority since it's currently unreachable via `BassModel` (fitting with `factor=np.array([1.,2.,3.])` fails earlier with `TypeError: TensorVariable cannot be converted to Python integer`), but an explicit `np.asarray(...).tolist()` + restore, or a scalar check, would stop it being silent.

**`new_shape` hardcodes T as axis 0.** `new_shape = (len(new_t), *old_value.shape[1:])` holds for everything `to_bass_dataset` constructs, but `_from_xarray` returns a user `Dataset` as-is, so dims order is user-controlled. Deriving it from the declared dims is the same effort and self-documents:

```python
dims = self.model.named_vars_to_dims["y_obs"]
new_shape = tuple(len(new_t) if d == "T" else s for d, s in zip(dims, old_value.shape))
```

Not a regression from this PR — a transposed `(product, T)` Dataset already dies at `fit` with "All initialization points failed", so the real fix is probably a `transpose` in `_from_xarray`. The adjacent `pm.set_data(..., coords={"T": new_t})` carries the same implicit assumption.

**Organizational thought (no action needed to merge):** the `Scaled` registration sits in `special_priors.py`, whose module docstring scopes it to priors defined there that don't inherit from `Prior`. `Scaled` is a pymc-extras class and this is a serialization shim, so `serialization.py` might read as the better home — or ideally `to_dict`/`from_dict` upstream on `Scaled`. It works today because `pymc_marketing/__init__.py:29` imports `special_priors` unconditionally.

## Tests

`test_save_load_round_trip` asserts only `isinstance(..., Prior)`, which would still pass if parameters got dropped in the round-trip. `assert loaded.model_config == model.model_config` is strictly stronger, and I confirmed it holds on your branch.

Separately, the `Scaled` round-trip is currently only covered end-to-end through a model fit (~10.5s). A direct `serialization.serialize` → `deserialize` unit test in `tests/test_special_priors.py` would run in milliseconds and pin the exact contract being fixed.

---

Nothing here blocks merge on the source side — the fixes are solid and well-targeted. The `p` prior and the leaked local path are the two I'd like sorted before this lands as public documentation; the rest is at your discretion. Happy to help with any of it.

juanitorduz:
thanks @anevolbap 🚀 . I let our internal PyMC Labs agent review the PR ;) Overall, it looks great, just some minor comments. Let me know what you think about them and if you need some support :)

daimon-pymclabs:
Round 2 — Pablo, this is looking great. I re-ran everything on a real v1 environment (pymc 6.0.1 / pytensor 3.0.7 / arviz 1.2.0, PR head `f966803`): `pytest tests/bass/test_model.py` → **41 passed**, and CI is fully green now (both notebook jobs, tests on 3.12 and 3.14, readthedocs, codecov patch). You addressed essentially every point from the last round, and a couple of them you did more thoroughly than I asked.

Things I verified individually:

- **`p` prior**: now `mu=0.02`, and I re-measured the achieved mass — exactly **0.9500** in [0.01, 0.03] with **3.9%** above 0.03, versus 0.528 / 47.2% before. The preliz warning is gone from the stored outputs.
- **Leaked local path**: gone — no `/home/pablo`, `repos-mios`, or `.claude/worktrees` anywhere in the notebook outputs.
- **Off-by-one**: fixed with `np.arange(len(T) + 26)`.
- **Parameter recovery**: this was the big one and you did it properly rather than cosmetically. `sample_prior_bass_data` returning the full prior draw, and cells 28/29/34 plotting the actual per-product `prior_draw["p"/"q"/"m"]` as reference lines, turns the notebook into a genuine recovery check. Cell 30's "true per-product values" wording, the dead `ref_val` becoming real reference lines with a "vs True Value" title, and cell 49 printing `loaded_model.model_config["m"]` so the fix is visible as living documentation — all exactly right.
- **Housekeeping**: `tempfile` for both the `.nc` and the mlflow store, autoreload magics dropped, `plot_peak` kwargs note at cell 42.
- **Source**: forecast shape derived from `named_vars_to_dims["y_obs"]` with `strict=True`, `_from_xarray` transposing so T leads, and `assert loaded.model_config == model.model_config` in the round-trip test. I confirmed `y_obs` always has declared dims (`('T',)` even for a 1-D array-built model), so the dims lookup is safe.

**A correction on my side first.** I previously claimed `Scaled(LogNormalPrior(...))` would break serialization, on the theory that `_serialize_scaled` writes the inner dist with a raw `to_dict()` carrying no `__type__`. I actually tested it this time: the inner dict *does* get `__type__` and the round-trip returns a real `LogNormalPrior`. That concern was simply wrong — the code is fine as written, sorry for sending you after a non-issue.

Two small things left, and then I think we're ready to merge:

**1. The old `p` prior is still in the test file.** `tests/bass/test_model.py:828` still has `Prior("Beta", mu=0.03).constrain(lower=0.01, upper=0.03)` — the exact thing you fixed in the notebook. It emits the same warning on every run of `test_save_load_round_trip_scaled_priors`:

```
UserWarning: The requested mass is 0.95, but the computed one is 0.528
```

That accounts for 3 of the 4 warnings in the suite. Same one-line fix: `mu=0.02`.

**2. A small hole in the new transpose support.** A `(product, T)` Dataset that *doesn't* carry a `T` coord still fails:

```
('product', 'T') -> ValueError: conflicting sizes for dimension 'T':
                    length 3 on 'T' and length 20 on {'product': 'observed', 'T': 'observed'}
('T', 'product') -> OK, dims=('T','product') Tlen=20
```

`_from_xarray` assigns the `T` coord with `n = next(iter(data.sizes.values()))` *before* the transpose, so it picks up the product size. `test_transposed_dataset_dims` passes only because `multi_product_ds` already has a `T` coord. Moving the transpose above the coord assignment (or deriving `n` from the observed variable's `T` position) closes it. It fails loudly rather than silently, so entirely your call whether to fix here or leave as a follow-up issue.

Nice work on this one — the notebook is a much better piece of documentation than when it started.

anevolbap:
Thanks for the thorough re-run, and no worries on the `Scaled` note, good to have it confirmed either way.

Both fixed and pushed (`eb360088`):

- `test_save_load_round_trip_scaled_priors`: `p` prior to `mu=0.02`, so the preliz mass warning is gone (the suite is down to just the arviz migration warning).
- `_from_xarray` now transposes before adding the `T` coord, so a `(product, T)` Dataset with no `T` coord takes its length from the `T` dim. Added a regression test for that case, the earlier one only covered the with-coord path, as you spotted.

juanitorduz (approved):
Thanks @anevolbap !

Issue #2726, opened 2026-07-09
Remove the Bass DiracDelta pymc.dims fallback once a pymc release ships dims.DiracDelta


anevolbap:
`create_bass_model` migrated to `pymc.dims` in #2724. `pymc.dims` did not support `DiracDelta` at the time, so `_create_dim_variable` (in `pymc_marketing/bass/model.py`) has a narrow fallback: for `DiracDelta` priors only, it creates a regular variable and wraps it with `pmd.as_xtensor` instead of `create_variable(xdist=True)`.

`DiracDelta` is now in `pymc.dims` upstream (pymc-devs/pymc#8350), shipping in pymc 6.2.0 (the PR merged two days after the 6.1.0 release). Once pymc-marketing requires a pymc release that includes it, remove the fallback so `_create_dim_variable` is just `create_variable(name, xdist=True)`, and re-run the bass tests and the `DiracDelta` docstring example.

anevolbap:
For whoever picks this up: `_create_dim_variable` is not on main yet, it arrives with #2771. The pymc floor also has to move to >=6.2.0 first (#2803), since `dims.DiracDelta` missed the 6.1.0 cut by two days. #2803 is in turn waiting on a pymc release that includes pymc-devs/pymc#8388.

anevolbap:
#2771 merged without the fallback. The lean rewrite dropped `_create_dim_variable` along with the `hasattr(pmd, ...)` branch, so `m`, `p` and `q` go straight through `create_variable(xdist=True)` and `DiracDelta` resolves natively on the `pymc>=6.2` floor. On `main` the name now appears only in the `create_bass_model` docstring example (`pymc_marketing/bass/model.py:117`).

Nothing left to remove here. Closing as superseded.

PR #731, opened 2024-06-10
Fix some ParetoNBDModel docstring typos

Mews:
## Description

Fixed some typos in the docstring for the `ParetoNBDModel` class

## Related Issue

- [x] Related to #727 

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [x] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Modules affected

- [ ] MMM
- [x] CLV

## Type of change

- [ ] New feature / enhancement
- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--731.org.readthedocs.build/en/731/

PR #2771, opened 2026-07-28
refactor(bass): migrate dim handling to pymc.dims

### LINKED ISSUE #2598: Migrate Bass Model dimension handling from create_dim_handler to pymc.dims

anevolbap:
Closes #2598

Migrates the Bass model's dimension handling from `create_dim_handler` (`pymc_extras.prior`) to `pymc.dims`, for consistency with the rest of the codebase.

Reopens #2724, which GitHub auto-closed when the `v1.0.0` branch was deleted and merged into `main`. Same work, retargeted to `main`.

In `create_bass_model`:

- `m`, `p`, `q` and the likelihood via `create_variable(xdist=True)`; deterministics via `pmd.Deterministic`.
- The public `F` and `f` take xtensor inputs or scalars, so there is no second copy of the formulas for the model graph.
- `t` and `y_obs` stay `pm.Data` wrapped with `pmd.as_xtensor`, so the data-setter and out-of-sample forecasting keep working.
- An `observed` with no labels (a plain array, or a `pm.Data` registered without dims) is labelled positionally in `(T, ...)` order, with the extra dims taken from the `likelihood` prior first, then `p`, `q` and `m`. An array laid out the way the likelihood declares it is read that way.
- Deterministics carry the dims their parameters have, transposed to `(T, ...)`. They are not broadcast up to the likelihood's dims: pooled parameters give pooled curves, and a per-product curve is the user's to build outside the model.
- This is a user-visible change: on `main` the deterministics carried the union of the `p`, `q` and `m` dims.
- `pm.set_data` on a tensor passed as a likelihood parameter now reaches `y` when there is no observed data. On `main` it did not, because `pymc_extras.Prior.create_likelihood_variable` deepcopies the parameters. The observed path still goes through upstream and still detaches.
- Without data (prior predictive), the outcome variable is built with `create_variable` and `mu` set, keeping the `UnsupportedDistributionError` and `MuAlreadyExistsError` guards, since upstream refuses `observed=None` (pymc-devs/pymc-extras#731).
- Priors must use distributions `pymc.dims` implements. A custom `VariableFactory` must take the `xdist` keyword, which the `pymc_extras` 0.14 protocol already declares, though the example in its docstring does not implement it.

`DiracDelta` is native in `pymc.dims` on the pymc>=6.2 floor, so #2726 has nothing left to remove and can be closed.

Verified: full bass test suite, mlflow autolog, `adopters`/`peak` matching the closed-form formula to machine precision, and the bass notebook mock runner.

----
📚 Documentation preview 📚: https://pymc-marketing--2771.org.readthedocs.build/en/2771/


williambdean:
Updating the base branch. Should work now with pymc version

williambdean:
Looking at the CI, the oldest-deps shard installs pymc==6.0.0 which lacks `pmd.Poisson` and `pmd.NegativeBinomial`. Bump the floor to `6.0.1` in pyproject.toml and it should be green.

anevolbap:
Done, bumped to `pymc>=6.0.1,<6.1.0` in `pyproject.toml` and re-locked (`01f0ee87`). The resolved version was already 6.0.1, so only the recorded specifier changed and `uv lock --check` passes. The oldest-deps shard should pick up `pmd.Poisson` now.

pymc_marketing/bass/model.py
```diff
@@ -236,9 +248,22 @@ def f(
...
+
+    Uses ``xdist=True`` so the variable lives in xtensor space from the
+    start. ``pymc.dims`` does not yet support ``DiracDelta``; for that case
+    only, create a regular variable and wrap it with ``pmd.as_xtensor``.
+    This wrap is applied to data, not to a supported prior, so it does not
+    reintroduce the two-step conversion pattern for the general case.
+    """
+    if getattr(prior, "distribution", None) == "DiracDelta":
+        return pmd.as_xtensor(prior.create_variable(name), dims=prior.dims or ())
+    return prior.create_variable(name, xdist=True)
```
williambdean: This is with 6.0.1, right? Let's just bump version like with the other PR

williambdean: Or isn't this going to not be needed anymore then with the bump?

anevolbap: `pmd.DiracDelta` isn't in 6.0.1, only `Poisson`/`NegativeBinomial` landed there; it ships in 6.1.0 (pymc-devs/pymc#8350). So the 6.0.1 bump does not remove this, the fallback stays until the floor can move to `>=6.1.0`, a bigger jump than the 6.0.1 one. Left as the follow-up (#2726), but happy to do the 6.1.0 bump here if you'd rather.

anevolbap:
Done, p and q now use `plot_dist` faceted per product, so each posterior sits next to its own true value (like the m plot). Re-ran.

daimon-pymclabs (commented):
Thanks for taking this on, @anevolbap. This is a solid migration and the hard parts are already right: moving `combined_dims` off set iteration order is a real bug fix that stands on its own, the `core_op` unwrap in the mlflow logger does exactly what it claims (`pmd.Poisson` -> `"Poisson"`, plain `pm.Normal` unaffected), and the rewritten `f()` is algebraically identical to the previous `pt.square` version. I ran the branch against `main` to compare behavior, and `tests/bass/test_model.py` passes for me apart from four `netCDF4`/`h5netcdf` import failures that are environmental and unrelated to this change.

There are two things I think should change before merge, plus a couple of smaller ones.

### 1. The likelihood has no `pymc.dims` fallback (`model.py:296-306`)

`_supports_xdist` gives `m`, `p` and `q` an escape hatch, but `_create_likelihood_variable` is called with `xdist=True` unconditionally, so the likelihood distribution gets none. `pymc.dims` in 6.0.1 exposes 25 distributions, and it has no `Wald`, `Binomial`, `ZeroInflated*`, `Hurdle*`, `BetaBinomial` or `Mixture`. So a likelihood that worked on `main` now fails.

This is worth fixing rather than documenting, because it contradicts the intent stated in your own docstring on `TestBassModelXdistFallbacks` ("must not narrow what `create_bass_model` accepts"). It also slips past the current tests: `test_prior_missing_from_pymc_dims` puts `Wald` on `m` and keeps `Poisson` as the likelihood, so it exercises the parameter path but not this one. Routing the likelihood through `_supports_xdist` as well, falling back to the non-xdist `create_likelihood_variable` with `pmd.as_xtensor`, should be enough, and a test with a non-`pymc.dims` likelihood would lock it in.

### 2. `_align_to_dims` raises `KeyError` on a dim the model has but `coords` omits (`model.py:217`)

```python
missing = {dim: len(coords[dim]) for dim in ...}
```

This reads the `coords` argument rather than the model. Pass a model that already carries a dim together with a `coords` dict that omits it and you get `KeyError: 'product'`. `main` was immune because `pm.Deterministic(dims=...)` resolves names against the model. `BassModel.build_model` always passes full coords, so nothing in-repo trips it, but `create_bass_model` is public API and someone will. `model.dim_lengths[dim]` fixes it and lets you drop the `coords` parameter from the helper entirely, which is a nice simplification.

### 3. `peak` is the one deterministic that skips `_align_to_dims` (`model.py:281`)

Its dim order therefore comes from xtensor broadcast rules instead. With `m`, `p`, `q` declaring `('product','country')`, `('product',)`, `('country',)` I get `peak` -> `('country','product')` while `adopters` -> `('T','product','country')`. This is not a regression (`main`'s set iteration order was worse), but it's the same class of inconsistency the `declared_dims` change was written to eliminate, and nothing currently tests `peak`'s dim order. Transposing to the subset of `combined_dims` present in `peak.dims` would close the loop.

### 4. `observed`'s own `.dims` are never consulted (`model.py:293-296`)

The lookup goes only through `model.named_vars_to_dims[observed.name]`, so an `xr.DataArray` that carries `.dims` directly is ignored. A DataArray laid out `('T','country','product')` against priors declaring `('product','country')` now raises `ValueError: xr.DataArray dims ... don't match requested specified ...`. Raising is arguably better than what `main` did (it transposed silently), so I'm fine either way, but a `getattr(observed, "dims", None)` check before the `named_vars_to_dims` lookup would just handle it. If you keep the hard error, please call it out in the PR description since it's user-visible.

### Smaller notes

- `_create_likelihood_variable` forks pymc_extras' `mu` guard and `deepcopy` to work around `observed=None` on the xdist path. Worth filing upstream and linking the issue in the comment so the fork has a clear deletion trigger.
- The docstring says `DiracDelta` lands "before pymc 6.2" while the PR body says 6.1.0 via pymc-devs/pymc#8350. Pick one.
- `assert not hasattr(pmd, "Wald")` will fail the day pymc adds `Wald` to `pymc.dims`, which is a dependency-upgrade signal rather than a regression. `monkeypatch.delattr` expresses the intent better.
- On the floor: keep the `pymc>=6.0.1` bump here. `uv.lock` already resolves pymc to 6.0.1, so it's a one-line `requires-dist` edit with no re-lock needed.

Nice work overall. Items 1 and 2 are the only ones I'd consider blocking, and both are small. Happy to look again as soon as you push.

anevolbap:
All four reproduce; pushed as 21a8ff61.

1. Fixed. The likelihood goes through `_supports_xdist` as well and falls back to the regular pymc path on the underlying tensors, in `combined_dims` order. Two tests cover it, with and without data. One correction on the list: `Binomial`, `BetaBinomial` and `Mixture` have no `mu` parameter, so `create_likelihood_variable` rejects them on `main` too (`Likelihood distribution 'Binomial' is not supported.`). The regression is `Wald`, `ZeroInflatedPoisson` and `HurdlePoisson`, which build on `main` and build again now.
2. Fixed with `model.dim_lengths[dim]`, and the `coords` parameter is gone from the helper. The trigger is narrower than a model dim missing from `coords`: a dim already on `m`, `p` or `q` is on the expression, so nothing is looked up. It takes a likelihood-only dim, which is what the test uses.
3. Fixed. `peak` is transposed to the subset of `combined_dims` it carries, so the example in the review gives `('product','country')` against `adopters` `('T','product','country')`.
4. Handled rather than documented. `observed.dims` is consulted before the `named_vars_to_dims` lookup, so a DataArray in any layout works and there is no user-visible error left to call out.

On the smaller notes: the `observed=None` xdist gap is filed as pymc-devs/pymc-extras#731 and linked in the comment, `monkeypatch.delattr` is applied, and the `pymc>=6.0.1` bump stays here.

On the version mismatch, the docstring is the correct one. `DiracDelta` is absent from `pymc/dims/distributions/scalar.py` at `v6.1.0` and present at `v6.2.0`, and 6.1.0 was released on 2026-07-07, before pymc-devs/pymc#8350 merged. Fixed in the PR body.

daimon-pymclabs:
Checked 21a8ff61 and 4d3c971. All four are closed as described.

The likelihood now goes through `_supports_xdist` and falls back to `create_likelihood_variable` on the underlying tensors in `dims` order, with `test_likelihood_missing_from_pymc_dims` and the `observed=None` twin covering both branches. `_align_to_dims` reads `model.dim_lengths` and the `coords` parameter is gone. `peak` is transposed to the subset of `combined_dims` it carries. `observed`'s own `.dims` are consulted first, with the `named_vars_to_dims` lookup as the fallback, so a DataArray in any layout works and there is no user-visible error left to document.

Your correction on item 1 is right and mine was sloppy: `Binomial`, `BetaBinomial` and `Mixture` have no `mu`, so `create_likelihood_variable` rejects them on `main` too. The regression was `Wald`, `ZeroInflatedPoisson` and `HurdlePoisson`. Same conclusion, narrower blast radius than I claimed.

On `DiracDelta`, you are right and the docstring is the correct one. I was reading the PR body's 6.1.0 as the shipped version rather than the merge target.

Nothing further from me. Items 1 and 2 were the only ones I would have blocked on and both are fixed.

daimon-pymclabs (commented):
Read through the diff. The migration looks careful and the fallback paths are well covered by the new tests. A few non-blocking notes:

1. **`peak` dims are a user-visible change.** Previously `peak_dims` was the union of the `p`, `q` and `m` dims, while `peak` itself was built from `p` and `q` after they had been broadcast to `combined_dims`. Now `peak` carries only the dims that `p` and `q` actually have, ordered like the other deterministics (covered by `test_peak_dims_follow_combined_dims_order`). That reads like a fix rather than a regression, but anyone indexing `idata.posterior["peak"]` by an `m`-only dim will see a shape change, so it is probably worth a line in the release notes.

2. **`observed` dim inference is duck-typed.** In `create_bass_model`:

   ```python
   observed_dims = getattr(observed, "dims", None) or (
       model.named_vars_to_dims.get(getattr(observed, "name", None), combined_dims)
   )
   ```

   This covers `xr.DataArray`, a registered `pm.Data`, and a bare array. The silent branch is a `pm.Data` registered *without* dims: `named_vars_to_dims` has no entry, so it falls back to `combined_dims` and the data is labelled positionally. That is the same behaviour as before this PR, so not a regression, but since `test_observed_is_labelled_with_its_own_dims` exists precisely because positional labelling can transpose data without saying so, a `pm.Data`-without-dims case might be worth an explicit test or a note in the docstring.

3. **`_supports_xdist` reads the signature.** `"xdist" not in signature(prior.create_variable).parameters` returns `False` for any factory that accepts `**kwargs` and dispatches internally, which sends it down the `as_xtensor` wrapping path. Harmless, just slightly conservative. Worth a sentence in the docstring so a future factory author knows why their `xdist` support is not being picked up.

4. The `f()` rewrite from `pt.square(...)` to `(...) ** 2` with the shared `exp_t` is algebraically identical to the previous expression and saves a duplicate `exp`. No numerical concern from my side; both forms overflow at the same place for large `t * (p + q)`.

Nothing here blocks. Points 1 and 2 are the ones I would want a second pair of eyes on.

anevolbap:
Addressed in 1bc024c7:

1. Added a bullet to the PR description flagging the `peak` shape change (dims narrowed to what `p` and `q` declare, `m`-only indexing stops working).
2. Added `test_observed_data_without_dims_is_labelled_positionally`, checking that a `pm.Data` without dims gets the same labelling and logp as one registered with explicit dims, plus a note in the `observed` docstring about the positional fallback.
3. Added a sentence to the `_supports_xdist` docstring: a factory that hides `xdist` behind `**kwargs` takes the wrapping path, declaring the parameter explicitly takes the native one.

9f7af31a merges `main` to resolve the `pyproject.toml`/`uv.lock` conflict from #2883; the dependency files now match `main` and dropped out of the diff.

daimon-pymclabs:
Checked 1bc024c and 9f7af31. All three notes are closed.

The PR description now flags the `peak` dim narrowing and the loss of `m`-only indexing, which was the only item anyone downstream needed warning about. `test_observed_data_without_dims_is_labelled_positionally` pins the fallback the way I wanted it pinned, comparing labelling and logp against the explicitly-dimmed `pm.Data`, and the `observed` docstring now states where labels are read from and what the positional `(T, ...)` order means. The `_supports_xdist` docstring says the check reads the literal signature, so `**kwargs` gets the wrapping path.

The merge drops `pyproject.toml` and `uv.lock` out of the diff and CI is green apart from the two skipped PyPI upload jobs. Nothing further from me.

pymc_marketing/bass/model.py
```diff
@@ -237,9 +257,121 @@ def f(
...
+    not supporting it and gets the ``as_xtensor`` wrapping instead. Declare
+    the parameter explicitly to take the native path.
+    """
+    if "xdist" not in signature(prior.create_variable).parameters:
+        # A VariableFactory written against the ``create_variable(name)``
+        # signature the protocol documents.
+        return False
+
+    if isinstance(prior, Censored):
+        return _supports_xdist(prior.distribution)
```
anevolbap: The signature check is what keeps custom `VariableFactory` objects working: anything written against the documented `create_variable(name)` protocol takes the `as_xtensor` wrapping path, anything declaring `xdist` takes the native one. This is why the fallback branch stays even though every distribution the default Bass model uses exists in `pymc.dims` at the pinned floor.

pymc_marketing/bass/model.py
```diff
@@ -237,9 +257,121 @@ def f(
...
+        return prior.create_likelihood_variable(
+            name,
+            mu=mu.transpose(*dims).values,
+            observed=observed.transpose(*dims).values,
+        )
+
+    # Censored keeps its parameters on the wrapped distribution.
+    inner = prior.distribution if isinstance(prior, Censored) else prior
+    if "mu" not in signature(inner.pymc_distribution.dist).parameters:
+        raise UnsupportedDistributionError(
```
anevolbap, line 296: `create_likelihood_variable` is only called with data: pymc-extras refuses `observed=None` there since pymc-devs/pymc-extras#732. The unobserved node goes through `create_variable` with `mu` attached instead, so the two guards that method applies (`UnsupportedDistributionError` for a mu-less distribution, `MuAlreadyExistsError`) are mirrored here to keep both paths raising the same typed errors. Covered by `test_mu_less_likelihood_without_observed_raises` and `test_unobserved_skips_create_likelihood_variable`.

pymc_marketing/bass/model.py
```diff
@@ -237,9 +257,121 @@ def f(
...
+    ``pmd.Deterministic`` takes the dims off the graph and only transposes
+    them, so a quantity built from parameters that do not carry every model
+    dim would be stored without those dims. Sizes come from the model, which
+    knows every dim it has registered, coords argument or not.
+    """
+    unknown = [
+        dim for dim in dims if dim not in var.dims and dim not in model.dim_lengths
+    ]
+    if unknown:
+        raise ValueError(
```
anevolbap: `pmd.Deterministic` reads dims off the graph, so a quantity built from parameters that do not carry every model dim needs an explicit `expand_dims` before storing. The `ValueError` matches the message pymc itself gives when a parameter dim is missing from the coords; before this it surfaced as a bare `KeyError` out of `dim_lengths`.

pymc_marketing/mlflow.py
```diff
@@ -561,8 +561,11 @@ def log_model_graph(model: Model, path: str | Path) -> None:
 
 
 def _get_random_variable_name(rv) -> str:
-    # Taken from new version of pymc/model_graph.py
-    symbol = rv.owner.op.__class__.__name__
+    op = rv.owner.op
```
anevolbap, line 564: Removed the `# Taken from new version of pymc/model_graph.py` line here: it dated to a900f976 (2024) and pymc's `random_variable_symbol` has since diverged (it now prefers `op.name` with a class-name fallback, and has no `core_op` unwrap), so the reference no longer described this helper. The `core_op` line is local to this PR.

williambdean:
Thanks for all of the iteraetion on this, @anevolbap! I think we are over complicating a bit of this. We can leverage xtensor and the checks that come from pymc, xtensor, pymc_extras, etc. Here are some items:

**All 6 helper functions are redundant with existing APIs.**

`_supports_xdist` (`inspect.signature` check) — `Prior`, `Censored`, and `VariableFactory` all accept `xdist: bool = False`. pymc_extras handles the introspection internally. Just pass `xdist=True`.

`_exp` — `ptx.math.exp` works on floats, xtensor, and xtensor scalars. No dispatch needed.

`_create_dim_variable` — `prior.create_variable(name, xdist=True)` is one line.

`_align_to_dims` — `pmd.Deterministic(name, value, dims=...)` already broadcasts and transposes.

`_check_dims_known` — `pmd.Deterministic` / `pmd.Data` raise `ValueError` for missing dims.

`_create_likelihood_variable` — `Prior.create_likelihood_variable` already has all guards (`UnsupportedDistributionError`, `MuAlreadyExistsError`). For observed: `prior.create_likelihood_variable(xdist=True)`. For unobserved: attach mu manually + `create_variable` (recommended workaround for pymc-extras#731, future-proof for upcoming deprecation).

The MMM codebase uses `Prior.create_variable(xdist=True)`, `ptx.math.*`, and `pmd.Deterministic` with `dims=` throughout (`transformers.py`, `link.py`, `fourier.py`, `additive_effect.py`). The Bass model should match.

`F()` and `f()` need `ptx.math.exp` and `ptx.math.square` — `pt.exp` / `pt.square` raise `TypeError` on xtensor variables.

`test_derivative` — `pytensor.grad` cannot work on xtensor graphs. `@pytest.mark.xfail` until pytensor#2337 merges.

~6 tests in `TestBassModelXdistFallbacks` exercise removed code paths and should be dropped.

`combined_dims` declaration-order fix, `peak` dims change, `core_op` unwrap in mlflow, and all `TestBassModelDims` tests are correct and should stay.

anevolbap:
Thanks for the detailed pass, it made me run every item against the released stack (pymc 6.2.0, pymc-extras 0.14.0). Good news first: the observed path does collapse to `create_likelihood_variable(xdist=True)`, `_check_dims_known` partly goes: 6.2 raises `ValueError` at `dims=` registration (m/p/q), though the observed path and the deterministic broadcast still need it, the bare failure there is a `KeyError` from inside pymc, and the unobserved branch is already the #731 mu-attach workaround. Three items break when run, though:

- `Prior("Wald", mu=10, lam=1).create_variable("w", xdist=True)` raises `UnsupportedDistributionError` (nested case too), and a `VariableFactory` without the kwarg raises `TypeError`. 0.14 has no internal fallback; that is what `_supports_xdist` covers.
- `ptx.math.exp` raises `TypeError` on numpy arrays and dim-less tensor vectors, and `pytensor.grad` fails on the result. `_exp` keeps the public `F`/`f` working on the inputs they accept today, which also keeps `test_derivative` green with no xfail.
- `pmd.Deterministic(dims=...)` transposes but does not broadcast: a value on `("T",)` with `dims=("T", "product")` raises `ValueError`. The broadcast is what `_align_to_dims` adds.

So the full cleanup means narrowing the contract to pymc.dims-supported distributions and xdist-aware factories, same as MMM. Is that the intent? If yes, the fallback helpers and the ~6 tests go and Wald-style priors stop working; happy to push that version.

williambdean:
We don't need to support everything. We want to leverage xtensor mainly like with the MMM implementation.
No need to "broadcast" a Deterministic. If it doesn't have that dimension, we can leave it to the user to do what they would like with it outside of the model. 

The model block should be lean.

anevolbap:
Pushed the lean version (42039a87): `_supports_xdist`, `_create_dim_variable`, `_check_dims_known` and `_align_to_dims` are gone, m/p/q and the likelihood go straight through `create_variable(xdist=True)`, and the deterministics are no longer broadcast. 108 bass tests green, model block down ~90 lines.

Two things stayed, with the reason:

- `_exp` in `F`/`f`. `ptx.math.exp` raises `TypeError: Cannot convert [...] to XTensorType` on a plain ndarray, so `f(0.03, 0.38, np.arange(5))` and every fixture built on it stops working, and `pytensor.grad` then fails on the xtensor graph. Keeping the dispatch is also what avoids the `test_derivative` xfail.
- The deterministics still get a transpose into `(T, ...)`. Without it `m * f(p, q, t)` comes out as `('product', 'T')` and the stored curves flip layout. No broadcasting, just the order.

One tradeoff to flag: without `_check_dims_known`, an unknown dim on `observed` now surfaces as `KeyError: 'geo'` from inside pymc_extras instead of a message naming the coord. The likelihood-dims case is unaffected, pymc's own error there is already clear. Happy to inline a two-line guard if that message is worth keeping.

williambdean:
Why the _exp 🥲

Just update the fixtures 😅

anevolbap:
As Claude usually says, "You're absolutely right!" — `_exp` is gone (cbaed2f2).

On `test_derivative`: skipped the xfail. Lowering the graph first (`rewrite_graph(..., include=("lower_xtensor",))`) lets `pytensor.grad` walk it, so the dF/dt against `f` check still runs instead of going dormant until pytensor#2337 lands.

daimon-pymclabs:
Review notes. Two of these I'd want closed before merge, two are cheap and belong in this PR rather than a follow-up. Everything else I found is listed at the bottom as non-blocking.

---

## 1. `copy.deepcopy(priors["likelihood"])` detaches tensor-valued parameters (blocker)

`model.py:409`. The comment above it is right about the problem (mutating the caller's prior leaks this model's dims into a reused config) but `deepcopy` is a heavier instrument than the problem needs, and it has a side effect.

`Prior.__deepcopy__` does `copy.deepcopy(self.parameters)`, so any pytensor variable sitting in a parameter is cloned. Confirmed on pytensor 3.3.0:

```python
>>> x = pt.scalar("x"); copy.deepcopy(x) is x
False
>>> copy.deepcopy(x * 2).owner.inputs[0] is x
False
```

So for a likelihood like `Prior("Normal", sigma=some_pm_data)`, the `y` that gets built is wired to a clone of `some_pm_data` that is not in the model. `pm.set_data` on the original then silently stops affecting `y` — no error, just a posterior that ignores the update. `Prior.deepcopy()` is not an escape hatch, it is `copy.deepcopy(self)`.

The mutation being guarded against is exactly one attribute, so scope it to that attribute. `Censored.dims` is a property whose setter forwards to `self.distribution.dims`, so the same code covers both types:

```python
from contextlib import contextmanager

@contextmanager
def _borrow_dims(prior: Prior | Censored, dims: tuple[str, ...]):
    """Set ``dims`` for the duration of the block, leaving the caller's prior as found.

    A copy is not usable here: ``Prior.__deepcopy__`` deep-copies ``parameters``,
    which clones any pytensor variable it finds and detaches it from the model
    graph, so a later ``set_data`` would no longer reach the likelihood.
    """
    original = prior.dims
    prior.dims = dims
    try:
        yield prior
    finally:
        prior.dims = original
```

and at the call site:

```python
with _borrow_dims(priors["likelihood"], combined_dims) as likelihood:
    observed_xt = ...  # unchanged
    _create_likelihood_variable(likelihood, "y", mu=adopters, observed=observed_xt)
```

`test_likelihood_prior_is_untouched` passes either way, which is why it did not catch this. The regression test that would:

```python
def test_shared_likelihood_parameter_stays_connected():
    """A tensor in a likelihood parameter must remain the model's own node."""
    with pm.Model(coords={"T": np.arange(5)}) as model:
        sigma = pm.Data("sigma", 1.0)
        create_bass_model(
            t=np.arange(5),
            observed=np.arange(5),
            priors={..., "likelihood": Prior("Normal", sigma=sigma, dims="T")},
            coords={"T": np.arange(5)},
            model=model,
        )
    assert sigma in pytensor.graph.basic.ancestors([model["y"]])
```

Same hazard applies to `inner.deepcopy()` at `model.py:284`. There the copy is genuinely needed (you are adding `mu` to `parameters`), so it cannot just be dropped — but it means the `observed=None` path detaches tensor parameters too. Minimum viable: copy the container and rebind the original parameter objects,

```python
unobserved = inner.deepcopy()
unobserved.parameters = {**inner.parameters, "mu": mu}
```

which keeps every original parameter object identical while still leaving `inner` untouched.

## 2. The `observed` dim lookup can pick up an unrelated variable's dims (blocker)

`model.py:417-420`:

```python
observed_dims = getattr(observed, "dims", None) or (
    model.named_vars_to_dims.get(getattr(observed, "name", None), combined_dims)
)
```

By the time this runs, `named_vars_to_dims` holds `adopters`, `innovators`, `imitators`, `peak`, `m`, `p`, `q`. A plain `pt.TensorVariable` that happens to be named `peak` — not registered with the model, just named — gets labelled with `peak`'s dims and transposed without complaint. Name equality is being used as a proxy for identity.

Also, `pm.Data` can be registered with a partially-`None` dims tuple, and `pmd.as_xtensor(obs, dims=(None, "product"))` fails from deep inside with an error that does not point back here.

Both are covered by one helper:

```python
def _observed_dims(observed, model: Model, combined_dims: tuple[str, ...]) -> tuple[str, ...]:
    """Axis labels for ``observed``: its own, else the model's, else positional."""
    own = getattr(observed, "dims", None)
    if own:
        return tuple(own)

    name = getattr(observed, "name", None)
    # Name equality is not identity: an unregistered variable may share a name
    # with a model variable, and would otherwise borrow its dims.
    if name is not None and model.named_vars.get(name) is observed:
        registered = model.named_vars_to_dims.get(name)
        if registered and all(dim is not None for dim in registered):
            return tuple(registered)

    return combined_dims
```

Test:

```python
def test_unregistered_variable_sharing_a_name_is_labelled_positionally():
    """Name equality must not be mistaken for identity."""
    # a bare variable named "peak", never registered on the model
    ...
    assert model["y"].dims == ("T", "product")  # not peak's dims
```

## 3. `F` / `f` lost plain-array support with no deprecation path (should fix here)

These are public and documented as `array-like`. `F(p, q, np.arange(10))` worked on `main`; now every call site needs `pmd.as_xtensor`, and every test in this diff was updated to do exactly that. The module docstring records the change, but nothing in the code does, and there is no test pinning what a user hitting this actually sees.

The signatures also disagree with the tests: `t: XTensorVariable`, but `test_derivative` still passes `pt.scalar`.

Two options, either is fine, but pick one explicitly:

**a. Accept array-like and wrap** — no break, one line each:

```python
def _as_time(t) -> XTensorVariable:
    """Accept an xtensor, or label a 1-d array-like positionally as ``("T",)``."""
    if isinstance(t, XTensorVariable):
        return t
    return pmd.as_xtensor(pt.as_tensor(t), dims=("T",))
```

**b. Keep the break and make it legible** — raise at the boundary and test the message:

```python
if not isinstance(t, XTensorVariable):
    raise TypeError(
        f"`t` must be an XTensorVariable, got {type(t).__name__}. "
        "Wrap plain arrays with `pymc.dims.as_xtensor(t, dims=('T',))`."
    )
```
plus `pytest.raises(TypeError, match="as_xtensor")`, and fix the annotations so the 0-d/scalar cases the tests rely on are actually covered by the declared type.

## 4. The deterministics dim change needs to be in the release notes (should fix here)

Dropping the broadcast so pooled `p`/`q`/`m` produce pooled curves is the right call and is well tested, but it changes posterior shapes and turns `plot_*` output for existing multi-series users from N subplots into one. The PR body says so; nothing in the repo does. With no CHANGELOG file, that means the release notes explicitly, along the lines of:

> **Breaking:** `adopters`, `innovators`, `imitators` and `peak` now carry only the dims their parameters carry. Previously they were broadcast to all combined dims, so a pooled `p`/`q`/`m` produced N identical curves; they now produce one. Posterior shapes and `plot_*` subplot counts change accordingly for models where any of `p`, `q`, `m` is pooled. Custom `VariableFactory` implementations for `m`/`p`/`q` must now accept an `xdist` keyword.

Worth an eyeball on the bass example notebook to confirm its narrative still matches the new shapes.

---

## Non-blocking

- `_create_likelihood_variable` reimplements pymc-extras internals (`signature(inner.pymc_distribution.dist).parameters`). The docstring cites pymc-extras#731; add a `# TODO(pymc-extras#731): drop once observed=None is supported upstream` at the code so it gets deleted rather than fossilised.
- `observed: pt.TensorLike | None` (`model.py:305`) should include `xr.DataArray` in the annotation and docstring — several new tests depend on that path.
- The `deterministic()` closure builds `order` by filtering `combined_dims`, so a dim on `value` that is not in `combined_dims` is dropped from the label rather than raising. Unreachable today; `assert set(value.dims) <= set(combined_dims)` keeps it that way.
- `BassPriors` types `m`/`p`/`q` as `VariableFactory`, but factories must now accept `xdist`. One that does not gets a bare `TypeError` from deep inside; a checked error naming the parameter would be kinder.
- Plotting error text: when the user passes an explicit unknown `dim`, "give `p`, `q` or `m` that dim" is the wrong advice for what is usually a typo'd dim name.
- Test gaps: a `Censored` likelihood with extra dims and `observed=None` never exercises dim propagation through the reconstructed `Censored(unobserved, ...)` — the existing test only covers `combined_dims == ("T",)`. And there is no coverage of the custom-`VariableFactory` path this PR flags as breaking.
- `lower()` depends on the internal rewrite name `"lower_xtensor"`. Annotated with pytensor#2337, which is the right thing to do; just noting it is a fragile hook.

Nice work overall — the migration is clean, and `test_observed_is_labelled_with_its_own_dims` and `test_combined_dims_follow_declaration_order` are exactly the right regression guards for the two subtle bugs this touches.

anevolbap:
**1.** This is not new in the PR. On `main` there is no copy at all, the code changes the prior in place, and `sigma_data` is still not an ancestor of `y`. Same for `pytensor.shared` and for a `pt` expression that is not constant. The values become constants inside `pymc_extras` on the `xdist=True` path, not in the deepcopy here. `_borrow_dims` is in anyway (eb0e2c42): it lends the dims and then restores the old ones, which is simpler than copying.

**2.** Right, and fixed. `priors["likelihood"].dims` was changed in place. That reached `model_config` and changed `bm.id`. New test: `test_building_does_not_mutate_the_caller_s_priors`.

**3.** Chose the explicit error (18a2e782). `F` and `f` now raise `TypeError` and name `pymc.dims.as_xtensor` when `t` has no dims. Tests check the message for an ndarray, a list and a vector with no dims.

**4.** Release notes come from labels, see `.github/release.yml`. `major` is the label that puts this under Major Changes.

Also merged main to get the `pandas<3` pin from #2951. The red CI here was the pandas 3.0.5 break, not the Bass code.

daimon-pymclabs:
Re-checked all four against `a47eb58` rather than the descriptions. Three are closed. The fourth is closed in mechanism but not in effect.

**1 and 2.** `_borrow_dims`, the parameter rebinding in `_create_likelihood_variable` (`unobserved.parameters = {**inner.parameters, "mu": mu}`) and `_observed_dims` with the `model.named_vars.get(name) is observed` identity check and the partially-`None` dims guard are all in, and `test_variable_sharing_a_name_does_not_borrow_its_dims`, `test_observed_data_without_dims_is_labelled_positionally` and `test_building_does_not_mutate_the_caller_s_priors` are the right guards. Taking your correction on 1: you are right that the detachment is not introduced here, and `test_building_does_not_mutate_the_caller_s_priors` catching the `bm.id` change is a better regression test than the one I proposed, since it pins the consequence users actually see.

That said, if the values really do become constants inside `pymc_extras` on the `xdist=True` path, then `pm.set_data` on a tensor passed into a likelihood parameter silently stops reaching `y` for every model built this way, not just Bass. That is a live footgun independent of this PR and there is no test anywhere pinning the behaviour either way. Happy to file it upstream separately if you agree with the reading, so it is tracked rather than folk knowledge.

**3.** `_check_time` is a better resolution than either option I offered: scalars of every type still work, arrays and dim-less vectors raise and name `pymc.dims.as_xtensor`, and the signatures now say `float | XTensorVariable | pt.TensorVariable` so the scalar cases the tests exercise are inside the declared type. `test_plain_array_time_points_point_at_as_xtensor` covers the message.

**4. Still open in effect.** The mechanism is right and I had missed `.github/release.yml`, but the `major` label is not on this PR. Current labels are `docs`, `enhancement`, `maintenance`, `tests`, `priority: medium`, `mlflow`, `Bass model`, and the categories in `release.yml` are matched in order, so `enhancement` claims it first and this lands under **New Features** rather than **Major Changes**. Auto-generated notes carry only the PR title, and `refactor(bass): migrate dim handling to pymc.dims` does not tell a multi-series user with a pooled `p`/`q`/`m` that their posterior shapes and `plot_*` subplot counts change. Adding `major` fixes the section; whether the title alone is enough warning is your call.

CI is fully green on `a47eb58` (55 success, 2 skipped), and I confirmed the earlier red was the pandas 3.0.5 break rather than the Bass code. #2953 lifts that pin properly, so the merge of main here is only a stopgap and this branch will want a rebase once that lands.

Nothing else blocking from me.

daimon-pymclabs:
Checked `fd7f3ac`. The change is right and the docstring correction is the better half of it: the old wording promised "whatever its type" while the code held a whitelist, so naming the whitelist removes a real trap for anyone reading the guard to work out what `t` accepts.

Two follow-ons, both cheap, neither blocking.

**The annotation drifted again.** `xr.DataArray` is now in `scalar_like` and in `test_scalar_time_points_need_no_labels`, but `F` and `f` still declare `t: float | XTensorVariable | pt.TensorVariable` (`model.py:195` and `:233`). That is the same mismatch as before, just one type smaller: a checker reading the signature rejects the call the test suite now asserts works. `float | int | np.number | npt.NDArray | xr.DataArray | XTensorVariable | pt.TensorVariable` is a mouthful, so a `TimeLike` alias next to `_check_time` would keep the guard and the signatures in one place and stop this recurring.

**A labelled 1-d `DataArray` is still rejected.** The commit message argues from `observed` already accepting an `xr.DataArray`, and that reasoning goes one step further than the patch does. `_observed_dims` takes `observed`'s own `.dims` when it has them, so a `DataArray` with `dims=("T",)` is labelled without the caller doing anything. For `t` the same object hits `np.ndim(t) == 0` and raises, telling the user to call `as_xtensor` with the dims the object is already carrying. It is the one input where the labels are unambiguous and available, so if you want the `observed` parity the message claims, that is the case to admit:

```python
if isinstance(t, xr.DataArray) and t.dims:
    return  # labels are on the object; `pmd.as_xtensor` reads them
```

Not asking for it in this PR, but it is a smaller gap than the one just closed.

**Status.** CI is green on `fd7f3ac` (55 pass, 2 skipping). #2953 merged, so the pandas pin is properly lifted on `main` and the branch is 2 commits behind, which makes this the moment for the rebase I mentioned and lets you drop the stopgap merge. `major` is still not on the labels, so the auto-generated notes will file this under New Features; leaving that with you as before.

anevolbap:
**1.** Right about the effect, and my earlier note was wrong twice. It is not the `xdist` path, and the values do not become constants.

`Prior.create_likelihood_variable` opens with `distribution = self.deepcopy()`, and `Prior.__deepcopy__` rebuilds with `**copy.deepcopy(self.parameters)`. A `pm.Data` parameter comes out as a second `TensorSharedVariable` holding the same value, so `set_data` updates the original while the graph keeps the clone. `xdist` is only forwarded on the last line, so both values behave the same.

Logp of `y` before and after `pm.set_data({"sigma_data": 50.0})`, on `Prior("Normal", sigma=pm.Data("sigma_data", 2.0), dims="T")`, pymc_extras 0.14.0:

```
create_variable(xdist=False)             wired     -4.8363 -> -14.4929
create_variable(xdist=True)              wired     -4.8363 -> -14.4929
create_likelihood_variable(xdist=False)  detached  -4.8363 ->  -4.8363
create_likelihood_variable(xdist=True)   detached  -4.8363 ->  -4.8363
```

So it hits every `Prior` likelihood, not only the dims ones. Please file it, scoped to `create_likelihood_variable` rather than to `xdist`.

The same probe on a Bass model whose likelihood `sigma` is a `pm.Data`:

```
                   main                          this branch
observed=None      detached, -6.4483 -> -6.4483  wired, -6.4483 -> -19.3238
observed=<data>    detached                      detached
```

The unobserved branch rebinds the parameters instead of mutating the copy, and that alone keeps the wiring, so the upstream fix is small. Added to the PR body, since it is a behaviour change nobody asked for.

**Annotation drift.** Fair, and it was mine: the guard moved in fd7f3ac4 and the signatures did not. Fixed in 048b1ad3 with a `TimeLike` alias, with the `isinstance` tuple next to it, so the guard and both signatures read from one place.

**Labelled 1-d DataArray.** Agreed, and `as_xtensor` does read the labels: given a `DataArray` with `dims=("T",)` it returns an `XTensorConstant` with dims `('T',)` and no arguments. Leaving it out of this PR as you suggest. The `observed` parity argument in fd7f3ac4's message went further than the patch did, that was overreach.

**4.** The order runs the other way. `Major Changes` is the first category in `.github/release.yml` (line 11) and `New Features` the third (line 17), and GitHub uses the first category a PR matches, so `enhancement` never claims it. `major` on its own is enough and nothing has to come off. Adding it needs maintainer rights, the labels API returns 403 here.

**Rebase.** Merged the new `main` instead (21a5ba89), so the pin is gone and the branch resolves pandas 3. The branch is 67 commits ahead with several update-branch merges pushed by others, so a rebase means a force push over a shared branch. The stopgap merge staying in the history costs nothing now that #2953 is in. CI is green on 21a5ba89.

daimon-pymclabs:
Taking the correction on 4: you are right and I read `release.yml` backwards. `Major Changes 🛠` is the first category (line 11) and GitHub stops at the first match, so `enhancement` was never going to claim it. **`major` is on the PR now** — my token has label rights on this repo, so that one is closed.

**1. Filed, or rather: I cannot file it.** `gh issue create` against `pymc-devs/pymc-extras` returns `403 Resource not accessible by personal access token`, so it needs someone with upstream rights. I verified the mechanism against `pymc_extras/prior.py` at 0.14.0 and it is exactly as you describe: `create_likelihood_variable` (line 1435) opens with `distribution = self.deepcopy()`, `Prior.__deepcopy__` rebuilds through `**copy.deepcopy(self.parameters)` (1328-1334), and `Prior.create_variable` (968) reads `self.parameters` directly with no copy at all. That asymmetry is the whole bug, and scoping it to `create_likelihood_variable` rather than to `xdist` is right. Body below, ready to paste:

---

**Title:** `Prior.create_likelihood_variable` deep-copies `pm.Data` parameters, detaching them from `set_data`

`Prior.create_likelihood_variable` deep-copies the prior before building the variable, so a `pm.Data` passed as a distribution parameter is cloned into a second shared variable. The model graph holds the clone, `pm.set_data` updates the original, and the likelihood silently stops responding. `Prior.create_variable` does not deep-copy, so the same parameter stays wired there. The two entry points disagree and only the likelihood one is affected.

Reproduction, on `Prior("Normal", sigma=pm.Data("sigma_data", 2.0), dims="T")`, logp of `y` before and after `pm.set_data({"sigma_data": 50.0})`, pymc_extras 0.14.0:

```
create_variable(xdist=False)             wired     -4.8363 -> -14.4929
create_variable(xdist=True)              wired     -4.8363 -> -14.4929
create_likelihood_variable(xdist=False)  detached  -4.8363 ->  -4.8363
create_likelihood_variable(xdist=True)   detached  -4.8363 ->  -4.8363
```

`xdist` makes no difference; it is forwarded on the last line, after the copy has already happened.

Cause:

```python
# pymc_extras/prior.py:1435-1438
distribution = self.deepcopy()
distribution.parameters["mu"] = mu
distribution.parameters["observed"] = observed
return distribution.create_variable(name, xdist=xdist)
```

Deep-copying a `TensorSharedVariable` yields a second shared variable holding the same value rather than the same object.

Impact: any model whose likelihood parameter is a `pm.Data` freezes the build-time value into the posterior and ignores later `set_data` calls. Nothing raises and the graph looks correct, so it fails quietly. Out-of-sample prediction and scenario code that varies a likelihood parameter this way gets the old value back with no warning.

Suggested fix, the same rebinding this PR already does on the unobserved branch:

```python
parameters = {**self.parameters, "mu": mu, "observed": observed}
```

That keeps the method non-mutating with respect to the caller's `Prior`, which is why the copy is there, while leaving shared variables identical. A test asserting the likelihood's logp moves after `pm.set_data` on a parameter would pin it for both entry points; neither has one today.

---

**Annotation drift.** `TimeLike` at `model.py:176` with the `isinstance` tuple beside it and the comment explaining why the tuple has to be spelled separately is the right shape, and both signatures now read from it. That closes it properly rather than moving the mismatch one type along.

**Rebase.** Agreed, withdraw the suggestion. A force push over a branch carrying other people's update merges is not worth removing a merge commit that costs nothing now #2953 is in.

CI is green on `dc80a6d` (6 runs, all success). Nothing blocking from me.

williambdean:
I'm good with this if you are @juanitorduz !

Thanks @anevolbap !

juanitorduz (approved):

PR #2784, opened 2026-07-30
perf: PyTensor rewrite Sum(ElemwiseMul) → Dot for faster MCMC sampling

williambdean:
## Summary

Introduce `local_sum_mul_to_dot`, a PyTensor rewrite that converts `Sum(ElemwiseMul(ExpandDims(w), X), axis=-1)` → `Dot(X, w)` during logp compilation. Opt-in via `use_db_name_as_tag=False` — only activated during `pm.sample()`; `compute_deterministics` uses the default mode.

## Performance impact

Measured on compiled `logp+dlogp` (what MCMC evaluates at each step). 3yr/10ch/Fourier(5) model.

| Linker | Without | With | Speedup |
|---|---|---|---|
| Numba | 0.061ms | 0.059ms | **1.03x** |
| CVM | 0.205ms | 0.204ms | ~1.0x |
| JAX | — | — | [upstream](https://github.com/pymc-devs/pytensor/issues/2317) |
| MLX | — | — | [upstream](https://github.com/pymc-devs/pytensor/issues/2316) |

JAX and MLX can't compile the full MMM model due to upstream PyTensor backend limitations (not our rewrite). Raw op tests pass on all 4 linkers.

## Why ship this?

The rewrite infrastructure (`_get_sampling_rewrites()`, mode-split in `RegressionModelBuilder.fit`, opt-in registration) establishes the pattern for future graph-level optimizations. The rewrite itself is safe, tested (23 tests, 4 linkers), and has zero impact on non-MMM models.


ricardoV94:
Why here and not in PyTensor? And numbers should show the logp+dlogp function that mcmc actually uses.

williambdean:
That is a possibility as well

williambdean:
Good point — updated the PR body with logp+dlogp numbers on a realistic 3yr/10ch/Fourier(5) model. ~9% with the default Numba linker.

daimon-pymclabs:
Reviewed the rewrite logic and the `model_builder` wiring. A few correctness concerns in `local_sum_mul_to_dot`, plus two questions about where the mode gets applied.

**1. `_unwrap_weight` doesn't check the DimShuffle pattern (possible miscompile)**

It accepts *any* `DimShuffle` whose input is 1D, then assumes the vector lands on the trailing axis. That holds for the implicit broadcast Elemwise inserts (`new_order == ('x', 0)`), but not for an explicitly written shuffle:

```python
X = pt.matrix("X")        # (T, C)
w = pt.dvector("w")       # (T,)
expr = (X * w[:, None]).sum(axis=-1)   # DimShuffle new_order == (0, 'x')
```

Here `weight` unwraps to `w` (length T), `data` is `X`, `axis == (1,) == data.ndim - 1`, so the guards pass and the rewrite emits `pt.dot(X, w)` — which contracts C against T. Silent wrong answer when `T == C`, shape error otherwise. Suggest requiring the inner axis to be last and everything else to be broadcast:

```python
new_order = var.owner.op.new_order
if new_order[-1] != 0 or any(o != "x" for o in new_order[:-1]):
    return None
```

**2. The axis guard is checked against the wrong ndim**

`node.op.axis` indexes the `Elemwise` output, but the guard compares against `data.type.ndim - 1`. When `data` is itself broadcast up (fewer dims than the mul output) those two differ, and `pt.dot(data, weight)` then has the wrong output shape. Safer to gate on the Sum's input and require no broadcasting of `data`:

```python
if data.type.ndim != inp.type.ndim:
    return None
if node.op.axis != (inp.type.ndim - 1,):
    return None
```

That last line also subsumes the `len(axis) != 1` check.

**3. Import-time mutation of the global `optdb`**

`optdb["specialize"].register(...)` runs at module import, and `mmm/__init__.py` imports `rewrites`, so merely importing `pymc_marketing.mmm` mutates PyTensor global state for the whole process. It also makes the module non-reimportable (re-registering the same name raises). This is the strongest argument for @ricardoV94's point — as a PyTensor rewrite it's registered once, in the right place, and gets PyTensor's own graph-rewrite test infrastructure. If it stays here, consider registering lazily inside `create_sampling_mode` guarded on the name not already being present.

**4. `pytensor.config.change_flags(mode=custom_mode)` — is a `Mode` object valid there?**

`config.mode` is a string flag (`"Mode"`, `"FAST_RUN"`, `"NUMBA"`, ...) with a validating filter; passing a `Mode` instance looks off. The documented path is `pm.sample(..., compile_kwargs={"mode": custom_mode})`, which also scopes precisely to logp/dlogp compilation rather than everything compiled inside the `with` block. Worth switching to that.

Related: `compile_kwargs`/`config.mode` only reaches the PyMC-compiled logp. With `nuts_sampler="nutpie"` or `"numpyro"` the gradient function is built through a different path, so the rewrite silently no-ops for a good share of real MMM runs. Worth stating in the `fit` Notes.

**5. Enabling it unconditionally for all MMM users, given the numbers**

`MMM._get_sampling_rewrites` returns the rewrite with no opt-out, and the PR body reports ~0.89× on the CVM linker. Anyone not on Numba takes a measured regression they can't turn off. Suggest a `fit` kwarg (or a class attribute) defaulting to off, or gating on the active linker.

**6. Minor**

- `model_builder.fit` rebuilds the mode inline, duplicating `create_sampling_mode`. Call it — that also keeps the two in sync.
- `create_sampling_mode` reconstructs a plain `Mode` from `base_mode.linker` and drops anything else the base carried (e.g. a `NanGuardMode` wrapper). `base_mode.including(*names)` does the same job and preserves the subclass.
- `pytest>=9.1.1` in `[dev]` looks unrelated to this PR.

Points 1 and 2 are worth a regression test each — an explicit `w[:, None]` shuffle and a broadcast `data` — since both currently slip past the guards.

juanitorduz:
@williambdean will you keep this here or move it to PyTensor?

williambdean:
Can move it to pytensor.

daimon-pymclabs:
Moving it to PyTensor resolves my point 3 outright — the `optdb["specialize"].register(...)` at import time is only a problem because `pymc_marketing.mmm.__init__` imports `rewrites`, so importing the MMM subpackage mutates PyTensor global state for the whole process and makes the module non-reimportable. Registered inside PyTensor it happens once, in the right place, and it inherits PyTensor's graph-rewrite test infrastructure (`rewrite_graph`, the `local_*` test conventions) instead of hand-rolled `dprint` assertions here.

Two things worth keeping in mind for the split, since the findings don't all travel together:

**Goes with the rewrite (still open):** points 1 and 2. `_unwrap_weight` accepts any `DimShuffle` over a 1-D input and assumes the vector lands on the trailing axis, so an explicit `w[:, None]` shuffle rewrites to a `dot` that contracts the wrong axis; and the axis guard compares `node.op.axis` (indexing the Elemwise output) against `data.type.ndim - 1`, which diverge when `data` is itself broadcast up. `7bd1a933` tightened the axis to a single last-dim sum but neither case is guarded yet. Both want a regression test, and PyTensor is a better home for them than here.

**Stays in pymc-marketing:** points 4, 5 and 6 are about the wiring, not the rewrite. `config.change_flags(mode=<Mode instance>)` vs `compile_kwargs={"mode": ...}`, the fact that neither reaches nutpie's or numpyro's gradient path, the unconditional enablement despite the ~0.89× CVM number, and the `create_sampling_mode` duplication in `model_builder.fit`. Those need deciding either way — even with the rewrite upstream, something here still has to opt into it, and on a non-Numba linker that opt-in is currently a regression users can't turn off.

Practical note if it does move: the rewrite then arrives via a PyTensor release, so `_get_sampling_rewrites` would gate on a version floor rather than a locally-registered name. Might be worth landing the guard fixes upstream first and keeping this PR as the (much smaller) wiring change.

juanitorduz:
> Can move it to pytensor.

sounds like the right call :) Thanks!

PR #2785, opened 2026-07-31
Support multidimensional inputs in to_xarray

hadibly9866:
## Description

This PR extends the `to_xarray` helper function to support multidimensional inputs.

Previously, `to_xarray` assumed all inputs were one-dimensional vectors. This change keeps the existing behavior for 1D arrays while automatically creating additional dimensions for higher-dimensional inputs.

## Changes

- Added support for multidimensional arrays in `to_xarray`
- Preserved existing customer dimension handling
- Added tests covering 2D array inputs

## Testing

Passed:

----
📚 Documentation preview 📚: https://pymc-marketing--2785.org.readthedocs.build/en/2785/

### DIFF
--- a/pymc_marketing/clv/utils.py
+++ b/pymc_marketing/clv/utils.py
@@ -32,11 +32,16 @@ __all__ = [
 
 def to_xarray(customer_id, *arrays, dim: str = "customer_id"):
     """Convert vector arrays to xarray with a common dim (default "customer_id")."""
-    dims = (dim,)
     coords = {dim: np.asarray(customer_id)}
 
     res = tuple(
-        xarray.DataArray(data=array, coords=coords, dims=dims) for array in arrays
+        xarray.DataArray(
+            data=array,
+            coords=coords,
+            dims=(dim,)
+            + tuple(f"dim_{i}" for i in range(1, np.asarray(array).ndim)),
+        )
+        for array in arrays
     )
 
     return res[0] if len(arrays) == 1 else res
--- a/tests/clv/test_utils.py
+++ b/tests/clv/test_utils.py
@@ -57,6 +57,18 @@ def test_to_xarray():
     assert new_y.dims == ("test_dim",)
     np.testing.assert_array_equal(new_y.coords["test_dim"], customer_id)
 
+    # Test multidimensional input
+    matrix = np.arange(20).reshape(10, 2)
+
+    new_matrix = to_xarray(customer_id, matrix)
+
+    assert isinstance(new_matrix, xarray.DataArray)
+    assert new_matrix.dims == ("customer_id", "dim_1")
+    np.testing.assert_array_equal(
+        new_matrix.coords["customer_id"], customer_id
+    )
+    np.testing.assert_array_equal(new_matrix.values, matrix)
+
 
 @pytest.fixture(scope="module")
 def fitted_gg(test_summary_data) -> GammaGammaModel:


daimon-pymclabs:
Solid, back-compatible change: `extra_dims` is opt-in, the 1D path is untouched, and deliberately refusing auto-generated trailing names (so unrelated axes of two arrays don't silently align) is the right call and well argued in the docstring. Four things I'd raise, one of them a real bug.

**Bug: a bare string passed as `extra_dims` silently splits into per-character dim names.** `Sequence[str]` accepts `str`, so this succeeds with no error:

```python
>>> to_xarray(np.arange(4), np.ones((4, 2, 3)), extra_dims="ab").dims
('customer_id', 'a', 'b')
```

Checked against xarray 2026.7.0. `extra_dims="channel"` happens to trip the uniqueness check (repeated `n`), so the failure mode is inconsistent as well as silent. An `isinstance(extra_dims, str)` guard, or narrowing the annotation to `tuple[str, ...] | list[str]`, would close it.

**`_resolve_dims` rejects intermediate ranks.** Only `ndim == 1` or `ndim == len(dims)` are accepted, so with `extra_dims=("channel", "lag")` you cannot pass a `(customer, channel)` array and a `(customer, channel, lag)` array in the same call. That may well be intentional given the alignment argument above, but the docstring doesn't state it, and the error text ("Pass 'extra_dims' naming the trailing axes of multidimensional arrays") misdirects when `extra_dims` was already supplied.

**Trailing dims get names but no coordinates.** You can label an axis `channel` but not give it values, so `.sel(channel="tv")` isn't available downstream, and a `DataFrame` input drops its column labels. Not a blocker for this PR, but an `extra_coords` argument seems like the natural follow-up.

**Minor points**

- A 0-d input now raises `Cannot label an array of 0 dimensions with ('customer_id',). Pass 'extra_dims' ...`, where the advice is unactionable. Worth a separate branch with a clearer message.
- The 2D case is appended to `test_to_xarray` while every other new case is a standalone test. I'd split it out for consistency.
- No test covers the string-`extra_dims` case above; worth adding alongside the fix.

juanitorduz:
Thanks — the string bug was real. Fixed in ab6bae3e, along with the two error-message complaints. The `extra_coords` suggestion I'm deliberately leaving out; reasoning below.

**String `extra_dims` — fixed.** Guarded at runtime rather than by narrowing the annotation, since the annotation doesn't run:

```
>>> to_xarray(np.arange(4), np.ones((4, 2, 3)), extra_dims="ab")
TypeError: extra_dims must be a sequence of dimension names, not a string;
pass extra_dims=('ab',) to name a single trailing axis.
```

Same message for `extra_dims="channel"`, so the inconsistency you spotted is gone too. `Sequence[str]` stays, so lists and tuples both still work. Covered by `test_to_xarray_rejects_string_extra_dims`.

**Intermediate ranks — intentional, now documented.** Keeping the restriction: with `extra_dims=("channel", "lag")` a `(customer, channel)` array is genuinely ambiguous — nothing says whether the missing axis is `lag` or `channel`, and silently assuming a prefix is exactly the class of guess this PR removed. Arrays with different trailing axes need one call each. That rule is now in the docstring, and the message no longer misdirects:

```
ValueError: Cannot label an array of 2 dimensions with ('customer_id', 'channel', 'lag'):
every array must have either 1 or 3 dimensions. Arrays whose trailing axes differ need
one call each, so that every axis is named unambiguously.
```

**0-d — separate branch.**

```
ValueError: Cannot convert a 0-dimensional array: every array needs a leading 'customer_id' axis.
```

**`extra_coords` — declining for now.** You're right that named-but-unpopulated axes are a real limitation, but `to_xarray` still has no multidimensional caller anywhere in the repo (every one of the six callsites passes 1-D pandas Series), and `.assign_coords(channel=df.columns)` is a one-liner. Shipping a second new parameter to serve a hypothetical consumer is how the anonymous `dim_1` scheme got here. I've documented the `assign_coords` idiom in the Examples instead; happy to add `extra_coords` the moment something in the codebase needs it.

**Test layout.** The 2-D case is now its own `test_to_xarray_names_trailing_axes`, and `test_to_xarray_invalid_dims_raise` gained rows for the intermediate-rank and 0-d branches.

`tests/clv/test_utils.py` is 87 passed; `ruff format`/`ruff check`/`mypy` clean on both files.

daimon-pymclabs (approved):
All four points from my earlier review are addressed. Re-checked the updated code against xarray 2026.7.0:

- A bare string now raises `TypeError` with the fix in the message (`extra_dims=('ab',)`), and `test_to_xarray_rejects_string_extra_dims` covers it.
- The 0-d case has its own message instead of the misdirecting "pass 'extra_dims'" advice.
- The intermediate-rank error states the accepted ranks and why, and the docstring gained a paragraph on the rule.
- The coords gap is answered with a documented `assign_coords` example. Verified: `.sel(channel="radio")` on a `DataFrame` input returns the right column.

No regressions on 1D, multi-array, custom `dim`, or pandas `Series` inputs, and all five parametrized error cases match their regexes.

One cosmetic nit, not worth another round: only the string case is guarded, so `extra_dims=(3,)` still yields dims `('customer_id', 3)`. xarray accepts non-string hashables, so nothing breaks.

LGTM once CI is green.

daimon-pymclabs:
Read ab6bae3 — all four resolved, and the `extra_coords` call is the right one. Six 1-D callsites do not justify a second parameter, and the `assign_coords` line in the Examples is the better answer until something in the repo needs more.

Two small things on the new code, neither blocking:

- The `TypeError` guard sits before the uniqueness check, so `extra_dims="channel"` now reports the string problem rather than the incidental repeated-`n` one. That is the ordering I would have picked.
- `_resolve_dims` reaches the 0-d branch only because `len(dims) >= 1` always holds, which is true today but implicit. A `ndim == 0` check ahead of the `ndim == 1` return would make it independent of that.

LGTM as it stands.

PR #2773, opened 2026-07-29
docs(var): Add MMM-VAR joint optimisation routine.

thomaspinder:
## Description

This PR demonstrates how the independent posterior artefacts of a MMM and a VAR may be considering simultaneously within a budget optimisation routine. 

## Checklist
- [x] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [ ] Included tests that prove the fix is effective or that the new feature works
- [x] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--2773.org.readthedocs.build/en/2773/

PR #2786, opened 2026-07-31
Remove merge queue support from CI workflows

daimon-pymclabs:
Drops the `merge_group:` trigger from `test.yml`, `test_notebook.yml`, `pypi.yml`, and `docs.yml` so CI no longer runs in merge-queue context.

**This PR alone does not disable the merge queue.** The queue is a branch-protection setting on `main`, not a file in the repo, so it has to be turned off in the repo settings ("Require merge queue" under the `main` protection rule).

Order matters: turn the setting off first, then merge this. If the workflows lose their `merge_group` triggers while the queue is still required, the required checks (`all_tests`, `build source distribution check`) never report in queue context and every entry stalls until it times out. There are 3 PRs in the queue right now.

----
📚 Documentation preview 📚: https://pymc-marketing--2786.org.readthedocs.build/en/2786/


juanitorduz:
@williambdean do you also want to remove this from the branch rule?

daimon-pymclabs:
@juanitorduz yes — and it has to come first, not after. The branch rule is the half of this that isn't in code, and Bugbot's medium-risk note above is the same point: while "Require merge queue" is still on for `main`, dropping the `merge_group:` triggers means a queued PR builds a `gh-readonly-queue/...` ref that no workflow reacts to, so `all_tests` and `build source distribution check` never report in queue context and the entry sits there until it times out.

The queue is definitely still live, so this isn't hypothetical — #2773 ran Test/Docs/PyPI on `gh-readonly-queue/main/pr-2773-8de81b8` at 07:30 today, and #2770 was merged by `github-merge-queue[bot]` at 07:42.

So the order is: turn off "Require merge queue" on the `main` branch rule (and `v1.0.0` if it's set there too), let the queue drain, then merge this. I don't have admin on the repo so I can't flip the setting — @williambdean or you would need to. Happy to leave this PR sitting until that's done.

juanitorduz (approved):

williambdean:
let's try it out!

daimon-pymclabs:
@williambdean this got auto-removed from the queue at 12:33 UTC "due to no response for status checks," which is the failure mode from my comment above rather than a flake. The branch still has the `merge_group:` triggers deleted, so when the queue built the `gh-readonly-queue/main/pr-2786-...` ref no workflow reacted to it, `all_tests` and `build source distribution check` never reported in queue context, and the entry timed out.

Requeuing will do the same thing every time. The order has to be:

1. Turn off "Require merge queue" in the `main` branch rule.
2. Merge this PR normally. The `pull_request` runs are green and are the ones branch protection will be reading once the queue requirement is gone.

Doing it the other way round is what we just watched. Happy to rebase if it has gone stale by the time the rule is off.

daimon-pymclabs:
Third auto-removal just now (18:17 UTC), same cause as the two before it. Enabling auto-merge does not change anything here: auto-merge still routes the PR through the queue, so it will keep timing out on status checks that no workflow can report while `merge_group:` is deleted on this branch. Nothing in CI can fix this from the PR side.

The one blocking step is a repo setting nobody has flipped yet: turn off "Require merge queue" in the `main` branch ruleset, then merge normally against the green `pull_request` runs. @juanitorduz @williambdean, one of you has admin on the ruleset.

Separately, the failing `build` (docs) check is unrelated to this PR. The artifact shows a single warning, `failed to reach any of the inventories`, i.e. a transient intersphinx fetch under `-W`. Docs are green on `main` and this branch only edits workflow triggers, so it is a flake and a rerun should clear it.

Issue #2732, opened 2026-07-14
Remove `model_config.py` Prior class `DeprecationWarning`?


ColtAllen:
https://github.com/pymc-labs/pymc-marketing/blob/v1.0.0/pymc_marketing/model_config.py#L154

This deprecation warning resides in the last conditional block of the following code snippet:

```python
        # Skip deserialization for other non-dict types (strings, numbers, etc.)
        # These are not distribution configurations
        if not isinstance(prior_config, dict):
            return prior_config

        try:
            dist = deserialize(prior_config)
        except Exception as e:
            parse_errors.append(f"Parameter {name}: {e}")
        else:
            msg = (
                f"{name} is automatically converted to {dist}. "
                "Use the Prior class to avoid this warning."
            )
            warnings.warn(msg, DeprecationWarning, stacklevel=2)

            return dist
```

It raises errors when the old condition is specified for priors:

`new_prior = {"a": {"dist": "HalfNormal", "kwargs": {"sigma": 2}}}`

These warnings are still occasionally being raised despite some tests which specifically check for them, and may be related to https://github.com/pymc-labs/pymc-marketing/issues/846.

There are significant API changes in `v1.0.0` Should this `DeprecationWarning` logic be removed?

ColtAllen:
Closed as completed by https://github.com/pymc-labs/pymc-marketing/pull/2789

PR #2789, opened 2026-07-31
Remove deprecated `Prior` module

ColtAllen:
## Description

The `Prior` class for defining priors was originally created in `pymc-marketing`, but now resides in `pymc-extras`. This PR removes the dead code as well as support for the old model config spec, which the library at large moved away from about two years ago. If a user tries to use the old spec, an informative error message is raised. 

## Related Issue

- [ ] Closes https://github.com/pymc-labs/pymc-marketing/issues/2732
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2789.org.readthedocs.build/en/2789/


daimon-pymclabs:
Summary of my earlier review as a concrete checklist.

### Must fix

1. **`pymc_marketing/model_builder.py:412`**: guard the new `class`/`data` branch in `_model_config_formatting` with `try/except DeserializableError` and fall back to `_format(value)`. Right now any saved `model_config` containing an unrelated dict with a string `class` key *and* a `data` key gets routed to `deserialize()` and makes the model unloadable instead of falling through to recursion. `test_non_prior_mapping_with_class_key_unchanged` uses `{"class": "premium", "tier": 2}`, which has no `data` key, so it doesn't cover this.
2. **`pymc_marketing/mmm/builders/factories.py`**: wrap `resolve`/`deserialize` in `build` and re-raise with the offending config key attached. `tests/mmm/builders/test_yaml.py` now expects a bare `DeserializableError` where it previously expected `ModelConfigError`, which named every bad key (`Parameter alpha: ...`). YAML is the entry point most likely to be hand-edited, so it's the worst place to lose the parameter name.
3. **`pymc_marketing/pie/model.py:219-222`**: call `parse_model_config` here, or move `check_legacy_prior_spec` into the base `__init__`. `check_legacy_prior_spec` only runs inside `parse_model_config`, and this path passes `model_config` straight to `super().__init__`, so a legacy dict prior still dies with the opaque downstream failure this PR exists to replace.
4. **`parse_model_config` / `CLVModel.__init__`**: add a `**kwargs` guard that raises the migration-style message for `non_distributions`. Currently callers get a bare `TypeError: unexpected keyword argument`, which is inconsistent with the PR's own thesis of "fail with a migration hint, not an opaque error."

### Should fix

5. Use `deserialize` uniformly in `_model_config_formatting`. The `dist` branch uses `Prior.from_dict` while the `class` branch uses `deserialize`; `deserialize` handles both and also picks up the `special_priors` registrations. Related: move the function-local `deserialize`/`Prior` imports to module top unless there's a circular-import reason.
6. Move `register_deserialization(is_alternative_prior, ...)` out of `mmm/builders/factories.py` into `special_priors.py` or `serialization.py`. Import-time mutation of the registry from a build-machinery module is awkward when those two modules already own that role, and `tests/test_prior.py` importing `factories` "to register custom deserializers" makes it visible. That test file is also now misnamed: it tests the YAML alt-prior deserializer and belongs under `tests/mmm/builders/`.
7. Finish the locale cleanup. The top-level `pymc_marketing.prior.*.po` files are deleted, but 18 files remain under `locales/es/LC_MESSAGES/api/generated/classmethods/pymc_marketing.prior.*`, and `locales/es/LC_MESSAGES/api/index.po:146` still carries the `prior` autosummary entry that `docs/source/api/index.md` just lost.
8. **`pymc_marketing/clv/models/gamma_gamma.py:416`**: `{dist: 'HalfCauchy', kwargs: {}}` became `Prior("HalfCauchy", beta=1)`, silently gaining a parameter. Confirm that's intended. Both rewritten docstring examples also use `Prior` without an import line in the snippet.
9. Restore the `Prior class` rule in `.github/labeler.yml`, or delete the label from the repo. The label still exists and is applied to this PR.

### PR description / release notes

10. Call out that `parse_model_config` is now a much weaker validator. `{"dis": "Normal"}`, `"gamma": "Completely wrong"`, and any typo'd key pass through silently and fail later. This looks intentional (the "Parsing with an error" docstring section and `test_parse_model_config_catches_errors` both go away), but `ModelConfigError` now has exactly one trigger and that's worth stating.
11. Call out that `from pymc_marketing.prior import Prior` never emitted a runtime `DeprecationWarning`, so users go straight from a working import to `ModuleNotFoundError`. Defensible for a major, but there's no migration or breaking-changes page under `docs/source/`, and this PR removes three public surfaces at once: the module, the dict config format, and `non_distributions`.

### Verified, not problems

- The JSON round-trip holds. `Prior.from_dict` converts `dims` list to tuple and kwarg lists to `np.array` (pymc-extras `prior.py:1140-1153`), so the new `dist` branch preserves the semantics the old generic-dict recursion provided. No equality or dtype regression for multi-dim priors with array kwargs.
- The `mv_its` fix is the right target: it builds `attrs_to_init_kwargs` from scratch rather than calling `super()`, making it the only override that was missing `_model_config_formatting` (`mmm.py` and `maxdiff.py` already call it, `pie` goes through `super()`).
- No deserializer clash from the tightened `is_alternative_prior`: every `special_priors` predicate keys on `"special_prior"`, not `"distribution"`.
- No `non_distributions` or `pymc_marketing.prior` references left in code, docs, or notebooks. `"removed in v1.0.0"` matches `__version__ = "1.0.0.dev0"`. No unused imports left behind.

ColtAllen:
All nine code/docs items in https://github.com/pymc-labs/pymc-marketing/pull/2789#issuecomment-5170574978 are done and verified. Summary:

**Must fix**
1. **`model_builder.py`** — the `class`/`data` branch is now guarded with `try/except DeserializableError` and falls back to recursion. `{"class": "premium", "data": {"tier": 2}}` survives unchanged instead of making the model unloadable.
2. **`factories.py`** — a `_naming(key)` context manager wraps every `resolve`/`deserialize` call in `build`, re-raising as `ModelConfigError`. The YAML fixtures now fail with `Parameter intercept: …` and `Parameter likelihood: …`; `test_yaml.py` asserts on that message again rather than a bare exception type.
3. **`pie/model.py`** — calls `parse_model_config` after `super().__init__`, following the `CLVModel` precedent. A legacy dict prior now raises the migration hint at construction.
4. **`model_config.py` / `CLVModel.__init__`** — a shared `_reject_removed_parameters` guard raises `ModelConfigError` naming `non_distributions`; unrelated stray keywords still get `TypeError`.

**Should fix**

5. Both `dist` and `class` branches go through `deserialize` now, imported at module top. This also picks up the `special_priors` registrations that `Prior.from_dict` rejects with `KeyError` — covered by a new test.
6. `is_alternative_prior` / `deserialize_alternative_prior` / `register_deserialization` moved to `special_priors.py`; `factories.py` imports it for the side effect. `tests/test_prior.py` merged into `tests/test_special_priors.py` and deleted.
7. 18 locale files deleted, plus the `prior` autosummary entry at `api/index.po:146`. Left the `*sample_prior*` / `special_priors` files alone — unrelated APIs matching on substring.
8. `Prior("HalfCauchy", beta=1)` → `Prior("HalfCauchy")` in the `GammaGammaModelIndividual` example; `from pymc_extras.prior import Prior` added to both snippets. The sibling `GammaGammaModel` keeps `beta=1` — that one matches its original `kwargs: {"beta": 1}`.
9. No labeler change (the rule's only glob pointed at the deleted file).

**Verification:** 379 tests pass across the affected suites, plus the full 295-test CLV suite; `pre-commit` clean (ruff, ruff-format, mypy). An end-to-end save/load round trip with `Censored` + a multi-dim `Prior` with array kwargs + `LogNormalPrior` + a plain `class`/`data` mapping in one config restored all four correctly (`dims=('channel',)`, `mu=array([1., 2.])`) — that scratch script is deleted.

**Two things need you:**
- [[sandbox/pr-notes-addendum.md](https://claude.ai/epitaxy/sandbox/pr-notes-addendum.md)](sandbox/pr-notes-addendum.md) has the items 10/11 text to paste into the PR description — no `gh` CLI in this environment, so I couldn't post it.
- Delete the "Prior class" label from repo settings.

juanitorduz (approved):
thanks @ColtAllen !

ColtAllen:
Thanks @juanitorduz !

@carlosagostini a lot of `locales/` translation files get deleted in this PR; is this worth posting a follow-up issue?

Issue #2779, opened 2026-07-30
decomposition.waterfall calls fig.tight_layout() unconditionally, overriding the ambient layout engine


juanitorduz:
### Summary

`MMMPlotDecomposition.waterfall` creates a plain figure and then calls `fig.tight_layout()` unconditionally:

- figure created with a bare `plt.subplots` — `pymc_marketing/mmm/plotting/decomposition.py:355`
- `fig.tight_layout()` — `pymc_marketing/mmm/plotting/decomposition.py:394`

When the ambient matplotlib style has `figure.constrained_layout.use: True`, that call **replaces the layout engine the style installed** and matplotlib warns:

```
UserWarning: The figure layout has changed to tight
```

`arviz`'s `arviz-darkgrid` sets that rcParam, and it is applied in the setup cell of many of our docs notebooks, so the warning shows up in rendered documentation output rather than only in user code.

### Why this framing

It is tempting to call this the notebook's problem (it is the notebook's style that turns constrained layout on) but the defect is on the library side: a plotting helper should lay its own figure out **without** overriding a layout engine chosen by the caller's environment. `tight_layout()` is the one call that does override it.

### Suggested fix

```python
- fig.tight_layout()
+ fig.set_layout_engine("tight")
```

`set_layout_engine` hands the engine over rather than fighting it: no warning, and the resulting axes positions are identical (checked to four decimal places on the notebook figures in #2769, which made the same change at eight of its own call sites). The alternative — calling `tight_layout()` only when `fig.get_layout_engine()` is None — preserves a caller's constrained layout instead, and is the better behaviour if we would rather respect the ambient engine than replace it.

Grepping `mmm/plotting/` shows no other layout handling, so this is the only site.

anevolbap:
Checked both options on matplotlib 3.10.9. `set_layout_engine("tight")` drops the warning and gives identical axes bounds, but it still replaces a constrained engine the caller installed, just quietly. The conditional version (`if fig.get_layout_engine() is None: fig.tight_layout()`) leaves an ambient engine alone; a bare `plt.subplots()` returns `None`, so behaviour without a style is unchanged. That one does change rendered output under `arviz-darkgrid`, from tight to constrained. Which is preferable here?

Also, `tight_layout` is called at 10 more sites outside `mmm/plotting/`, all in `mmm/plot.py`. Should the fix cover those too, or stay on `decomposition.waterfall`?

anevolbap:
[...]

PR #2793, opened 2026-07-31
fix(mmm): keep the ambient layout engine in decomposition.waterfall

anevolbap:
Addresses #2779

`MMMPlotDecomposition.waterfall` called `fig.tight_layout()` unconditionally, which replaces a layout engine the caller's style installed. Under `arviz-darkgrid` (which sets `figure.constrained_layout.use = True`) matplotlib warns `The figure layout has changed to tight`, and that warning shows up in rendered docs.

This takes the second option from the issue: lay the figure out only when no engine is set. A bare `plt.subplots()` returns `None`, so behaviour without a style is unchanged; a constrained engine now survives instead of being replaced. The other option, `set_layout_engine("tight")`, silences the warning but still overrides the caller's engine.

Scope is the single site named in the issue. `tight_layout` is called at 10 more places in `mmm/plot.py` with the same problem; open question on the issue whether they belong here.

The `get_layout_engine` API needs matplotlib 3.6, while `pyproject.toml` declares `matplotlib>=3.5.1`. In practice the floor is higher: `arviz-plots[matplotlib]>=1.2.0` requires `matplotlib>=3.9`. Left the pin alone here.

Note for docs: `mmm_quickstart`, `mmm_example` and `mmm_case_study` call `decomposition.waterfall` under `arviz-darkgrid`, so those figures now render constrained instead of tight. Nothing fails, the notebooks re-execute in CI, but stored outputs change on the next run.

Tests: `test_keeps_ambient_layout_engine` runs `waterfall()` under `figure.constrained_layout.use = True` with the layout warning as an error and checks the `ConstrainedLayoutEngine` survives, for both the single-panel and multi-panel cases. It fails on the unpatched source. `test_lays_out_figure_without_ambient_engine` covers the default path.


juanitorduz:
Thanks @anevolbap ! I will keep the tight layout for the waterfall plot (as you are doing here).... but maybe remove them completely in the other cases? WDYT?

anevolbap:
I measured this before answering, in a fresh env matching main (pymc 6.0.1, arviz 1.2.0, arviz-plots 1.2.0, matplotlib 3.10.9). Method: rebuild the same figures from the existing test fixtures twice, once as they are and once with `tight_layout` patched to a no-op, then compare the drawn geometry.

Removing the call sends every figure back to matplotlib's default subplot rectangle, so the panels cover 59.7% of the canvas in all four methods I measured, against 82.4% in `posterior_distribution`, 74.3% in `budget_allocation`, 69.1% in `cv_predictions` and 61.3% in `sensitivity_analysis` today. The plot area shrinks and the margins grow:

posterior_distribution

`cv_predictions` is the one case where something actually collides. The legend currently overlaps a single text object (the `date` axis label, 12.6px). With the `rect=[0, 0.07, 1, 1]` band gone it overlaps six, including four date tick labels, worst overlap 14.0px:

cv_predictions

Beyond that I found no new panel to panel collisions and nothing falling off the canvas, so removal is not broken, but it is a visible change to every figure for anyone on default rcParams rather than a cleanup. The guard keeps the current rendering when no style is active and still stops the override when one is, one line per site.

Independent of that choice: three of those calls are `plt.tight_layout()` rather than `fig.tight_layout()`, so they act on whatever figure happens to be current instead of the one being built.

Happy to apply the guard to the 10 sites in `mmm/plot.py` too. Better as a follow-up PR, or added to this one?

anevolbap:
Flipped to ready. Scope here stays on `decomposition.waterfall`. Happy to do the 10 `mmm/plot.py` sites as a follow-up, or to add them here if that is preferable.

juanitorduz (approved):

PR #2677, opened 2026-07-02
v1.0.0

williambdean:
## v1.0.0 — The stable foundation for Bayesian marketing analytics

After over four years of development (2,848 commits, 110 contributors, 42 releases),
this release marks the transition to a stable, well-covered API built on PyMC 6
and arviz 1.x (`pymc>=6.0.0`, `arviz>=1.2.0`, `pytensor>=3.0.0`).

### What is PyMC-Marketing?

Bayesian tools for marketing analytics — Media Mix Modeling (MMM), Customer
Lifetime Value (CLV), Customer Choice Analysis (CSA), Bass Diffusion, and more.
Developed and maintained by [PyMC Labs](https://www.pymc-labs.com).

### Major Milestones Since 0.1.0

- **MMM (2022)** — Adstock, saturation, budget optimization, lift test calibration,
  time-varying media, Fourier seasonality, custom splines
- **CLV (2023)** — Pareto/NBD, BetaGeo, Gamma-Gamma, ShiftedBetaGeo, MBG/NBD
- **Customer Choice (2024–2025)** — Multinomial/Nested Logit, Mixed Logit,
  MV-ITS, MaxDiff, Consideration Sets, Bayesian BLP
- **Bass Diffusion Model (2025)** — New product adoption forecasting
- **Infrastructure (2026)** — PyMC 6 migration, unified DataTree interface,
  interactive Plotly plots, MMMPlotSuite, MLflow autologging

### Core Upgrade

- **PyMC 6 / arviz 1.x**: All modeling migrated to `xr.DataTree`. A stable and
  well-covered API for every module.

### Breaking Changes

- **CLV**: Removed `ShiftedBetaGeoModelIndividual` and all pre-v1.0.0 API
  surface (deprecated `data` in `__init__`, `fit_method`, `_prior` suffix shims).
- **MMM**: Unified data input into `xr.Dataset` representation.
- **Budget optimizer**: Collapsed `default_constraints` into `constraints`.
- **Serialization**: Removed deprecated Legacy MMM shims.

### New Features

- **Bayesian BLP**: Structural demand estimation (`customer_choice.bayesian_blp`).
- **Bass model**: Plotting methods (diffusion/adoption curves).
- **MLflow**: Autolog support for BassModel.
- **Model builder**: Warns on unused `model_config` keys.

### Fixes

- CLV date normalization, dims restoration, metadata preservation
- Plot: `n_samples=0`, `hdi_prob` support
- Xarray `combine_by_coords` FutureWarning
[...]

PR #2795, opened 2026-07-31
feat(mmm): allow constant tensors in adstock sample_prior

shivamlalakiya:
## Description

This replaces #2304, which was auto-closed on 2026-07-28 when the `v1.0.0` base branch was deleted after #2677 was squash-merged into `main` — not by a review decision. GitHub disables the reopen button while the base ref is missing, so this is a clean PR from `main`, per the same guidance @ColtAllen gave on #2305. Review history from @ricardoV94, @williambdean and Cursor lives on #2304; the branch is the same one, rebased onto `main` with `git rebase --onto upstream/main <old-v1.0.0-head>`. No conflicts.

`Transformation.sample_prior` raised when a parameter was supplied as a constant tensor instead of a `Prior`, because `pm.sample_prior_predictive` had no free random variable to sample. Constants are now wrapped in `pm.DiracDelta` — the workaround @williambdean noted in #1749 — so PyMC produces a valid prior dataset carrying the constant value for every draw.

## Changes

* `sample_prior` wraps constant parameters in `pm.DiracDelta`. Constants are detected by mirroring `_create_distributions`' own dispatch — anything without a `create_variable` method registered no random variable — so nothing changes when every prior is a distribution, and the mixed constant/`Prior` case is covered.
* Second commit: `pm.DiracDelta` refuses `XTensorVariable` input, so a constant carrying dims raised `TypeError: To avoid subtle bugs, PyTensor forbids automatic conversion of XTensorVariable to TensorVariable`. It is now unwrapped with `.values` and its dims are passed through, so the prior keeps the named dim rather than an anonymous `adstock_alpha_dim_0`.
* Third commit (review follow-up): a dims-carrying constant passed without the matching `coords` now raises a `ValueError` naming the missing dims instead of PyMC's bare `KeyError`, matching the message quality of the `Prior` path.
* Tests in `tests/mmm/components/test_base.py` and `test_adstock.py` for the all-constant, mixed constant/`Prior`, dims-carrying-constant, and missing-coords cases.

## Scope against #1749

Where each input now lands, using the issue's `GeometricAdstock` example:

| `priors={"alpha": ...}` | `sample_prior` | `sample_curve` |
|---|---|---|
| `as_xtensor(np.array([0.5, 0.3, 0.2]), dims=("channel",))` | works, keeps `channel` | works |
| `pt.as_tensor_variable([0.5, 0.3, 0.2])` (the issue's snippet) | works, dim is `adstock_alpha_dim_0` | still raises `TypeError: non-scalar TensorVariable cannot be converted to XTensorVariable without dims` |
| `np.array([0.5, 0.3, 0.2])` | works, dim is `adstock_alpha_dim_0` | still raises `TypeError: Cannot convert [0.5 0.3 0.2] to XTensorType` |
| `[0.5, 0.3, 0.2]` (list) | works, dim is `adstock_alpha_dim_0` | still raises `TypeError: Cannot convert [0.5, 0.3, 0.2] to XTensorType` |
| scalar constant | works | works |

@williambdean — your [June comment](https://github.com/pymc-labs/pymc-marketing/issues/1749#issuecomment-4694026745) said the sampling not working was acceptable but evaluating the curve would be nice. Curve evaluation now works for a constant that declares its dims. Bare vectors — `pt.as_tensor_variable([...])`, a numpy array, or a list — still fail in `sample_curve` at `as_xtensor` inside `geometric_adstock`, because a bare constant carries no dim names to recover. Those last two are the likelier shapes in real user code, so the limitation is worth calling out explicitly: they succeed at `sample_prior` and then fail one call later. The fix would be to require or attach dims for constant parameters, which felt like a separate call to make.

So this is marked "related to" rather than "closes" #1749; happy to take the remaining piece here or in a follow-up, whichever you prefer.

## Related Issue

- [x] Related to #1749
- [x] Replaces #2304

## Checklist

- [x] If you are a pro: each commit corresponds to a [relevant logical change](https://www.pymc-marketing.io/en/latest/contributing/index.html)

----
📚 Documentation preview 📚: https://pymc-marketing--2795.org.readthedocs.build/en/2795/


daimon-pymclabs:
Thanks for picking this one up @shivamlalakiya, and nice work. This is close to mergeable. I went through it properly rather than just eyeballing the diff: merged your branch into current `main` (16 commits behind, merges clean), installed the stack and ran `tests/mmm/components/`, which gave 113 passed and 1 skipped. CI is green.

The approach is right. `pm.DiracDelta` is exactly the workaround suggested in #1749, and it addresses the three things raised on the superseded #2304: the "there's no prior being sampled" concern, the rejection of the `pm.Deterministic` route, and the catch that the mixed constant/`Prior` case silently dropped constants. That last one is genuinely fixed and covered by a test now, which is the part that makes this PR worth merging.

A few things I would tighten before it goes in. None of them are blockers on the core logic.

**1. The scope table understates the `sample_curve` limitation.**

You document that a bare `pt.as_tensor_variable([...])` still fails in `sample_curve`. The same failure hits plain numpy arrays and Python lists, which are documented as supported prior types in the class docstring and are far more likely than `pt.as_tensor_variable` in real user code:

```
priors={"alpha": np.array([.5, .3, .2])}
  -> sample_prior OK, dims ('chain', 'draw', 'adstock_alpha_dim_0')
  -> sample_curve TypeError: Cannot convert [0.5 0.3 0.2] to XTensorType
```

So the common path now succeeds at `sample_prior` and then dies one call later. Adding those two rows to the table would make the limitation discoverable to whoever hits it.

**2. The identity check is indirect.**

`if variables[parameter_name] is self.function_priors[parameter_name]` infers "this was a constant" from object identity, which relies on `create_variable`'s `except AttributeError: var = dist` returning the exact same object. Mirroring the dispatch directly reads better and drops the need for the `variables` round-trip entirely:

```python
prior = self.function_priors[parameter_name]
if hasattr(prior, "create_variable"):
    continue
```

Not a bug today, just coupling to an implementation detail that is one refactor away from silently skipping constants.

**3. Error quality on the recommended path.**

A dims-carrying constant with no `coords` passed raises a bare `KeyError: "Dimensions {'channel'} are unknown to the model..."`. The equivalent `Prior` path raises a `ValueError` with an actionable hint ("Add them at initialization time or use `model.add_coord`"). Same class of failure, so not a regression, but this is the path the PR recommends, so the worse message is a little unfortunate.

**4. Two nits.**

The docstring says "fixes #1749" while the PR body deliberately says "related to". Worth reconciling, and issue numbers in a public API docstring do render in the readthedocs output. Also, `import pytensor.tensor as pt` sits inside the test function in `test_adstock.py`; every other import in that file is module-level and `pt` is not used anywhere else.

One thing I checked and cleared, so you do not need to chase it: a length-mismatched constant under declared coords silently returns the constant's length with a positional index instead of the coord labels. That is PyMC accepting the mismatch at graph-build time (`pm.Normal` behaves the same), not something this PR introduced, so I would not hold the PR for it.

Sort out the table rows and the two nits, optionally swap the identity check, and this looks good to go from my side. Thanks again for the contribution and for the tests, they are the right ones.

shivamlalakiya:
Thanks for the thorough pass @daimon-pymclabs — all four points are addressed, and the branch is rebased onto current `main` (it was 18 behind; no conflicts).

**1. Scope table.** Confirmed both rows you raised, and added them plus the exact `TypeError` for each:

| input | `sample_prior` | `sample_curve` |
|---|---|---|
| `np.array([0.5, 0.3, 0.2])` | works, dim `adstock_alpha_dim_0` | `TypeError: Cannot convert [0.5 0.3 0.2] to XTensorType` |
| `[0.5, 0.3, 0.2]` | works, dim `adstock_alpha_dim_0` | `TypeError: Cannot convert [0.5, 0.3, 0.2] to XTensorType` |

You are right that these are the likelier shapes in real user code, so the PR body now says so explicitly rather than only naming `pt.as_tensor_variable`.

**2. Identity check.** Swapped for your version, which also drops the `variables` round-trip:

```python
constant = self.function_priors[parameter_name]
if hasattr(constant, "create_variable"):
    continue
```

`_create_distributions` is still called for its side effect of registering the distributions; only its return value is no longer needed. Worth noting the old check was in fact already load-bearing on a second implementation detail: `create_variable` can index the variable (`var = var[{idx_dim: idx}]`), and an indexed constant would no longer be the same object, so it would have been skipped. `sample_prior` passes no `idx`, so that path was unreachable — but it is one more reason the identity check was the wrong hook.

**3. Error quality.** Fixed rather than left, since it is on the path the PR recommends. A dims-carrying constant with missing coords now raises:

```
ValueError: Not all dims ('channel',) of the constant parameter 'alpha' are part of
the model coords. Missing: ['channel']. Pass them in the `coords` argument of `sample_prior`.
```

The hint points at `sample_prior`'s own `coords` argument rather than `model.add_coord`, since the model here is created internally and `add_coord` is not reachable for the caller. Covered by `test_sample_prior_constant_with_unknown_dims`.

**4. Nits.** Docstring now says "related to #1749" to match the PR body, and `import pytensor.tensor as pt` is module-level in `test_adstock.py`.

Also noted your cleared finding on the length-mismatched constant — agreed that is PyMC accepting the mismatch at graph-build time, so I left it alone.

Verification: `uv run pytest tests/mmm/components/` → 221 passed, 1 skipped. Full `uv run pytest tests/mmm` → 2591 passed, 3 skipped, 2 xfailed.

daimon-pymclabs (approved):
Thanks @shivamlalakiya, all four are addressed and the PR is in better shape than my review asked for. Re-read the diff at `5085767`: the identity check is gone, `_create_distributions()` is now called for its side effect only, the scope table carries the numpy and list rows with their exact `TypeError`, and the missing-coords case raises an actionable `ValueError` covered by `test_sample_prior_constant_with_unknown_dims`. `XTensorVariable` was already imported at module level, so the new `isinstance` branch has no import gap. Codecov shows all nine added lines covered, project coverage flat at 94.39%.

Your note on the indexed-constant path is a good catch and a better argument against the identity check than the one I made — the object-identity hook would have silently skipped a constant that `create_variable` had indexed, and only the fact that `sample_prior` passes no `idx` kept that unreachable. Worth having in the history.

One nit, not worth another round: `constant.values` in the `XTensorVariable` branch reads the value off the constant directly, so it only holds for constant-folded input. A symbolic `XTensorVariable` (say the output of an op rather than an `as_xtensor` of an array) would fail there rather than at a checked boundary. Every documented way of supplying a constant parameter produces the folded kind, so this isn't reachable from the public API today — flagging it only so the next person to widen the accepted types knows the branch assumes it.

Approving. LGTM.

PR #2796, opened 2026-07-31
fix(mmm): use original model when clone_model=False in sample_posterior_predictive

### LINKED ISSUE #2788: MMM.sample_posterior_predictive(clone_model=False) raises UnboundLocalError when no deterministics are frozen
### Summary

`MMM.sample_posterior_predictive(..., clone_model=False)` raises `UnboundLocalError` (a `NameError`) whenever the model has no frozen deterministics, because `model` is only bound inside the two conditional branches ([`pymc_marketing/mmm/mmm.py:2576-2584`](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/mmm.py#L2576-L2584)):

```python
if names := self.frozen_deterministics:
    # This always clone
    model = deterministics_to_flat(self.model, names=names)
elif clone_model:
    model = self.model.copy()
model = self._set_xarray_data(       # <- UnboundLocalError when neither branch ran
    dataset_xarray=dataset_xarray,
    model=model,
)
```

### Reproduction sketch

Any fitted `MMM` with `frozen_deterministics` empty (the default):

```python
mmm.sample_posterior_predictive(X, clone_model=False)
# UnboundLocalError: cannot access local variable 'model' where it is not associated with a value
```

### Suggested fix

```python
else:
    model = self.model
```

which is what `clone_model=False` advertises — sampling against the original model in place. A regression test with `clone_model=False` and no frozen deterministics would pin it.

### Context

Found while auditing library behaviour for the advanced funnel notebook (#2769); the notebook itself always uses the default `clone_model=True`, so it is unaffected.

shivamlalakiya:
## Description

`MMM.sample_posterior_predictive(..., clone_model=False)` raised `UnboundLocalError` whenever the model had no frozen deterministics (the default), because `model` was only bound inside the `if`/`elif` branches:

```python
if names := self.frozen_deterministics:
    # This always clone
    model = deterministics_to_flat(self.model, names=names)
elif clone_model:
    model = self.model.copy()
model = self._set_xarray_data(...)  # <- UnboundLocalError when neither branch ran
```

This adds the missing `else` branch, which uses the original model in place — what `clone_model=False` advertises:

```python
else:
    model = self.model
```

Also adds a regression test, `test_sample_posterior_predictive_clone_model`, parametrized over `clone_model=True/False`. It asserts `frozen_deterministics == []` (so the `else` branch is the one exercised), that the posterior predictive is drawn on the new dates, and that the original model's `date` coord is mutated only when `clone_model=False`. The test fails with `UnboundLocalError` on `main` and passes with the fix.

## Related Issue
- [x] Closes #2788
- [ ] Related to #

## Checklist

### Note for reviewers

`_set_xarray_data` has a `clone_model: bool = True` parameter (`pymc_marketing/mmm/mmm.py:2475`) that is never read in the body and never passed by any caller. Left untouched here to keep this fix minimal — happy to remove it in a separate PR if you want.

----
📚 Documentation preview 📚: https://pymc-marketing--2796.org.readthedocs.build/en/2796/

### DIFF
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -2578,6 +2578,8 @@ class MMM(RegressionModelBuilder):
             model = deterministics_to_flat(self.model, names=names)
         elif clone_model:
             model = self.model.copy()
+        else:
+            model = self.model
         model = self._set_xarray_data(
             dataset_xarray=dataset_xarray,
             model=model,
--- a/tests/mmm/test_mmm.py
+++ b/tests/mmm/test_mmm.py
@@ -1415,6 +1415,48 @@ def test_sample_posterior_predictive_same_data(single_dim_data, mock_pymc_sample
     )
 
 
+@pytest.mark.parametrize("clone_model", [True, False])
+def test_sample_posterior_predictive_clone_model(
+    single_dim_data, mock_pymc_sample, clone_model
+):
+    """
+    Test that sampling from the posterior predictive works with both clone_model
+    values when no deterministics are frozen, and that clone_model=False sets the
+    new data on the original model in place.
+    """
+    X, y = single_dim_data
+    X_train = X.iloc[:-5]
+    X_new = X.iloc[-5:]
+    y_train = y.iloc[:-5]
+
+    mmm = MMM(
+        date_column="date",
+        target_column="target",
+        channel_columns=["channel_1", "channel_2", "channel_3"],
+        adstock=GeometricAdstock(l_max=2),
+        saturation=LogisticSaturation(),
+    )
+
+    mmm.build_model(X_train, y_train)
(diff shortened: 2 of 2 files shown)


shivamlalakiya:
CI is green on this branch: 56 checks passed, 0 failed (the two skipped ones are the PyPI upload jobs, which don't run on PRs).

Locally, the full `tests/mmm` suite also passes: `2517 passed, 3 skipped, 2 xfailed`. Reverting just the `mmm.py` hunk makes the new `test_sample_posterior_predictive_clone_model[False]` case fail with the reported `UnboundLocalError`, so the test pins the regression.

Ready for review whenever you get a chance.

juanitorduz (approved):
thank you @shivamlalakiya

PR #2797, opened 2026-07-31
fix(mmm): repair and test diversification_ratio (closes #1388)

### LINKED ISSUE #1388: `diversification_ratio` not tested
The helper is not tested anywhere. It's unclear how a user was supposed to use as part of an optimization cost because it requires 2D samples, whereas the optimizer was always giving 1D samples before #1357 

https://github.com/pymc-labs/pymc-marketing/blob/32f44c78b40e2630661b23d6ffe6b696da0a5c72/pymc_marketing/mmm/utility.py#L504-L551

CC @cetagostini

shivamlalakiya:
## Description

Closes #1388.

The issue reports that `diversification_ratio` is not tested directly. Writing that test turned out to be enough to show that the function raised on **every** call, for any input:

```python
samples = as_xtensor(rng.normal(size=(50, 3)), dims=("sample", "channel"))
budgets = as_xtensor(np.array([100.0, 200.0, 300.0]), dims=("channel",))
diversification_ratio(samples, budgets)
# ValueError: Function expected samples to be a 1D tensor variable. Got 2 dimensions.
```

Two independent bugs:

1. It called `_check_samples_dimensionality(samples)`, which enforces **1D** samples, while the function's own maths (and docstring) require **2D** `(sample, asset)` samples. So 2D input was rejected outright.
2. A misplaced parenthesis in the `w'Σw` computation:

   ```python
   portfolio_var = ptx.dot(
   )
   ```

   `ptx.dot(cov_matrix, weights)` was passed as a *second positional argument to* `.rename(...)` rather than as the second argument of the outer `ptx.dot`. So even with 1D input the function failed, with a confusing `Cannot rename channel to channel': channel not in ('sample',)`.

## Changes

- `_check_samples_dimensionality` takes an `ndim` argument, defaulting to `1`. All existing callers (and the two uses in `tests/mmm/test_budget_optimizer.py` and `mmm_budget_allocation_example.ipynb`) are unaffected; the error message for `ndim=1` is byte-identical.
- `diversification_ratio` asks for `ndim=2` and computes `w'Σw` correctly.
- Docstring: states that `samples` needs a `"sample"` dim plus an asset dim matching the `budgets` dim, notes that — unlike the other utility functions, which consume samples of the *total* response — this one needs the response broken down per asset (e.g. `channel_contribution` summed over `date`), and adds a runnable example. This is the second half of the issue ("it's unclear how a user was supposed to use [it] as part of an optimization cost").

## Tests

New tests in `tests/mmm/test_utility.py`:

- `test_diversification_ratio_matches_numpy` — parametrized over three datasets, compared against a NumPy reference `(w · σ) / sqrt(w'Σw)` at `rtol=1e-8`.
- `test_diversification_ratio_perfectly_correlated_assets` — DR is exactly `1.0` when all assets are scalar multiples of a common factor.
- `test_diversification_ratio_rewards_uncorrelated_assets` — uncorrelated assets score above perfectly correlated ones at equal weights.
- `test_diversification_ratio_requires_2d_samples` — 1D samples raise a clear `ValueError` instead of the old rename error.

```
$ uv run pytest tests/mmm/test_utility.py -q
52 passed in 5.97s

$ uv run pytest tests/mmm/test_budget_optimizer.py -q
25 passed, 22 warnings in 44.42s

$ uv run pytest tests/mmm/test_budget_optimizer_mmm.py -q
36 passed, 45 warnings in 91.89s

$ uv run pytest tests/ -q
9 failed, 3681 passed, 46 skipped, 2 xfailed, 850 warnings in 1289.60s
```

The 9 failures are all in `tests/test_mlflow.py` (logged-artifact set mismatch) and are pre-existing — they reproduce identically on an unmodified checkout of this branch's base commit.

`pre-commit run --files pymc_marketing/mmm/utility.py tests/mmm/test_utility.py` passes.

## Notes / out of scope

`BudgetOptimizer` still cannot drive this utility out of the box: the default `response_variable`, `total_media_contribution_original_scale`, is 1D over `sample`, and there is no shipped per-channel aggregated deterministic to point at. Using `diversification_ratio` as an optimization cost requires the user to register a 2D `(sample, channel)` response themselves. Happy to follow up with such a deterministic (or an integration test) in a separate PR if maintainers want it — it felt like a larger API decision than this fix.

----
📚 Documentation preview 📚: https://pymc-marketing--2797.org.readthedocs.build/en/2797/


pymc_marketing/mmm/utility.py
```diff
@@ -586,14 +593,28 @@ def diversification_ratio(
...
+
     References
     ----------
     - Choueifaty, Y., & Coignard, Y. (2008). Toward Maximum Diversification. *Journal of Portfolio Management*.
     - Meucci, A. (2009). Managing Diversification. *Risk*, 22(5), 74-79.
     """
     samples = as_xtensor(samples)
     budgets = as_xtensor(budgets)
-    samples = _check_samples_dimensionality(samples)
+    samples = _check_samples_dimensionality(samples, ndim=2)
```
daimon-pymclabs, line 621: The new guard checks rank but not dim names, so the exact confusing error this PR sets out to eliminate can still fire.

The PR body cites `Cannot rename channel to channel': channel not in ('sample',)` as the symptom. With budgets whose dim doesn't match the samples asset dim, you get the same class of message:

```
ValueError: Cannot rename asset to asset': asset not in ('sample', 'channel')
```

And 2D samples with a misnamed sample dim give:

```
ValueError: Reduced dims ('sample',) not found in array dimensions ('draw', 'channel')
```

Both confirmed by execution on this branch. The function needs a `"sample"` dim plus an asset dim matching `budgets`, and `ndim == 2` verifies neither. The docstring addition at line 583 ("Its dim must match the asset dim of ``samples``") documents the contract but nothing enforces it.

Checking `"sample" in samples.dims` and that the budgets dim is present in `samples.dims` would cover both, with a message that names the offending dim. `test_diversification_ratio_requires_2d_samples` only exercises the rank path, so neither case currently has a test.

daimon-pymclabs (commented):
Verified the core fix by execution in an isolated pytensor env, not just by reading.

Both changes are correct. The `ndim` parameter lets `diversification_ratio` ask for 2D instead of being rejected by the hard 1D check, and the misplaced closing paren in the `w'Σw` line is fixed. I ran the patched function against an independent NumPy reference `(w·σ)/sqrt(w'Σw)` and got `1.5738396618306048` from both; the perfectly-correlated single-factor case returns exactly `1.0`. The `ndim=1` error string is byte-identical to the old one, so the existing call sites and the notebook are unaffected.

Three inline notes below. The open design question is the one already flagged: `BudgetOptimizer` still can't drive this utility, since the default `total_media_contribution_original_scale` response is 1D over `sample` and no per-channel aggregated deterministic ships. That makes the docstring example the only currently working usage path, and it decides whether this stays a fix or becomes a fix plus a new deterministic. Reasonable to defer, but worth an explicit call.

pymc_marketing/mmm/utility.py
```diff
@@ -586,14 +593,28 @@ def diversification_ratio(
...
     weights = budgets / budgets.sum()
     individual_volatilities = samples.std(dim="sample", ddof=1)
```
daimon-pymclabs: Optional, and pre-existing rather than introduced here: `samples.std(dim="sample", ddof=1)` is mathematically redundant with `_covariance_matrix` below, whose diagonal is those same `ddof=1` variances. Since this is an optimizer cost function it recomputes on every iteration. Taking the volatilities from `ptx.math.sqrt(cov_matrix.diagonal(...))` would drop one pass over the samples. Fine to leave if you'd rather keep the two terms reading independently, but this function is being repaired anyway.

tests/mmm/test_utility.py
```diff
@@ -399,6 +400,82 @@ def test_general_functions(samples, budgets, func):
         pytest.fail(f"Function {func.__name__} raised an unexpected exception: {e!s}")
```
daimon-pymclabs, line 404: Worth noting explicitly: `diversification_ratio` stays out of this `test_general_functions` smoke parametrization, and it can never join it, since the harness hardcodes `dims=("sample",)` and `dims=("budget",)` while this function needs 2D samples and a budgets dim matching the asset dim.

That makes it a permanent carve-out from the shared utility-function contract, so regressions in that contract won't be caught for this function. The new dedicated tests are good coverage of the math; this is about the shared-contract gap. A short comment near the parametrization recording why it's excluded would keep the next reader from assuming it was an oversight.

shivamlalakiya:
@daimon-pymclabs thanks for executing the fix rather than only reading it. All three notes plus the open design question are addressed in 6a9fb19, and the branch is rebased onto `main` (`d402988`).

**1. Dim names, not just rank ([comment](https://github.com/pymc-labs/pymc-marketing/pull/2797#discussion_r3702611835))** — you're right that `ndim == 2` verifies neither half of the contract. Both cases now fail with a message that names the offending dim:

```python
>>> diversification_ratio(as_xtensor(s, dims=("draw", "channel")), as_xtensor(b, dims=("channel",)))
ValueError: Function expected samples to have a "sample" dim. Got dims ('draw', 'channel').

>>> diversification_ratio(as_xtensor(s, dims=("sample", "channel")), as_xtensor(b, dims=("asset",)))
ValueError: Function expected samples to have the budgets dim "asset". Got dims ('sample', 'channel').
```

`test_diversification_ratio_requires_sample_dim` and `test_diversification_ratio_requires_matching_budgets_dim` cover them. I left `budgets` itself unguarded beyond the `[asset_dim] = budgets.dims` unpack, since no other utility function validates `budgets` and that felt like separate scope — say the word if you'd rather it raise there too.

**2. Redundant volatility pass ([comment](https://github.com/pymc-labs/pymc-marketing/pull/2797#discussion_r3702611838))** — taken. `individual_volatilities` now comes off the covariance diagonal:

```python
cov_matrix = _covariance_matrix(samples, asset_dim=asset_dim)
individual_volatilities = ptx.math.sqrt(
    as_xtensor(pt.diagonal(cov_matrix.values), dims=(asset_dim,))
)
```

`xtensor` has no `diagonal`, so this uses the same `.values` escape hatch `_compute_quantile` already uses a few hundred lines up in this file. Result is unchanged: `1.5738396618306048` against your NumPy reference, and the perfectly-correlated case still returns exactly `1.0`.

**3. Permanent carve-out from `test_general_functions` ([comment](https://github.com/pymc-labs/pymc-marketing/pull/2797#discussion_r3702611841))** — comment added above the parametrization recording that the exclusion is structural (the harness hardcodes 1D `("sample",)` samples and a `("budget",)` dim) and pointing at the dedicated tests.

**On the open question: `BudgetOptimizer` can drive this today, and no new deterministic is needed.** I checked instead of deferring. `total_media_contribution_original_scale` is a full scalar sum per draw (`link.py:227`, `link.py:299`), so it can't work — but `extract_response_distribution` returns whatever dims the named variable has, so a per-channel deterministic registered on the model gives exactly `("sample", "channel")`:

```python
with mmm.model:
    pmd.Deterministic(
        "channel_contribution_total", mmm.model["channel_contribution"].sum(dim="date")
    )

BudgetOptimizer(
    model=wrapper,
    num_periods=4,
    response_variable="channel_contribution_total",
    utility_function=diversification_ratio,
).allocate_budget(total_budget=100)
```

That converges, and gradients flow through the whole graph. `test_allocate_budget_with_diversification_ratio` in `tests/mmm/test_budget_optimizer.py` asserts the response dims, `res.success`, budget conservation, and `DR >= 1`. The `samples` docstring now spells out this requirement and states explicitly that the default response variable will not work, so the docstring example is no longer the only working path.

Verification: `uv run pytest tests/mmm/test_utility.py tests/mmm/test_budget_optimizer.py` gives 80 passed, and `pre-commit` is clean on the touched files.

juanitorduz:
thank you @shivamlalakiya !

juanitorduz:
@shivamlalakiya can you please resolve the conflicts so that we can merge this one? If you need help I could do it :)

shivamlalakiya:
@juanitorduz happy to, but I don't think this one should be rebased: #2995 merged into `main` earlier today (8cf77c84) and fixes the same two bugs this PR does, so there's almost nothing left here.

The conflict is my hunk against theirs in the same lines of `diversification_ratio`. Both branches replace the 1D `_check_samples_dimensionality` guard with a 2D check plus `"sample"`/budgets-dim validation, both close the misplaced parenthesis in `w'Σw`, both add Notes and an Examples recipe to the docstring, and both add a `BudgetOptimizer` integration test. Their test coverage matches or exceeds mine (`matches_numpy`, the `sqrt(n)` property, the DR ≥ 1 oracle, parametrised bad-dims, and `test_diversification_ratio_through_optimizer`). Resolving the conflict would mean re-litigating code that just landed, for no behavioural gain.

The only thing on this branch that isn't on `main` is a guard rejecting a `budgets` dim literally named `"sample"`. On `main` that input still raises, just with a less helpful message (`ValueError: Size of dim 'sample' does not match`). If you'd like that message improved I'm glad to open a small follow-up PR against `main` for it, but it didn't seem worth keeping this branch alive.

Closing in favour of #2995. Thanks for the reviews on this one, and #1388 is covered by what merged.

Issue #277, opened 2023-05-03
Investigate survival model alternatives to existing lifetime models


ricardoV94:
Could be more flexible:
  1. Easier to integrate covariates
  2. Easier to customize (priors, likelihoods, arbitrary structure)
  3. More similar approach to MMM (have on robust model that can be customized)

ricardoV94:
ya incorporate!

ColtAllen:
I'm planning to work on https://github.com/pymc-labs/pymc-marketing/issues/134 soon. The methodology for adding time-invariant covariates is rooted in survival analysis:

https://www.brucehardie.com/notes/019/time_invariant_covariates.pdf

Issue #273, opened 2023-05-03
Advertise GPU sampling in docs


ricardoV94:

wd60622:
Maybe this can be called out explicitly in https://www.pymc-marketing.io/en/stable/notebooks/general/other_nuts_samplers.html
Thoughts @juanitorduz? Or maybe this is completed?

wd60622:
Some good items here: 

- [ ] Add new or to existing notebook
- [ ] Add to the features list in the README / index

Any thing else @cluhmann?

cluhmann:
I think a small notebook demonstrating one or more (maybe it makes sense to demonstrate all) of the GPU-aware samplers and maybe some of the diagnostics one performs to make sure the GPU is registered/available? If the latter is omitted, it could probably just be some raw markdown I guess.

PR #2800, opened 2026-08-01
docs(mmm): tutorial on long-term brand effects with brand metrics and a Bayesian VARX

juanitorduz:
## Description

NB Review https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2800/changes/?expand=docs%2Fsource%2Fnotebooks%2Fmmm%2Fmmm_brand_metrics_long_term.ipynb

This PR adds a new example notebook, `docs/source/notebooks/mmm/mmm_brand_metrics_long_term.ipynb`, a tutorial on measuring the **long-term effect of brand marketing** by integrating brand-tracking metrics with an MMM through a two-stage Bayesian workflow:

1. **Stage 1**: an `MMM` with a time-varying intercept (HSGP, Matern 5/2) and a `TruncatedNormal(lower=0)` likelihood recovers the short-term channel contributions and an explicit *dynamic baseline*, after a prior predictive check against the observed sales.
2. **Stage 2**: a Bayesian VARX (via [`impulso`](https://github.com/thomaspinder/impulso), already a docs dependency) models {baseline, awareness, consideration} with brand spend exogenous. Its cumulative dynamic multiplier quantifies the long-run transmission of brand spend into base sales, with full posterior uncertainty.

The point of the pairing is that the total return on brand is the short-term effect, which is genuinely zero here, plus a 26-week multiplier that a standard MMM cannot see at all.

## What the notebook does

- **Everything is simulated, so everything is validated.** The data come from PyMC generative models following the `mmm_data_generator.ipynb` pattern: a spend model, a structural funnel (`pytensor.scan` recursions for awareness -> consideration -> base sales), and a generator `MMM` whose parameters, including the whole time-varying `intercept_contribution` path, are fixed with `pymc.do`. The true 26-week multiplier is 5.15 by construction.
- **The naive MMM is the foil.** With brand as a third channel it recovers a direct brand ROI of about zero, which is correct and useless. The two-stage headline recovers a median of 4.71 with a 94% HDI of [2.33, 6.31].
- **Stage-1 uncertainty is propagated rather than plugged in.** The headline is a mixed posterior that refits the VARX across 40 Stage-1 baseline draws. The single plug-in fit is kept as the over-confident comparison: its interval is half the width of the honest one (1.95 against 3.98).
- **Oracle refits localize the error.** Swapping in the true baseline, and then the latent surveys as well, shows the coefficient distortion in the plug-in fit is almost entirely Stage-1 smoothing error rather than survey noise or the VAR specification, which is exactly the generating equation. A fourth row refits with the default Minnesota prior to show what over-shrinking the funnel cross-lags costs.
- **Diagnostics get read, not just printed.** The fitted system's spectral radius is checked per posterior draw, lag-order criteria are run on both the estimated and the clean frame, and the stationarity tests (ADF, KPSS, Johansen, each with how to read it) sit in an appendix so the main flow stays on the measurement story.
- Registered in the gallery under MMM > Causal Inference.

## Notes for reviewers

- The dynamic multiplier is built from impulso's public `compute_ma_phi` and `lag_matrices` exports rather than `FittedVAR.dynamic_multiplier`, which is incompatible with ArviZ >= 1.0 (upstream: [thomaspinder/Impulso#273](https://github.com/thomaspinder/Impulso/issues/273)).
- The notebook runs end to end with real sampling in roughly six minutes, dominated by the 40 VARX refits in the mixed-uncertainty cell. Notebook CI mocks `mmm/**`, so that path stays fast.
- Known limitation: the stability ladder pools `keep_per_fit` draws per refit, so the 10-draw rung is a 1000-sample pool against 4000 at the 40-draw rung. Its tail quantiles therefore mix baseline-mixing with Monte Carlo error from the smaller pools.
- References: Cain (2022, IJRM), R. O'Sullivan's "Capturing the long-term causal effect of brand marketing", 1749.io's "Measuring Marketing Effectiveness Over the Long-Term", and Dekimpe and Hanssens (1999).

## Related Issue

- [ ] Closes #
- [x] Related to # (none; new example notebook)

## Checklist

- [x] Included tests that prove the fix is effective or that the new feature works (the notebook runs end to end with real sampling and validates recovery against simulated ground truth; the CI mock runner passes)


juanitorduz:
https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2800/changes/?expand=docs%2Fsource%2Fnotebooks%2Fmmm%2Fmmm_brand_metrics_long_term.ipynb

juanitorduz:
I think to use `FittedVAR.dynamic_multiplier()` we need https://github.com/thomaspinder/Impulso/pull/277 🤔

juanitorduz:
Thanks a lot for the review, @thomaspinder: genuinely useful!

Here some auto generated notes on the changes:

**Granger causality for the funnel ordering + ROPE** — done. There is a new "Granger Causality and the Funnel Ordering" subsection inside the validation section, running `fitted.granger_causality(...)` over all six ordered pairs of the endogenous block against the true `A` matrix. Since the generator has exactly two non-zero cross-lags (awareness → consideration and consideration → baseline) it doubles as a recovery check. The ROPE is set explicitly at 0.05 sd of the effect per sd of the cause, and the prose states what `p_rope` is not — following your `GrangerCausalityResult` docstring, which is the clearest statement of that distinction I have read anywhere.

One note: I build the HDI column by hand from `result.norm_draws` rather than calling `result.summary()`, because `summary()` and `hdi()` both pass `hdi_prob=` to `az.hdi`, which ArviZ 1.x renamed to `prob=`. Same root cause as #273.

**`integration_order` for the stationarity appendix** — done, and it is a real simplification. The separate `adf_test` / `kpss_test` cells are gone, replaced by one `integration_order(varx_df, variables=endog_vars)` call; the two-by-two ADF/KPSS verdict grid I was explaining by hand is now just the `joint_status` column, and `d_max` is the number the levels-VAR decision actually turns on. Johansen stays, since integration order does not speak to cointegration.

**`FittedVAR.dynamic_multiplier()` and the VAR posterior predictive check** — I would like both, and both are blocked on #277. This repo is on ArviZ 1.2 / PyMC 6 / Python 3.14, where:

```
fitted.dynamic_multiplier(horizon=5)
  -> TypeError: DataTree.__init__() got an unexpected keyword argument 'posterior_predictive'
fitted.posterior_predictive(seed=0)
  -> TypeError: DataTree.__init__() got an unexpected keyword argument 'posterior_predictive'
```

Both die on the same line pattern, `az.InferenceData(posterior_predictive=...)` (`fitted.py:254` and `:608`), since ArviZ 1.x aliases `InferenceData` to `xarray.DataTree`. I checked 0.0.11 hoping the fix had ridden along, but that line is unchanged there — so the hand-rolled `posterior_dynamic_multiplier` (built on your public `compute_ma_phi` / `lag_matrices`) stays for now, and there is no VAR PPC cell yet.

As soon as #277 lands in a release I will swap the helper body for `fitted.dynamic_multiplier(...)`, add the posterior predictive check with the coverage diagnostic from your how-to, and simplify the Granger table to `result.summary()`. Happy to hold this PR for that if you would rather the notebook ship using the public API throughout — just say the word. (Incidentally, the named-index fix in 0.0.11 is load-bearing for the PPC here: `varx_df`'s index is named `date`.

juanitorduz:
Unblocked by [impulso v0.0.12](https://github.com/thomaspinder/Impulso/releases/tag/v0.0.12) — thanks for turning #277 around so quickly, @thomaspinder. Pushed in 46207144, with the floor raised to `impulso>=0.0.12` (0.0.11 raises `TypeError` on all three paths below). All three of your remaining suggestions are in.

**`FittedVAR.dynamic_multiplier`.** The 25-line `posterior_dynamic_multiplier` helper over `compute_ma_phi` / `lag_matrices` is gone, replaced by a two-liner that calls the public method and reaches into `result.idata["posterior_predictive"]["dynamic_multiplier"]` for the draws. I kept a named wrapper rather than inlining because three call sites need the raw posterior rather than a `median()`/`hdi()` summary — the plug-in fit, the oracle refits, and the 40 baseline-draw refits that get pooled into the mixed posterior. `compute_ma_phi` drops out of the imports; `lag_matrices` stays, since the spectral-radius check still needs the companion matrix and there is no public helper for that.

The draws come back with exactly the dims and coords the hand-rolled version built, so this is a numerically exact swap. I re-executed the whole notebook and diffed every output cell against the previous run: **39 common cells, 1 differing** — the watermark, `impulso 0.0.11 -> 0.0.12`. Granger table, spectral radius, oracle comparison, ROI ladder, headline (`median 4.71, 94% HDI [2.33, 6.31]`) all identical to the digit. Also gone: the `arviz.MigrationWarning` that used to be the first output of the imports cell.

**Posterior predictive check.** New `### Posterior Predictive Check` subsection sitting between the convergence diagnostics and the `B_exog` read, using `fitted.posterior_predictive(seed=...)` and the coverage recipe from your predictive-checks how-to, broken out per series:

```
95% predictive band covers 96.3% of observations
  baseline: 97.8%
  awareness: 95.6%
  consideration: 95.6%
```

The figure is a three-panel time-series overlay (central 95% band vs the observed path, with the flight windows shaded) rather than `plot_ppc_dist`, since the three series live on very different scales. The prose is explicit that at 181 dates the binomial standard error at nominal 95% is ~1.6pp, so none of those splits is distinguishable from nominal — and that the observations are autocorrelated anyway, which makes even that generous.

What I found worth writing down is what this check *cannot* see. Two cells later the baseline equation turns out badly distorted (own lag pushed to ~0.95, consideration loading collapsed to zero, a spurious awareness loading in its place), and coverage comes out fine regardless: one-step-ahead replication conditioned on observed lags is blind to a reallocation of predictive power among coefficients. So the section says plainly that a model can replicate its own sample and still be wrong about where the effect travels — which is the same warning your `p_rope` docstring delivers, arriving through a third door. It also notes this is the only check in that whole section needing no ground truth, which is what makes it the one a practitioner can actually run.

**Granger table via `summary()`.** The hand-built HDI over `norm_draws` and the comment explaining why it existed (#273) are both deleted; each row is now `result.summary(prob=0.94).loc["norm"]`. I added a sentence noting `summary()` reports per-lag rows plus the aggregating `norm` row, and that at a single lag they carry the same information modulo sign. Values unchanged, as expected now that `hdi_bounds` normalises the keyword.

The `_arviz_compat` seam is doing exactly what it says on the tin from the consumer side — nothing in the notebook has to know which ArviZ line is installed.

juanitorduz:
@thomaspinder I think I am ready for another review round :)

thomaspinder (approved):
With the recent changes, this is looking great! Approved, with excitement

juanitorduz:
Thank you for your review @thomaspinder ! 

Exited to see impulso in action!

PR #2801, opened 2026-08-01
refactor(budget-optimizer): allocate_budget returns BudgetOptimizationResult

PabloRoque:
## Description

Stacked on #2654 (base: `refactor/budget-optimizer-pm-model`). First of two follow-up refactors preparing the ground for #2621.

`allocate_budget` now returns a `BudgetOptimizationResult` instead of a variable-arity tuple (2-tuple, or 3-tuple when `callback=True`):

```python
@dataclass
class BudgetOptimizationResult:
    budgets: DataArray                     # optimized allocation, monetary units
    scipy_result: OptimizeResult           # raw scipy diagnostics
    optimized_vars: dict[str, DataArray]   # {} until #2621
    callback_info: list[OptimizationIterationInfo] | None  # was the 3rd tuple element
```

**Backward compatible:** iterating the result yields `(budgets, scipy_result)`, so every existing `optimal, res = optimizer.allocate_budget(...)` call site works unchanged. `callback=True` no longer changes the return arity — the per-iteration diagnostics land in `result.callback_info`, now typed via `TypedDict`s (`OptimizationIterationInfo`, `ConstraintIterationInfo`).

Also:
- **Remove the `mu_effects` field.** Its only behaviour was raising `NotImplementedError` when an effect had `replace_for_optimization`, and #2621 fills that gap with a different mechanism (`optimizable_vars`) under which the field is dead state. A `DeprecationWarning` is emitted (and the argument ignored) if a caller still passes it. `MMM.budget_optimizer` no longer forwards it.
- **Collapse `_compiled_functions`** — a dict keyed by the fixed `utility_function` field that could only ever hold one entry — into a plain `_objective_and_grad` attribute.

### Motivation

`optimized_vars` gives #2621's co-optimized lever values a typed home instead of attaching attributes onto scipy's `OptimizeResult`, and the fixed return arity removes the `callback=` wart before a third consumer appears. Deleting `mu_effects` now means the placeholder doesn't outlive the design it was placed for.

## Related Issue

- Related to #2654 and #2621 (this PR sits between them)

## Checklist

- [x] Code passes all tests: `test_budget_optimizer.py` (27), `test_budget_optimizer_mmm.py` + `test_cost_per_unit.py` (80), all passing locally
- [x] New tests: result-object contract (attributes + iteration), fixed arity with/without callback, `mu_effects` deprecation warning
- [x] pre-commit (ruff, mypy) clean

## Modules affected

[...]

PR #2802, opened 2026-08-01
refactor(budget-optimizer): OptimizationVariable abstraction for the decision vector

PabloRoque:
## Description

**Stacked on #2801** (which stacks on #2654). GitHub cannot target a fork branch as base, so this PR targets `refactor/budget-optimizer-pm-model` and *includes* #2801's commit — review only the top two commits:

- `d9063460` — the `OptimizationVariable` abstraction (behaviour-neutral code move)
- `483793aa` — the zero-gradient warning (separable; drop it if contentious)

### The problem

The forward map (flat decision vector → model tensors: mask scatter, `channel_scales`, temporal distribution, `cost_per_unit`, adstock zero-padding) and the inverse map (solution → labelled `DataArray`) were written independently, ~150 lines apart, agreeing only by convention. Consequences: `x0` was effectively unusable (nobody can hand-construct boolean-mask-flattening order over transposed dims for a `(channel, geo)` budget), and any second decision variable (#2621's levers) must fork the flat-layout/bounds/x0/unpacking path in four places.

### Commit 1 — `OptimizationVariable` abstraction (behaviour-neutral)

New module `pymc_marketing/mmm/optimization_variables.py`:

- **`OptimizationVariable`** (ABC): one named, contiguous segment of the flat vector — `to_model(z)` (forward), `unpack(x)` / `pack(da)` (inverse pair, stated once), `default_x0`, `default_bounds`.
- **`MediaVariable`**: a faithful relocation of the existing media path. No arithmetic changed.
- **`OptimizationVariables`**: owns the flat symbolic input, the slice layout (tiles `[0, size)` exactly), and the substitution dict for a single joint `do()` call.

`BudgetOptimizer` shrinks by ~120 lines; construction is now `do(freeze_dims_and_data(...), self._variables.substitutions())`, and bounds/x0/result-unpacking each go through the same layout. `_budgets_flat`/`_budgets` remain as attributes (constraints and downstream code reach into them).

**`x0` now accepts a labelled `DataArray` or `{name: DataArray}` dict** — warm-starting from a previous solution or a current spend plan finally works for multi-dimensional budgets.

**Behaviour-neutrality evidence:** all 127 pre-existing tests in `test_budget_optimizer.py`, `test_budget_optimizer_mmm.py`, and `test_cost_per_unit.py` pass unchanged, plus 16 new property tests (`pack(unpack(x)) == x`, `unpack(pack(da)) == da` over 1-D/2-D shapes with full/partial masks and shuffled coords; slice tiling; error paths). One existing test's expected error message was updated (`test_allocate_budget_custom_minimize_args` asserted an incidental numpy indexing message from the old hand-written unpack; the layout now raises a deliberate shape-validation error).

One pre-existing oddity preserved deliberately: when `budget_distribution_over_period` is set, the distribution branch rebuilds from the unscaled flat variable, so `channel_scales` is effectively ignored on that path. `MediaVariable` replicates this exactly; flagging it here for a separate fix if it is indeed a bug.

### Commit 2 — zero-gradient warning (separable)

With a configurable `response_variable`, no decision variable — media included — is guaranteed to reach the objective; the solver would silently return the seed as the "optimum". `allocate_budget` now evaluates the compiled gradient once at `x0`; any variable whose whole slice is exactly zero (excluding coordinates pinned by `low == high` bounds) is re-probed at a perturbed in-bounds point and reported by name if still zero. Pinned coordinates are skipped so deliberately fixed cells are not misreported as inert. Cost: at most two extra function evaluations, the second only when a variable's slice is exactly zero.

### What this does to #2621

Its rebase becomes: implement the lever as an `OptimizationVariable` subclass, translate `OptimizableMuEffect` → variable in `MMM.budget_optimizer`, and fill `result.optimized_vars`. The `_var_slices` bookkeeping, the media-head/lever-tail special-casing in four places, and the lever-only gradient guards all dissolve into the shared layout.

## Related Issue

- Related to #2654, #2801, #2621 (sequence: #2654 → #2801 → this → #2621 rebase)

## Checklist

- [x] All optimizer suites pass locally: 127 pre-existing + 20 new (property tests, labelled-x0, probe)
- [x] pre-commit (ruff, mypy) clean

## Modules affected

- `mmm.optimization_variables` (new)
- `mmm.budget_optimizer`

## Type of change

- [x] Refactor (behaviour-neutral) + one new warning

----
📚 Documentation preview 📚: https://pymc-marketing--2802.org.readthedocs.build/en/2802/


daimon-pymclabs:
### Blocker: the inverse map disagrees with the forward map

`MediaVariable` takes its `dims`/`coords` from the mask (`optimization_variables.py:192-193`), but the forward map is purely positional: `scattered()` does `budgets_zeros[self._bool_mask].set(z)` into the model's tensor layout. Step 4 only transposes the mask (`budget_optimizer.py:1413`) and never reindexes it to `_budget_coords`, so when a user passes a `budgets_to_optimize` mask whose coord order differs from the model's, the two maps drift and `unpack()` labels each value with the wrong channel.

Reproduced on the `mmm_wrapper` fixture with a mask coord order of `[channel_2, channel_1]` against a model ordered `[channel_1, channel_2]`, bounds capping `channel_1` at 5 and `channel_2` at 95:

```
forward-map channel_data (first row): [1000.  1.]
=> flat entry 0 (value 1000) lands in model column 0, which the MODEL calls channel_1
=> unpack() labels the value 1000 as channel: channel_2
```

End to end, base and this branch produce the identical flat solution (`4.99999999999935`, `95.0`); only the labels flip. `budget_bounds` *is* reindexed to model coords (`budget_optimizer.py:1889-1895`), so flat entry 0 correctly carries `channel_1`'s bound and lands in `channel_1`'s model column. Base labels it `channel_1`; this branch labels it `channel_2`. That makes it a regression in `d9063460`, not a pre-existing bug, and a silent one: the returned allocation is self-consistent and simply attributes spend to the wrong channel.

**Fix:** either source `MediaVariable.coords` from `_budget_coords` (restores base behaviour exactly), or better, reindex the mask to model coord order at step 4 so the forward and inverse maps cannot drift.

**Why the tests miss it:** the 16 new property tests check `pack(unpack(x)) == x` and `unpack(pack(da)) == da`. Both assert self-consistency of the inverse pair with itself; neither pins the inverse to the forward map, which is the invariant that broke. A test that evaluates `to_model` on a distinguishable flat vector and asserts the model column it lands in matches the label `unpack` returns would catch it (it fails on this branch). No existing test passes a reordered mask, so the full suite staying green does not speak to this.

PabloRoque:
Confirmed, and thank you — reproduced exactly as you described, fixed in `8547e0bd` with your option 2 (reindex the mask at step 4).

**Your diagnosis was right, and the fix reaches one layer further than the report.** Because the mask is consumed positionally *and* supplies the labels, a reordered mask doesn't only mislabel — it also selects the wrong cells. That half is pre-existing (base has it too), so option 1 (`coords` from `_budget_coords`) would have restored base's labels while leaving it in place; option 2 fixes both. Measured, model ordered `[channel_1, channel_2]`, user mask meaning "optimize channel_2, freeze channel_1" written as `[channel_2, channel_1]`:

```
before: optimized value lands in model column 0 == channel_1   (the channel the user froze)
after:  lands in model column 1 == channel_2, unpacks as channel_2
```

The nastier variant is that with a partial mask the two shifts can cancel in the labelled output: the returned allocation looks correct while the objective behind it was computed with the spend in the other channel's saturation curve.

**Two regression tests, both verified failing without the fix** (I checked each independently, including that the label assertion alone catches it — `channel_1` came back with 95.0 against a 5.0 cap):

- `test_shuffled_mask_labels_match_model_coords` — your prescription: distinguishable per-channel bounds, assert the value attributed to a channel is the one its own bounds produced.
- `test_partial_mask_result_is_invariant_to_coord_order` — the same intent written in two coord orders must agree on the allocation *and* the objective value, catching the cancelling case.

Your point about why the property tests missed it is the part I'd want to keep on the record: `pack(unpack(x))` and `unpack(pack(da))` pin the inverse pair to *itself*, so a shared-frame error is invisible to them by construction. The new tests pin the inverse map to the forward map instead.

Also worth noting the asymmetry that made this possible: `cost_per_unit` (`:1742`) and `budget_bounds` (`:1994`) were already reindexed to model coords; only the mask was merely transposed. `budget_distribution_over_period` still is — it's transposed but not reindexed, so a user passing it in a different coord order hits the same class of drift. That one is pre-existing and out of this PR's diff; happy to take it here if you'd rather not leave it, otherwise I'll file it as a follow-up.

The stacked PR (#2804) has been rebased onto the fix. Full suites green: 129 on this branch, 185 with #2804 on top.

PabloRoque:
Fair correction on "addressed your feedback" — that should have read "addressed the blocker". The rest is now in, in `3cb32698`.

**Taken, with your reasoning:**

- **Probe direction.** You convinced me. The midpoint is `total_budget/2` *per cell* for media, and for a response already flat at `x0` a more distant point is flatter still, so the probe was capable of confirming its own "inert" verdict. It now steps a small amount **relative to `x0`**, staying inside the box, with the far end of the box as a fallback only when the box is narrower than the step. Both existing cases still separate correctly: a quadratic at its optimum reads non-inert, a genuinely dead coordinate stays zero. All branches — including the low-only/high-only/unbounded ones you noted are unreachable for `MediaVariable` — now have direct unit tests.
- **`check_gradients: bool = True`** on `allocate_budget`. And since you flagged the cost: the finiteness check and the gradient probe now share **one** evaluation at `x0` instead of taking one each, so with levers the overhead is halved as well as switchable.
- **`flat_dim`** — implemented your version rather than my first instinct (a shared constant, which wouldn't stop a caller passing a custom value to one and not the other): the container now owns the name and pushes it onto its variables, and it's declared on the protocol. Test asserts realignment.
- **Multi-variable coverage.** A second synthetic variable now exercises the `isel` branch end to end — layout, `x0`/bounds concatenation, `pack`/`unpack`, `substitutions`, and `to_model` on the isel'd slice reading the right flat entries. Independently valuable: it's the branch every future variable takes, and it was invisible to CI.
- **`pack` with an extra dim**, **`OptimizationVariable(s)`/`MediaVariable` exports**, and the **stale TypeError comment** (now says what `filter` actually still does — a shape check, since the dtype is coerced above it).

**Passing on one, deliberately:** the duplicate `scattered()` subgraph behind `self._budgets`. You're right that it undercuts the "stated once" framing, and I'd rather concede that than fix it — every clean version means caching symbolic slices or restructuring graph construction, for values that already agree and a subgraph pytensor merges. Noting it as a known blemish; happy to revisit if it ever stops being cosmetic.

**On the neutrality claim** — reasonable not to take my word for it given your box. For the record I did complete the runs: 140 green on this branch, 196 with #2804 stacked on top. And your point stands regardless: no existing test passed a reordered mask, so a green suite never spoke to the bug.

**No need for the probe PR** — your forward-vs-inverse probe is in as `test_shuffled_mask_labels_match_model_coords`, plus `test_partial_mask_result_is_invariant_to_coord_order` for the case where the labelling and selection shifts cancel in the output while the model optimizes the other channel's curve. I verified each fails without the fix, including that the label assertion alone catches it.

Thanks — the blocker was a genuine silent-wrong-answer bug and the test-gap diagnosis (round-trip properties pin the inverse pair to itself, not to the forward map) is the part I'd most want future readers of this module to see.

PabloRoque:
All round-2 items landed in `61a3342f` (152 tests green here, 208 with #2804 stacked):

- **Two-point probe**, as you proposed: local step + furthest reachable bound, inert only when both read zero, second evaluation lazy. Your dead-zone case is a self-calibrating test on the repo fixture — verified it fails with local-only probing, reproducing your exact warning. The one-sided local swing was my overcorrection, not your ask.
- **Narrow boxes**: relative-tolerance-vs-absolute-room fixed, scale floored at unit scale (your `x0=1e-12` case), and coordinates with no distinct probe point now carry no evidence either way — they can no longer convict a variable via a re-evaluation at `x0`. Tested at `x0=1e6` with boxes of width 1 and 5.
- **`flat_dim`**: contract is now *reject*, not realign — you're right that silently renaming a coherent variable is worse. The commit message overstating the old bug is corrected (`scattered()` is positional; the uniform branch never read `flat_dim`).
- **The two open items**: the `isnull` branch and the bool coercion now have tests. On the pydantic envelope — real, but uniform across every `model_post_init` raise, so I'd convert them all in a separate pass rather than make one inconsistent.
- **Not taking**: the duplicate `scattered()` subgraph (conceded as a blemish; every clean fix restructures graph construction for values that already agree).

**Scope note for this PR:** the inert-variable probe is a best-effort *warning* — it tries to answer "is this function flat everywhere or just here?" from point evaluations, which no finite probe set can settle; two points cover the two failure modes found so far, and results are never affected (as you noted). I'm freezing it at this level for #2802: the load-bearing guarantees remain the exact ones (construction-time ancestry check, `check_gradients` escape hatch), and further probe refinements go to a follow-up issue rather than this thread.

Thanks for verifying each item against the code — two of the last three findings were introduced by my fixes to the previous ones, which is exactly the failure mode only this kind of review catches. No need for the offered PR; both your repros are in as tests.

PabloRoque:
Superseded by #2879, which rebuilds this from current `main` as a single commit.

`main` has moved a long way since this branch was cut: #2654 merged, the `BudgetOptimizationResult` work landed via #2853, and #2852 restructured `model_post_init`. Rebasing would have meant resolving duplicate-content conflicts across six commits for code that is already upstream, so #2879 carries only what is genuinely new.

What changed in the rebuild:

- **The inert-variable gradient probe is gone** (with its `check_gradients` kwarg and tests), tracked in #2807 instead. It never affected results, and it accreted an edge case per review round — it does not belong in an abstraction PR.
- **The mask coordinate-order fix is kept**, both halves, with the two regression tests. Worth noting this bug is still live in `main`: `budgets_to_optimize` is transposed but never reindexed, while `cost_per_unit` and `budget_bounds` are.
- Everything else is the same abstraction, now **5 files / +932−148 in one commit** rather than 7 files / +1574−231 across six.

@daimon-pymclabs — your findings all carried over; the review history that produced them is here, so I am leaving this thread intact rather than deleting the branch. Closing in favour of #2879.

Issue #2648, opened 2026-06-26
Drop pymc-extras git pin when >=0.12.2 is released on PyPI


williambdean:
Currently `pymc-extras` requires a git source pin in `pyproject.toml`:

```toml
"pymc-extras>=0.11.0,<0.13.0",

[tool.uv.sources]
pymc-extras = { git = "https://github.com/pymc-devs/pymc-extras.git", rev = "ee8cc37df25f177ae7aea8a76587fb26da261a80" }
```

The git pin resolves to `0.12.2.dev1`. The latest PyPI release at time of writing is `0.12.1` (2026-06-09), which does not include all the DataTree-compatible changes needed by the PyMC 6 migration (https://github.com/pymc-labs/pymc-marketing/pull/2616).

Once `pymc-extras>=0.12.2` (or newer) lands on PyPI:

1. Test that the PyPI release works with this codebase
2. Drop the `[tool.uv.sources]` block for `pymc-extras`
3. Update the version range if needed
4. Run `uv lock`

anevolbap:
`pymc-extras` 0.12.2 was never released. The line went 0.12.1 (2026-06-09) to 0.13.0 (2026-07-13), then 0.13.1 and 0.14.0 (2026-07-28). The pinned commit `ee8cc37` is an ancestor of `v0.13.1`, so the DataTree changes are on PyPI.

The blocker moved. 0.13.x requires `pymc>=6.1,<6.2` and 0.14.0 requires `pymc>=6.2,<6.3`, while `pyproject.toml` pins `pymc>=6.0.0,<6.1.0`. Dropping the git source needs a PyMC bump first.

I installed the package against pymc 6.1 + pymc-extras 0.13.1 and against pymc 6.2 + pymc-extras 0.14.0. Both fail at import for the same reason: `pymc.model.fgraph.extract_dims` and `pymc.pytensorf.StringConstant` were removed in 6.1, since `ModelVar` now carries `name` and `dims` as Op props. That hits `pymc_marketing/model_graph.py` (one line) and `pymc_marketing/pytensor_utils.py` (`_prefix_model` renames dims by replacing the string constants, so it needs rebuilding around the new Op props). Every other pymc-internal import in the package still resolves under 6.2.

The draft PR #2803 already does the pymc 6.2 bump and vendors `extract_dims`, so it covers that code side. The failing jobs I checked there are CLV logprob errors (`MeasurableClip` dtype), not dependency resolution.

[...]

Issue #2625, opened 2026-06-12
CI broken: matplotlib 3.11.0 breaks arviz-plots 1.1.0 at import


anevolbap:
All CI jobs currently fail at import time with:

```
AttributeError: module 'matplotlib.style' has no attribute 'core'
```

Example run: https://github.com/pymc-labs/pymc-marketing/actions/runs/27405050613/job/80992101658 (from #2624, but any PR triggers it).

**Cause**

matplotlib 3.11.0 (PyPI, 2026-06-12) moved the style implementation into `matplotlib.style` and left `matplotlib.style.core` as a deprecated shim that is no longer imported automatically, so plain attribute access on it raises `AttributeError`. `arviz_plots` 1.1.0 does `mplstyle.core.USER_LIBRARY_PATHS.append(...)` at import time, and `arviz` imports `arviz_plots`, so any `import pymc` fails. Jobs that ran before the matplotlib release (main, 2026-06-11) are green; everything after is red.

arviz-plots fixed this on main in arviz-devs/arviz-plots#466 (merged 2026-05-26), but their latest release 1.1.0 (2026-04-23) predates the fix.

**Possible fix**

`arviz-plots` is a direct dependency here and `matplotlib>=3.5.1` has no upper bound, so the resolver picks 3.11.0. Pin `matplotlib<3.11` (same pattern as the pymc-extras upper bound in #2623) until the arviz devs publish an arviz-plots release that includes the fix, then drop the pin. matplotlib 3.10 with arviz_plots 1.1.0 works (test suite passes locally on 3.10.9).

PR #2803, opened 2026-08-02
chore(deps): bump pymc to 6.2, drop explicit pytensor dependency

PabloRoque:
## Description

Bump `pymc>=6.2.0,<6.3.0` and drop the explicit `pytensor` dependency — pymc 6.2 pins `pytensor>=3.2.2,<3.3` transitively, so the direct requirement was redundant (and its `>=3.0.0` floor admitted broken versions, see below).

### Why now

pytensor ≥ 3.2.0 fixes second derivatives through `Blockwise{Convolve1d}` (the batched geometric adstock): on 3.0.x/3.1.x every Hessian construction from pytensor's own gradients tutorial (grad-of-grad, `pushforward`, `hessian_vector_product`, the scan recipe, the `hessian` macro) fails at runtime with `ValueError: Scalar check failed (npy_bool)` on the C/numba path whenever shapes are symbolic. Same op/backend family as pytensor#2228 (fixed by pytensor#2234). This unblocks exact Hessian-vector products for `BudgetOptimizer` (planned trust-constr support). Verified: the exact HVP now matches finite differences to 6e-12 on the default backend.

### Migration: pymc#8340 (fgraph representation change)

pymc 6.1 moved variable names and dims from symbolic `StringConstant` inputs onto `ModelVar` op `__props__` — an acknowledged breaking change ("attaching it to var.name is not reliable as PyTensor has no obligation to preserve it"). This repo's `pymc<6.1.0` pin dated from that release. The migration:

- **`extract_dims` vendored** into `pytensor_utils.py` (removed upstream), reading the new `op.dims` representation.
- **`_prefix_model` simplified ~95 → ~35 lines**: op props are now the single renaming channel — verified empirically that `model_from_fgraph` reconstructs RV names, `named_vars`, dims and coords purely from the props. The only extra step is renaming value variables (e.g. `x_log__`), which carry their own names and must stay unique across merged models. Ops are copied rather than mutated (name/dims are hash-bearing `__props__`). Deleted: two variable-name rename loops and the fuzzy `<base>__`-suffix matching heuristic, all built on the representation pymc#8340 deprecated.
- **`model_graph.py`**: `model_free_rv` gained an explicit `name` parameter.
- One white-box test updated to assert on the rebuilt model instead of fgraph variable names (the deprecated channel).

Transitive cleanup from the pymc bump: zarr/toolz no longer pulled in.

## Related Issue

- Upstream: pymc-devs/pymc#8340, pymc-devs/pytensor#2228 / #2234

## Checklist

- [x] `tests/test_pytensor_utils.py` + `tests/test_model_graph.py`: 14 passed
- [x] `tests/mmm/test_budget_optimizer*.py` + `test_cost_per_unit.py`: 103 passed
- [x] pre-commit (ruff, mypy) clean

## Modules affected

- `pyproject.toml` / `uv.lock`
- `pymc_marketing.pytensor_utils` (merge/prefix machinery)
- `pymc_marketing.model_graph`

## Type of change

- [x] Dependency bump with required migration; net −93 lines

----
📚 Documentation preview 📚: https://pymc-marketing--2803.org.readthedocs.build/en/2803/


anevolbap:
This bump is also the unlock for the `pymc-extras` git pin in #2648. pymc-extras 0.14.0 requires `pymc>=6.2,<6.3`, so it becomes installable only with `pymc>=6.2.0,<6.3.0` here. Dropping the `[tool.uv.sources]` git source then means moving `pymc-extras>=0.11.0,<0.13.0` to `>=0.14.0,<0.15.0`.

That in turn clears the `matplotlib<3.11` upper bound added for #2625. pymc-extras 0.14.0 requires `preliz>=0.27`, and preliz 0.27.0 is the release that guards the `style.core` access matplotlib 3.11 removed. arviz-plots, the original trigger in #2625, fixed its own copy in 1.2.0, which `pyproject.toml` already requires.

Measured locally with the matplotlib bound removed: matplotlib 3.11.1 with preliz 0.26.0 gives 8 failures in `tests/mmm`, all `AttributeError: module 'matplotlib.style' has no attribute 'core'` through `create_constrained_inverse_gamma_prior`, and they pass once preliz 0.27.1 is installed.

Would you rather fold the pymc-extras and matplotlib bounds into this PR, since they touch the same `pyproject.toml` lines, or keep this focused on pymc and leave them to a follow-up?

juanitorduz:
I think most of this one landd in https://github.com/pymc-labs/pymc-marketing/pull/2883, however feel free to push improvements and better type hints :)

PabloRoque:
Done in #2883

PR #2804, opened 2026-08-03
feat(mmm): optimizable_vars + OptimizableMuEffect — effect levers on the OptimizationVariable stack

PabloRoque:
## Description

**Stacked on #2802** (which stacks on #2801 → #2654). GitHub cannot target a fork branch as base, so this PR targets `refactor/budget-optimizer-pm-model` and includes the stack's commits — **review the top 2**:

- `baafd5bd` — `LeverVariable` + `optimizable_vars` (optimizer layer)
- `3b29c804` — `OptimizableMuEffect` + MMM wiring (marketing layer)

### What this is

The successor to #2621's *mechanism*, ported onto #2802's `OptimizationVariable` abstraction. #2621's design survived five review rounds but its diff was written against optimizer internals that #2802 superseded; this PR carries the reviewed design forward on the new architecture and supersedes #2621 (a closing summary will map every review finding to where it landed).

**Scope note:** this PR is the *mechanism only*. `DiscountedEventEffect` — the first concrete optimizable effect, with its closed-form-optimum validation (`d* = (β−1)/(β+1)`), serialization, and tutorial notebook — follows in a separate PR, so the application ships together with its documentation. The mechanism here is exercised end-to-end by synthetic in-test effects instead.

### Commit 1 — optimizer layer (`optimizable_vars` via `LeverVariable`)

A lever is one `pm.Data` node with a single non-date dim, co-optimized in native units alongside the media budgets, never entering the budget-sum constraint. `LeverVariable` implements #2802's variable protocol in ~60 lines (identity forward map, warm start at the current model value clipped to bounds, declared native bounds); the flat layout, single joint `do()`, bounds/x0 assembly and result unpacking all come from the shared container — replacing the ~290 lines of head/tail bookkeeping #2621 needed for the same behaviour. Optimal lever values land in `result.optimized_vars` (the #2801 result field).

Guards ported from #2621: construction-time reachability of every lever from the response variable; a warning when levers are scored against the media-contribution objective; an objective-finiteness check at the warm start; duck-typed lever inference on the legacy-wrapper path (the optimizer still imports zero marketing code — the AST guard test enforcing that is ported verbatim). The zero-gradient probe is upgraded to **per-coordinate** reporting: a single out-of-window event is named (`promo_data['early_event']`) while its in-window siblings optimize — uniform across all variables, media included.

### Commit 2 — marketing layer (`OptimizableMuEffect`)

A thin marker ABC (`lever_var_name`, `lever_bounds`) that `MMM.budget_optimizer` translates into `optimizable_vars` entries. Wiring ported with the reviewed design intact: **gated** `total_response_original_scale` deterministic (registered only when an optimizable effect is present — no silent additions to existing users' posteriors), per-effect `_mu_baseline` refresh so repricing effects compose multiplicatively, and the `None` / dict-merge / `{}`-opt-out semantics on both optimizer entry points.

@williambdean — requesting your eyes on the marker-ABC design specifically: it follows your #2621 comment (minimal protocol, concreteness in subclasses), and the docstring notes it as a stepping stone toward generic X-dataset variable referencing. Your sign-off on this shape was the one thing #2621's five rounds never captured.

### Verification

- 183 tests green across the five affected suites; the 14 ported integration tests passed on their first run under the new architecture after only mechanical adaptations.
- Mechanism coverage without the concrete effect: a windowed synthetic lever exercises the out-of-window per-coordinate warning end-to-end; a mixed media+lever objective asserts a warm start at the lever optimum does not false-alarm **and** media moves off its uniform seed (the round-5 counterexamples, both directions).
- Complete #2621 feedback ledger swept: every finding is ported, superseded-by-stack, fixed-here, or explicitly deferred to the effect+notebook PR — no orphans. (Deferred: the notebook narrative items and juanitorduz's style comments; the `set_data` planned-calendar fix travels with `DiscountedEventEffect`.)

## Related Issue

- Supersedes #2621 (to be closed with a mapping comment once this lands)
- Sequence: #2654 → #2801 → #2802 → this → effect+notebook PR

## Modules affected

- `mmm.optimization_variables` (`LeverVariable`)
- `mmm.budget_optimizer` (`optimizable_vars`, guards, `optimized_vars`)
- `mmm.additive_effect` (`OptimizableMuEffect`), `mmm.link`, `mmm.mmm` (wiring)

## Type of change

- [x] New feature (backward compatible — no levers declared means behaviour identical)

----
📚 Documentation preview 📚: https://pymc-marketing--2804.org.readthedocs.build/en/2804/


PabloRoque:
Closing in favour of two smaller PRs, now that #2879 has landed the abstraction this was built on.

This branch is 62 commits behind the current design: it still carries the superseded result-type commit (landed as #2853), the pre-review version of `OptimizationVariable`, and all four probe commits that #2879 deliberately dropped. Its two genuinely new commits sit on an abstraction that has since changed (shared coordinate alignment, `FLAT_DIM`, scaling moved after the spreading branch, `optimized_vars` now populated). Rebuilding is less work than untangling, and it keeps each PR to one reviewable concern.

Splitting into:

1. **optimizer layer** `LeverVariable` + `optimizable_vars`, graph-only, no marketing imports. This also closes a gap #2879 left behind: no test there builds `OptimizationVariables` with two variables, so the `variable_slice` `isel` path that every additional variable depends on has never run in CI. A lever is the first real second variable.
2. **marketing layer** `OptimizableMuEffect` plus the `MMM`/`link` wiring, including the warning for levers scored against the media-contribution objective (that check pattern-matches an MMM variable name, so it belongs at the adaptor layer rather than in the optimizer).

Two things deliberately not carried over: the inert-variable probe (tracked in #2807) and the `f(x0)` finiteness check. The latter only improved an error message, since a non-finite objective already fails loudly through scipy, and the case it was written for (a discount depth of 1.0 under a log link) is caught earlier by `DiscountedEventEffect` at build time.

No content is lost; the branch stays for reference.

PR #2805, opened 2026-08-03
Create `ModelFitter` class

### LINKED ISSUE #1871: Consolidate model fitters into `ModelBuider`
**https://github.com/pymc-labs/pymc-marketing/pull/1870 should be merged before starting work on this issue.**

MMMs are the only models using `ModelBuilder.fit` directly. Other models have to override with their own fitters & samplers due to different data-loading requirements, which has led to code redundancies for idata extensions and prior/predictive checks. The CLVModels also provide fitters other than NUTS and a wrapper for `arviz.fit_summary` that could easily be moved into `ModelBuilder` for shared functionality.

Let's consolidate all shared boilerplate into `ModelBuilder.fit`, and pare down all the various `model.fit(**specific_data_kwargs)` methods into the bare minimum needed to call `super().fit(**kwargs)`.  We can start with this code created by @williambdean in https://github.com/pymc-labs/pymc-marketing/pull/1467:

<details>
<summary>Model fit starter code</summary>

```python
def create_fit_data(self) -> xr.Dataset:
	"""Create the fit_data group based on the input data."""
	if isinstance(self.data, pd.DataFrame):
		return self.data.to_xarray()
	return self.data

def fit(
	self,
	progressbar: bool | None = None,
	random_seed: RandomState | None = None,
	**kwargs: Any,
) -> az.InferenceData:
	"""Fit a model using the data provided at initialization.
	
	Sets attrs to inference data of the model.

	Parameters
	----------
	progressbar : bool, optional
		Specifies whether the fit progress bar should be displayed. Defaults to True.
	random_seed : Optional[RandomState]
		Provides sampler with initial random seed for obtaining reproducible samples.
	**kwargs : Any
		Custom sampler settings can be provided in form of keyword arguments.
	
	Returns
	-------
	self : az.InferenceData
		Returns inference data of the fitted model.
	
	Examples
	--------
>>> 	model = MyModel()
>>> 	idata = model.fit()
	Auto-assigning NUTS sampler...
	Initializing NUTS using jitter+adapt_diag...
	"""
	if not hasattr(self, "model"):
		self.build_model()
		
	sampler_kwargs = create_sample_kwargs(
		self.sampler_config,
		progressbar,
		random_seed,
		**kwargs,
	)
	with self.model:
		idata = pm.sample(**sampler_kwargs)
	
	if self.idata:
		self.idata = self.idata.copy()
		self.idata.extend(idata, join="right")
	else:
		self.idata = idata
	
	if "fit_data" in self.idata:
		del self.idata.fit_data
	
	fit_data = self.create_fit_data()

	with warnings.catch_warnings():
		warnings.filterwarnings(
			"ignore",
			category=UserWarning,
		)
		self.idata.add_groups(fit_data=fit_data)
	self.set_idata_attrs(self.idata)
	return self.idata # type: ignore
```
</details>

ColtAllen:
## Description

Many models in the library have their own respective `fit` methods, creating inconsistency and a lot of redundant code. This PR creates a new Mixin class in `model_builder.py` to standardize fitting across all PyMC-Marketing models. It makes additional samplers much easier to add in the future, and as currently written, adds support for MAP, ADVI, and other samplers for models that previously only supported MCMC fits.

## `model_builder.py` UML Diagram
```mermaid
classDiagram
    direction TB

    class ModelIO {
        <<mixin>>
        +_model_type: str
        +version: str
        +idata: DataTree
        +model_config: dict
        +sampler_config: dict
        +id: str
        +save(fname)
        +load(fname, check)$
        +load_from_idata(idata, check)$
        +create_idata_attrs() dict
        +set_idata_attrs(idata) DataTree
        +attrs_to_init_kwargs(attrs)$ dict
        +idata_to_init_kwargs(idata)$ dict
        +build_from_idata(idata)*
        +_serializable_model_config: dict*
    }

    class ModelFitter {
        <<mixin>>
        +model: Model
        +is_fitted_: bool
        +approx: Approximation
        -_recompute_deterministics: bool
        +fit(data, *, method, progressbar, random_seed, sample_kwargs, kwargs) DataTree
        #_prepare_fit(data) «hook»
        #create_fit_data_group() Dataset «hook»
        #_get_sampling_model() Model «hook»
        #post_sample_model_transformation() «hook»
        -_sample_with_deterministics(sampler_kwargs, step_factory) DataTree
        -_fit_mcmc(sampler_kwargs) DataTree
        -_fit_demz(sampler_kwargs) DataTree
        -_fit_map(kwargs) DataTree
        -_fit_approx(method, progressbar, random_seed, sample_kwargs, kwargs)
    }

    class ModelBuilder {
        <<abstract>>
        -_skipped_config_keys: set
        +__init__(model_config, sampler_config)
        +default_model_config: dict*
        +default_sampler_config: dict*
        +build_model()*
        +fit_result: Dataset
        +graphviz(kwargs)
        +table(kwargs) Table
    }

    class RegressionModelBuilder {
        <<abstract>>
        +output_var: str*
        +build_model(X, y)*
        +build_from_idata(idata)
        +fit(X, y, *, method, progressbar, random_seed, sample_kwargs, kwargs) DataTree
        +approximate_fit(X, y) «deprecated»
        +predict(X) ndarray
        +predict_proba(X)
        +predict_posterior(X)
        +sample_prior_predictive(X, y)
        +sample_posterior_predictive(X)
        +create_fit_data(X, y) Dataset
        -_data_setter(X, y)*
        -_validate_fit_inputs(X, y)
        -_prepare_fit(data) «override»
        #create_fit_data_group() «override»
    }

    class MMM {
        #_get_sampling_model() «freeze_dims_and_data»
    }
    class BassModel {
        -_prepare_fit(data) «rebuild every fit»
    }
    class CLVModel {
        -_prepare_fit(data) «build once, reject new data»
    }
    class MaxDiff {
        -_recompute_deterministics = False
    }

    ModelIO <|-- ModelBuilder
    ModelFitter <|-- ModelBuilder
    ModelBuilder <|-- RegressionModelBuilder
    ModelBuilder <|-- CLVModel : clv/models/basic.py
    ModelBuilder <|-- BassModel : bass/model.py
    RegressionModelBuilder <|-- MMM : mmm/mmm.py
    RegressionModelBuilder <|-- MaxDiff : customer_choice/

    note for ModelFitter "fit() dispatches on method:
mcmc / demz -> _sample_with_deterministics
map -> _fit_map (find_MAP)
advi / fullrank_advi -> _fit_approx
then merges idata, dropping groups
derived from the previous posterior"
```

## Related Issue

- [ ] Closes https://github.com/pymc-labs/pymc-marketing/issues/1871
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2805.org.readthedocs.build/en/2805/


daimon-pymclabs:
Reviewed the full diff (19 files, +994/-788), the removed per-model `fit` implementations, and the relevant PyMC internals (5.28) to check the kwarg-filtering and deterministics claims. All 57 checks are green; it's still a draft. Net effect is real: five samplers available on every model, 84 fewer lines, one pipeline instead of six.

## Blocking-ish

**1. Refitting with a different method leaves stale groups behind** (`model_builder.py:932-938`)
`CLVModel.fit` used to do `self.idata = idata` (wholesale replace). The base now merges: `self.idata.copy()` then `.update(idata)`. `update` only replaces same-named groups, so `fit(method="mcmc")` followed by `fit(method="map")` keeps the MCMC `sample_stats` (and `log_likelihood`) attached to a 1-draw MAP posterior, plus any `posterior_predictive` from the earlier run. That's a silent inconsistency, and a regression specifically for CLV. Suggest dropping the derived groups (`sample_stats`, `log_likelihood`, `posterior_predictive`, `predictions`) on refit, the same way `fit_data` is already dropped.

**2. `_get_sampling_model` is only honoured by two of the five methods**
`_sample_with_deterministics` calls it (`:722`), but `_fit_MAP` (`:756`) and `_fit_approx` (`:774`) use `self.model` directly. `MMM` overrides it with `freeze_dims_and_data` (`mmm/mmm.py:2084`), so `mmm.fit(method="map")` and `method="advi"` run against the unfrozen graph. The class docstring says the hook returns "the model actually handed to the sampler", which isn't true. Not a wrong-answer bug, but it undercuts the standardization the PR is for.

**3. Base `fit(data=...)` silently discards `data`** (`:665`)
`ModelFitter._prepare_fit` takes `data` and never uses it; it only builds the model. Bass raises a clear error when data is missing, CLV forwards it to `build_model`, and the base drops it on the floor. Any subclass that doesn't override the hook will fit stale data with no warning. Related: CLV's `_prepare_fit` calls `build_model(None)` when nothing is built and no data is passed, which dies inside `_validate_data` with `'NoneType' has no attribute 'columns'` instead of "data is required".

**4. `method` sits in a different position in every signature**
Base and Bass: `(data, method, progressbar, random_seed, ...)`. `RegressionModelBuilder`, `MMM`, `MaxDiffMixedLogit`, `MixedLogit`: `(..., progressbar, random_seed, method, ...)`. Bass in particular moved the second positional from `progressbar` to `method`, so `bass.fit(data, False)` now resolves to `method=False`. Make `method` keyword-only (`*` before it) across the board while the API is still pre-1.0.

## Worth fixing

**5. The deterministics recompute runs even when there are none** (`:722-736`)
Most CLV models have no `Deterministic`. They now get a `var_names`-restricted trace plus a `compile_fn` and a full pass over draws (with progress bar) computing an empty output list. Cheap guard:
```python
if not self._recompute_deterministics or not model.deterministics:
    return pm.sample(**kwargs)
```

**6. Non-int seeds are silently dropped on the MAP path** (`:913`)
`isinstance(random_seed, int)` is False for `np.int64`, and Generator/list seeds are dropped without a word, so `fit(method="map", random_seed=rng)` is quietly unseeded. Convert or warn.

**7. `approximate_fit` doesn't need a deprecation cycle.** It landed 2026-06-01 in #2477, after the last release (0.19.4, 2026-05-06), and `version.py` says `1.0.0.dev0`. It has never shipped. Deleting it in 1.0.0 is cleaner than shipping a method that arrives already deprecated, plus a `FutureWarning` shim to carry forward.

**8. The Bass `m` prior is two changes in one, and the scale looks wrong** (`bass/model.py:504`)
Giving market potential positive support is correct and the docstring explaining the `-inf` initial point is a good catch. But `m` is "total number of eventual adopters" and `adopters(t) = m·f(t)` under a Poisson likelihood, so `HalfNormal(sigma=10)` implies roughly 8 adopters ever. The module's own example uses `DiracDelta(c=10_000)` (`:115`), and the test data is `Poisson(lam=100)` over 20 periods, about 2000 adopters. `test_fit_map` only asserts `m > 0`, so it won't catch a posterior pinned at the prior. Consider a data-scaled default, or requiring the user to set `m`, and split this out of the refactor PR so it gets its own changelog line.

**9. Lost warning suppression.** `maxdiff`, `mixed_logit`, `mnl_logit` and `nested_logit` all wrapped `self.idata["/fit_data"] = ...` in a `catch_warnings` block filtering "The group fit_data is not defined in the DataTree scheme". The base assigns without it. Intentional, or does that warning come back for those four?

**10. ADVI odds and ends.** `random_seed` coming from `sampler_config` reaches `pm.fit` but not `approx.sample` (only the explicit parameter gets `setdefault`), so seeded-via-config ADVI draws aren't reproducible. The `method`-in-`sampler_config` error message hardcodes `'advi'` (`:786`) even under `fullrank_advi`. And since `chains` is in most real `sampler_config`s, a normal ADVI fit now emits two warnings (the chains one plus the ignored-`tune`/`chains` list); fine as designed, just noisy.

**11. Test gaps.** No `map` or `demz` coverage on a `RegressionModelBuilder`/`MMM`, which is exactly the class where `_get_sampling_model` diverges (finding 2). Nothing asserts what base `fit(data=...)` does (finding 3). And `fit(method="map", draws=100)` forwards `draws` into `scipy.optimize.minimize` for an opaque `TypeError`; a validation pass on the MAP path would be friendlier.

Nits: `_fit_MAP`/`_fit_DEMZ` casing vs `_fit_mcmc`/`_fit_approx`; `sampler_kwargs` is computed then discarded on the map and advi paths; `_sample_kwargs.setdefault("draws", config.get(...))` is redundant since `draws` already came through `merged`; the demz warning says NUTS-only keys "will be ignored", but `nuts_sampler` would actually have raised, not been ignored; `MMM.fit` accepts `sample_kwargs` only via `**kwargs` and doesn't document it.

## What's good

The kwarg filter derived from the `pm.fit` / `Inference.fit` / `ObjectiveFunction.step_function` signatures is a real improvement over the hard-coded 18-name list it replaces, and `test_approx_fit_parameters_are_derived` locks it in. Calling `step_factory` inside the model context is a subtle correctness detail that the comment explains well. `_recompute_deterministics = False` with a per-model justification on BLP/maxdiff/mixed logit shows the memory tradeoff was actually thought about rather than flattened. The `demz` `target_accept` strip is necessary: I confirmed `pm.sample` routes unknown kwargs into `step_kwargs` and raises when an explicit `step` is supplied.

Two side wins worth calling out. The ADVI path was previously computing deterministics twice (draw-by-draw inside `Approximation.sample`'s `NDArray.record`, then again vectorized via `compute_deterministics`); dropping the second one is why the GAM notebook went 7.63s to 1.98s. And `mixed_logit.create_and_fit` had `self.fit(extend_idata=True, kwargs=fit_kwargs)`, which passed a dict as a kwarg named `kwargs`; that's now fixed. Consolidating the duplicated `create_sample_kwargs` out of `mmm.py` is overdue cleanup.

ColtAllen:
UML Diagram for `model_builder.py` added to https://github.com/pymc-labs/pymc-marketing/pull/2805#issue-5049910343. 

All review items in https://github.com/pymc-labs/pymc-marketing/pull/2805#issuecomment-5179524012are addressed in five commits on `modelfitter`, every affected suite passes (597 + 40 + 28 + MMM fit subset, all green), and the working tree is clean. Here's what landed:

**Blocking fixes** — [[babef1aa](https://github.com/pymc-labs/pymc-marketing/commit/babef1aa)](https://github.com/pymc-labs/pymc-marketing/commit/babef1aa) and [[38b9bc12](https://github.com/pymc-labs/pymc-marketing/commit/38b9bc12)](https://github.com/pymc-labs/pymc-marketing/commit/38b9bc12):
- Refitting with a different method now drops the previous fit's derived groups (`sample_stats`, `log_likelihood`, `posterior_predictive`, `predictions`, warmup groups) before merging, so a MAP refit can't carry MCMC baggage.
- `_fit_map` and `_fit_approx` now sample the model from `_get_sampling_model()`, so MMM's frozen graph is used on every path. A spy test locks this in.
- Base `fit(data=...)` raises `NotImplementedError` instead of silently discarding data, and CLV models raise a clear "data is required" error instead of the `NoneType.columns` crash.
- `method`, `progressbar`, `random_seed`, and `sample_kwargs` are keyword-only in all eight `fit` signatures (including `ParetoNBD`, which the review missed), in one consistent order. A repo-wide grep found no positional call sites to fix.

**Worth-fixing** — [[813cb0e6](https://github.com/pymc-labs/pymc-marketing/commit/813cb0e6)](https://github.com/pymc-labs/pymc-marketing/commit/813cb0e6) and [[7189c7de](https://github.com/pymc-labs/pymc-marketing/commit/7189c7de)](https://github.com/pymc-labs/pymc-marketing/commit/7189c7de): the no-deterministics guard, MAP seed normalization (`np.integer`/`Generator`/`RandomState` convert, unsupported types warn), restored `fit_data` warning suppression, config-seeded ADVI reproducibility, the `fullrank_advi` message fix, the MAP kwarg filter (derived from `pm.sample` vs `pm.find_MAP` signatures, so `fit(method="map", draws=100)` warns instead of dying in scipy), plus all five nits including the `_fit_map`/`_fit_demz` renames.

**Bass `m` prior** — [[e9b355dc](https://github.com/pymc-labs/pymc-marketing/commit/e9b355dc)](https://github.com/pymc-labs/pymc-marketing/commit/e9b355dc): when the user hasn't overridden `m`, `build_model` sets `sigma = 2 × observed cumulative adoptions`. `test_fit_map` now asserts the MAP estimate lands within 4× of the data's scale (~2000), not merely `> 0`.

**Pushback on finding 7** (worth replying on the PR): the reviewer's claim that `approximate_fit` "never shipped" is wrong — it landed in **#1923 on 2025-09-04**, not #2477, and `git tag --contains` shows it in releases **0.17.0 through 0.18.2**. The `FutureWarning` deprecation shim is the correct path and stays.

juanitorduz:
@ColtAllen here is a PR to improve this one : https://github.com/ColtAllen/pymc-marketing/pull/13 (after a review with Damino)

juanitorduz (approved):

juanitorduz:
@ColtAllen feel free to merge once you are done :)

Issue #2544, opened 2026-05-08
Use incrementality module for ROAS computation in budget plots


isofer:
From [PR #2531 discussion](https://github.com/pymc-labs/pymc-marketing/pull/2531#discussion_r3208844543) with @juanitorduz.

The budget allocation plotting code (`BudgetPlots.allocation_roas`) computes ROAS directly via `sample_response_distribution()`. For consistency, we should route this through the incrementality module instead.

This could also open the door to computing ROAS at different time points within an allocation period (e.g., monthly within a quarterly budget), which would require updating the incrementality module to support arbitrary allocations and adding a `roas_over_time` plot.

**Scope:**
- Refactor ROAS computation in budget plots to use the incrementality module
- Explore support for time-varying ROAS within an allocation period

Issue #2484, opened 2026-04-07
Make interventional contributions the default recommendation


daimon-pymclabs:
## Motivation

pymc-marketing now has two approaches for computing channel contributions:

1. **Traditional decomposition** — multiply each channel's transformed values by its estimated coefficient. This is the older, more prominent approach in docs and examples.
2. **Interventional (g-computation)** — compare model predictions under actual spend vs. a counterfactual with spend removed. This is implemented in the `Incrementality` class (merged Feb 2026, `pymc_marketing/mmm/incrementality.py`).

The interventional approach is strictly more correct for causal attribution and should be the recommended default.

## Problem with traditional decomposition

Traditional decomposition gives correct causal attribution **only when the DAG is a flat regression** with no paths between predictors (i.e., all inputs are independent causes of the outcome). This assumption breaks down in common real-world scenarios:

- **Seasonality as a confounder/mediator**: time-of-year drives both baseline demand (season → outcome) and budget allocation (season → media → outcome). Traditional decomposition attributes the media-mediated seasonal effect to media rather than season, overstating media ROI in high-spend periods like Q4.
- **Marketing funnel structure**: upper-funnel awareness channels drive lower-funnel conversion channels. Traditional decomposition double-counts or misallocates contributions across funnel stages.
- **Any DAG with paths between predictors**: whenever one model input causally affects another, coefficient × observed-value decomposition conflates direct and indirect effects.

Interventional contributions (g-computation) handle all of these correctly by construction — they answer "what would have happened without this channel's spend?" rather than "how much of the fitted value is mechanically attributable to this input."

## What should change

1. **Documentation and examples should recommend interventional contributions as the primary approach.** The `Incrementality` class currently exists but is not prominently featured in the main MMM example notebooks. Traditional decomposition methods appear first and more often in user-facing materials.

2. **The `Incrementality` class docs should explicitly name g-computation** as the underlying methodology and explain why it produces more accurate causal attribution than traditional decomposition — especially as pymc-marketing pushes users toward more realistic DAG structures (upper funnel models, marketing funnels).

3. **Traditional decomposition methods should carry a note** explaining their assumptions (flat DAG, independent predictors) and pointing users to `Incrementality` for causal attribution in models with more involved structure.

## Related issues

- #2085 — Proposes adding a full SCM + g-computation example to the `mmm_upper_funnel_causal_approach` notebook
- #2330 — Decoupling incrementality logic from MMM implementation details for generality

[...]

Issue #2330, opened 2026-02-24
Decouple incrementality logic from MMM implementation details via a general model setup entrypoint


isofer:
$(cat <<'EOF'
## Context

In PR #2290 (Incrementality module), @ricardoV94 [noted](https://github.com/pymc-labs/pymc-marketing/pull/2290#discussion_r2833032194) that the `Incrementality` class directly references MMM internals (e.g. `self.model.model[channel_data]`, time-varying media/intercept shared variables) when setting up the vectorized evaluator. This couples the incrementality logic to the specific MMM implementation, making it fragile when the default model changes and impossible for subclasses with custom models to participate correctly.

## Problem

The current incrementality code manually handles model-specific shared variables:

```python
data_shared = self.model.model[channel_data]
# ...
if hasattr(self.model, media_temporal_contribution):
    # handle time_varying_media
if hasattr(self.model, intercept_temporal_contribution):
    # handle time_varying_intercept
```

This pattern:
- Breaks when MMM subclasses have custom model structures
- Requires updating incrementality code every time the default model changes
- Duplicates the same coupling problem that was already solved for the budget optimizer

## Proposed Solution

Follow the precedent set by `_set_predictors_for_optimization` in the budget optimizer. Create a general-purpose entrypoint method on the MMM class (e.g. `_prepare_model_for_evaluation` or similar) that:

1. Allows the model to declare which shared variables need to be replaced and how
2. Lets subclasses override this method to handle their custom model structures
3. Is reusable across incrementality, sensitivity analysis, budget optimization, and any future module that needs to define functions from existing models

This way, `Incrementality` (and other consumers) call a single model method instead of reaching into model internals.

## References

- PR #2290: Incrementality module (where the coupling was introduced)
- Review comment: https://github.com/pymc-labs/pymc-marketing/pull/2290#discussion_r2833032194
- Existing precedent: `_set_predictors_for_optimization` in the budget optimizer
EOF
)

PR #2769, opened 2026-07-28
docs(mmm): add advanced geo-level funnel MMM example via DataVarMuEffect

juanitorduz:
## Summary

Adds `docs/source/notebooks/mmm/mmm_funnel_mueffect_advanced.ipynb`, an advanced companion to `mmm_funnel_mueffect.ipynb`. Same core idea, but a funnel DAG that looks much more like a real application:

- **Two upper-funnel channels** (`tv_spend`, `social_spend`) feeding one shared latent demand pool, each with its own carryover and saturation on the way there.
- **A geo panel** (`dims=("geo",)`, 3 regions × 130 weeks), with some parameters pooled across geos and others free per geo.
- **A mediator that is itself media**: structural lower-funnel spend `M* = D + λ_g·B` (demand plus an exogenous budget), observed with reporting noise, converting through its own adstock/saturation — plus branded search volume as a **pure indicator** of demand (loading κ, no causal role).
[...]

PR #2813, opened 2026-08-04
Make incrementality correct beyond the additive case: link-aware reduction, mediated mu_effects, and pm.do interventions

### LINKED ISSUE #2811: Incrementality module assumes an identity link: ROAS/mROAS are silently wrong for multiplicative (log-link) MMMs
## Summary

`pymc_marketing/mmm/incrementality.py` implements the ROAS / mROAS estimators of Jin, Wang, Sun, Chan & Koehler (2017), *Bayesian Methods for Media Mix Modeling with Carryover and Shape Effects* ([paper](https://storage.googleapis.com/gweb-research2023-media/pubtools/3806.pdf)), and it does so correctly **for identity-link models**. Under `link="log"` (the multiplicative MMM, see `docs/source/notebooks/mmm/mmm_multiplicative.ipynb`) it runs without error and returns numbers that are not incremental revenue. There is no guard, no warning, and no test coverage for that path.

Two independent things break. Neither is a bug in the paper mapping; both are consequences of the module assuming the response is additive in the linear predictor.

## Current behaviour (identity link, correct)

For reference, the mapping to the paper as implemented today:

| Method | Paper |
|---|---|
| `contribution_over_spend` | eq (10), ROAS |
| `marginal_contribution_over_spend` | eq (11), mROAS (`1.01` factor, `0.01 * spend` denominator, sign flip at `incrementality.py:879`) |
| `spend_over_contribution` | `1 / eq(10)`; a reciprocal for CAC-style reporting, not itself a paper formula |

Carryover handling follows the paper's Figure 3 recipe: a pre-change window of length `l_max` held at actual spend for carry-in, and the summation extended past `t1` for carry-out.

One cosmetic divergence: the paper sums to `t1 + L - 1`, the code masks `dates <= t1 + l_max * freq_offset`, one period further. Because `geometric_adstock` uses `lags = arange(l_max)`, spend at `t1` has zero weight by `t1 + l_max`, so the extra period contributes nothing. Harmless, but it isn't the paper's index and is worth aligning.

## Problem 1: the counterfactual is evaluated in linear-predictor space

`compute_incremental_contribution` extracts the response graph at `response_variable="channel_contribution"` (`incrementality.py:363`), differences baseline vs counterfactual, sums over dates, and multiplies by `target_scale` (`incrementality.py:922`).

But `channel_contribution` lives in the linear predictor: `mu_var = intercept + channel_contribution.sum(dim="channel")` (`mmm.py:2335`). Under the log link the response is `exp(mu) * target_scale`. A sum of differences of log-space contributions, scaled by `target_scale`, is not a difference in predicted sales. `incrementality.py` never references `link` and no inverse link is applied anywhere in the module.

The paper flags this exact failure mode:

> If a transformation of sales, such as log transformation, is used as the response variable in (7), a corresponding inverse transformation should be applied to get the predicted sales.

`Incrementality` is constructed unconditionally at `mmm.py:2960`, regardless of link.

## Problem 2: the per-channel readout relies on additive separability

`_build_counterfactual_scenarios` applies `counterfactual_spend_factor` to **all channels at once** (`incrementality.py:719`, indexing only the date axis), then reads per-channel differences off the `channel` dim of the result. That shortcut is valid only because `channel_contribution[:, c]` depends solely on `x[:, c]` and the response is additive in those columns, so one perturbation yields every channel's answer.

Under a non-linear link this collapses. `exp(mu) - exp(mu - v_m)` depends on the intercept, seasonality, controls, and every other channel. A single all-channels perturbation cannot be decomposed per channel, so the log-link version needs one counterfactual per channel.

The paper's own caveat is the thing that stops holding:

> As the model in (7) is additive in media effects, ROAS_m does not depend on other media channels except the mth channel or on control variables.

Consequence to document, not to fix: per-channel ROAS under a log link is baseline- and control-dependent, and per-channel incrementals will not sum to total media incrementality. That overlap is already documented for `compute_mean_contributions_over_time` in the multiplicative notebook.

## The machinery for the fix already exists

This is not a from-scratch problem. The repo already implements the correct estimand elsewhere:

- `LogLinkSpec.create_media_contribution_deterministic` (`link.py`) registers `total_media_contribution_original_scale` as `exp(mu) * s - exp(mu - mu_media) * s` — the response-scale counterfactual, differenced per draw.
- `MMM.compute_counterfactual_contributions_dataset` (`mmm.py:1282`) does the per-component version, `inv(mu) * s - inv(mu - v_j) * s`, inside each draw before averaging.
- `LinkSpec.inverse_link` / `original_scale_transform` give the link-generic hook.
- `LinkSpec.mean_correction` handles median vs mean: `exp(mu)` is the LogNormal median, the paper's `Y_hat` is predicted sales.
- `add_original_scale_contribution_variable` already warns that per-component `*_original_scale` variables are multiplicative factors, not additive contributions, under a log link.
- `BudgetOptimizer` already treats the response variable as configurable (`response_variable="total_media_contribution_original_scale"`), rather than hardcoding a linear-predictor node.

So the work is mostly wiring `incrementality.py` into conventions the codebase has already settled on.

## Proposed change

**Step 1, immediate:** fail loudly. Raise `NotImplementedError` in `Incrementality` (or at the `mmm.incrementality` property, `mmm.py:2960`) when `link != identity`, with a message pointing at `compute_counterfactual_contributions_dataset` for the log-link case. Silent wrong numbers in a ROAS estimator are worse than a hard error.

**Step 2, generalize:**

1. Stop hardcoding `response_variable="channel_contribution"`. Evaluate on the response scale via `link_spec.inverse_link(mu) * target_scale`, mirroring `BudgetOptimizer`'s configurable `response_variable`.
2. One counterfactual evaluation per channel under a non-linear link, keeping the current single-pass fast path for the identity link where separability holds.
3. Difference inside each posterior draw, then aggregate, so credible intervals stay correct (same argument as the `compute_counterfactual_contributions_dataset` docstring).
4. Expose `central_tendency: {"median", "mean"}` and route "mean" through `LinkSpec.mean_correction`.
5. Document the non-additivity: per-channel ROAS is baseline-dependent and per-channel incrementals need not sum to the total.
6. Add log-link cases to `tests/mmm/test_incrementality.py`. A useful invariant: with a single channel, per-channel incremental should equal the total media counterfactual from `total_media_contribution_original_scale`.
7. Optionally align the carry-out mask with the paper's `t1 + L - 1`.

## On `pymc.do`

Scoping claim, not a rejection: `pm.do` is the right abstraction for interventions on endogenous nodes, and it is not what fixes the link bug this issue is about. Those are separate things and an earlier version of this section conflated them. Revised after reviewer feedback.

**What `pm.do` does not fix.** The defect here is the readout scale: differences are taken in linear-predictor space and never passed through the inverse link. `do(channel_data=...)` returns the same wrong-scale number with a nicer API. Separately, `channel_data` is a `pm.Data` container, so swapping it *already is* the intervention on that node. `extract_response_distribution` conditions the graph on posterior draws, swaps `channel_data` for a batched `xtensor` input, and `vectorize_graph` compiles one evaluator reused across all windows and both baseline and counterfactual. For an exogenous input, `pm.do` buys no correctness there.

**Correction to an earlier claim in this issue.** This section previously said a `pm.do` rewrite "would cost performance and break the vectorized window machinery." That is too strong. `pm.do` defaults to `make_interventions_shared=True`, so a constant intervention lands as a shared variable in the returned model, and the returned graph can be vectorized over intervention values exactly as today. `pm.do` and `vectorize_graph` compose; the choice is not one or the other.

**Where `pm.do` earns its keep.** Cutting the parents of an *endogenous* node, which the `pm.Data` swap structurally cannot express. Setting spend to zero is not the same estimand as setting a channel's contribution to zero: with `time_varying_media`, `channel_contribution = baseline_channel_contribution * media_temporal_latent_multiplier` (`mmm.py:2301-2322`), and more generally a saturation function need not map zero spend to zero contribution. `pm.do` on `channel_contribution` states "remove this channel's effect" directly, which is the estimand `compute_counterfactual_contributions_dataset` targets. The same applies to mediation: holding a mediator or a latent demand pool at its factual value while perturbing upstream spend is a direct/indirect decomposition that no amount of `channel_data` swapping can state, and it is the case the DAG-aware work in #2814 walks into.

That belongs in a follow-up, tracked separately, because it changes *which* counterfactual is computed rather than whether the result is on the response scale.
## Related

- #2330 — decouple incrementality from MMM implementation details
- #2484 — make interventional contributions the default recommendation

Neither addresses link functions.

Line references are against `d402988`.

daimon-pymclabs:
## Description

`Incrementality` implements ROAS (eq. 10) and mROAS (eq. 11) of [Jin, Wang, Sun, Chan & Koehler (2017)](https://storage.googleapis.com/gweb-research2023-media/pubtools/3806.pdf) by generating counterfactuals. Both the reduction and the per-channel readout assumed the response is **additive in the media contributions**. Under `link="log"` — the multiplicative MMM — and under a mediated `mu_effect` — the funnel MMM — that assumption is false, and the module ran without error and returned numbers that were not incremental response. No guard, no warning, no test coverage.

This PR makes the module correct outside the additive case, and separates the three concerns that were tangled in the single-file implementation: *which node is intervened on*, *how far in time the intervention reaches*, and *how the perturbed graph is reduced to an increment*.

It has grown well past its original scope. Three PRs have been merged into this branch, each reviewed on its own:

| | scope |
|---|---|
| #2815 | mediated `mu_effects` — spend reaching the response through a funnel |
| #2817 | spend-probe correctness, direct-path measurement, `spend_reach.py` split |
| #2829 | route every intervention through `pymc.do`; admit endogenous targets |

The sections below describe the merged result. The individual PRs remain the readable commit-by-commit history.

---

## 1. Link-aware reduction

Two things were wrong, and only the first is obvious:

1. **The counterfactual was reduced in linear-predictor space.** The module differences `channel_contribution` and multiplies by `target_scale`, but `channel_contribution` lives in `mu = base + sum_c v_c`. Under a log link the response is `exp(mu) * s`, so a sum of log-space differences is not a difference in predicted sales. The paper flags exactly this: *"If a transformation of sales, such as log transformation, is used as the response variable in (7), a corresponding inverse transformation should be applied to get the predicted sales."*

2. **The per-channel readout relied on additive separability.** One counterfactual scales all channels at once and per-channel increments are read off the `channel` dim. That is valid only because the response is additive in those columns.

The key observation is that (2) does **not** require per-channel counterfactual evaluations for a plain MMM. Because the linear predictor is additive in the channel contributions and `v_c` depends only on channel *c*'s spend, perturbing channel *m* alone gives

```
exp(mu_cf) = exp(mu) * exp(delta_m),    delta_m = v_cf_m - v_m
```

so the increment is

```
sum_t [Yhat_cf - Yhat] = sum_t Yhat_t * (exp(delta_tm) - 1)
```

where `Yhat = exp(mu) * s` is the baseline response-scale prediction. Every ingredient is already available: `delta` from the single existing all-channels evaluation, `Yhat` from the model's own `{output_var}_original_scale` deterministic. For a plain separable model nothing extra is compiled and the scenario count is unchanged; what changes is the *reduction*, not the counterfactual. The same expression covers the identity link, where `inv = id` makes the base term cancel and it collapses to `s * sum_t delta`.

The link-dependent step is isolated behind one small abstraction rather than threaded through the counterfactual code as conditionals:

```
IncrementalReducer (ABC)
  └── counterfactual_minus_baseline(delta) -> sum_t [Yhat_cf - Yhat]
        ├── IdentityLinkReducer   delta.sum("date") * scale
        └── LogLinkReducer        (Yhat * expm1(delta)).sum("date")
```

- Subclasses map one-to-one onto `LinkSpec` implementations. `_build_reducer` runs **before** the graph is compiled, so an unsupported link raises `NotImplementedError` in milliseconds rather than silently reusing the additive reduction.
- The reducer owns the return to original units, so the blanket `result * target_scale` at the end of the period assembly is gone. The ratio methods (`contribution_over_spend`, `spend_over_contribution`, `marginal_contribution_over_spend`) only divide by spend and needed no changes.
- The increment is formed per posterior draw and only then aggregated, so credible intervals are correct.
- `expm1` rather than `exp(x) - 1`: marginal incrementality perturbs spend by 1 %, which makes `delta` small and the subtraction cancellation-prone.

### `central_tendency`

Added to the public methods, defaulting to `"median"`. Under a log link `exp(mu) * s` is the *median* of the `LogNormal` response; `"mean"` routes through `LinkSpec.mean_correction` (`exp(sigma**2 / 2)`), which multiplies the baseline weight linearly and so factors straight through the sum. Ignored under an identity link. A fixed-sigma `LogNormal` has no `y_sigma` to build the correction from and fails with a clear message.

## 2. Mediated effects (channel-dependent `mu_effects`)

Spend does not always reach the response through `channel_contribution` alone: a funnel `mu_effect` can read `channel_data` itself, and then part of the incremental response travels through the effect. This PR includes that path rather than reporting the direct path as if it were the total.

- **Effects opt in** by returning an `IncrementalitySpec` from `MuEffect.incrementality_spec()`. An effect that depends on `channel_data` and has not opted in raises `NotImplementedError`; spend-independent effects (trends, events, seasonality) are part of the baseline and are skipped without being asked anything.
- **One counterfactual per channel** when a mediator is present: a funnel sums channels *inside* a nonlinear transform, so no per-channel column survives a single all-channels perturbation. The separable single-scenario fast path is kept for plain models.
- **The window is measured, not declared.** A mediated path that chains a second adstock behind the model's own outlives it, so the evaluation window is sized from a probe of the compiled graph (§4). Declarations are honoured when wider than the measurement and rejected when narrower.
- `Δmu = Δv_m + Σ_j Δe_j` is handed to the same reducers, which stay link-only and know nothing about mediation.
- `compute_joint_incremental_contribution` was added for the "how much does media drive in total" question, which under a non-additive response is a **different estimand** from summing the unilateral per-channel numbers. The module docstring spells out why neither is a bug and which question each answers.

## 3. Interventions via `pymc.do` (#2829)

The counterfactual is now stated explicitly as an intervention. The evaluator composes three steps, in this order:

```
do(model, {target: intervention})            # the intervention
  -> extract_response_distribution(...)      # condition on posterior draws
  -> vectorize_graph(...)                    # batch over scenarios
```

`do` states *what* is intervened on; the batch vectorization states *how many* values it takes. The split is forced by `pm.do` itself: an intervention must carry the target's own dims, so the scenario axis can only be introduced afterwards.

Two keyword-only parameters on `CounterfactualEvaluator`, both defaulting to the pre-existing behaviour:

- `intervention_target: str = "channel_data"` — any data variable or deterministic with a leading `date` dimension. Random variables are refused: observed ones because `do` would silently delete the likelihood, free ones because a parameter is conditioned on, not intervened on.
- `intervention_mode: "replace" | "scale"` — `"replace"` substitutes the evaluator's input for the target (the spend path; the scenario array *is* the intervention). `"scale"` multiplies the target's factual value by the input, symbolically, because an endogenous target's factual values differ per posterior sample: a mask of ones is the baseline, a mask of zeros removes the target's effect.

Which node is intervened on decides the estimand: `channel_data` answers the **spend** counterfactual (total effect through adstock, saturation and mediators); a zero mask on `channel_contribution` answers **effect removal**. The module docstring documents when the two coincide (`f(0) = 0`, no time-varying media, factor 0) and why they must differ for fractional factors. Endogenous targets are not yet wired into the `Incrementality` public API — that is the remaining half of #2822.

## 4. Module layout

The work split one 1.2k-line file into three modules with one-way dependencies:

| module | lines | owns |
|---|---|---|
| `counterfactual.py` | 1167 | the intervention and its evaluation |
| `spend_reach.py` | 851 | how far a change in spend reaches, and whether the accounting closes |
| `incrementality.py` | 1731 | periods, spend, estimands, link dispatch, public API |

- **`counterfactual.py`** — `EvaluationWindows`/`PeriodWindow` (which dates each period is evaluated over and which are summed), `CounterfactualScenarios` (perturbed spend plus row bookkeeping), and `CounterfactualEvaluator` (`do` on the intervention target, then one compiled batched evaluator over every node the counterfactual reaches, with date-indexed `pm.Data` inputs discovered from the graph and cut to windows in lockstep, and memory-bounded batching). Knows nothing about estimands, links or ROAS.
- **`spend_reach.py`** — `SpendProbe`: single-date spend perturbations on the untruncated axis, from which two guards are read. `measure()` establishes how far in time a change in spend moves each evaluated node (a custom adstock/saturation that reduces over `date` forces full-axis evaluation; a mediated tail widens the window), choosing probe dates **per channel** so flighted campaigns and disjoint spend supports cannot blind it. `assert_increment_is_complete()` verifies the evaluated nodes account for the whole move in the linear predictor, so a misreported or unattributable spend path fails loudly instead of silently shrinking the increment. When nothing is probeable (spend only on the first date), evaluation falls back to the full axis with a warning that also says the completeness check went unverified.
- **`incrementality.py`** — keeps everything it already owned: period grouping, spend aggregation, the ratio methods, plus the new reducers and link dispatch.

`extract_response_distribution` (`pytensor_utils`) now accepts a list of variables and extracts them in a single pass, so `channel_contribution`, each effect's contribution and the linear predictor share one conditioned, vectorized graph.

### Backward compatibility

`Incrementality` is the only public name `incrementality.py` exported before this PR, and all of its public methods keep their names, parameter names, order and defaults. The one signature change is an appended keyword with a default (`central_tendency="median"`). `counterfactual.py` and `spend_reach.py` are new files, not extractions of anything that existed, so no import path that worked before stops working. All eight notebooks that use the module go through `mmm.incrementality.contribution_over_spend(...)`; none imports from the module directly.

`pymc_marketing.mmm.__all__` gains `Incrementality` and `IncrementalitySpec`.

## Validation

The algebra was checked against an oracle that shares no code with the module: one `sample_posterior_predictive` per (period, channel) on `y_original_scale`, i.e. a genuine per-channel spend counterfactual evaluated on the response scale.

| check | result |
|---|---|
| formula vs per-channel oracle (2 channels, zero-out) | max rel. err. **1.1e-15** |
| single channel vs `total_media_contribution_original_scale` | max rel. err. **3.9e-16** |
| marginal (factor 1.01) vs oracle | max rel. err. **1.2e-13** |
| mediated funnel path vs oracle, after window clamping | max rel. err. **2.0e-15** |
| pre-fix behaviour vs oracle | wrong by orders of magnitude |

**Identity link is algebraically unchanged.** Summing differences and differencing sums are the same operation, so the only difference is float accumulation order; worst observed relative error across the 14 pre-existing ground-truth parametrisations is 2.2e-11, and the ground-truth tolerance is tightened from `1e-4` to `1e-8` accordingly. All of them pass untouched.

Two behaviours that were approximate before are now exact rather than merely closer:

- **Window clamping.** Evaluation windows were zero-padded on the left when they reached before the start of the data. For spend that padding is inert, which is why it never mattered; a funnel's contribution at zero spend is its own intercept, so the padding injected synthetic history. Clamping to the fitted dates is provably a no-op on the separable path.
- **`effective_l_max` is measured for every model**, not only mediated ones, so a custom adstock or saturation that reaches further than `adstock.l_max` no longer silently truncates the sum.

## Tests

The suite mirrors the package's module structure, with the probe helpers defined in `test_spend_reach.py` and imported by `test_incrementality.py` — the same one-way direction as the modules themselves.

- **`tests/mmm/test_incrementality.py`** — the reducers in isolation (`TestIncrementalReducers`, including the `expm1` precision case); `TestLogLinkIncrementality` against the per-(period, channel) oracle across fixtures × frequency × carryover × factor, including `monthly` with `include_carryover=False`, `num_samples` against the oracle restricted to the same selected draws on a **two-chain** posterior (a draw misalignment between the baseline weights and the evaluated graph would be a silently wrong number), `central_tendency="mean"` against the corrected oracle, monthly and panel ROAS wrappers, `start_date`/`end_date` under a log link, and negative controls (the pre-fix linear-predictor computation must disagree; a single channel must reproduce `total_media_contribution_original_scale`). `TestMediatedMuEffects`/`TestMediatedEdgeCases` cover the funnel path: oracles under both links, joint estimand, refusals (no opt-in, duck-typed spend effect, misreported contribution, under-declared carryover), full-axis effects, dark-start/dark-channel probe placement, two mediators, `panel × log link × mediation`, and a materiality check that dropping the mediated path both trips the completeness guard and understates the increment.
- **`tests/mmm/test_spend_reach.py`** — `SpendProbe` against analytic nodes: reach measurement, probe placement (dark dates, per-channel probes, unprobeable axes), completeness accept/refuse cases, and the fallback warning admitting that completeness goes unverified.
- **`tests/mmm/test_counterfactual.py`** — windows, scenario building, batching invariance, date-indexed auxiliary inputs cut along the axis they call `date` regardless of layout, plus the `do`-specific suites: `TestInterventionValidation` (every refusal — unknown target, observed RV, free RV, frozen target, integer dtype, non-leading `date` dim, unknown mode, response independent of target), `TestEndogenousIntervention` (toy model with `f(0) != 0`: replace-mode identity, ones-mask == baseline, zeros-mask == exactly zero, and the exact algebraic divergence between spend-zeroing and effect-zeroing with the direction pinned), and `TestSpendMaskEquivalence` (real MMM fixtures: with identity link, `f(0) = 0` and no time-varying media, `do(channel_contribution * 0)` reproduces the spend-zeroing increment to `rtol=1e-6` — #2822's sharpest test — while at factor 0.5 the two estimands separate in the direction concavity dictates).
- **`TestRoasRecovery`** (`@pytest.mark.slow`) — end-to-end ROAS recovery: data simulated from the `mmm_example` notebook's data-generating process (library `geometric_adstock` and `logistic_saturation`), true ROAS computed from the DGP's own counterfactual (the `mmm_roas` notebook construction), and the three `mmm_multiplicative` variants (linear, log, log-log) each fit with real NUTS and required to recover it via `contribution_over_spend`, agree with each other, and — as an end-to-end negative control — the pre-fix computation required *not* to.

Fixtures are module-scoped where read-only, and the probe-derived `effective_l_max` is cached per model, which together remove most of the repeated model builds and graph compilations. The pre-`do` counterfactual suites pass **without a single edit** after #2829, which is the behaviour-preservation gate for the default spend path.

## Notebooks

- **`mmm_multiplicative.ipynb`** — gains a link-aware ROAS comparison across the three model variants, a "how contributions are computed" explainer, and a note on reading the multiplicative decomposition plot; summary metrics now go through `pymc_marketing.metrics`.
- **`mmm_funnel_mueffect.ipynb`** — direct/indirect/total effect section now uses the mediated incrementality path.

Both were re-executed for real (nutpie, full fits) after the `do` refactor with no content changes: all ROAS and counterfactual numbers reproduce the committed outputs, the only numeric movement being one metrics line in the multiplicative summary within sampling noise (nmae 0.0385 → 0.0384, crps 140.41 → 140.36). Both also pass the mock-mode CI runner.

## Deliberately not in this PR

- **The `t1 + L - 1` off-by-one.** The paper sums to `t1 + L - 1`; the code masks `dates <= t1 + l_max * freq_offset`, one period further. `geometric_adstock` uses `lags = arange(l_max)`, so that extra period contributes exactly zero. Harmless, but it is not the paper's index. Changing it would alter existing numbers for a non-reason, so it should be its own PR.
- **Holding a mediator at its factual value** (the direct-effect estimand). Needs simultaneous interventions, which generalizes the single `(target, mode)` pair to a mapping. Tracked in #2822.
- **Endogenous targets in the `Incrementality` public API.** The evaluator accepts them; the windowed, period-wise public methods still only intervene on spend. Also #2822.
- **Induced-spend denominators for a mediated ROAS**, where a channel's own spend is not the only spend its counterfactual moves.
- **A guard for channel-separable `channel_contribution`.** The separable fast path assumes `v_c` depends only on channel *c*'s spend. A custom transform that mixes channels would pass the completeness check (which compares the channel-*sum*) while corrupting the per-channel readout; the assumption is documented, and a per-channel probe guard is a candidate follow-up.
- **`compute_counterfactual_contributions_dataset` hardening.** Its `central_tendency` is unvalidated (a typo silently yields median) and its non-log branch is a bare `else` — the same silent-additive hazard `_build_reducer` now fails loudly on. Worth its own small PR.
- **`validate_likelihood_compatibility` under the identity link.** `MMM(link="identity", model_config={"likelihood": Prior("LogNormal", ...)})` is accepted, on the grounds that "the additive decomposition does not depend on the distributional form" — which is false when `mu` is not the response mean. Pre-existing, and it breaks the contribution decomposition generally rather than incrementality specifically, so it wants its own issue.

## Notes for review

- Two comments in `incrementality.py` claimed `channel_data`'s non-date axes are `(channel, *custom_dims)`. For panel models the layout is `(date, *custom_dims, channel)` — channel is **last**. Nothing depended on the wrong ordering, but the comment cost a debugging cycle while writing the panel test, so it is now accurate.
- `mypy` reports one error in `incrementality.py` (`self.idata = idata`, line 478). It is pre-existing and untouched by this diff.
- Related, none superseded: #2330 (decouple incrementality from MMM internals), #2484 (make interventional contributions the default), #2544 (use incrementality for ROAS in budget plots).

## Related Issue
- [x] Closes #2811
- [x] Related to #2822

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2813.org.readthedocs.build/en/2813/


juanitorduz:
Pushed a follow-up in 924dbb44 addressing the review and the Bugbot comment.

- **Panel mean-scale fix:** `central_tendency="mean"` raised for panel log-link models, because the LogNormal correction carries the model's custom dims but sample stacking was flattening it to `(sample,)` alone. Stacking now infers and preserves every non-sample dimension and its coordinates, so the correction lands per custom-dim cell.
- **Wider log-link oracle:** the per-channel ground-truth test now covers controls, time-varying media, panel monthly/marginal/no-carryover, and spend factors of `0.5` and exactly `1.0`. Added a test pinning the direction of the multiplicative overlap, so per-channel increments are expected to overstate the joint knock-out rather than sum to it.
- **Sharper subsampling test:** it previously used a duplicated posterior, which made it blind to a chain-major/draw-major mix-up. The chains now differ and the selected draws are checked explicitly.
- **Docs:** corrected the dimension-order claims that still said `(channel, *custom_dims)`, and spelled out that a zero spend factor is a spend intervention, not removal of a channel's effect.

Incrementality tests and pre-commit are green locally.

juanitorduz:
From https://github.com/pymc-labs/pymc-marketing/pull/2815

---

Builds on #2813; targets `claude/incrementality-link-functions`. Follows up the question raised there about `mu_effects` that depend on `channel_data`.

## The gap

`Incrementality` assumed a spend counterfactual reaches the response only through `channel_contribution`. A `mu_effect` can read `channel_data` itself — the funnel mediators in #2769 do, via `mmm.channel_data_scaled` — and then part of the incremental response travels through the effect rather than through `channel_contribution`. That part was silently dropped, so the direct path was reported as if it were the total.

On the funnel fixture added here, ignoring the mediated path understates the per-channel increment by **6.3%** and **17.7%** (10.4% of the total). Nothing warned; the numbers just came back smaller.

## What changed

**Mediated effects join the increment, additively in the linear predictor.** Since `mu` is the sum of its components, so is the perturbation:

```
Δμ_t = Δv_{t,m} + Σ_j Δe_{t,j}
```

which goes to the same `IncrementalReducer` as before. The reducer hierarchy from #2813 is untouched: it converts a change in the linear predictor into a change in the response and does not care how many nodes that change was collected from. The log-link algebra survives verbatim.

**Effects opt in, or incrementality refuses.** New `MuEffect.incrementality_spec()` returns an `IncrementalitySpec` or `None`. An effect whose contribution depends on `channel_data` and returns `None` raises `NotImplementedError` rather than being dropped. Effects that do *not* depend on spend — trends, events, seasonality — are part of the baseline, cancel in the difference, and are skipped without being asked anything, so no existing effect has to change.

The spec carries one field, `additional_carryover_lags`. Date-indexed `pm.Data` an effect reads is discovered from the graph rather than declared, so an effect cannot forget to mention one. How far in time the effect propagates a change in spend is not discoverable that way — a funnel chains a second adstock behind the model's own — so it is declared, and the evaluation window is sized for the longest path spend can take.

**One counterfactual per channel when an effect is present.** Without mediation a single all-channels perturbation suffices, because `v_{t,c}` depends on channel *c*'s spend alone and column *m* of that one evaluation *is* channel *m*'s counterfactual. A funnel sums over channels *inside* a nonlinear transform, so no per-channel column survives. Scenarios become (period, channel) pairs; in the separable case every channel of a period still points at the same row, so that path costs exactly what it did before.

**Evaluation windows are clamped to the fitted dates instead of padded out.** This was a real bug, not a tidy-up. Windows were zero-padded on the left when they reached before the start of the data. For spend that is inert — zero in, zero out, and the graph's own adstock pads with zeros there anyway — but a funnel's contribution at zero spend is its own intercept, so the padding injected synthetic history that carried into the window through the effect's adstock. It cost 0.9% of relative error, roughly a tenth of the mediated effect being measured. Clamping instead makes the windowed evaluation exact: every evaluated date sees the same inputs it would on the full axis. The separable path is unaffected (padding there was already inert), and windows get shorter, so it is marginally cheaper.

**A joint estimand.** `compute_joint_incremental_contribution` perturbs every channel at once. Per-channel numbers are leave-one-out and do not sum to it whenever the response is not additive in the channels — under a log link, or with an effect that mixes channels — because interaction mass is counted by every channel that touches it. Both are now first-class and documented as answering different questions; summing the per-channel ones answers neither. Under `link="identity"` with no mediation they coincide exactly, and there is a test asserting that.

**`extract_response_distribution` accepts a sequence of names.** `channel_contribution` and a mediated effect read the same spend through the same adstock; extracting them together keeps that subgraph shared instead of conditioning, rewriting and vectorizing it once per node. A single name keeps the old scalar return type, and `@overload`s make the distinction visible to callers.

## Validation

Against an independent oracle — one `sample_posterior_predictive` per (period, channel) on the response scale, which never forms a linear-predictor difference:

| check | result |
| --- | --- |
| funnel, `link="log"`, max relative error per draw | 1.98e-15 |
| same, before the window fix | 8.9e-03 |
| funnel, `link="identity"` | matches to 1e-6 |
| joint estimand vs all-channels-at-once oracle | matches to 1e-6 |
| identity link, no effects: `loo.sum("channel")` vs joint | matches to 1e-10 |
| 14 pre-existing ground-truth parametrisations | unchanged, all pass |

New tests cover the mediated algebra under both links and both `counterfactual_spend_factor` regimes, at `all_time` and `monthly` frequency (the latter is what exercises interior windows); the joint estimand; the leave-one-out overcount; a magnitude assertion that the mediated path is material, so a regression that quietly dropped it again would fail rather than pass; the `NotImplementedError` for an effect that has not opted in; the skip for an effect that ignores spend; window clamping; per-channel scenario construction; and the multi-output extraction.

## Notes

- `mypy` still reports one pre-existing error in `incrementality.py` (`self.idata = idata`). Untouched by this diff, so left alone.
- The estimand question is settled by exposing both rather than by choosing: per-channel stays leave-one-out and keeps its name, the joint gets its own method.
- Not in scope: interventions on a channel's *contribution* rather than its spend (`pm.do` territory, where zero spend need not mean zero contribution), and induced-spend denominators for a mediated ROAS, where a channel's spend is not the only spend its counterfactual moves.

----
📚 Documentation preview 📚: https://pymc-marketing--2815.org.readthedocs.build/en/2815/

juanitorduz:
Pushed a follow-up in ce614679 addressing both review blockers, each verified before fixing.

- **Docs build fix:** the recursive api autosummary picked up `ScenarioKey`, a bare generic-tuple alias autodoc cannot format (`Estimand`, a `Literal` alias beside it, renders fine). Dropping it from `__all__` turned out to be insufficient — autosummary collects public module attributes by name, not from `__all__` — so the alias is now private (`_ScenarioKey`); it is the key type of an internal mapping, not a user-facing name. A from-scratch `-W` build emits no warning for the module and the generated page lists only `Estimand`.
- **Per-channel spend probe:** `_select_probe_index` chose one date by total spend, so a `mu_effect` reading a subset of channels could sit dark at the probed date — reach measured as none, window sized from the model's own adstock alone, mediated tail silently cut, completeness comparing zero against zero. `SpendProbe` now selects per channel: a single covering probe when one date moves every spending channel (dense spend stays exactly as cheap as before), otherwise one probe per channel, with each node's reach taken as the widest any probe demonstrates and the completeness identity checked under every probe. A channel whose only spend sits at the excluded first date disables the probe and falls back to full-axis evaluation, as an all-dark axis already did.
- **Regression tests:** unit tests for the selection (covering fast path, disjoint supports, all-zero channel ignored, first-date-only spend refused), plus the A/B distilled into a fixture pair — a funnel mediator reading `channel_1` alone with an empty `IncrementalitySpec` (the documented "measure it" default), control with dense spend vs. treatment with disjoint channel supports. The two arms now agree on measured reach, window length, and the oracle.

Incrementality tests (175), pre-commit, and a clean `sphinx-build -W` are green locally.

juanitorduz:
The `mmm_multiplicative` notebook now closes with a **ROAS Comparison** section — the docs-facing counterpart of `TestRoasRecovery`: three links, one estimand, evaluated with `mmm.incrementality.contribution_over_spend` and checked against the true ROAS of the data-generating process.

Since the notebook's dataset is the `mmm_example` DGP (scaled by 1000), the true ROAS is derivable in-notebook by applying the known adstock/saturation transforms to the observed spend: **x1 = 4967.8, x2 = 2274.9**. Against that truth, from the committed run (median, 94% HDI):

| Model | x1 | x2 |
|---|---|---|
| True ROAS | 4967.8 | 2274.9 |
| Additive | 4916.8 [4307.3, 5495.0] (−1.0%) | 2320.7 [2105.8, 2510.3] (+2.0%) |
| Log (Multiplicative) | 4744.2 [4276.0, 5302.4] (−4.5%) | 2476.8 [2221.5, 2718.7] (+8.9%) |
| Log-Log | 3900.2 [3511.5, 4342.3] (−21.5%) | 2367.9 [2113.7, 2633.5] (+4.1%) |

The section shows per-channel forest plots of the three posteriors against the truth line, plus this table. The prose makes the two points the PR exists for: the estimates are comparable across links *because* the counterfactual is reduced on the response scale, and a comparable estimand doesn't repair misspecification — it exposes it (the log-log model's `β·log(1+x)` can't reproduce the logistic saturation of the DGP, and its x1 interval honestly excludes the truth, consistent with the tolerances in `TestRoasRecovery`).

Workflow: developed via a jupytext round-trip, smoke-tested with the mock runner, then a full clean re-execution (clear outputs → `jupyter execute --inplace`, 4 chains × nutpie per model, zero divergences in all three fits). Pre-commit (ruff, watermark, nbformat, gallery sync) is green.

pymc_marketing/mmm/incrementality.py
```diff
@@ -215,6 +378,145 @@ def __init__(
...
+        """
+        if central_tendency == "median":
+            return xr.DataArray(1.0)
+
+        correction = self.model._link_spec.mean_correction(
+            posterior, self.model.output_var
+        )
+        if "chain" not in correction.dims:
+            return correction
+        return self._stack_samples(correction, dims=())
```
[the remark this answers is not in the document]
daimon-pymclabs: Addressed in the follow-up (head `48b6b34`). `_stack_samples` no longer takes a `dims=()` argument: it derives `other_dims` from the array itself, transposes `("chain", "draw", *other_dims)`, and reshapes only the leading two axes, carrying each remaining dim's coordinates through. A panel `y_sigma` with a `country` dim survives the flatten as `("sample", "country")`, so the correction broadcasts by name against the perturbation per cell rather than raising.

Covered by `test_panel_mean_central_tendency_corrects_each_custom_dim_cell`, which asserts each custom-dim cell gets its own sigma rather than a pooled one.

docs/source/notebooks/mmm/mmm_funnel_mueffect.ipynb
```diff
@@ -109,10 +109,10 @@
    "id": "f5f77799",
```
timbo112711, line 70: <p>The second sentence here, </p><p> </p><blockquote>Returning an &nbsp;{class}<code>~pymc_marketing.mmm.additive_effect.IncrementalitySpec</code> from &nbsp;<code>incrementality_spec</code> says so, and the incrementality module then includes the &nbsp;mediated path in its ROAS numbers instead of reporting the direct path alone.</blockquote><p> </p><p>may need some adjustment. It's explaining what the return value signals and what the module does with it<em>, </em>but the linkage between the two clauses feels mechanical ("says so, and then..."). "says so" is vague and colloquial after a technical reference. It also forces the reader to mentally track back to what "so" refers to. </p><p> </p><p>Something like this could enhance the flow on the paragraph.</p><p> </p><blockquote>When <code>incrementality_spec</code> returns an {class}<code>~pymc_marketing.mmm.additive_effect.IncrementalitySpec</code>, the incrementality module includes the mediated path in its ROAS numbers rather than reporting the direct path alone.</blockquote>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3771943443'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='6'/>

timbo112711, line 70: <p>Is the indenting off here or does it just look that way in how ReviewNB renders it 😁</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3771943448'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='57'/>

docs/source/notebooks/mmm/mmm_funnel_mueffect_advanced.ipynb
```diff
@@ -73,7 +73,10 @@
     "data-generating process, a d-separation analysis of every specification, the six\n",
```
timbo112711: <p>Looks like this started on a new line when it should continue on from "cannot"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3771943497'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='1'/>

timbo112711: <p>Is this suppose to be all formatted in code or just plain text?? Looks to be rendering in code formatting 😄</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3771943502'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='121'/>

timbo112711: <p>Looks like the formatting here cuts to a new line right after "to"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3771943504'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='129'/>

timbo112711: <p>Same thing here, looks like the formatting cut to a new line mid sentence </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3771943506'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='160'/>

timbo112711: <p>First bullet needs to follow the indentation </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3771943510'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='161'/>

docs/source/notebooks/mmm/mmm_funnel_mueffect.ipynb
daimon-pymclabs: Agreed, taken verbatim. Rewritten as "When \`incrementality_spec\` returns an {class}\`~pymc_marketing.mmm.additive_effect.IncrementalitySpec\`, the incrementality module includes the mediated path in its ROAS numbers rather than reporting the direct path alone." Pushed in `6ad3ccb`.

daimon-pymclabs: Rendering artifact. The bullet continuations are indented two spaces, which is what markdown needs under a `- ` marker, and Sphinx/MyST reflows them into single paragraphs. ReviewNB shows each source line as its own line, so the 80-column hard wrap and the two-space continuation both become visible. No change needed.

docs/source/notebooks/mmm/mmm_funnel_mueffect_advanced.ipynb
daimon-pymclabs: Same ReviewNB rendering artifact. The markdown source is hard-wrapped at 80 columns, and a single newline is a soft break, so Jupyter and the Sphinx docs build render "A model that cannot see the funnel has to put that mass somewhere" as one continuous sentence. ReviewNB renders each source line literally, which makes the wrap look like a break.

daimon-pymclabs: Plain text in the built docs. That is a MyST `{note}` directive, which Sphinx renders as an admonition box. ReviewNB has no MyST support so it falls back to showing the fenced block as code. It is the house convention here, 29 notebooks in `docs/source/notebooks/` use the same directives.

daimon-pymclabs: Hard wrap again, after "a geo panel exists to / expose". Renders as one flowing paragraph in the docs build.

daimon-pymclabs: Same, this is the 80-column wrap showing through ReviewNB rather than a break in the rendered output.

daimon-pymclabs: Checked the raw source: every continuation line in that first bullet is indented exactly two spaces, matching the rest of the list, so it renders as one bullet. What you are seeing is ReviewNB printing the leading indentation literally instead of reflowing.

docs/source/notebooks/mmm/mmm_funnel_mueffect.ipynb
juanitorduz: Reworded in 6ad3ccb3 with the suggested phrasing: the cell now leads with "When `incrementality_spec` returns an ... `IncrementalitySpec`, the incrementality module includes the mediated path in its ROAS numbers rather than reporting the direct path alone."

juanitorduz: The indentation itself was valid markdown, but the cell was hard-wrapped at a fixed column and ReviewNB renders those wraps literally, which is what made it look off. Fixed in 7a573cf3: all prose in both funnel notebooks is now unwrapped (one line per paragraph/bullet), so the cell renders continuously.

docs/source/notebooks/mmm/mmm_funnel_mueffect_advanced.ipynb
juanitorduz: That was a hard line wrap in the markdown source, which ReviewNB renders as a literal break. Fixed in 7a573cf3 by unwrapping all wrapped prose in the notebook, so the sentence now flows through "cannot see".

juanitorduz: Intentional: the cell is a MyST `{note}` directive (triple-backtick fence), which renders as a note admonition box in the built docs; ReviewNB shows the raw fence as code. This matches the convention used across the other mmm notebooks. Rendered version: https://pymc-marketing--2813.org.readthedocs.build/en/2813/notebooks/mmm/mmm_funnel_mueffect_advanced.html

juanitorduz: Same hard-wrap artifact as above; fixed in 7a573cf3 by unwrapping the prose, so the sentence no longer breaks after "to".

juanitorduz: Same hard-wrap artifact; fixed in 7a573cf3 by unwrapping the prose in this cell.

juanitorduz: The first bullet had hard-wrapped continuation lines with ragged indentation. In 7a573cf3 every bullet is a single line, so the indentation issue is gone.

daimon-pymclabs:
On the two remaining ReviewNB threads in \`mmm_multiplicative.ipynb\` (cells 74 and 76, the **Summary Metrics** intro and the closing link-choice paragraph): no change needed there, and unlike the funnel notebooks this one was never hard-wrapped.

I checked the raw source at your review commit \`e1081597\` and again at head \`7a573cf3\`: both cells store one line per paragraph (cell 74 is 3 source lines, cell 76 is 9, and every one of those is a whole paragraph, blank line, or numbered bullet). There is no mid-sentence newline and no line holding a lone period, so what you are seeing is ReviewNB soft-wrapping the long line in its own preview column rather than anything in the notebook.

Rendered version for comparison: https://pymc-marketing--2813.org.readthedocs.build/en/2813/notebooks/mmm/mmm_multiplicative.html

The hard-wrap issue you flagged was real in the two funnel notebooks, and that is what \`7a573cf3\` unwrapped.

juanitorduz (approved):

docs/source/notebooks/mmm/mmm_funnel_mueffect_advanced.ipynb
```diff
@@ -3,19 +3,14 @@
   {
```
jamespooley, line 3: <p>To help readers keep track of things, it might be better to have informative names always used for the naive models; so instead of <code>Naive B++</code> etc. you'd have something like:</p><p> </p><ul><li><code>Naive A (Mediator-as-Channel)</code></li><li><code>Naive B (Mediator-Omitted)</code></li><li><code>Naive B+ (Mediator-Omitted + Demand)</code></li><li><code>Naive B++ (Mediator-Omitted + Trend)</code></li><li><code>Naive C (Proxy-as-Channel)</code> or <code>Naive C (Search Volume-as-Channel)</code> or <code>Naive C (Indicator-as-Channel)</code></li></ul><p> </p><p>Something like this is used in tables later in the notebook, but keeping it consistent throughout should also help the flow by reducing a reader's memory tax.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3786156331'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jamespooley, line 3: <p>In general, my thinking is that since the notebook is so dense, <code>{important}</code>-style tl;dr "reorientation" boxes should start major sections (e.g., Head-to-head, ROAS, and Parameter recovery) with the key findings/takeaways at a high level.</p><p> </p><p>The mid-section ones are good for emphasizing when an argument/finding, but starting the appropriate sections with these would help further guide the reader</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2813/discussion/#comment-3786156337'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='145'/>

daimon-pymclabs:
@jamespooley both of these landed just after the merge, so I've opened #2880 rather than amending here.

On the naming: the notebook already had canonical labels that every table and figure prints verbatim (`Naive A (mediator as channel)`, `Naive B (mediator omitted)`, `Naive B+ (demand as control)`, `Naive B++ (trend as control)`, `Naive C (indicator as channel)`), and the prose was the part that had drifted to bare letters. I reused those exact strings rather than introducing the hyphenated variants, so prose and output labels now match character for character; that seemed closer to what you were after than a third scheme. They're in the model-structures table, the five fitting-section headings, and the first mention of each model in the head-to-head and bias discussions. Bare letters remain only as shorthand within a paragraph that has already named the model, since spelling out `Naive B++ (trend as control)` four times in one paragraph reads worse than the memory tax it saves. Say the word if you'd rather it were absolute.

On the reorientation boxes: added an `{important}` at the top of Parameter recovery, Head-to-head, and ROAS, each giving the section's finding before the detail. Mid-section boxes untouched.

Markdown only, no re-run needed.

PR #2814, opened 2026-08-04
docs(mmm): add advanced geo-level funnel MMM example (six models, one ground truth)

juanitorduz:
## Description

Adds a new MMM example notebook, **`docs/source/notebooks/mmm/mmm_funnel_mueffect_advanced.ipynb`**, and registers it in the docs gallery.

It is the follow-up to the existing [Funnel-Aware MMM via a custom `MuEffect`](https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_funnel_mueffect.html) notebook, scaled up to the setting practitioners actually meet: several upper-funnel channels feeding one shared demand pool across a geo panel, a mediator that is *itself* paid media, and a separate observational proxy for demand that causes nothing.

### The question it answers

Upper-funnel media creates demand that later converts through lower-funnel spend. When an MMM treats that lower-funnel spend as just another channel, or leaves it out, or controls for a *symptom* of demand instead, what happens to its ROAS estimates?

### What the notebook does

1. **Builds a synthetic geo panel with known ground truth.** Two upper-funnel channels (`tv_spend`, `social_spend`), a lower-funnel mediator that is paid media and converts (`lower_spend`), a branded-search indicator that tracks demand but causes nothing (`search_volume`), and media plans confounded with category demand through three separate strands: an annual seasonal cycle (absorbed by every model's Fourier basis), a growth trend (observable but carried by no standard specification), and correlated noise (observable by nobody).

2. **Interrogates the causal graph before fitting anything.** A `networkx` d-separation analysis enumerates all 43 backdoor paths, computes which conditioning set each specification implies, and *derives* rather than asserts each model's identification verdict, including all minimal adjustment sets.

3. **Fits six models to the same data**, differing in exactly one thing: what else they condition on and how:

   | Model | Beyond the two upper channels | How it treats the funnel |
   |---|---|---|
   | `Naive A` | `lower_spend` as a third **channel** | conditions on the mediator, blocking the indirect path |
   | `Naive B` | nothing | ignores the mediator; a backdoor stays open |
   | `Naive B+` | `category_demand` as a **control** | blocks that backdoor via the observed demand index |
   | `Naive B++` | the growth trend as a **control** | blocks only the trend strand |
   | `Naive C` | `search_volume` as a third **channel** | conditions on a *symptom* of demand |
   | `Funnel` | a custom `FunnelEffect` (`DataVarMuEffect`) | models the mediation explicitly |

4. **Scores all six against the truth** on parameter recovery, direct/indirect/total effect decomposition, ROAS nationally and per geo-channel cell, and predictive fit.

### Headline results

- All six models fit the target almost equally well (in-sample $R^2$ spans **0.979 to 0.987**) while their ROAS conclusions differ by tens of percent. **Predictive fit does not diagnose causal bias.**
- Mean absolute error in total ROAS: `Naive A` 18.7%, `Naive B` 33.2%, `Naive B+` 32.9%, `Naive B++` 29.2%, `Naive C` 24.9%, **`Funnel` 8.7%**.
- The funnel model is the only specification combining full coverage of the true total ROAS (6/6 geo-channel cells) with the narrowest intervals (mean 94% HDI width 0.397, against 0.455–0.703 for the naive models).
- Each naive model fails in its own characteristic way, and the d-separation analysis predicts most of it in advance. Where the graph's ranking and the data disagree, the notebook says so and explains why: a valid adjustment set tells you *which* variables you need, never *how* they must enter the model.

### Library features exercised

- `DataVarMuEffect` (`pymc_marketing.mmm.additive_effect`): the custom effect reads its data straight out of the training `xarray` Dataset, so only the genuinely custom part has to be written.
- `pymc.dims` xtensors: named broadcasting lets `("geo",)`, `("channel",)` and `("date", "geo", "channel")` combine without a single `reshape`.
- The incrementality module (`contribution_over_spend`), whose output the notebook cross-checks against a hand-rolled counterfactual to every printed digit.

The final section is explicit about the framework's limits: no built-in decomposition knows what is upstream of what, and the notebook documents exactly how the data-layer and counterfactual decompositions each handle a custom `MuEffect`.

### Notes for reviewers

- Everything is seeded off a single `seed` / `rng` pair at the top; the notebook is fully re-executed, with 0 error outputs and 1 divergence in 4000 draws (`Naive C`), max $\hat{R}$ = 1.00 across all six models.
- Every number quoted in the prose has been checked against the printed output it refers to.
- This supersedes #2769, which carries the review history but has become hard to follow after many rounds of back-and-forth. The content here is that branch's final state, rebased onto `main` as a single commit.

## Related Issue

- [ ] Closes #
- [x] Related to #2769

## Checklist


timbo112711:
@juanitorduz just reviewed the geo notebook. Overall really liked how you structured it. Very detailed and concise! 🙌🏻. Left a few minor comments in the notebook

juanitorduz:
@timbo112711 Thank you for your review! I addressed your comments https://github.com/pymc-labs/pymc-marketing/pull/2814/commits/b4f82baea8462e16ee7db78fc7881881a022ffab 

Would you mind taking a last look :)

timbo112711 (approved):
This is great work! Really enjoyed the detail. At the end I thought using `forest_panel` to compare the 6 models was super nice

juanitorduz:
Thank you @timbo112711 🫶 ! I will merge as it is and based on the feedback iterate on smaller PRs :)

PR #2750, opened 2026-07-21
fix(clv): correct daily lifetime value scaling

Sanjays2402:
## Description

Daily CLV horizons now use the average Gregorian month length (365.25 / 12) instead of 30 days, preventing annual estimates from being scaled to 360 days. The regression verifies that a 12-month daily horizon requests 365.25 days.

## Related Issue

- [x] Closes #2749
- [ ] Related to #

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [x] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [x] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--2750.org.readthedocs.build/en/2750/

PR #2820, opened 2026-08-05
fix(clv): render the CLV likelihood equations and correct daily/hourly CLV scaling

### LINKED ISSUE #2749: Revise `customer_lifetime_value` "Daily" Time Unit Scaling Factor
When estimating expected CLV, `future_t` is always specified in months, but can be scaled to weeks, days or hours with the `time_unit` parameter.  The value for "Daily" is currently 30 days per month, which will inflate CLV estimates because it only scales by 360 days per year instead of the full 365.25 (accounting for leap years.)

This is easily fixed by changing the value for "D" to 30.4375. "H" value should also be changed to 30.4375 * 24:

https://github.com/pymc-labs/pymc-marketing/blob/228b9bfe08a752ab64599ff0bf0c2b5e25aeefb6/pymc_marketing/clv/utils.py#L123

### LINKED ISSUE #2798: Fix LaTeX expressions in the CLV Documentation
The equations for these CLV likelihood functions have not been rendering properly since v0.18.0:

- [pymc_marketing.clv.distributions.BetaGeoNBD](https://www.pymc-marketing.io/en/latest/api/generated/pymc_marketing.clv.distributions.BetaGeoNBD.html#pymc_marketing.clv.distributions.BetaGeoNBD)
- [pymc_marketing.clv.distributions.ModifiedBetaGeoNBD](https://www.pymc-marketing.io/en/latest/api/generated/pymc_marketing.clv.distributions.ModifiedBetaGeoNBD.html#pymc_marketing.clv.distributions.ModifiedBetaGeoNBD)
- [pymc_marketing.clv.distributions.ParetoNBD](https://www.pymc-marketing.io/en/latest/api/generated/pymc_marketing.clv.distributions.ParetoNBD.html#pymc_marketing.clv.distributions.ParetoNBD)
- [pymc_marketing.clv.distributions.ShiftedBetaGeometric](https://www.pymc-marketing.io/en/latest/api/generated/pymc_marketing.clv.distributions.ShiftedBetaGeometric.html#pymc_marketing.clv.distributions.ShiftedBetaGeometric)

daimon-pymclabs:
Closes #2798.
Closes #2749.

## #2798 — the CLV equations render as raw TeX

`sphinx.ext.mathjax` wraps any multi-line `.. math::` block in `\begin{split}`, so the `\begin{align}` that four CLV docstrings opened ended up nested inside it. The generated markup looks like this (taken from the live page for `BetaGeoNBD`):

```
\[\begin{split}...\\
\begin{align}
D_1 &= ...
\end{align}\end{split}\]
```

MathJax rejects `align` nested inside `split` with **"Erroneous nesting of equation structures"**. The docs load MathJax **4** (`cdn.jsdelivr.net/npm/mathjax@4/tex-mml-chtml.js`); MathJax 3 tolerated the same nesting, which is why this only appeared at the version bump in v0.18.0 rather than when the docstrings were written.

The fix is to use `aligned`, which nests legally inside `split`, and to fold each lead-in line into the same environment so the `&` alignment points actually line up.

Affected classes, all four named in the issue: `ParetoNBD`, `BetaGeoNBD`, `ModifiedBetaGeoNBD`, `ShiftedBetaGeometric`.

### Malformed TeX fixed along the way

Independent of the nesting, these blocks contained TeX that was wrong on its own terms:

- `\right)^(r+x)` — unbraced superscript, so only the `(` was raised. Now `^{r+x}`.
- `\frac{\Gamma(r+x) \alpha^r)}{...}` — stray closing paren in the MBG/NBD `A_1` numerator.
- `_2F_1` and `{_2}{F}{_1}` — subscript with no base. Now `{}_2F_1`.
- `>=` and `|` — now `\geq` and `\mid`.

### Docstrings that disagreed with the code

Two equations did not match the `logp` they document. Flagging these explicitly since they go beyond rendering — please sanity-check:

- **`ParetoNBD`**: docstring had `\frac{\Gamma(r+x)\alpha^r\beta}{\Gamma(r)+(\alpha+t_x)^{r+s+x}}`. `logp` computes `gammaln(r+x) - gammaln(r) + r*log(alpha) + s*log(beta) - rsx*log(larger+t_x)`, i.e. `\Gamma(r)` **multiplies** the bracket rather than adding to it, and `\beta` carries exponent `s`. Corrected to `\frac{\Gamma(r+x)\alpha^r\beta^s}{\Gamma(r)(\alpha+t_x)^{r+s+x}}`, which also matches eq. (22) of the Fader & Hardie note the docstring cites.
- **`ModifiedBetaGeoNBD`**: docstring had `A_1 = \frac{\Gamma(r+x)\alpha^r}{\Gamma(x)}`; `logp` uses `gammaln(r+x) - gammaln(r)`. Corrected to `\Gamma(r)`.

No `logp` was touched — only the prose describing it.

## #2749 — daily and hourly CLV estimates are inflated

`customer_lifetime_value` takes `future_t` in months and scales it into `time_unit` periods. The factor for `"D"` was a flat `30`, so a 12-month horizon covered 360 days instead of 365.25, and `"H"` inherited the same error via `30 * 24`. Expected purchases were accumulated over a short year, inflating CLV.

Now scaled by the average calendar month, `365.25 / 12 = 30.4375`, via a named `_DAYS_PER_MONTH` constant so the year-length identity is visible at the definition rather than buried in a dict literal.

I left `"W"` at its existing `4.345` since the issue scopes the change to `"D"` and `"H"`. Worth noting it is very slightly inconsistent: `4.345 * 7 = 30.415`, where `_DAYS_PER_MONTH / 7` would be `4.3482`. Changing it would shift existing weekly results, so I did not fold it in unasked — happy to if you want the three units strictly consistent.

Added `test_time_unit_scaling`, parametrized over all four units, pinning a 12-month horizon to `12`, `12 * 4.345`, `365.25` and `365.25 * 24` periods respectively.

Credited @Sanjays2402 as co-author — they had opened #2750/#2754 for this before the v1.0.0 merge broke them.

## Verification

- Extracted every `.. math::` block and `:math:` role from `clv/distributions.py`, applied the exact wrapping from `sphinx/ext/mathjax.py::html_visit_displaymath`, and rendered each through MathJax 4. **Before: 4 failures**, all "Erroneous nesting of equation structures", exactly the four classes in the issue. **After: 0 failures of 18.** Also rasterized the results and eyeballed all four equations for sane layout, not just absence of errors.
- Confirmed the generated markup my harness fed MathJax is byte-identical to what the published page currently serves.
- `test_time_unit_scaling` fails on the old constant (`assert [360] == [365.25 ± 3.7e-04]`, `assert [8640] == [8766.0 ± 0.008766]`) and passes on the new one, so it is not vacuous.
- `tests/clv/test_utils.py` 65 passed; `tests/clv/test_distributions.py` 50 passed.
- `pre-commit` clean on all three files (ruff, ruff format, mypy, API-doc completeness).

----
📚 Documentation preview 📚: https://pymc-marketing--2820.org.readthedocs.build/en/2820/


juanitorduz (approved):

daimon-pymclabs:
Follow-up per review: the `"W"` factor is now derived from `_DAYS_PER_MONTH` as well, so all four time units are consistent.

`30.4375 / 7 = 4.348214…` rather than the literal `4.3482` — using the expression keeps a 12-month `"W"` horizon at exactly 365.25 days (the literal lands on 365.2488). The test now pins `"W"` to `365.25 / 7` alongside the other units, and all 65 tests in `tests/clv/test_utils.py` pass.

Note this does shift existing weekly CLV results slightly (~0.07% longer horizon).

PR #2828, opened 2026-08-06
feat(mmm): ROAS calibration via add_cost_per_target_calibration + tutorial notebook

### LINKED ISSUE #2738: Docs: demonstrate ROAS/CPA priors as a first-class calibration feature
From a discussion on making priors over ROAS a first-class feature (#2264, #905). Per @juanitorduz this is essentially possible today via `MMM.add_cost_per_target_calibration` (#2407) — the gap is documentation showing users how.

**What exists:** `MMM.add_cost_per_target_calibration(data, calibration_data)` adds an observed likelihood per calibration row, with `mu = mean(spend)/mean(contribution)` over the date dim (numerator and denominator meaned separately for numerical stability, not `mean(cost/target)`). The underlying `add_cost_per_target_observations` also supports `target_per_cost=True` to flip the ratio to `mean(contribution)/mean(spend)` = **ROAS**.

**What's missing:**
1. No docs example walking through calibrating an MMM to experiment-derived ROAS/CPA priors.
2. `target_per_cost` (the ROAS direction) isn't exposed on the public `MMM.add_cost_per_target_calibration` wrapper — only on the lower-level helper (call site never passes it). Worth surfacing so a ROAS demo can use the one-line method. (Small API addition; could fold into #2737 or be its own task.)

**Docs — options:**
- **(1) Extend `mmm_roas.ipynb`** with a section using `add_cost_per_target_calibration`, contrasting it with the manual `pm.Normal`-on-ROAS / Google-paper approach the notebook already shows.
- **(2) New notebook** `mmm_roas_calibration.ipynb` focused solely on the calibration API (build → `add_cost_per_target_calibration` → prior/posterior predictive on ROAS → interpretation), cross-linked from `mmm_lift_test.ipynb` and `mmm_roas.ipynb`.
- **(3) Extend `mmm_lift_test.ipynb`**, since CPT/ROAS calibration is conceptually adjacent to lift-test calibration and #2407 already touched it.

**Recommendation: option 1.** `mmm_roas.ipynb` already frames the ROAS-calibration problem and the Google-paper motivation; adding a "first-class API" section there keeps one canonical ROAS page, avoids notebook sprawl, and lets us directly contrast the manual approach with the built-in method. Add a cross-link from `mmm_lift_test.ipynb`.

**Depends on #2737** (positive-support likelihood) so the demo doesn't teach a Normal likelihood on a strictly-positive quantity.

daimon-pymclabs:
Closes #2738.

## What this does

`lift_test.add_cost_per_target_observations` already supported `target_per_cost`
(flipping `mean(cost)/mean(target)` to `mean(target)/mean(cost)`), but
`MMM.add_cost_per_target_calibration` did not pass it through, so ROAS
calibration was not reachable from the `MMM` API. This PR plumbs it through:

- keyword-only `target_column` (default `"cost_per_target"`) and
  `target_per_cost` (default `False`), so existing behaviour is unchanged
- a `KeyError` when `target_column` is missing from `calibration_data`,
  matching the existing `channel` validation
- docstring example for the ROAS case

## Tests

- `test_add_cost_per_target_observations_ratio_direction[False/True]` checks the
  ratio direction analytically, comparing the calibration term's logp against a
  hand-computed Normal logp for constant cost/target.
- `test_add_roas_calibration_target_per_cost` covers the `MMM`-level path.
- `test_add_cost_per_target_calibration_missing_target_column` covers validation.

## Still to come on this branch

A tutorial notebook (`docs/source/notebooks/mmm/mmm_roas_calibration.ipynb`) using
the dataset from the existing `mmm_roas` notebook: same model config, a naive model
and a ROAS-calibrated one, comparison via `az.plot_forest` / `az.plot_dist` rather
than tables, and an explanation of likelihood calibration and the role of `sigma`.
It is drafted and executing now; the gallery entry lands with it, since the
`gallery-in-sync` hook requires the notebook to exist.

Draft until the notebook is committed with outputs.

----
📚 Documentation preview 📚: https://pymc-marketing--2828.org.readthedocs.build/en/2828/


daimon-pymclabs:
## Adversarial review

Reviewed against the code paths this notebook calls rather than against its own prose. Clustered by priority.

---

### P0 — Blocking

**1. `pre-commit.ci` fails on two real issues.**

- `ruff check` raises `B007` in notebook cell 55: `for i, (name, value) in enumerate(zip(names, values, strict=True))` never uses `name`.
- `ruff format` reformats the notebook, so the committed file is unformatted.

Both are one-line fixes, but they are hard failures.

**2. The notebook has not actually been validated by CI yet.**

Stating this plainly because it is easy to misread the check list: `notebooks (cvm, --notebooks mmm --start-idx 10)` and `notebooks (numba, --notebooks mmm --start-idx 10)` are the two jobs that would execute this notebook (it lands there alphabetically). The `cvm` one was **cancelled** by matrix fail-fast, not failed, and the `numba` one is still pending. The other red checks (`changes`, and the clv/bass/customer_choice/general notebook jobs) died in `Set up job` with `Failed to resolve action download info. Error: Service Unavailable` before running anything, i.e. GitHub infra, not this PR.

So the notebook currently has one clean end-to-end execution outside CI and zero inside it. Worth a re-run once P0.1 is pushed.

---

### P1 — Correctness of the tutorial's argument

**3. The notebook plots one quantity, calibrates a second, and compares both to a third.**

Three distinct code paths are presented to the reader as if they were one number computed once:

- `get_roas` uses `incrementality.contribution_over_spend(frequency="all_time")`, a zero-spend counterfactual `Y(actual) − Y(spend=0)` with `include_carryover=True`. That flag prepends `l_max` observations before the period **and extends the evaluation window by `l_max` periods** to catch trailing adstock. The denominator, `_aggregate_spend`, covers the fitted range only.
- The calibration term constrains `mean(channel_contribution_original_scale) / mean(spend)` over exactly the fitted dates, decomposed rather than ablated, with no window extension.
- `true_roas` is `(y − y01).sum() / x1.sum()` on the raw frame.

For an additive model with `f(0) = 0` these coincide in principle, and with `l_max=4` the window mismatch is far too small to explain the ~16% gap, so I do not think this is the cause of the shortfall. But a reader new to the library will assume the dashed line, the calibration target and the posterior density are the same quantity. That needs a sentence, and the "3 sigma above target" claim should say **which** of the two ratios it is 3 sigma above.

**4. The headline ranking claim rests on a difference of posterior means.**

The takeaway is that 15% leaves the channel ranking inverted, at 155 vs 150. That is a 3% gap between two posteriors with SDs around 15, reported to three significant figures with no MCSE. The defensible statement is `P(ROAS_x2 > ROAS_x1)` from the joint posterior, which is already stacked in `combined` and costs one line. For a conclusion this load-bearing, this is the weakest link in the argument.

(The shared `rng` across the four fits matches what `mmm_roas` does, so that is not itself a finding; it only makes cross-model differences noisier than the presentation implies.)

**5. In-sample CRPS is the wrong instrument, and the library already ships the right one.**

The section caveats itself twice (in sample, small spread) and then asks the reader to trust the ranking anyway. `pymc_marketing.mmm.time_slice_cross_validation.TimeSliceCrossValidator` exists, and `mmm_roas` already uses it to make this exact point out of sample (its prose: the baseline "consistently outperforms the calibrated model with respect to the CRPS on the test set"). Either do it properly with the existing tool, or cut the section down to a pointer. As written it teaches a beginner to draw a causal conclusion from an in-sample score.

---

### P2 — API and test quality

**6. `target_column` collides with an existing `MMM` attribute.**

`MMM.__init__` already takes `target_column`, meaning the model's target series. The new keyword on `add_cost_per_target_calibration` means the calibration frame's value column. The new test shows the collision within four lines:

```python
MMM(target_column="target", ...)
...
add_cost_per_target_calibration(target_column="roas")
```

Same keyword, same object, two unrelated meanings. `value_column` or `calibration_column` avoids it. The name is inherited from `lift_test`, but the collision only becomes direct at the `MMM` level, which is what this PR adds.

**7. The MMM-level test does not cover the behaviour the PR fixes.**

`test_add_roas_calibration_target_per_cost` asserts the RV name is present and the coord length is 2. It never inspects `mu`. Removing `target_per_cost=target_per_cost` from the pass-through in `mmm.py` leaves that test green, which is precisely the bug this PR exists to fix.

The analytic logp check in `test_lift_test.py` is good, but it exercises the function that already worked on `main`. It also uses constant cost and target, so `mean(a)/mean(b)` and `mean(a/b)` are equal there and the ratio-of-means behaviour is untested. A non-constant fixture would separate them.

---

### P3 — Prose, naming, and docs

- **Title/gallery card say "ROAS Priors"** while cell 4 spends a paragraph explaining that you cannot put a prior on a deterministic. Rename.
- **`{meth}` xref in cell 0** points at `pymc_marketing.mmm.multidimensional.MMM`, a deprecated shim that emits a `FutureWarning` on import. Should be `pymc_marketing.mmm.mmm.MMM`. (All five `{ref}` labels do resolve, including `mmm_tv_intercept`.)
- **Cell 10 says `x2` "should come out roughly right"**; cell 25 shows it off by 17% with the truth outside the interval, and cell 43 later concedes it is "less of a clean control than the setup suggests". Fix cell 10 rather than apologising for it downstream.
- **Cell 51 warns that converting a standard error by reciprocal is wrong**, then cell 50 writes `sigma = 0.15 / true_roas`. That is relative-sigma-preserving and defensible, but it reads as contradicting the box directly above it. One code comment resolves it.
- **`mmm.py` docstring** lists the calibration columns as ``` ``target_column``: desired CPT (or ROAS) value ```, which renders as a column literally named `target_column`. Separately, the "numerically unstable when spend is spiky" justification is carried over from `lift_test.py`; the actual reason for ratio-of-means is that it is the definition of aggregate ROAS, not numerical stability. Both pre-existing on `main`, now duplicated here.
- Removing the stale `cpt_variable_name` param doc was correct.

---

### Supporting numbers

Posterior means with 94% HDIs from the clean run. Of the eight intervals, exactly one covers its true value.

| Channel | True | Naive | Tight (5%) | Moderate (15%) | Loose |
|---|---|---|---|---|---|
| `x1` | 93 | 198 `[173, 224]` | 108 `[99, 116]` | 155 `[140, 171]` | 193 `[168, 216]` |
| `x2` | 171 | 142 `[120, 165]` | **164 `[150, 179]`** | 150 `[128, 169]` | 143 `[121, 165]` |

CRPS (in sample): naive 6.96, loose 6.96, moderate 7.30, tight 9.07.

juanitorduz:
Review response, pushed in 2c2b43b9.

**ReviewNB comments (juanitorduz):** all five were addressed in bcdd71ef — the comments were anchored to the older f00e0d0 snapshot. Tip box comparing the three ROAS notebooks, `roas_naive` display removed, naive-vs-calibrated comparison spelled out in numbers, legend on the density plot, and the in-sample CRPS section all exist at HEAD.

**Adversarial review:**

- **P0.1** Fixed: B007 loop variable removed, notebook reformatted; `ruff check` and `ruff format --check` pass, as does the full pre-commit suite.
- **P0.2** Notebook re-executed end to end after the edits (no errors, zero divergences in all four fits). CI notebook jobs should now get a clean run to validate.
- **P1.3** Added a note distinguishing the three computations (plotted counterfactual ROAS, calibration ratio over fitted dates, simulation ground truth), and the "three sigma" sentence now says it is the plotted posterior mean relative to the calibration target.
- **P1.4** Fixed: the ranking claim is now backed by `P(ROAS_x2 > ROAS_x1)` computed per model from the joint posterior (naive/loose ≈ 0.2%, moderate ≈ 35%, tight ≈ 100%), with prose noting that the difference of posterior means alone would overstate the precision.
- **P1.5** Not changing: the in-sample CRPS section stays, per the author's explicit request. The reading of the section as teaching a causal conclusion from an in-sample score is backwards — its stated lesson is that in-sample fit does *not* identify the causal model (selecting on it picks the naive model). Out-of-sample CRPS via `TimeSliceCrossValidator` would multiply runtime and duplicate `mmm_roas`, which the section already links for exactly that demonstration.
- **P2.6** Not renaming: `target_column` is kept for consistency with the released `lift_test.add_cost_per_target_observations(target_column=...)`. The collision with `MMM(target_column=...)` is conceptual, not functional, and introducing a second name for the same concept across the two levels trades one confusion for another. Author's call.
- **P2.7** Fixed: the MMM-level test now builds the same model with `target_per_cost` flipped and asserts the calibration RV logps differ, so deleting the pass-through fails the test. The lift_test ratio-direction test now uses non-constant series, separating ratio-of-means from mean-of-ratios.
- **P3** All fixed: title renamed to "Calibrating an MMM with ROAS Estimates" (cell 0 + gallery); xref points at `pymc_marketing.mmm.mmm.MMM`; the `x2` setup paragraph now says unconfounded ≠ recovered (direct vs inherited bias); the CPA `sigma` line has a comment explaining it is a fresh 15% relative choice on the CPA scale, not a reciprocal-converted SE; the docstring bullet no longer renders as a literal `target_column` column and the ratio-of-means rationale now states the real reason (definition of the aggregate ratio, not numerical stability). `lift_test.py` wording left alone — pre-existing on main, out of scope here.

Quoted numbers (intervals, CRPS, posterior means) were re-checked against the new execution; sampling is not bit-for-bit reproducible across runs, so the CRPS list now quotes one decimal, matching the plot annotations.

juanitorduz (approved):
hehe YOLO

PR #2831, opened 2026-08-06
refactor(mmm): Add compute/transform/inverse_transform contract to VariableScaling

williambdean:
## What

Scaling objects now own their behavior. Before, MMM had ~110 lines of helper methods that orchestrated scaling (`_compute_scale_for_variable`, `_build_fixed_scale_from_dict`, etc.), with in-graph division and NaN/Inf guards in `build_model()`.

After, each `VariableScaling` subclass implements `compute()`, `transform()`, and `inverse_transform()`. MMM just calls them.

## Why

The old approach scattered scaling logic across `mmm.py` — computing scales, dividing in-graph, guarding against NaN/Inf, and reconstructing fixed scales from dicts. Adding a new scaling type required touching `mmm.py` in 3+ places.

The new approach encapsulates everything in the scaling object itself. To add a new scaling type, implement 4 methods:

```python
from pymc_marketing.mmm.scaling import VariableScaling

class ZScoreScaling(VariableScaling):
    """Future example — not yet wired end-to-end."""

    @property
    def artifact_keys(self) -> tuple[str, ...]:
        return ("mean", "std")

    def compute(self, data):
        reduce_dims = ("date", *self.dims)
        return Dataset({
            "mean": data.mean(dim=reduce_dims),
            "std": data.std(dim=reduce_dims),
        })

    def transform(self, data, artifacts):
        return (data - artifacts["mean"]) / artifacts["std"]

    def inverse_transform(self, data, artifacts):
        return data * artifacts["std"] + artifacts["mean"]

# Usage — no changes to MMM needed:
mmm = MMM(scaling={"target": ZScoreScaling(dims=())}, ...)
```

## What changed

| File | Before | After |
|---|---|---|
| `scaling.py` | Data holder with `method` field | Abstract `compute`/`transform`/`inverse_transform` contract |
| `mmm.py` | ~110 lines of `_build_fixed_scale*` helpers | Calls `scaling.compute()` directly |
| `build_model()` | In-graph division + NaN/Inf switches | Pre-scaled data enters graph as-is |
| `_set_xarray_data()` | No scaling | Pre-scales via `scaling.transform()` |
| `BudgetOptimizer` | `_channel_scales = 1.0` (hardcoded) | Property reading actual scales |

## Concrete subclasses added

- **`MaxAbsScaling(dims=())`** — divides by max absolute value (default, replaces `DataDerivedScaling(method="max")`)
- **`MeanAbsScaling(dims=())`** — divides by mean absolute value (replaces `DataDerivedScaling(method="mean")`)
- **`DataDerivedScaling`** — kept for backward-compat serialization; deserialization maps to the new classes

## What's NOT in this PR (follow-up issues)

- `artifact_keys` property — support multi-key artifacts (e.g., z-score needs mean + std, not just scale)
- `inverse_transform_graph()` — wire `add_original_scale_contribution_variable()` through the scaler's inverse instead of hardcoded `var * target_scale`
- Decomposition helpers — accept full artifacts instead of bare `target_scale`
- Custom scaling example — end-to-end notebook with serialization

----
📚 Documentation preview 📚: https://pymc-marketing--2831.org.readthedocs.build/en/2831/

### DIFF
--- a/pymc_marketing/mmm/__init__.py
+++ b/pymc_marketing/mmm/__init__.py
@@ -73,6 +73,8 @@ from pymc_marketing.mmm.preprocessing import (
 from pymc_marketing.mmm.scaling import (
     DataDerivedScaling,
     FixedScaling,
+    MaxAbsScaling,
+    MeanAbsScaling,
     Scaling,
     VariableScaling,
 )
@@ -107,6 +109,8 @@ __all__ = [
     "LogisticSaturation",
     "MMMBuilder",
     "MMMPlotSuiteFacade",
+    "MaxAbsScaling",
+    "MeanAbsScaling",
     "MediaConfig",
     "MediaConfigList",
     "MediaMuEffect",
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -235,10 +235,9 @@ from pymc_marketing.mmm.plot import MMMPlotSuite
 from pymc_marketing.mmm.plotting import MMMPlotSuiteFacade
 from pymc_marketing.mmm.plotting.budget import BudgetPlots
 from pymc_marketing.mmm.scaling import (
-    DataDerivedScaling,
     FixedScaling,
+    MaxAbsScaling,
     Scaling,
-    VariableScaling,
     panel_channel_fixed_scaling_remaining_dims,
     validate_fixed_scaling_keys,
 )
@@ -507,15 +506,15 @@ class MMM(RegressionModelBuilder):
             scaling = deepcopy(scaling)
 
             if "channel" not in scaling:
-                scaling["channel"] = DataDerivedScaling(method="max", dims=self.dims)
+                scaling["channel"] = MaxAbsScaling(dims=self.dims)
             if "target" not in scaling:
-                scaling["target"] = DataDerivedScaling(method="max", dims=self.dims)
+                scaling["target"] = MaxAbsScaling(dims=self.dims)
 
             scaling = Scaling(**scaling)
 
         self.scaling: Scaling = scaling or Scaling(
-            target=DataDerivedScaling(method="max", dims=self.dims),
-            channel=DataDerivedScaling(method="max", dims=self.dims),
+            target=MaxAbsScaling(dims=self.dims),
+            channel=MaxAbsScaling(dims=self.dims),
         )
 
         if set(self.scaling.target.dims).difference([*self.dims, "date"]):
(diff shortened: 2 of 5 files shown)


daimon-pymclabs (changes requested):
Reviewed the full diff plus the consumers of `channel_data` / `_channel_scales` outside it. The encapsulation is the right shape and I like that `compute`/`transform`/`inverse_transform` puts the NaN guard next to the division that causes it. One blocking issue, then a naming problem I think matters more than it looks, then smaller things.

## P0 — Blocking

**1. `BuildMergedModel` still hardcodes `_channel_scales = 1.0`, and pre-scaling breaks its shared-`channel_data` premise.**

`budget_optimizer.py:444` sets `self._channel_scales = 1.0` in `BuildMergedModel.__init__`, and `_replace_channel_data_by_optimization_variable` reads `self.mmm_model._channel_scales` to convert monetary budgets into graph units. That 1.0 was *correct* before this PR: `channel_data` held raw spend and each sub-graph divided by its own `channel_scale`, so budgets went in unscaled and the graph handled it. After this PR `channel_data` holds pre-scaled values and no division happens in-graph, so monetary budgets are substituted directly into a node whose units are spend/scale. The optimizer then solves the wrong problem silently — no error, just a wrong allocation. This hits the `num_models == 1` path too, which wraps a single `BudgetOptimizerWrapper` whose new property returns the real scales while the merged wrapper above it reports 1.0.

The deeper problem is that a fix cannot just mirror the new property. `merge_on="channel_data"` deliberately keeps **one** unprefixed `channel_data` shared across all merged sub-models. A shared *raw* node is coherent when each sub-graph applies its own scale. A shared *pre-scaled* node cannot be, because it would have to equal `spend / scale_1` and `spend / scale_2` simultaneously. Merging is only safe now when every merged model happens to have identical channel scales, which for per-region models fitted on different data is exactly when it is false.

So this needs a decision, not a one-liner: either keep the in-graph division for the optimizer path, or make `merge_models` prefix `channel_data` per model and have the optimizer broadcast one budget vector into N pre-scaled nodes. Worth saying explicitly in the PR description either way, because "`_channel_scales = 1.0` hardcoded → property reading actual scales" in the table reads as though the whole class of bug was swept up, and one instance was left.

**2. No test covers the `_channel_scales` change.**

`_channel_scales` appears in the diff exactly twice, both on the removed/added lines themselves. Nothing asserts the property returns the fitted scale, that it falls back to 1.0 pre-fit, or that a budget optimization on a model with a non-unit channel scale produces the same allocation before and after this refactor. That last one is the test that would have caught finding 1. Given the change converts a constant into a value that now actually feeds an optimizer, it is the highest-value test in the PR.

## P1 — The new class names assert something the code does not do

**3. `MaxAbsScaling` computes `data.max()`, not `max(abs(data))`. Same for `MeanAbsScaling`.**

```python
def compute(self, data: xr.DataArray) -> xr.Dataset:
    """Compute ``max(abs(data))`` over date and configured dims."""
    reduce_dims = ("date", *self.dims)
    scale = data.max(dim=reduce_dims)
```

There is no `abs` anywhere in either class, and `scaling_description()` returns `"max-absolute"` / `"mean-absolute"`. The *behaviour* is unchanged from `DataDerivedScaling(method="max")`, so this is not a regression, but the PR is what promotes an accurate name (`DataDerivedScaling(method="max")`, which promised only a max) into a public class name and docstring that promise absolute values. That matters where the difference bites: an all-negative or mixed-sign series (a target expressed as a delta, a control-like channel) gets a negative or near-zero scale under `.max()`, which flips signs or explodes, whereas the documented `max(abs(·))` would not. `MeanAbsScaling` is worse in the mixed-sign case because a mean can sit near zero while `mean(abs(·))` cannot.

Pick one: add the `abs()` (a behaviour change, needs a note and probably its own PR), or name them `MaxScaling` / `MeanScaling` and drop "absolute" from the descriptions. Shipping the name without the semantics locks in the wrong contract, since these are now the public, recommended classes.

## P2

**4. `_safe_scale` is a no-op on tensors, and the in-graph guard it replaced was not.**

The old code put `switch(isnan | isinf, 0, ·)` *inside* the PyMC graph, so it protected any value that ever flowed through `channel_data`, including tensors substituted later. `_safe_scale` only branches on `isinstance(result, xr.DataArray)` and returns tensors untouched, with the docstring "tensor inputs are assumed to already contain finite values". That assumption does not hold on the optimizer path: `do(model, {"channel_data": budgets / channel_scales})` divides by the scale in-graph, and a zero scale (the all-zero-channel case the old comment called out explicitly) now yields `inf` reaching the likelihood instead of a clamped 0.

Either the abstract `transform` docstring should stop promising unconditional clamping, or the tensor branch should apply `pt.switch`. I lean toward the latter, since the guard existed precisely for a case that still exists.

**5. `_channel_scale` is now dead.**

`mmm.py:2141` binds `_channel_scale = pmd.Data("channel_scale", ...)` and nothing reads it (`_target_scale` is still used at 2281). Keep the `pmd.Data` call — `constant_data["channel_scale"]` is public surface, `get_channel_scale()` and the saturation-curve plots read it — but drop the unused binding. Worth a one-line comment that the `Data` exists for downstream consumers rather than for the graph, otherwise the next reader deletes it.

**6. `DataDerivedScaling`'s deprecation warning fires from a `model_validator`, so it triggers on validation, not just construction.**

`deserialize_variable_scaling` correctly suppresses it in the paths it owns, but any user code that round-trips a saved config through `Scaling.model_validate` / `Scaling(**...)` outside those helpers now emits a `DeprecationWarning` for a class they never named. Since both remaining legacy branches in `deserialize_variable_scaling` are unreachable for `method in {"max", "mean"}` and `"fixed"` — i.e. every value the `Literal` permits — the fall-through `DataDerivedScaling(...)` construction only runs for a `method` that would fail validation anyway. If that is right, those three `catch_warnings` blocks are dead code and the class can keep the warning on `__init__` instead.

## Nit

`_set_xarray_data` now casts to the model's dtype and *then* divides, so the model's dtype is no longer the last operation: if the scale is wider than `original_dtype`, the division promotes the result back. Reordering (transform, then `astype`) restores the intent of the existing `test_set_xarray_data_preserves_dtypes`. The current test does not catch it because it exercises float32 input into a float64 model, which is the direction where the promotion is a no-op.

PR #2812, opened 2026-08-04
Case study: R2D2 for controls

TeemuSailynoja:
## Summary

Adds a new MMM case-study notebook, `mmm_control_dimensionality.ipynb`, under the gallery title **When More Controls Affect ROAS Estimation: An R2D2 Case Study**.

The notebook translates the prior-geometry question from [Riha et al. (2026), *To select or not to select*](https://arxiv.org/abs/2606.22850) into a media-mix setting: when stakeholders ask for a broad list of neutral controls, how does the control-prior choice affect total media ROAS estimation?

The case study compares two prior systems across nested control sets:

- an independent coefficient prior for controls and routine Fourier terms;
- a split R2D2 prior that shares one control-and-seasonality budget across nuisance coefficients while leaving media priors untouched.

It separates three levels of evidence:

1. the prior-predictive mechanism as nuisance dimensionality grows;
2. one fitted-dataset illustration;
3. a 50-replication repeated experiment over $K \in \{5, 10, 25\}$ controls.

The main practical conclusion is narrow: split R2D2 gives broad-control MMMs a dimension-aware starting point by preventing the nuisance prior allowance from expanding solely because the control list is longer. In the primary $K=25$ stress test, split R2D2 produced shorter ROAS intervals and lower RMSE in this simulation, while the notebook explicitly treats coverage and paired error differences as Monte-Carlo-limited rather than definitive.

## Stacked dependency

This PR is stacked on #2986 because the notebook uses the native R2D2 implementation introduced there.

## Included changes

- Adds `docs/source/notebooks/mmm/mmm_control_dimensionality.ipynb`
- Adds the notebook to the MMM gallery `Case Studies` section
- Updates the generated gallery markdown entry

No library code, `pyproject.toml`, or `uv.lock` changes are included.

## Test plan

- [x] Full notebook execution completed and outputs are committed
- [x] Repeated-study diagnostics are reported in the notebook
- [x] Gallery source and generated markdown are updated
- [ ] Docs build renders the notebook and gallery entry

----
📚 Documentation preview 📚: https://pymc-marketing--2812.org.readthedocs.build/en/2812/

PR #2833, opened 2026-08-06
feat: add R2-D2-M2 prior for variance decomposition

williambdean:
## R2-D2-M2 Prior for Variance Decomposition

Implements the R2D2M2 prior (Zhang et al., 2022) for automatic shrinkage and variable selection via Dirichlet variance decomposition.

### Standalone PyMC Example

```python
import pymc as pm
import pymc.dims as pmd
import numpy as np
import xarray as xr
from pymc_marketing.r2d2 import R2D2Decomposition

ds = xr.Dataset({
    "controls": (("obs", "control"), np.random.randn(100, 2)),
    "y": ("obs", np.random.randn(100)),
}, coords={"obs": range(100), "control": ["a", "b"]})

r2d2 = R2D2Decomposition(
    r2=Prior("Beta", mu=0.8, sigma=0.4),
    total_sigma=Prior("LogNormal", mu=0, sigma=1),
    dims={"control": "control"},
)

with pm.Model(coords={"obs": range(100), "control": ["a", "b"]}) as model:
    controls = pmd.Data("controls", ds["controls"])
    beta_control = r2d2.split("control").create_variable("beta_control")
    intercept = pmd.Normal("intercept", mu=0, sigma=1)
    mu = intercept + controls @ beta_control
    sigma = r2d2.error_sigma.create_variable("sigma")
    pmd.Normal("y_obs", mu=mu, sigma=sigma, observed=ds["y"])
```

### MMM Integration

```python
from pymc_marketing.mmm.components.adstock import GeometricAdstock
from pymc_marketing.mmm.components.saturation import LogisticSaturation

r2d2 = R2D2Decomposition(
    r2=Prior("Beta", mu=0.8, sigma=0.4),
    total_sigma=Prior("LogNormal", mu=0, sigma=1),
    dims={"control": "control", "fourier": "fourier_mode"},
)

mmm = MMM(
    model_config={
        "likelihood": Prior("Normal", sigma=r2d2.error_sigma),
        "gamma_control": r2d2.split("control"),
        "gamma_fourier": r2d2.split("fourier"),
    },
)
```

### Key Features

- **Lazy auto-build**: Decomposition created on first `create_variable()` call
- **Shared decomposition**: `split()` and `error_sigma` share the same `R2D2Decomposition` object — changing one affects all
- **Type validation**: `r2` and `total_sigma` must be `Prior` instances; `total_sigma` must be scalar
- **Dims validation**: Dimension values must be strings; uses actual variable names for `split_dim` coord
- **Informative errors**: Unknown component names list available components

### Serialization

**Design**: Instead of generic $ref reference tracking, shared decomposition identity is restored via a post-deserialization merge step. `serialize_model_config` serializes each R2D2 object independently (embedding the full decomposition inline in each split/sigma). `deserialize_model_config` then walks the config tree and merges identical `R2D2Decomposition` instances so all splits and sigmas reference the same object.

This approach:
- Removes ~200 lines of reference-tracking infrastructure (`$ref`, memos, `ContextVar`, batch methods)
- Is contained to a ~35-line `_merge_shared_decompositions()` utility
- Keeps `to_dict`/`from_dict` simple on all three R2D2 classes (no memo interaction, no context variables)
- All 58 R2D2+serialization tests pass; all MMM save/load tests pass

### References

- Zhang et al. (2022): https://arxiv.org/abs/2208.07132


daimon-pymclabs (changes requested):
Thanks for the ping. I focused on the `serialization.py` changes, since that's where CI is going red — the failures in `tests/mmm/test_mmm.py` (`test_save_load_equality_with_all_effects`, the three `test_time_varying_*_save_load` cases) are caused by this PR, not flakes.

### 1. `ReferenceTracker` is process-global state on the `serialization` singleton, and it is never reset

`TypeRegistry.__init__` creates one `ReferenceTracker`, and `serialization = TypeRegistry()` is a module-level singleton. Nothing clears `_seen` or `_deserialized` between top-level `serialize()` / `deserialize()` calls, so identity tracking leaks across completely unrelated operations for the lifetime of the process.

That is what produces:

```
TypeError: HSGP.parameterize_from_data() got an unexpected keyword argument '$ref'
```

The second time an object with the same identity passes through `_serialize_with_refs`, it emits `{"$ref": N}` instead of its real payload. The HSGP consumer has no idea what `$ref` means and splats it in as a kwarg. The `DifferentModelError` failures are the same root cause one layer up: the serialized config no longer round-trips, so the model id hash stops matching.

The autouse fixture that resets the tracker in the test suite is the tell here. If tests need to reset global state to pass, production needs the same reset and doesn't have one.

Suggested direction: scope the tracker to a single top-level call rather than to the registry. Either a context manager that creates a fresh tracker and tears it down, or thread an explicit `memo` dict through the recursion the way `copy.deepcopy` does. Deduplication is a property of one serialization pass, not of the registry.

### 2. The two sides assign reference IDs by different rules, so `$ref` can silently resolve to the wrong object

On the write side, `track()` increments `_ref_counter` for **every** object that reaches `_serialize_with_refs`.

On the read side:

```python
result_id = len(self._tracker._deserialized)
self._tracker.store_deserialized(result_id, result)
```

but `deserialize` returns early — before that line — for the `$ref` branch, the `__deferred__` branch, and the custom-`entry.deserializer` branch. Any object taking one of those paths is counted by the serializer and skipped by the deserializer, so the two numberings drift apart. Combined with `_deserialized` never being cleared, `len()` also keeps growing across unrelated loads, so a `$ref: 0` written by a fresh serialize can resolve to whatever object happened to be number 0 in some earlier, unrelated deserialize.

Unlike issue 1, this fails silently — you get a wrong object, not a `TypeError`. Worth making the ID explicit rather than positional: emit it on the first occurrence (`{"__type__": ..., "__id__": N, ...}`) so the reader never has to independently reproduce the writer's ordering.

### 3. `$ref` payloads carry no `__type__`

Any call site that inspects `__type__` before dispatching to `deserialize` breaks on a bare `{"$ref": N}`. That is the concrete path in the HSGP failure. If the ID is emitted inline as in the suggestion above, this problem goes away with it.

### 4. `track()` runs after the recursive call, so cycles are not protected

```python
serialized = obj.to_dict()   # recurses
self._tracker.track(obj)     # only marked afterwards
```

A self-referential object recurses forever rather than emitting a ref. The R2D2 objects aren't cyclic today, so this is latent, but marking before recursing costs nothing and is the usual invariant for memo-based traversal.

## On `r2d2.py`

The design reads well and the docstrings are genuinely good. Three things I'd want resolved:

**`_built` is per-instance and never scoped to a model.** `create_variable` short-circuits on `self._built` and returns the cached `self._error_sigma`. If the same `R2D2Decomposition` is used to build a second `pm.Model` — which MMM does on load and on rebuild — you silently get tensors belonging to the *previous* model graph, with no error. Consider keying `_splits` / `_error_sigma` on `pm.modelcontext(None)`, or asserting the context matches the model that was built into.

**Dirichlet coord length.** `add_coord(f"{name}_component", list(self.dims.keys()))` gives a coord of length `n_components`, but the Dirichlet is `np.ones(K)` where `K = sum(dim_sizes)` — one entry per coefficient, not per component. For `dims={"control": "control"}` with two controls that's a coord of length 1 against a variable of length 2. The tests build the model but never sample or assert the shape, so nothing currently catches it. Could you confirm what this does at InferenceData conversion time? Either the coord should enumerate all K coefficients, or the Dirichlet should be per-component.

**`__deepcopy__` discards `memo`.** Two references to the same `R2D2Split` inside one deepcopied structure become two distinct objects. The shared `decomposition` is preserved, which is the case you're testing, so this is minor — but populating `memo` is the cheap fix.

## Coverage

Patch coverage is 48.5% with 87 lines uncovered. Given that issues 1 and 2 above are both "wrong behaviour on the second call, correct on the first," I'd specifically want tests that serialize and deserialize twice in the same process without a reset fixture, plus one that builds two models from a single decomposition. Those are the shapes the current suite can't see.

Happy to look again once the tracker is scoped.

juanitorduz:
@williambdean This is great! Maybe you can join forces with @TeemuSailynoja ? See https://github.com/pymc-labs/pymc-marketing/pull/2812

TeemuSailynoja:
I'm happy to help here. I'm working on a R2D2 ecxample case study and this might make it much simpler!

williambdean:
@daimon-pymclabs please review this again.

juanitorduz:
@williambdean  and @TeemuSailynoja , shall we port this upstream (pymc-extras)? How does this compare to https://github.com/pymc-devs/pymc-extras/blob/main/pymc_extras/distributions/multivariate/r2d2m2cp.py ? :)

williambdean:
Will take a look and compare. I just worry a bit with the serialization since that is from here and not pymc-extras. However, I think we can work through that. Just would take a bit longer ...

juanitorduz:
> Will take a look and compare. I just worry a bit with the serialization since that is from here and not pymc-extras. However, I think we can work through that. Just would take a bit longer ...

Ok! We can have it here if we need it and the port it to extras if necessary :)

juanitorduz:
@williambdean here is a guided review with Claude (I am just experimenting with this flow), see if you find this useful to iterate or ignore if it does not make sense (again, a bit of an experiment) :)

## Detailed review (Claude Code, multi-agent verified)

**TL;DR:** the PR's headline feature — serializing R2D2 objects so that `R2D2Split`/`R2D2Sigma` reconnect to a single shared `R2D2Decomposition` after save/load — is never actually wired into the MMM save path, and it fails end-to-end in both directions. A saved model with an R2D2 likelihood crashes on `MMM.load`, and the split path rebuilds with a duplicate-variable error. Three additional bugs in the new serialization machinery were confirmed by executed reproductions. Findings ranked below, each with a direct fix.

### Blockers (feature does not work end-to-end)

**1. R2D2 classes are never registered with pymc_extras' deserializer — `MMM.load` crashes** (`pymc_marketing/r2d2.py:91`)
The PR documents using `Prior('Normal', sigma=r2d2.error_sigma)` as the likelihood. On load, that dict routes through pymc_extras' `deserialize → Prior.from_dict`, and the nested `{'__type__': 'pymc_marketing.r2d2.R2D2Sigma', ...}` payload has no matching pymc_extras `Deserializer` (their matchers look for `dist`/`special_prior` keys, not `__type__`). Result: `DeserializableError` at `model_builder.py:438` — the saved model cannot be loaded. The `# Register R2D2 deserializers` comment in `__init__.py` claims a registration that doesn't exist.
**Fix:** dual-register exactly the way `special_priors.py` does (see its lines 64, 383, 454, 891): call pymc_extras' `register_deserialization` for `R2D2Split`, `R2D2Sigma`, and `R2D2Decomposition` with a matcher on `data.get('__type__')`, alongside the local `TypeRegistry` registration. Add a round-trip test: `fit → save → MMM.load` with an R2D2 likelihood.

**2. Shared-reference dedup never engages on the real save path — rebuild after load raises `Variable name r2d2_r2 already exists`** (`pymc_marketing/r2d2.py:156`)
`mmm/mmm.py:1007` serializes each `model_config` entry with a **separate** `serialize()` call (fresh memo each time), and `model_builder.py:423` deserializes per entry. The only dedup entry points, `serialize_batch`/`deserialize_batch`, have zero production callers. So after load, `gamma_control`'s `R2D2Split` and the likelihood's `R2D2Sigma` each hold an independent decomposition copy; during `build_model` the first auto-builds `r2d2_r2`/`r2d2_total_sigma`/`r2d2_split`, and the second's `create_variable('r2d2')` collides. Verified by executed repro.
**Fix (two parts):**
- Wire the batch API into the actual path: make `_serializable_model_config` serialize the whole `model_config` through one shared memo (e.g. `serialize_batch` over the values, or a `serialize(mapping)` variant that threads one memo), and mirror that in `model_builder`'s load.
- Make the auto-build prefix configurable (`R2D2Decomposition(name="r2d2_control")` → variable names derived from it) so two legitimate independent decompositions in one model don't collide either.

**3. Serialize/deserialize ref-id desync when a custom `deserializer` (or `DeferredFactory`) is involved — refs silently resolve to the wrong object** (`pymc_marketing/serialization.py:454`)
`serialize` tracks *every* object in the memo (including the custom-serializer branch), but `deserialize` returns early on the `entry.deserializer` and `DeferredFactory` branches **without** calling `memo.store`. The id sequences drift: in the executed repro, `{'$ref': 1}` intended for the decomposition resolved to the `R2D2Split` instead — `restored_sigma.decomposition` was silently the wrong object; a same-instance-twice batch raised `Reference $ref=0 not found`.
**Fix:** every deserialize branch that produces an object must call `memo.store(result)` before returning, mirroring serialize's tracking exactly. Add a symmetry test: serialize a batch containing a custom-deserializer object plus shared references and assert `is`-identity after round-trip.

**4. Within-payload `$ref`s are unreadable — `_resolve_nested_refs` runs before children are deserialized** (`pymc_marketing/serialization.py:457`)
`serialize()` itself emits `{'a': {full child}, 'b': {'$ref': 0}}` when one object appears twice inside a single payload, but `deserialize()` calls `_resolve_nested_refs` on the whole dict *before* `from_dict` has deserialized and stored `'a'` — so `memo.has_ref(0)` is `False` and it raises `Reference $ref=0 not found in nested data`. Every payload with an internal duplicate is unreadable. Verified by repro.
**Fix:** don't pre-resolve top-down. Either (a) deserialize depth-first and resolve `$ref`s lazily at the point they're encountered (store each constructed child in the memo as you go), or (b) do a two-pass resolve: first pass deserializes full occurrences and records ids, second pass patches ref sites. Option (a) is the standard design and also fixes finding 8.

### Correctness bugs beyond the main path

**5. Any nested dict containing a `'$ref'` key is treated as an internal reference — legitimate user data now fails to round-trip** (`pymc_marketing/serialization.py:472`)
A registered class with a `dict[str, Any]` field holding e.g. JSON-schema content `{'schema': {'$ref': '#/definitions/thing'}}` serialized fine pre-PR and now raises `SerializationError` on deserialize. Also, `from_dict` now receives a rebuilt copy of the payload, silently changing mutation semantics.
**Fix:** namespace the ref marker so it can't collide with user data — e.g. `{'__pymc_ref__': N}` — and only treat a dict as a ref when it matches that exact shape (marker key present *and* value is an int *and* no other keys).

**6. `__deepcopy__` on `R2D2Split`/`R2D2Sigma` ignores the memo and aliases the original decomposition** (`pymc_marketing/r2d2.py:115`)
`time_slice_cross_validation.py:822` deep-copies the MMM per fold precisely so fitted state doesn't leak — but with R2D2 in `model_config`, every fold and the original share one decomposition's `_splits`/`_error_sigma`/`_built_model_id`, so folds stomp each other's build state.
**Fix:** delete the custom `__deepcopy__` methods entirely. Default deepcopy semantics with the memo dict already do the right thing: copies within one `deepcopy` call share one new decomposition, while the original stays isolated.

**7. `_built_model_id == id(model)` caching is unsound — CPython reuses addresses** (`pymc_marketing/r2d2.py:430`)
Experiment showed consecutively created `pm.Model` objects reuse the previous address almost immediately. A new model at a dead model's address is mistaken as "already built", `create_variable` skips building, and `compile_logp` fails with "mixing variables from different models" — or worse, stitches freed-graph tensors in.
**Fix:** hold `self._built_model_ref = weakref.ref(model)` and compare `self._built_model_ref() is model`. A dead ref returns `None`, which can never be the new model, so address reuse is harmless.

**8. `_resolve_nested_refs` misses refs that are list elements and never recurses into nested lists** (`pymc_marketing/serialization.py:485`)
`[{'$ref': 0}]` and `[[{'$ref': 0}]]` pass through untouched, so `from_dict` receives literal ref dicts — silent corruption. Latent today, but the first class that serializes a list of shared children hits it.
**Fix:** if eager resolution is kept at all (finding 4's fix supersedes this), make the walker uniformly recursive: one function that handles dict values, list elements, and nested lists identically, resolving any element that *is* a ref dict, not just dicts that *contain* one.

### Design / maintenance

**9. `ReferenceTracker` (~50 lines) is dead code** (`pymc_marketing/serialization.py:139`)
Instantiated once as `TypeRegistry._tracker` and never read; all live dedup uses `_SerializeMemo`/`_DeserializeMemo`, which have subtly different semantics (a future fix applied to the dead class does nothing).
**Fix:** delete `ReferenceTracker` and the `_tracker` field. One dedup mechanism, one set of semantics.

**10. `R2D2Decomposition` hand-rolls math that `pymc_extras.distributions.R2D2M2CP` already ships** (`pymc_marketing/r2d2.py:277`)
The r2 prior, sqrt-variance split, and Dirichlet allocation duplicate the tested upstream implementation — which additionally handles input scaling, non-centered parameterization, and positive-probability masking that this version omits. Two implementations in one dependency stack will drift.
**Fix:** delegate the math to `R2D2M2CP` and keep only the lazy `VariableFactory`/serialization plumbing local. If a deliberate deviation is intended, document why in the class docstring.

### Suggested order of work

1. Findings **3, 4, 5, 8** are all symptoms of the same design: fix them together by moving to lazy, depth-first ref resolution with a namespaced marker and symmetric memo tracking (one refactor of `serialize`/`deserialize`).
2. Then **2**: thread one memo through `_serializable_model_config` and the load path, and make the R2D2 prefix configurable — this is the piece that makes the feature actually work.
3. Then **1** (pymc_extras dual registration) plus an end-to-end `fit → save → load → build` test with R2D2 in both `gamma_control` and the likelihood; that single test would have caught findings 1–4.
4. **6, 7, 9** are small, independent fixes. **10** is a scoping question worth discussing before merge rather than a mechanical change.

One candidate finding (fourier dims validation) was refuted during verification and dropped.

daimon-pymclabs (changes requested):
Re-reviewed at `b0a8c4d`. The four serialization issues from the last round are addressed, and I can confirm the fixes: the tracker is now a per-pass `_SerializeMemo` / `_DeserializeMemo` threaded through a `ContextVar` instead of registry-global state, `$ref` payloads carry `__type__`, `_built_model_id` scopes the cached tensors to the model that built them, and the `{name}_split` coord now enumerates all K coefficients rather than the components. CI is green. Nice work — the single-pass `serialize_model_config` is the right shape.

One blocking issue remains, and it is the one that fails silently.

## Reference IDs are still assigned by two different rules

The write side counts an ID for **every** object that reaches `_serialize_with_refs`, including ones with a custom serializer:

```python
entry = self._registry[type_key]
if entry.serializer is not None:
    serialized = entry.serializer(obj)
else:
    serialized = obj.to_dict()

memo.track(obj)          # <- always
```

The read side returns before `memo.store()` for the custom-deserializer branch:

```python
if entry.deserializer is not None:
    return entry.deserializer(data, context)   # <- no store()

resolved_data = self._resolve_nested_refs(data, memo)
result = entry.cls.from_dict(resolved_data)
memo.store(result)
```

So any registered type with a custom `deserializer` consumes an ID on write and none on read, and every subsequent `$ref` in that pass is off by one.

This is not hypothetical for `model_config`. `Scaled` is registered with exactly this pair:

```python
serialization.register(Scaled, serializer=_serialize_scaled, deserializer=_deserialize_scaled)
```

and `create_idata_attrs` now sends the whole `model_config` through `serialize_model_config`. So an MMM whose config holds a `Scaled` prior plus an `R2D2Decomposition` shared by two entries breaks on load. Because `serialize_model_config` walks `config.items()` in insertion order, whether it breaks depends on where the `Scaled` key sits relative to the second R2D2 reference, which makes it an intermittent bug rather than a reliable one.

### Repro

Driving `TypeRegistry` directly with a stand-in for `Scaled` (a registered type with a custom deserializer) sitting ahead of a shared object:

```python
reg.register("__main__.Shared", Shared)
reg.register("__main__.Custom", Custom, deserializer=_deser_custom)
reg.register("__main__.Holder", Holder)   # Holder.to_dict nests Shared via _serialize_with_refs

shared = Shared()
cfg = {"a": Custom(), "b": Holder(shared), "c": Holder(shared)}
back = reg.deserialize_model_config(reg.serialize_model_config(cfg))
```

Serializes to `{"a": {...}, "b": {..., "shared": {"__type__": "...Shared", "v": 1}}, "c": {..., "shared": {"$ref": 1, ...}}}` and then:

```
SerializationError: Reference $ref=1 not found in nested data.
The referenced object may not have been deserialized yet.
```

Drop the `Custom` entry and the IDs no longer collide with a missing slot, so it stops raising and starts returning the wrong object instead:

```python
cfg = {"b": Holder(shared), "c": Holder(shared)}
# b.shared -> dict
# c.shared -> Holder
# identity preserved -> False
```

Two separate things fall out of that second case. `$ref: 0` resolves to the `Holder` (the first object the reader happened to store) rather than to the `Shared` it pointed at on write, and `b.shared` never gets deserialized at all: `_resolve_nested_refs` substitutes `$ref` dicts but walks straight past a nested `__type__` dict, so `from_dict` receives a raw dict.

R2D2 itself is safe from that second part because `R2D2Split.from_dict` / `R2D2Sigma.from_dict` opt in via `get_current_deserialize_memo()` and call `_deserialize_with_refs` on the nested payload, which restores the post-order the writer used. That is the part that makes this hard to catch: the numbering happens to line up for R2D2 alone and drifts the moment anything else registered shares the pass.

### Suggested fix

Positional IDs require the reader to independently reproduce the writer's traversal, and there are now three registered types (`Scaled`, `EventAdditiveEffect`, plus anything future) where it cannot. I'd emit the ID explicitly on first occurrence instead:

```python
serialized["__id__"] = memo.next_id(obj)      # write
memo.store(data["__id__"], result)            # read, every branch
```

Then no branch ordering matters and `_resolve_nested_refs` stops being load-bearing. If you'd rather keep it positional, the minimum is a `memo.store()` on the custom-deserializer branch and on any other path that returns early, plus a test that puts a `Scaled` prior ahead of a shared `R2D2Decomposition` in the same `model_config` and round-trips it.

## Two minor ones, not blocking

`memo.track(obj)` still runs after `obj.to_dict()` rather than before, so a self-referential object recurses forever instead of emitting a ref. Nothing in R2D2 is cyclic today, so this stays latent, but marking before recursing is free.

`_built_model_id = id(model)` is vulnerable to id reuse: if the model a decomposition was built into is garbage collected, CPython can hand the same address to a new `pm.Model`, and the staleness check then passes against a graph the tensors do not belong to. Holding a `weakref.ref` to the model and comparing the referent sidesteps it.

pymc_marketing/r2d2.py
```diff
@@ -68,6 +68,20 @@
...
 ...     },
 ... )
+
+Gotchas
+-------
+- **Beta prior sigma must be small enough**: For ``r2=Prior("Beta", mu=M, sigma=S)``,
+  the sigma must satisfy ``S < sqrt(M * (1-M))``. For mu=0.8, this means sigma < 0.4.
+  Values too large produce invalid (alpha=0, beta=0) parameters. Use sigma=0.2.
+
+- **total_sigma must be scalar**: No dims allowed. This is per the R2D2M2 paper.
```
TeemuSailynoja: With multidimensional MMM we could look at R² per dimension. One could want to encode an assumption like "In any given country, we should have an R² of roughly 0.6."

From the API side, wouldn't this just require that the priors for `r2` and `total_sigma` share a target dimension?

pymc_marketing/r2d2.py
```diff
@@ -334,6 +348,14 @@ class R2D2Decomposition:
     ...         "gamma_fourier": r2d2.split("fourier"),
     ...     },
     ... )
+
+    Notes
+    -----
+    - The ``r2`` prior controls global shrinkage (how much variance the model explains).
```
TeemuSailynoja, line 333: should we highlight that this is residual variance?

pymc_marketing/r2d2.py
TeemuSailynoja, line 120: should we add an option for signed coefficients?

williambdean:
@daimon-pymclabs Please keep iterating on this PR until you feel sick

daimon-pymclabs:
Happy to keep going, but `b0a8c4d` is still the head, so there is nothing new for me to read since the last pass. Re-reviewing the same tree would just restate it.

Where it stands: the one blocking item is the reference-ID asymmetry. The write side calls `memo.track(obj)` for every object reaching `_serialize_with_refs`, including types with a custom serializer, while the read side returns from the custom-deserializer branch before `memo.store()`. Any registered type with a custom deserializer therefore consumes an ID on write and none on read, and every later `$ref` in that pass is off by one. `Scaled` is exactly that shape, and `create_idata_attrs` now routes the whole `model_config` through `serialize_model_config`, so a config holding a `Scaled` prior plus a shared `R2D2Decomposition` breaks on load, intermittently, depending on key insertion order.

Smallest thing that closes it: emit the ID explicitly on first occurrence (`serialized["__id__"] = memo.next_id(obj)` on write, `memo.store(data["__id__"], result)` on every read branch) so no branch ordering matters and `_resolve_nested_refs` stops being load-bearing. If you would rather keep positional IDs, then a `memo.store()` on the custom-deserializer branch and on every other early return, plus a round-trip test with a `Scaled` prior ordered ahead of a shared `R2D2Decomposition`.

Two non-blocking ones carry over: `memo.track(obj)` runs after `obj.to_dict()` rather than before (latent infinite recursion on a self-referential object), and `_built_model_id = id(model)` is vulnerable to CPython address reuse after the original model is collected, which a `weakref.ref` sidesteps.

Push whenever and I will pick it up on the next pass.

pymc_marketing/serialization.py
```diff
@@ -216,8 +216,7 @@ def deserialize(
 
         Three-tier dispatch:
```
TeemuSailynoja: One tier fell out from doc string but is still in the code?

pymc_marketing/serialization.py
```diff
@@ -250,6 +249,157 @@ def deserialize(
...
+
+        type_key = f"{value.__class__.__module__}.{value.__class__.__qualname__}"
+        if type_key in self._registry:
+            return self.serialize(value)
+
+        if hasattr(value, "to_dict"):
+            base = value.to_dict()
+            if isinstance(base, dict):
+                return {k: self._serialize_config_value(v) for k, v in base.items()}
+            return base
```
TeemuSailynoja, line 309: if `value.to_dict()` doesn't return a dictionary at this stage, what will it return?

pymc_marketing/serialization.py
```diff
@@ -250,6 +249,157 @@ def deserialize(
...
+        Returns
+        -------
+        dict
+            JSON-safe dict suitable for ``json.dumps``.
+        """
+        return {k: self._serialize_config_value(v) for k, v in config.items()}
+
+    def _serialize_config_value(self, value: Any) -> Any:
+        """Serialize a single config value."""
+        import numpy as np
```
TeemuSailynoja, line 270: did we have a style preference for inline imports?

pymc_marketing/r2d2.py
```diff
@@ -0,0 +1,511 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""R2D2M2 prior for variance decomposition in regression models.
+
+The R2D2M2 prior (Zhang et al., 2022) provides automatic shrinkage and
+variable selection by placing a Dirichlet distribution on the R-squared of a
+regression model, then allocating explained variance across coefficients.
+
+Reference: https://arxiv.org/abs/2208.07132
```
TeemuSailynoja: Should we drop the M2 as this module doesn't (yet) allow for defining hierarchies.

pymc_marketing/r2d2.py
```diff
@@ -0,0 +1,511 @@
...
+    dims: dict[str, str]  # component_name -> dim_name
+
+    def __post_init__(self) -> None:
+        """Validate parameters and initialize state."""
+        # Type hints already enforce Prior, but add runtime check for clarity
+        if not isinstance(self.r2, Prior):
+            raise TypeError(
+                f"r2 must be a Prior instance, got {type(self.r2).__name__}. "
+                f"Example: Prior('Beta', mu=0.8, sigma=0.4)"
+            )
```
TeemuSailynoja, line 352: We don't validate that `r2` is a scalar Prior.

pymc_marketing/r2d2.py
TeemuSailynoja: This would also let me represent a prior belief that a certain dimension (for example a certain customer segment) is more sensitive to some covariate, like media, that is left out of the R2D2 decomposition.

pymc_marketing/r2d2.py
TeemuSailynoja: If I'm understanding this correctly, now each control would receive a parameter that is pooled across target dimensions. Should we define an option for hierarchy so we get the `M2` of R2D2M2 into the class?

pymc_marketing/r2d2.py
williambdean: Something that I think would be great. However, maybe as an extensions. Just then a lot more checks in place. For example, vector r2 does that go into independent betas and or indendent sigmas. 

Was thinking that this implementation would be for the national model at the moment. 

Is you notebook national or add an dimension?

pymc_marketing/r2d2.py
williambdean: I had thought through this before and don't remember that being part of the other implementations that I've seen. Maybe follow up? Is there a way that it is done?

pymc_marketing/r2d2.py
williambdean: Fixed — now says "residual standard deviation" with R² formula reference in docstring.

pymc_marketing/serialization.py
williambdean: Fixed — "Three-tier dispatch" → "Two-tier dispatch".

pymc_marketing/serialization.py
williambdean: Avoids circular imports and defers heavy dependencies (numpy, pymc_extras). Follows existing codebase pattern.

pymc_marketing/r2d2.py
williambdean: Fixed — added `if self.r2.dims: raise ValueError(...)` with same pattern as total_sigma. Test added.

pymc_marketing/r2d2.py
williambdean: Done — renamed `R2D2Decomposition` → `R2D2` across the codebase. M2 varying effects follow-up: #2865.

pymc_marketing/r2d2.py
williambdean: Per-component importance weights → #2838. Per-segment pooled variance (M2 varying effects) → #2865.

pymc_marketing/r2d2.py
williambdean: Correct — currently single-level with no group pooling. M2 varying effects with shared pooled variance is follow-up #2865.

pymc_marketing/r2d2.py
williambdean: Yes — `r2` and `total_sigma` with dims would create independent per-group decompositions. Follow-up #2866.

pymc_marketing/r2d2.py
williambdean: Not from either paper — both use symmetric zero-mean distributions. Sign constraints break the variance decomposition theory. Created discussion issue #2867 with possible approaches.

pymc_marketing/serialization.py
williambdean: Defensive fallback for objects whose `to_dict()` returns non-dict types. All registered types return dicts in practice — could remove to enforce contract.

juanitorduz:
@daimon-pymclabs, what is missing to merge this one?

daimon-pymclabs (approved):
Re-reviewed at `6199c2a`. Short answer: nothing on my side any more — my earlier `CHANGES_REQUESTED` was the thing holding the merge box red, and it is stale. Clearing it with this approval.

## Why the old blocker is gone

I had flagged a reference-ID asymmetry: the write side counted an ID for every object reaching `_serialize_with_refs`, the read side skipped `memo.store()` in the custom-deserializer branch, so every later `$ref` in a pass was off by one.

That machinery no longer exists. The `$ref` / `_SerializeMemo` / `_DeserializeMemo` approach was dropped, and sharing is now restored after the fact: `serialize_model_config` writes each `R2D2` inline (so the same decomposition appears twice in the JSON), and `deserialize_model_config` calls `_merge_shared_decompositions`, which walks the config and collapses structurally identical `R2D2` objects onto one instance. No IDs, no ordering dependency, nothing to get out of sync. The failure mode I described is not reachable in this design.

It is also wired end to end now, which the earlier review round said it was not: `create_idata_attrs` writes `model_config` through `serialize_model_config`, `_model_config_formatting` reads it back through `deserialize_model_config`, and `test_r2d2_mmm_save_load_round_trip` asserts `gamma_control.decomposition is sigma.decomposition` after a real `MMM.save` / `MMM.load`, plus `r2d2_weights` in `loaded.model.named_vars`. That is the assertion that actually matters here.

## What is still red (not review items)

- **`codecov/patch` is failing** — 92.34% patch coverage, 18 uncovered lines of 235. Worth a glance, but note the head report had only 27 upload sessions against the base's 56 when it computed, because the run was still in flight, so it may settle on its own once all shards upload.
- **`test_slow (cvm, 2)` was still in progress** at the time of writing. Everything else on the rollup is green, including both notebook matrices and the full test grid.

## Non-blocking notes for later

1. `_deserialize_config_value` turns any all-numeric list back into an `np.ndarray`, so a config value that was a plain `[1, 2, 3]` returns as an array. Harmless for the round-trips under test, mildly surprising if someone compares configs for equality.
2. `_merge_shared_decompositions` keys on serialized content, so two *independent* `R2D2` objects with the same `r2`, `total_sigma` and `dims` would be collapsed into one after load. Not reachable today, since `R2D2Split.create_variable` and `R2D2Sigma.create_variable` both build with the hardcoded name `"r2d2"` and two decompositions in one model would already collide on variable names. If multiple decompositions per model ever becomes a goal, both of those need to move together.
3. `TypeRegistry.deserialize`'s docstring now says "Two-tier dispatch" but the body still has the custom-deserializer branch ahead of `cls.from_dict`. Docstring drift only.
4. `create_idata_attrs` overwrites `attrs["model_config"]` after `super()` already set it from `_serializable_model_config`, so there are two live serialization paths for the same attribute and the base one is now dead weight on `MMM`. Fine for this PR, but worth collapsing before a third path appears.

On @juanitorduz's earlier question about porting to `pymc-extras`: the parts worth porting are `R2D2` / `R2D2Split` / `R2D2Sigma` themselves. The `serialize_model_config` / `_merge_shared_decompositions` half is coupled to this repo's `TypeRegistry` and to `MMM.create_idata_attrs`, so it would not travel as-is. Comparing against `pymc_extras.distributions.multivariate.r2d2m2cp`, that one is a distribution factory returning variables; this one is a reusable decomposition object you hand to a `model_config` and pull lazy split/sigma references off, which is the shape the MMM path needs.

juanitorduz:
@williambdean @TeemuSailynoja, shall we merge this, since there don't seem to be any blockers? This way @TeemuSailynoja can continue with https://github.com/pymc-labs/pymc-marketing/pull/2812?

pymc_marketing/r2d2.py
```diff
@@ -71,12 +76,16 @@
 
 Gotchas
 -------
+- **r2 must be scalar**: ``r2`` Prior must not have dims. The R2D2 prior uses
+  a single global R² decomposition. For per-group R², see the follow-up issue.
```
TeemuSailynoja: Probably shouldn't reference an issue in the docstring. Although I will make sure to see that issue 😄

pymc_marketing/r2d2.py
```diff
@@ -319,9 +333,11 @@ class R2D2Decomposition:
     -----
     - The ``r2`` prior controls global shrinkage (how much variance the model explains).
     - The ``total_sigma`` prior controls the overall scale of coefficients.
-    - The Dirichlet prior (currently uniform) splits model variance across components.
+    - The Dirichlet prior (currently flat, a_π=1) splits model variance across components.
     - Use ``split("component_name")`` to get a lazy reference to a component's variance.
-    - Use ``error_sigma`` to get the error standard deviation.
+    - Use ``error_sigma`` to get the residual standard deviation.
+    - This is single-level R2D2 (no varying effects). For the M2 multilevel extension
+      with pooled variance across groups, see the follow-up issue.
```
TeemuSailynoja: Mentions issue without number.  Perhaps remove the mention to issue fully.

TeemuSailynoja: So I think this is "extra":
> For the M2 multilevel extension with pooled variance across groups, see the follow-up issue.

TeemuSailynoja (approved):
Looks great, I'll try to steal time to review the follow-ups too, but this should be a great basis.

I find doc strings mentioning follow-up issues odd. They introduce a pattern of describing what the function doesn't do instead of focusing on what it is doing.
Perhaps just personal (lack of) taste.

pymc_marketing/r2d2.py
juanitorduz: @daimon-pymclabs Can you push a fix to this branch directly?

pymc_marketing/r2d2.py
juanitorduz: @daimon-pymclabs Can you push a fix to this branch directly?

daimon-pymclabs: Fixed in a8e3e71 — dropped the follow-up-issue reference. The gotcha now just states the constraint: `r2` must be scalar because the decomposition is a single global R².

pymc_marketing/r2d2.py
daimon-pymclabs: Fixed in a8e3e71 — removed the unnumbered issue mention entirely rather than adding a number.

daimon-pymclabs: Agreed, that line was extra. Removed in a8e3e71. Also cleaned up two sibling cases you did not flag: the same "see the follow-up issue" tail in the `R2D2Decomposition` class docstring and in the scalar-`r2` ValueError message. The docstrings now describe what the class does, not what it defers.

daimon-pymclabs: Pushed: a8e3e71.

pymc_marketing/r2d2.py
daimon-pymclabs: Pushed: a8e3e71.

PR #2839, opened 2026-08-07
feat(mmm): parametrise GeometricAdstock by half-life

TeemuSailynoja:
## Summary

Reasoning about carryover as "the number of periods until the effect halves" is more intuitive for most stakeholders than reasoning about a retention rate, so `GeometricAdstock` can now take a prior on the half-life instead of on `alpha`.

```python
adstock = GeometricAdstock(l_max=10, priors={"halflife": Prior("Gamma", mu=3, sigma=1)})

# or, to use the default InverseGamma(2.6, 1) prior
adstock = GeometricAdstock(l_max=10, parametrization="halflife")
```

The default half-life prior is matched to the default `Beta(1, 3)` on alpha, so switching parametrisation is a change of coordinates rather than a change of model: it implies a median alpha of 0.207 against 0.206, and 5th and 95th percentiles of 0.019 and 0.651 against 0.017 and 0.632. Read as a half-life that default is short, a median of 0.44 periods, which the docstring and the guide both now say out loud.

Since the weight at lag $t$ is $\alpha^{t}$, a half-life $h$ gives $\alpha = 0.5^{1/h}$, which maps any positive half-life into $(0, 1)$, so no constraint can be violated. Alpha is computed inline rather than as a `Deterministic`, so the sampled variable is `adstock_halflife` and there is no `adstock_alpha`.

Everything is contained in the `GeometricAdstock` class; no changes to `Transformation` or `MMM` were needed:

- `__init__` sets `default_priors` as an instance attribute before delegating to the parent. Because the base class reads `self.default_priors`, that one assignment redirects the `function_priors` setter, `variable_mapping`, and `model_config` at once. The class attribute is untouched, so anything reading it off the class is unaffected.
- `_has_defaults_for_all_arguments` is overridden, since the base version requires an exact match between the function signature and `default_priors`, which cannot hold with two alternative parameters.
- `to_dict` records `parametrization` only when it is `"halflife"`, so existing serialised payloads are byte-identical.
- The `function_priors` setter rejects a prior for the inactive parameter with a `ValueError` rather than keeping one that would be silently ignored. Since `Transformation.__init__` assigns through that setter, this covers a contradictory `parametrization`, priors for both parameters with the parametrisation left to be inferred, and assignment after construction.

On the MMM side this works out of the box: `default_model_config` picks up `adstock_halflife` through `variable_mapping`, `with_default_prior_dims` attaches `(*dims, "channel")` to the half-life prior, and the existing unused-key warning in `ModelBuilder` catches a stale `model_config={"adstock_alpha": ...}`.

## Note for reviewers

Half-life models have no `adstock_alpha` in the trace, so helpers that name it explicitly (`mmm.plot.prior_vs_posterior(var="adstock_alpha")`, the example in `plotting/diagnostics.py`) need `adstock_halflife` instead. This is called out in the class docstring. Curve-based tooling (`sample_curve`, `plot_curve`) is unaffected. Happy to add the `Deterministic` back if we would rather keep `adstock_alpha` available everywhere.

The guide gained a "Parametrizing by Half-Life" section: the mapping and why it survives truncation and normalisation, decay curves for a range of half-lives, and how to choose a prior.

## Test plan

- [x] `tests/mmm/components/test_adstock.py` (67 passed, 1 skipped): parametrisation inference, `variable_mapping`/`model_config`, the class attribute staying untouched, rejected priors in both directions and for both at once, rejected assignment after construction, the default matching `Beta(1, 3)` in closed form across five quantiles, half-life exactness under truncation and normalisation, numerical equivalence of `halflife=1.0` with `alpha=0.5`, `sample_prior`/`sample_curve`, and serialisation round-trips
- [x] `tests/mmm/test_mmm.py::TestGeometricAdstockHalfLife`: `adstock_halflife` reaches the built model with `("country", "channel")` dims, `model_config` overrides it, a stale `adstock_alpha` warns, and an `MMM.save`/`load` round-trip keeps the parametrisation
- [x] `tests/mmm/test_mmm.py`, `tests/mmm/components`, `tests/mmm/builders` (557 passed, 1 skipped)
- [x] `pre-commit`

----
📚 Documentation preview 📚: https://pymc-marketing--2839.org.readthedocs.build/en/2839/


TeemuSailynoja:
@daimon-pymclabs would you have time to give this an initial review?

juanitorduz:
@TeemuSailynoja this is great! Now that you are it, would you mind adding a short explanation of this parametrization in the example notebook https://www.pymc-marketing.io/en/stable/notebooks/mmm/adstock_functions_guide.html ? We also do this in a different PR :)

juanitorduz:
> @daimon-pymclabs would you have time to give this an initial review?

I just started one from Discord :)

daimon-pymclabs:
Reviewed with a clean checkout, a full run of the touched test modules, and some edge-case probes of my own. The design is sound and I'd approve after two doc changes and one small API fix. I did not find a correctness bug in the math.

**Verified**

`tests/mmm/components/test_adstock.py` passes (45), as do `test_mmm.py::TestGeometricAdstockHalfLife`, `test_serialization_migration.py` and `test_media_transformation.py` (28). A NUTS fit on simulated data with true `alpha=0.7` (half-life 1.94) recovers the truth under both parametrizations, zero divergences, comparable ESS.

Half-life stays exact under truncation and normalization, since normalization is a constant divisor: `w[h]/w[0] == 0.5000` for `(l_max=10, h=2)`, `(10, 5)`, `(10, 8)`, `(30, 8)`. Worth stating in the docstring as a guarantee, since it is the first question a stakeholder-facing parameter invites.

The class-attribute shadowing holds up: after constructing a half-life instance, `GeometricAdstock.default_priors` is still `{"alpha": Beta(1, 3)}`, a fresh instance is still alpha, and `with_default_prior_dims` deepcopies without touching the original. Serialization round-trips with a custom prefix and alpha payloads are byte-identical.

**1. The two defaults are different models, not the same model in different clothes**

`Beta(1, 3)` on alpha implies a median half-life of about 0.44 periods. `Gamma(mu=2, sigma=1)` on half-life implies a median of 2. A user who flips `parametrization="halflife"` expecting a cosmetic reparametrization gets a substantially more persistent carryover prior, and on a weakly identified channel that moves ROAS and carryover conclusions. The docstring currently presents the two framings as interchangeable. It should say plainly that the defaults are not matched, and show the implied-alpha summary. Whether the half-life default should be retuned toward `Beta(1, 3)` is a maintainer judgement call, but the asymmetry needs to be visible either way.

**2. Gradient underflows to exactly zero for small half-life**

`d alpha / d h = alpha * ln(2) / h^2`, and alpha collapses faster than `h^2` grows:

```
h=0.1    alpha=0.0009765625  grad=6.77e-02
h=0.01   alpha=0.0000000000  grad=5.47e-27
h=0.001  alpha=0.0000000000  grad=6.47e-296
h=1e-4   alpha=0.0           grad=0.0
```

With the `Gamma(shape=4)` default this is harmless, since the prior has zero density at the origin. But `HalfNormal`, `Exponential`, or any `Gamma` with shape below 1 are natural choices for a positive parameter and all put mass near zero. A chain that wanders there sees a likelihood exactly flat in `h` and cannot recover on likelihood gradient alone. The docstring should state that the half-life prior must keep mass away from zero, since the flag invites users to supply their own.

**3. The contradiction check only fires one way**

`parametrization="halflife"` plus an `alpha` prior raises. The reverse silently wins for half-life:

```python
GeometricAdstock(l_max=5, parametrization="alpha",
                 priors={"halflife": Prior("Gamma", mu=3, sigma=1)})
# .parametrization == "halflife"
```

The user explicitly asked for alpha and got half-life with no signal. Small fix: default `parametrization` to `None`, only infer when it was left unset, and raise on an explicit contradiction. Currently untested in either direction.

**4. On the reviewer question about `adstock_alpha` as a `Deterministic`**

I'd add it back. The cost is one deterministic per channel-geo. The benefit is that `prior_vs_posterior(var="adstock_alpha")`, `plotting/diagnostics.py`, and the MMM modeling docs keep working unchanged, and users get the quantity they actually compare across models. A docstring note asking people to rewrite plotting calls will get missed.

One caveat: name it `f"{self.prefix}_alpha"` rather than a literal. Two adstock instances sharing a prefix in one model already collide today (`ValueError: Variable name adstock_halflife already exists.`), so a same-prefix deterministic inherits that constraint without making things worse, but it should not be hardcoded.

**5. Smaller notes**

`_has_defaults_for_all_arguments` now compares `default_priors` against `{self.parametrization}` and never inspects the function signature. Fine for `GeometricAdstock`, but a subclass overriding `function` loses the check entirely. Validating that `default_priors` is a subset of the signature parameters minus `x` and `dim` would keep the guard alive.

`function(self, x, alpha=None, halflife=None, *, dim)` with both `None` reaches `geometric_adstock(x, alpha=None, ...)` and fails obscurely. Only reachable by calling `function` directly, so a guard is optional.

Missing coverage worth adding: the explicit contradiction in point 3, `update_priors({"adstock_halflife": ...})` on a half-life instance, and an `MMM.save`/`load` round-trip. I traced `AdstockTransformation.from_dict` and it passes `parametrization` through correctly, but nothing tests it at the MMM level.

Unrelated to this PR: extra prior keys are silently accepted and dropped from `variable_mapping` for both parametrizations, so `priors={"bogus": ...}` is a no-op. Pre-existing, just noting it.

juanitorduz (approved):
@TeemuSailynoja some minor comments from daimon (I would leave the Deterministic out as we can easily conpute it back from the posterior). 

The notebook additions is optional. Otherwise, LGTM :)

TeemuSailynoja:
- New docs page for [GeometricAdstock](https://pymc-marketing--2839.org.readthedocs.build/en/2839/api/generated/pymc_marketing.mmm.components.adstock.GeometricAdstock.html#pymc_marketing.mmm.components.adstock.GeometricAdstock)

TeemuSailynoja:
@juanitorduz and @daimon-pymclabs I'm ready for a second round 😉

pymc_marketing/mmm/components/adstock.py
```diff
@@ -262,8 +307,61 @@ class GeometricAdstock(AdstockTransformation):
...
+    ) -> None:
+        prior_names = set(priors or {})
+
+        if parametrization is None:
+            parametrization = "halflife" if "halflife" in prior_names else "alpha"
+        elif (
+            conflicting := (set(self._alternative_parameters) - {parametrization})
+            & prior_names
+        ):
+            raise ValueError(
```
daimon-pymclabs: **Blocking: both priors passed with no explicit `parametrization` gives an unloadable model.**

The conflict check sits in the `elif`, so it only runs when `parametrization` was passed explicitly. On the inference path, `priors={"alpha": A, "halflife": H}` infers `"halflife"`, and the `function_priors` setter then merges to `{"halflife": H, "alpha": A}`, so `A` is silently ignored. That is exactly the failure mode this check is meant to prevent.

It gets worse on the round trip. `to_dict` emits both priors plus `parametrization="halflife"`, and `from_dict` calls `cls(**data)`, which now takes the explicit branch and raises `ValueError`. So a constructible transformation serialises into something that raises on load, and an `MMM.load` of a saved fit fails.

(Reasoned from the code paths, not executed.)

Fix is small: run the check unconditionally after inference rather than in the `elif`.

```python
if parametrization is None:
    parametrization = "halflife" if "halflife" in prior_names else "alpha"

if conflicting := (set(self._alternative_parameters) - {parametrization}) & prior_names:
    raise ValueError(...)
```

Worth a test for the both-priors case in each direction.

pymc_marketing/mmm/components/adstock.py
```diff
@@ -262,8 +307,61 @@ class GeometricAdstock(AdstockTransformation):
...
+                f"Priors for {conflicting.pop()!r} are not used when"
+                f" parametrization={parametrization!r}."
+                f" Provide a prior for {parametrization!r} instead."
+            )
+
+        if parametrization == "halflife":
+            # Shadows the class attribute so that variable_mapping, model_config
+            # and the function_priors setter all key off halflife.
+            self.default_priors = deepcopy(self.halflife_priors)
```
daimon-pymclabs: The `deepcopy` is redundant: the `function_priors` setter already deepcopies `default_priors`. Harmless, but it reads as if it is guarding something.

daimon-pymclabs (commented):
Read `components/base.py` and the full `adstock.py` at this commit, and checked every `adstock_alpha` reference in the library.

## What it does

Adds an optional half-life parametrization to `GeometricAdstock` via $\alpha = 0.5^{1/h}$. The implementation is contained: an `__init__` override that shadows `default_priors` on the instance, an override of `_has_defaults_for_all_arguments` to allow two alternative parameters, `to_dict` emitting `parametrization` only in the non-default case, and `function` accepting either parameter.

The design is right. Shadowing `default_priors` on the instance is the correct lever, since `variable_mapping`, `model_config` and the `function_priors` setter all read it, so one assignment redirects all three with no base-class changes. `apply` calls `self.function(x, dim=..., **kwargs)` with kwargs keyed off `variable_mapping`, so the half-life variable flows through untouched.

On the note for reviewers: every `adstock_alpha` left in `pymc_marketing/` is inside a docstring or example (`mmm.py:2190`, `plot.py:1412-1669`, `plotting/diagnostics.py:79-86`). Nothing functional breaks.

## Blocking

One real bug, commented inline: passing **both** `alpha` and `halflife` priors with no explicit `parametrization` silently drops the `alpha` prior and then produces an instance that cannot be reloaded. See the inline comment on the conflict check.

## Description vs code

The PR body says the default half-life prior is `Gamma(mu=2, sigma=1)`; the code has `InverseGamma(alpha=2.6, beta=1)`. Worth fixing before merge, it is what reviewers read first.

## On the Deterministic question

I would not add it back. Putting `pm.Deterministic("adstock_alpha", ...)` inside `function` makes the transform require a model context (it is currently callable bare) and creates a name collision the moment the transformation is applied more than once in a model. Cheaper fix is documentation: state the one-line recovery, `0.5 ** (1 / idata.posterior["adstock_halflife"])`, in the class docstring next to the existing warning and in the notebook section. That covers `prior_vs_posterior` and the diagnostics example at no modelling cost.

## Prior equivalence

The docstring's equivalence claim checks out, and holds better than stated. Sampling it: implied $\alpha$ median 0.2066 against Beta(1,3)'s 0.2063, and the match is tight across the whole distribution, with the 5/25/50/75/95 quantiles at 0.020/0.092/0.208/0.377/0.651 against Beta's 0.017/0.091/0.206/0.370/0.632.

Since that is a deterministic monotone transform of the InverseGamma quantiles, `test_default_prior_matches_alpha_default` can assert it in closed form on several quantiles instead of drawing 20,000 samples and pinning one median at `abs=0.01`. Faster and strictly stronger.

The right tail is well behaved, $P(h > 10) \approx 0.0007$, so the flat-likelihood risk at $h \gg l_{max}$ is not a real concern.

## Tests

Coverage is good: parametrization inference, the class attribute staying clean, the rejected cross-prior in both directions, exactness under truncation and normalization across `l_max` and `normalize`, numerical equivalence at h=1, and both serialization directions.

Three gaps:

1. The both-priors case in the inline comment.
2. A `to_dict`/`from_dict` round trip for the `parametrization="halflife"` default-prior instance. Only the custom-prior one is covered directly at the component level.
3. `update_priors({"adstock_alpha": ...})` on a half-life model, which should hit the base class's "No priors were updated" warning rather than doing nothing quietly.

## Nits

`__repr__` is inherited, so it shows the priors but never `parametrization`; adding it would make the alpha-vs-half-life distinction visible at a glance.

pymc_marketing/mmm/components/adstock.py
```diff
@@ -273,7 +371,43 @@ def function(self, x, alpha, *, dim: str):
...
+    def to_dict(self) -> dict:
+        """Convert the adstock transformation to a dictionary."""
+        data = super().to_dict()
+
+        if self.parametrization != "alpha":
+            data["parametrization"] = self.parametrization
+
+        return data
+
+    _alternative_parameters = ("alpha", "halflife")
```
daimon-pymclabs, line 420: `_alternative_parameters` sits between methods and the prior dicts. Move it up with the other class attributes, or down next to `default_priors`/`halflife_priors`.

pymc_marketing/mmm/components/adstock.py
```diff
@@ -234,16 +237,58 @@ def function(self, x, alpha, *, dim: str):
...
+
+    which maps any positive half-life into :math:`(0, 1)` and holds exactly for
+    every ``l_max`` and either setting of ``normalize``. Under the half-life
+    parametrisation the trace will contain ``adstock_halflife`` instead of
+    ``adstock_alpha``.
+
+    The two defaults are matched, implying a median ``alpha`` of 0.207 against
+    0.206. A custom half-life prior should keep its mass away from zero, where
+    the likelihood goes numerically flat. The
+    :ref:`adstock functions guide <adstock_functions_guide>` covers both points.
```
daimon-pymclabs, line 261: The default half-life has median 0.44 periods, with 57.5% of the mass below half a period. That is the correct consequence of matching Beta(1, 3) and I would keep it for consistency, but it reads oddly for a parametrization whose selling point is stakeholder intuition: the default says half the effect is gone inside half a week.

The docstring quotes the implied `alpha` (0.207) but never the implied half-life. Worth stating it in periods too, so anyone reaching for `parametrization="halflife"` sees what they are getting and knows to retune.

juanitorduz (approved):
LGTM :)

(optionally look into daimon's suggestion and merge when you are ready)

Thanks @TeemuSailynoja :)

pymc_marketing/mmm/components/adstock.py
TeemuSailynoja: Confirmed by running it, and it fails exactly as you traced: `priors={"alpha": A, "halflife": H}` infers `"halflife"`, keeps both, and then `from_dict` on its own `to_dict` raises.

Fixed in c61bd376, but in the `function_priors` setter rather than in `__init__`. `Transformation.__init__` assigns through that setter, so construction is still covered, and it closes a second route to the same unloadable state that `__init__` alone would leave open:

```python
adstock = GeometricAdstock(l_max=10, parametrization="halflife")
adstock.function_priors = {"alpha": Prior("Beta", alpha=1, beta=3)}
# before: {"halflife": ..., "alpha": ...}, and from_dict(to_dict()) raises
```

The check runs before the base setter, so a rejected assignment leaves the priors untouched. Tests cover both priors with the parametrisation inferred, both with it explicit either way, one prior against an explicit contradiction in both directions, and the assignment path.

pymc_marketing/mmm/components/adstock.py
TeemuSailynoja: Dropped in c61bd376, along with the now-unused `from copy import deepcopy`. The setter does `{**deepcopy(self.default_priors), **priors}`, and nothing mutates `default_priors` in place, so a plain assignment is enough. `test_class_default_priors_unchanged` is the guard, and I also checked that `update_priors` on a half-life instance leaves `GeometricAdstock.halflife_priors` at `InverseGamma(2.6, 1)`.

pymc_marketing/mmm/components/adstock.py
TeemuSailynoja: Grouped with `default_priors` and `halflife_priors` in c61bd376.

pymc_marketing/mmm/components/adstock.py
TeemuSailynoja: Numbers confirmed: median 0.4395 periods, 57.7% of the mass below half a period.

Keeping the matched default, and quoting the half-life in periods as you suggest. The docstring now reads "The priors imply a median half-life of 0.44 periods", and the guide adds that a channel expected to carry over for weeks needs a prior of its own, next to the alpha percentiles it already reports.

Issue #2821, opened 2026-08-05
Year-over-year target transform: multiplicative-by-seasonality MMM, and what it does to identification


daimon-pymclabs:
## Summary

For KPIs whose variance is dominated by an annual seasonal cycle, a useful reparameterisation is to stop modelling the level and model the **year-over-year ratio** instead. The prior cycle's observed level becomes the seasonal baseline, so seasonality enters multiplicatively and for free, with no Fourier basis, no trend component, and no per-unit seasonal profile to estimate.

The transform is cheap to implement and, in our experience on high-seasonality panels, effective. The reason to open an issue rather than a PR is that the transform is not just a preprocessing step: it changes what the likelihood can identify, and therefore what a "channel contribution" number means. Getting that part wrong is easy and quiet. This issue proposes the feature and writes down the identification consequences so they can be designed for rather than discovered later.

## The transform

Let $y_{u,t}$ be the KPI for unit $u$ (store, region, product, whatever the panel dimension is) at period $t$, and let $r(t)$ be a **reference map** sending each period to its counterpart one cycle earlier. Model the ratio

$$z_{u,t} \;=\; \frac{y_{u,t}}{y_{u,\,r(t)}}$$

with an additive linear predictor in which every driver enters as a **within-pair contrast**:

$$z_{u,t} \;=\; \alpha_u \;+\; \sum_c \beta_c \bigl[f_c(a_{c,t}) - f_c(a_{c,\,r(t)})\bigr] \;+\; \sum_k \gamma_k \bigl[x_{k,u,t} - x_{k,u,\,r(t)}\bigr] \;+\; h_t \;+\; \varepsilon_{u,t}$$

where $f_c$ is the usual adstock-then-saturation composition, $x_k$ are controls in raw driver units, and $h_t$ carries whatever calendar effects survive reference alignment. Levels are recovered exactly by $y_{u,t} = y_{u,r(t)} \cdot z_{u,t}$.

Three things fall out immediately:

- **Seasonality is absorbed multiplicatively.** Any component of the seasonal shape that repeats cycle over cycle is already in $y_{u,r(t)}$. The model only has to explain departures from last cycle.
- **The per-unit baseline collapses to a scalar.** $\alpha_u$ is that unit's typical growth rate. No per-unit seasonal curve, which is where most of the parameters in a level-space panel MMM go.
[...]

PR #2840, opened 2026-08-07
Add year-over-year target transform example notebook

daimon-pymclabs:
Closes part of the scope in #2821 (item 7, the synthetic benchmark) and prototypes item 1 (the reference map).

## What this adds

A new Custom Models gallery notebook, `docs/source/notebooks/mmm/mmm_yoy_target_transform.ipynb`, that builds the year-over-year target transform as a custom PyMC model on top of PyMC-Marketing components, in the style of `mmm_components` and `mmm_custom_splines`, with data generated from a generative PyMC model in the style of `mmm_data_generator`.

Instead of modelling the level, the notebook models the ratio $z_{u,t} = y_{u,t} / y_{u,r(t)}$ with every driver entering as a within-pair contrast, so seasonality is absorbed multiplicatively by the prior cycle's observed level and the per-unit baseline collapses to a scalar growth rate.

## Contents

1. **`build_reference_map`** — the self-contained piece from the issue: an event-anchored $r(t)$ with shoulder windows, returning the $k$ duplicated and $k$ orphaned reference periods that realigning a window against an unshifted base mapping necessarily produces. The notebook shows the seam explicitly and states the cycle-level roll-up consequence.
2. **Synthetic panel** — four stores, six years weekly, generated by a PyMC model: LKJ-correlated channel spend, a seasonal shape *inherited* through the reference map rather than re-drawn, a moving holiday whose week-of-year drifts between 12 and 15, one always-on channel held at a flat plan, and proportional noise. Ground-truth parameters are clamped with `pymc.do` and the target is forward-simulated.
3. **The custom model** — `GeometricAdstock` then `LogisticSaturation` applied to the **full contiguous series** and only then gathered at $t$ and $r(t)$, so the reference period's carryover is built from its own history. One `build_yoy_model` function serves data generation, inference, and out-of-sample prediction.
4. **Comparison** — a vanilla additive `MMM` with four yearly Fourier harmonics fitted on the same panel and the same scaled spend, scored against the year-over-year model on a 52-week holdout in KPI units.
5. **Identification** — the parts of the issue that are easy to get wrong quietly.

## Results

Holdout, last 52 weeks, KPI units:

| | year-over-year | vanilla + Fourier |
|---|---|---|
| RMSE | 36.8 | 147.1 |
| MAPE | 2.6% | 9.4% |
| 94% coverage | 0.894 | 0.942 |
| mean 94% HDI width | 116 | 457 |

The ratio model carries no Fourier basis, no trend and no per-unit seasonal profile, and still cuts holdout error by roughly a factor of four with intervals about a quarter as wide. Both land near nominal coverage, which is why the notebook insists on reading coverage and width together.

Three identification results the notebook demonstrates rather than asserts:

- **The always-on channel is invisible.** Its spend is identical cycle over cycle, its contrast is exactly zero, and its posterior standard deviation over prior standard deviation comes back at 1.04. The other two channels sit at 0.40 and 0.16.
- **Two contribution bases, two answers.** On the contrast basis the always-on channel contributes exactly 0. On the level basis with $a^\ast = 0$ it is the *largest* media contributor at ~92k. The regrouping is asserted to be exact; the leverage diagnostic reports 2.6 for `tv`, 0.97 for `search`, and `inf` for the always-on channel.
- **Closure is asserted at unit-period grain**, not on the national total.

The conclusions section states plainly that the panel was generated from the transform's own assumption, so the margin should be read as what a level model pays when that assumption holds, not as a general ranking, and lists when the level-space model is the better tool.

## Notes

- Runs end to end in about 100 seconds on 8 cores; executed with the outputs committed.
- `ruff check` and `ruff format` clean, watermark present, `nbformat` valid, `scripts/generate_gallery.py --check` in sync.
- Filed under **Custom Models** in the gallery.
- No library code is touched. This is deliberately an example rather than an API: the notebook's closing note argues the reference map is the piece that wants to become a first-class component, which is the open design question in #2821.

----
📚 Documentation preview 📚: https://pymc-marketing--2840.org.readthedocs.build/en/2840/


juanitorduz:
## Review: math, claims and comments

Reproduced the DGP and all three fits locally; the committed level-model numbers match to 3 s.f. (RMSE 147.05, MAPE 9.42, coverage 0.942, width 456). Findings ordered by severity. I am pushing fixes for all of them to this branch.

### 1. Blocking: the reference map duplicates the *prior holiday*, so the panel has ghost holidays

`build_reference_map` drags the current window `{c-pad..c+pad}` onto `{p-pad..p+pad}` and leaves the base mapping for every other period. Whenever the holiday moves by `|k| > pad`, the current period `p + 52` (a normal week) keeps its base reference `p`, which is the prior holiday. The seam table already shows it: row `t=117 -> r(t)=65`, `t is holiday=False`, `r(t) is holiday=True`. The duplicated list `[11, 64, 65, 120, 170, 171, 219]` contains both prior holidays (65, 170).

Because the DGP builds `y[t] = y[r(t)] * z[t]`, the spike is copied:

```
t=117: r(t)=65  (holiday)  y[t]/mean(y[t±1]) per unit = [1.77 1.75 1.72 1.59]
t=222: r(t)=170 (holiday)  y[t]/mean(y[t±1]) per unit = [1.63 1.37 1.34 1.43]
weeks with spike > 1.3x: [12 65 117 119 170 220 222 273]   -> 8 spikes in 6 cycles
```

Cycles 3 and 5 have two holiday spikes each. Consequences:

- "The moving holiday needs no separate term: the anchored reference map moves it" — it *copies* it when `|k| > pad`.
- "Period-level reads are unaffected" (warning box and conclusion 3) is false. On real data the same map divides normal week 117 by holiday 65, a crater exactly like the naive map's, just displaced. The seam is not only a roll-up problem.
- "Structural property rather than an implementation bug" — it is a consequence of *shifting* one window against a fixed base. Rotating the prior-cycle segment spanned by both windows keeps `r` a bijection: no duplicates, no orphans, every holiday paired with a holiday, cycle roll-ups exact. The cost moves from "normal week / holiday week" to "week / week three weeks away", which on a smooth seasonal shape is a few percent.

### 2. Major: the "factor of four" mostly measures the reference level, not the MMM

| holdout, KPI units | RMSE | MAPE | coverage | width |
|---|---|---|---|---|
| oracle (true `mu`) | 36.2 | 2.55% | | |
| analytic noise floor $\sigma_u\sqrt{2/\pi}$ | 34.5 | 2.49% | | |
| **YoY (notebook)** | 36.8 | 2.59% | 0.894 | 116 |
| seasonal naive `z = alpha_u` (4 params, no media) | 59.6 | 4.27% | | |
| seasonal naive `z = 1` (0 params) | 69.1 | 4.83% | | |
| level + Fourier + holiday dummy + linear trend | 88.6 | 6.73% | 0.928 | 338 |
| level + Fourier (notebook) | 147.1 | 9.42% | 0.942 | 456 |

- YoY is 2% from the oracle. It cannot be beaten on this panel; the number is the noise floor, not model quality.
- Seasonal naive with per-unit growth beats level+Fourier 2.5x. Of the 147 -> 37 gain, 147 -> 60 is the reference level; media and control buy 60 -> 37.
- Level model per-unit mean error is `[-66, +4, +85, -123]` for `[north, south, east, west]`, the sign pattern of the DGP growth rates `[1.03, 1.00, 0.98, 1.05]`. That is a missing trend, not a seasonality failure. A linear trend control removes it (bias ~ +10 everywhere).
- 32% of the level model's MSE is the single holiday week (4 of 208 points); RMSE excluding it is 122.6. A holiday indicator takes that share to 2.8%.
- Both level fits have convergence problems the notebook does not report: committed run `saturation_lam[sponsorship]` ESS 105, r-hat 1.039; on rerun min ESS 57, r-hat 1.10. Partly structural, see 6.

### 3. Framing: the DGP *is* the model, and it is a seasonal random walk

- `z_observed[t] = y[t]/y[r(t)] = z_true[t]` to floating point. The YoY model is fitted on its own prior-predictive draw; the level construction affects only the level model and the KPI scoring.
- `y[t] = y[r(t)] * z[t]` chains across cycles, so `log y` at each week-of-year is a random walk with step sd ~ 0.03, about ±7% after five cycles. "A seasonal shape that genuinely repeats" undersells this: it is the seasonal-random-walk case where seasonal differencing is *optimal*, and it contradicts the "when not to use it: if the seasonal shape is drifting" advice.
- The informative benchmark is missing: fixed seasonal shape, independent proportional noise, moving holiday. There `z = (1+e_t)/(1+e_r(t))` has about twice the noise variance and is negatively correlated across cycles, which the iid Normal likelihood ignores. "Heteroscedasticity is largely handled" is untested against that case.
- Issue item 7 is ratio vs log-difference vs level-with-Fourier; log-difference is not in the notebook.

### 4. Wrong comment: the coverage explanation

"lands slightly under nominal here, which is the price of a noise model that is tight where the level is low" — under this DGP the proportional noise model is exactly right, so there is no price. Actual cause:

```
true z_sigma:       [0.030 0.035 0.028 0.032]
train realised sd:  [0.0303 0.0337 0.0271 0.0274]   <- posterior z_sigma[west]=0.0274 is right for the draw
test realised sd:   [0.0288 0.0364 0.0294 0.0347]
YoY coverage/unit:  [0.942 0.923 0.923 0.846]
```

West's training noise came out low and its test noise high on this seed. 0.894 on n=208 is 2.8 SE below 0.94, so "near nominal" is a stretch too. Same cause behind `z_sigma[west]` sitting 3.3 posterior sd from truth while the text says "on target".

### 5. Wrong comment: the closure check cannot fail

`residual := (z - mu) * y_ref` and `sum(parts) = mu * y_ref`, so `sum(parts) + residual = z * y_ref` for *any* split of `mu`. A constant misplaced between baseline and control leaves the sum unchanged and the check passes at national and unit-period grain alike. The paragraph motivating the check describes a failure mode it does not detect. What it does verify is that the same `y_ref` defined `z` and dollarised the parts; asserting against the observed level `y` says that directly.

### 6. Claim not isolated: the always-on channel is unidentified in the level model too

`sponsorship` is constant per unit, so `f(sponsorship_u)` is a per-unit constant, exactly collinear with the level model's per-unit intercept (4 units, 4 intercepts). That is why its parameters have the worst ESS in the level fit. "Media appears only as a contrast" is a genuine property of the YoY design, but a flat channel is invisible to any model with unit intercepts, and any plan that repeats cycle over cycle is confounded with the seasonal basis in a level model. The YoY model makes it exact and detectable via the sd ratio; the level model makes it prior-dependent and silent. The text should say that instead of implying the level model identifies it.

### 7. Smaller

- "A spike-and-crater pair at *every* moving holiday" with the naive map: not on this data. The ghost fills the crater; naive `z` at 117 is `[1.12 1.11 0.99 1.06]`, at 222 `[1.06 0.99 1.01 0.99]`. Craters exist only at 64, 171, 272. Fixing 1 restores the claim.
- The burn-in argument for `obs_idx_model` is right, but the generator uses `obs_idx_all`, so the first eight modelled weeks were generated with a partially warmed contrast. Harmless; worth a sentence.
- The leverage diagnostic uses mean absolute values, the text formula uses norms.
- `saturation_beta` prior sd 0.116 vs truths 0.13/0.08/0.06: the prior is at the truth's scale, which flatters recovery.

The identification sections are the strongest part and mostly survive. Fix for each item follows in the next commit on this branch.

juanitorduz:
## Fixes pushed (c0dd778), branch rebased on `main`

Every item in the review above is addressed in the notebook; nothing in library code changes. Notebook re-executed end to end on `main`'s lock (pymc 6.2.0, pymc-marketing 1.1.0), about 130 s on 8 cores.

**1. Reference map.** `build_reference_map` now rotates the prior-cycle segment spanned by the current and prior event windows instead of shifting one window against the base lag. `r(t)` is a bijection between consecutive cycles: `duplicated = []`, `orphaned = []`, no holiday week paired with a non-holiday week or vice versa, maximum displacement `2*pad + 1 = 3` weeks. The panel now has exactly one spike per cycle at the anchored position (`[12, 65, 119, 170, 220, 273]`). The holiday has a three-week profile so that the shoulder is load-bearing rather than decorative, and the text says what `pad` costs and why it should match the event footprint.

**2. Comparison.** The holdout table now reads:

| panel 1 (ratio DGP), KPI units | RMSE | MAPE | coverage | width |
|---|---|---|---|---|
| noise floor (true `mu`) | 36.5 | 2.55% | | |
| year-over-year | 37.1 | 2.59% | 0.899 | 117 |
| seasonal naive, fitted growth | 59.6 | 4.21% | | |
| seasonal naive, `z = 1` | 69.3 | 4.78% | | |
| level + Fourier + holiday + trend | 88.4 | 6.57% | 0.856 | 252 |
| level + Fourier | 145.8 | 9.29% | 0.880 | 394 |

The text attributes the gain accordingly: reference level first, media and control second. Sampler diagnostics for the level models are reported split by the intercept/`sponsorship` ridge (ESS ~10, r-hat ~1.3) versus everything else (ESS 300-500, r-hat 1.01), with per-chain holdout RMSE showing the ridge is prediction-invariant (chains agree to 0.5 KPI units). That ridge is presented as the level-space face of the same non-identification the ratio model reports as `sd ratio = 1`.

**3. Second panel.** New section generating the same spend, control, media response and growth with a *fixed* seasonal shape, the moving holiday, and i.i.d. proportional noise:

| panel 2 (fixed shape, iid noise) | RMSE | MAPE | coverage | width |
|---|---|---|---|---|
| noise floor (true mean) | 43.7 | 2.60% | | |
| year-over-year | 63.9 | 3.68% | 0.942 | 241 |
| seasonal naive, fitted growth | 87.6 | 5.11% | | |
| level + Fourier + holiday + trend | 58.6 | 3.97% | 0.894 | 181 |

Off home turf it is a draw that the level model wins on RMSE. Two structural costs are measured: posterior `z_sigma` sits above `sqrt(2)` times the level noise (0.045-0.061 vs 0.042-0.049), and residuals at `t` and `r(t)` correlate at about -0.35 to -0.40. The notebook also names the reason the ratio model loses at the peak: anchoring realigns the event but not the seasonal level underneath it, so on a fixed shape the anchored ratio at the holiday is `S(h_c)/S(h_{c-1})`.

**4-7. Text.** Coverage shortfall traced to realised holdout noise per unit (west: fitted 0.0274, holdout 0.0347, coverage 0.83, which is what a fifth-too-narrow interval gives). Closure check now asserts against the observed level and is described as a dollarisation check, not a split check. Always-on channel stated to be unidentified in the level model too. "Period-level reads are unaffected", "spike-and-crater at every holiday", and "the anchored map moves the holiday" are corrected. Conclusions and the "when not to use it" list rewritten to match; log-difference explicitly listed as not covered.

Source edited as a jupytext percent script and converted back; `ruff`, watermark, nbformat and gallery hooks pass.

daimon-pymclabs:
## Checked the fixes: bijection holds on this calendar, but breaks when an event crosses the cycle boundary

Thanks for pushing these, the review is right on every count and the rewrite is a real improvement. I re-ran the reference map independently and cross-checked the committed outputs against the numbers in your comment; they agree. One new finding that the rotation introduces, plus a wording nit.

### Verified

- `build_reference_map` on the committed calendar (`holiday_week_of_year = [12, 13, 15, 14, 12, 13]`, `pad=1`): `duplicated = []`, `orphaned = []`, no holiday week paired with a non-holiday week, max displacement 3. Matches cell 10's stored output.
- Bijection also holds under stress for larger and alternating moves (`|k|` up to 5) and `pad` 0 to 3, as long as every event window stays inside its cycle.
- Panel 1 and panel 2 holdout tables, per-chain RMSE, the `z_sigma` / realised-noise table and the residual correlations in the committed outputs match the tables in your comment.

### New: the rotated segment is not constrained to the prior cycle

`low`/`high` are absolute indices, so when an event's window straddles a cycle boundary the segment spans two cycles and the rotation mixes them. The result is a reference that is not one cycle back at all, and the bijection guarantee in the docstring ("Empty by construction") no longer holds.

A monotone drift across new year, which is the Ramadan case the intro motivates:

```
holiday_week_of_year = [48, 50, 0, 2, 4, 6], pad = 1

duplicated reference periods: [51, 103]
orphaned reference periods:   [49, 100]
r(103) = 101      # two weeks back, not a year
```

And an event oscillating across the boundary:

```
holiday_week_of_year = [50, 51, 1, 2, 50, 51], pad = 1

duplicated reference periods: [52, 104]
orphaned reference periods:   [51, 101, 259]
# displacements of +/-3 smear across ~50 consecutive weeks
```

The `z` this produces at the affected weeks is a short-horizon ratio, not a year-over-year one, which is worse than the ghost-holiday failure the rotation replaced: silent, and not localised at the seam.

Two things would make this safe, and neither touches the committed panel:

1. Clamp the segment to the cycle of `prior` (`[(prior // period) * period, (prior // period + 1) * period - 1]`) and raise when `min(prior, prior + shift) - pad` and `max(prior, prior + shift) + pad` land in different cycles. A boundary-crossing event cannot be handled by a within-cycle rotation at all; it needs the reference cycle redefined anchor-relative, which is out of scope here but worth naming.
2. Add a sentence to the "when not to use it" list: fixed-length cycles with an event that crosses the cycle boundary between consecutive occurrences are not supported. The printed `duplicated` / `orphaned` / max-displacement diagnostics do catch it, so the notebook should say those prints are the check, not decoration.

### Nit on the displacement note

"The displacement is the whole cost of anchoring. Its size is `2*pad + 1`" is right for the displaced non-event periods, but the quantity cell 10 prints is `max(|k|, 2*pad + 1)`, which equals 3 here only because `|k| <= 2`. With `pad=1` and a five-week move the event block itself displaces by 5 while the displaced weeks still displace by 3:

```
k = 5, pad = 1: displacements = {+3, +3, +3, +3, +3, -5, -5, -5}
                                 displaced weeks   event block
```

Worth either relabelling the print as two numbers (event realignment vs displaced-week cost) or saying the printed max is the event's own move when the move exceeds the window width.

Happy to push the guard and the two text changes if you have not started on them.

juanitorduz:
## Addressed in b7114b0, with one correction to the diagnosis

Thanks for stress-testing it. Both of your examples do break the map, and the fix is pushed, but the cause is not the cycle boundary, and the clamp you proposed would reject a case that works. Details, because the distinction decides what the guard has to check.

### The failing anchors are occurrences two weeks (or two years) apart, not windows across a boundary

The anchors are built from `holiday_week_of_year` as one occurrence per cycle. Your first calendar encodes indices `[48, 102, 104, 158, 212, 266]`: the pair `(104, 102)` says the event happened in week 50 of cycle 1 and again two weeks later in week 0 of cycle 2, so `shift = (104 - 52) - 102 = -50`. The rotation is then asked to rotate a 53-period segment by 50 and does exactly that. Your second calendar has `(105, 103)` at two weeks and `(258, 158)` at 100 weeks (`shift = +48`). Neither pair is "the same event one cycle earlier"; the week-of-year convenience produced them, and the function had no check.

A genuine yearly event that passes new year is fine. Drifting two weeks earlier per occurrence, `[55, 105, 155, 205, 255, 305]` (weeks 3, 1, 51, 49, 47, 45, so cycle 2 has two occurrences and the windows sit across the index boundary at 103|104 and 155|156):

```
duplicated: [], orphaned: []
event -> reference: [(105, 55), (155, 105), (205, 155), (255, 205), (305, 255)]
t - r(t) ranges over [50, 55] weeks
```

The rotation works on runs of consecutive periods and never looks at which calendar year an index falls in. Clamping the segment to `prior`'s cycle, or raising when `low`/`high` land in different cycles, would reject this one. The Ramadan-in-2030 case you motivate with (two occurrences in one calendar year, ~50 weeks apart) is precisely this shape and is representable by passing index pairs directly; what cannot be represented is that calendar through the one-week-of-year-per-cycle list, and that is a limitation of the demo encoding, not of the map.

### What is now enforced

`build_reference_map` raises `ValueError` when

- `|shift| > period // 2`, i.e. the two occurrences are not about one cycle apart (your two calendars fail here: `anchor (104, 102) pairs occurrences 2 periods apart`),
- an event window starts before the series (`low < 0`), which previously assigned a reference to a first-cycle period, or
- two anchors' windows overlap, since composing two rotations on shared periods breaks the bijection.

The docstring states the contract; "empty by construction" now reads "empty whenever the anchors satisfy the contract", and the notebook says the printed diagnostics are the check. Two new cells show the boundary-straddling drift succeeding and the week-of-year misuse being refused. Conclusions and the "when not to use it" list name the boundary: overlapping events, or occurrences not about a cycle apart, need an event-relative reference definition, which is out of scope and the builder refuses rather than approximates.

### Displacement nit

Right, and taken. The cell now prints the two quantities separately, event realignment `|k| = [1, 2, 1, 2, 1]` and pushed-aside displacement `2 * pad + 1 = 3`, plus the overall maximum, and the note says the maximum is whichever is larger. Checked on a five-week move with `pad=1`: displacements are exactly `{+3, -5}` as you listed.

Committed panel, tables and figures are unchanged (this calendar satisfied the contract already). Notebook re-executed end to end, hooks pass.

daimon-pymclabs:
## Contract holds under fuzzing, with one condition left unstated

Checked out `b7114b0` and re-ran everything independently. Your correction is right and mine was wrong: the failure is the anchor pair, not the cycle boundary. Both of my calendars encode occurrences two weeks apart through the week-of-year list, `shift` is then ~50, and the clamp I proposed would have rejected `[55, 105, 155, 205, 255, 305]`, which works. Confirmed the drift case returns `duplicated = []`, `orphaned = []`, `t - r(t)` in `[50, 55]`, and both new cells' committed outputs match a fresh run.

Guard behaves as described: `(104, 102)` and `(105, 103)` both raise, `|shift| = 26` passes and `27` raises at `period = 52`, `low < 0` raises, overlapping windows raise. Displacement is now two numbers and the five-week check gives `{+3, -5}` as stated.

### One condition the docstring does not name

I fuzzed 20,000 random calendars (`period` in 4, 7, 12, 52; 1 to 2 events; `pad` 0 to 3; `shift` drawn inside the contract) and kept every case that passed the guard but returned a non-empty `duplicated` or `orphaned`: 1,463 of them. Every single one is the same shape, and none is silent, so the guard is not leaking short-horizon ratios. But the condition itself is not in the contract.

When the last event's rotated window reaches past the end of the series, `0 <= t < n_periods` drops part of the rotation and the reference for those weeks is never assigned:

```python
build_reference_map(n_periods=312, period=52, anchors=((311, 257),), pad=1)
# duplicated: []      orphaned: [258]
# r(308) = 259, r(309) = 260   <- both point forward of the naive lag
```

The dropped rotation shows up as an orphan, so the printed diagnostics catch it, and that is exactly the role the notebook now gives them. The problem is only that the docstring says `duplicated` and `orphaned` are "empty whenever the anchors satisfy the contract above", and this calendar satisfies the contract. A reader who takes the docstring at face value has no reason to look at the prints.

I would not raise here. A series that ends partway through an event's window is a normal data situation, and refusing it would be worse than returning a reference with a short tail. Two sentences in the docstring cover it: the current window must fit inside the series for the bijection to be exact, and when it does not the affected weeks appear in `orphaned` rather than raising. The "when not to use it" list already has the right neighbours for it.

Nothing else to flag. Happy to push the docstring edit if you would rather not re-execute for it, but nothing in the committed panel changes either way.

Issue #2835, opened 2026-08-07
Target values outside the likelihood support build fine and die as -inf logp


daimon-pymclabs:
Split out of #2818, where it was raised as a second gap and deliberately kept out of that fix's scope.

## Problem

A target containing values outside the likelihood's support builds without error and fails later as a `-inf` logp, with nothing pointing at the cause.

Reported instance: `link="identity"` with a `LogNormal` likelihood and a target containing negative values. `IdentityLinkSpec.validate_target` is a no-op, so nothing rejects it at build time.

## Why this is not a link concern

That specific pair goes away once #2818 rejects identity + `LogNormal`, but the general gap does not. `TruncatedNormal(lower=0)` with a negative target also builds and also dies at `-inf` logp, under a link/likelihood pair that stays legal after that fix.

`IdentityLinkSpec.validate_target` being a no-op is correct behaviour *for the link*: the identity link genuinely accepts any real value. The missing check is that we never validate the target against the **likelihood's** support, which belongs wherever the `Prior` is resolved rather than in `LinkSpec`.

Note that `LogLinkSpec.validate_target` currently rejects non-positive `y`. That check is doing double duty: it is stated as a link constraint but is really the `LogNormal` support check, which is why the identity side has no equivalent to inherit.

## Proposal

Add a likelihood-level support check that runs at build time alongside the existing `validate_target` call (`mmm/mmm.py:2216-2217`), covering at least:

- `LogNormal`, `Gamma`: `y > 0`
- `TruncatedNormal`: `lower <= y <= upper` for whichever bounds are set
- `NegativeBinomial`, `Poisson`: non-negative integers
- `Beta`: `0 < y < 1`
- unrecognised distributions: no check, no warning (the likelihood check in #2818 already warns for these)

Error messages should name the distribution, the violated bound, and how many target values violate it.

Open question worth settling in the implementation: whether `LogLinkSpec.validate_target` then delegates to the likelihood check rather than keeping its own hard-coded positivity rule, so the two cannot drift.

shivamlalakiya:
Picking this up. Before writing the check I measured what the likelihood actually sees, and it changes where the check has to go and how the open question about `LogLinkSpec.validate_target` should be answered.

## The likelihood does not observe the target

`build_model` observes the *scaled* target, not `y` (`mmm/mmm.py:2322-2331`, `2479`):

```python
target_data_scaled = _target / _target_scale
target_data_scaled = pmd.math.switch(
[...]

Issue #2834, opened 2026-08-07
IdentityLinkSpec.mean_correction is unconditionally 1.0, so TruncatedNormal contributions are silently off by the truncation factor


daimon-pymclabs:
Split out of #2818, where it surfaced while narrowing the identity-link likelihood check.

## Problem

`IdentityLinkSpec.mean_correction` returns `xr.DataArray(1.0)` unconditionally (`pymc_marketing/mmm/link.py`), with the docstring "for the Normal likelihood the mean equals the median". That is true for `Normal`, but the identity link accepts other likelihoods, and for `TruncatedNormal` the claim is false: truncation shifts `E[y]` off `mu`, so anything comparing contributions against the conditional mean is off by the truncation factor with nothing saying so.

This is not a hypothetical pairing. `TruncatedNormal(lower=0, sigma=...)` is what the in-repo guidance recommends for non-negative targets (`skills/mmm-modeling/SKILL.md`, `skills/mmm-modeling/references/model_specification.md`, including the table recommending it), and `tests/mmm/test_link.py` asserts the identity link accepts it. So the default advice produces the silently uncorrected case.

## Blocker: the signature cannot express the right answer

`mean_correction` is a method on `LinkSpec`, so it sees only the link. Under `link="identity"` the correct correction depends on the **likelihood**:

- `Normal`: no correction
- `Gamma`, `NegativeBinomial` in their `mu` parameterisations: no correction
- `TruncatedNormal`: truncation correction
- `LogNormal`: rejected outright once #2818 lands

The information that decides the answer is not in scope at the current signature. This needs the likelihood `Prior` passed in, or dispatch on the `(link, likelihood)` pair, before the rest of the fix is worth writing.

## Blocker: `mu` is not in the posterior

The correction is pointwise in `mu`, but `mu` is registered as a Deterministic only under the log link (`mmm.py:2393`); the identity branch only sets `mu_var.name = "mu"` (`mmm.py:2395`), which labels the variable without registering it, so it never reaches the posterior.

[...]

PR #2841, opened 2026-08-07
fix(mmm): reject non-response-scale likelihoods under link='identity'

### LINKED ISSUE #2818: link="identity" accepts a LogNormal likelihood and every per-component contribution is then wrong
## Summary

`MMM(link="identity", ...)` accepts a `LogNormal` likelihood without complaint, and every per-component contribution the model then reports is wrong. `validate_likelihood_compatibility` returns unconditionally for the identity link:

```python
# pymc_marketing/mmm/link.py:169-170
def validate_likelihood_compatibility(self, likelihood) -> None:
    """Accept any likelihood: the additive decomposition does not depend on the
    distributional form."""
    return
```

The stated reason is true only when `mu` is the response mean. Under `Prior("LogNormal", mu=..., sigma=...)` it is not: `mu` is the mean of `log(y)`, so the response is `exp(mu)` up to a scale and the decomposition is multiplicative, not additive. The identity `LinkSpec` nonetheless registers `channel_contribution_original_scale` as `channel_contribution * target_scale` and lets the contribution decomposition sum in the linear predictor.

## Reproducer

```python

mmm = MMM(
    channel_columns=["channel_1", "channel_2"],
    date_column="date",
    target_column="y",
    adstock=GeometricAdstock(l_max=4),
    link="identity",
    model_config={"likelihood": Prior("LogNormal", sigma=Prior("HalfNormal"))},
)
mmm.build_model(X=X, y=y)   # builds, no warning
```

## Why it matters beyond one method

This is not specific to incrementality, which is why it should not be fixed inside #2813. Anything that reads a `*_original_scale` contribution inherits it:

- `add_original_scale_contribution_variable` warns about exactly this hazard *under a log link* ("a per-component `*_original_scale` variable is a multiplicative factor rather than an additive contribution") and stays quiet here, because the link says identity.
- `compute_mean_contributions_over_time` and the contribution plots present numbers that do not add up to the response.
- `IdentityLinkReducer` sums log-scale deltas and multiplies by `target_scale`, so incrementality returns a quantity in no particular units.

In every case the number is plausible-looking and finite, so nothing surfaces.

## What the fix has to decide first

The mechanical part is small, but there is a real question underneath it: what is `link="identity"` with a `LogNormal` likelihood *supposed* to mean? Two coherent answers, and they lead to different fixes.

1. **It is a mistake.** The combination is a user specifying a multiplicative model through the additive API. `validate_likelihood_compatibility` should refuse it and point at `link="log"`, which already implements the correct decomposition (`LogLinkSpec.create_media_contribution_deterministic` computes `exp(mu)*s - exp(mu - mu_media)*s`) and the correct median/mean distinction (`LinkSpec.mean_correction`). Cheapest, and probably right.

2. **It is a legitimate "positive-response Normal-ish" model** where the user does not want the log link's multiplicative contribution semantics. Then the identity `LinkSpec` has to stop claiming its decomposition is valid: no `*_original_scale` per-component variables without a warning, and incrementality raising `NotImplementedError` rather than reducing additively.

My inclination is (1), narrowed to: refuse any likelihood whose mean is not `mu` under the identity link, listing the ones that are. That turns a silent wrong number into an error at build time and costs one method.

Either way the check belongs in `LinkSpec`, since it is the object that knows how `mu` maps to the response, and it already has the hook.

## Not urgent, but silent

Nothing about this is new — it predates #2813 and #2815 — and it needs a deliberate specification decision rather than a patch, which is why it is filed rather than folded into the branch that surfaced it.

anevolbap:
Fixes #2818

`MMM(link="identity")` accepted a `LogNormal` likelihood. `mu` is then the mean of `log(y)`, not of `y`, so every `*_original_scale` contribution was a log-scale delta multiplied by `target_scale`, which is not a contribution in any units. `validate_likelihood_compatibility` early-returned for the identity link, so nothing stopped it.

The check now applies the criterion agreed in the thread: `mu` has to be on the scale of the response. `RESPONSE_SCALE_LIKELIHOODS` (`Normal`, `StudentT`, `TruncatedNormal`, `Gamma`) passes, `NON_RESPONSE_SCALE_LIKELIHOODS` (`LogNormal`) raises, and anything else warns rather than raises so custom priors keep building.

The raise applies on the load path too, as decided in the thread, so the error message carries the `idata_to_init_kwargs` recipe for repairing a saved model without refitting. There is no dedicated load test, because once this lands you cannot produce a saved identity plus `LogNormal` file through the public API, so such a test would have to hand-build the idata. Say the word if you want one anyway.

One thing found while writing it: a `Censored` likelihood holds a `Prior` in `.distribution` instead of a name, and `Prior` is unhashable, so the set membership raised `TypeError` where identity used to build. `_distribution_name` unwraps until it reaches a name, so `Censored(Normal)` passes and `Censored(LogNormal)` raises. That also changes the log branch, where `Censored(LogNormal)` used to raise `TypeError` and now passes.

#2834 (`mean_correction` under truncation) and #2835 (target versus likelihood support) stay separate.


pymc_marketing/mmm/link.py
```diff
@@ -39,6 +40,30 @@ class LinkFunction(StrEnum):
     LOG = "log"
 
 
+#: Likelihoods whose ``mu`` parameter is on the scale of the response, so the
+#: additive decomposition under the identity link is in the units of the target.
+RESPONSE_SCALE_LIKELIHOODS = frozenset(
+    {"Normal", "StudentT", "TruncatedNormal", "Gamma"}
+)
```
anevolbap, line 64: Kept to the four names settled in the thread. `Laplace`, `Beta` and `InverseGamma` are also in `pymc.dims` and also take `mu` on the response scale, so they warn here today. Add them, or leave that for a separate change?

daimon-pymclabs (commented):
Read this against the trace in #2818. The check lands where it should, and the criterion in the docstring (`mu` on the scale of the response, unknown likelihoods warn rather than raise) is the one we settled on. `_distribution_name` unwrapping `Censored` is a good catch that I missed in the original trace.

I also checked the repair recipe in the error message, since "repair without refitting" is a strong claim. It holds: `create_likelihood_variable` is handed `mu=mu_var`, the linear predictor, and `inverse_link` is never applied to the likelihood's `mu`. So `link="identity"` with LogNormal and `link="log"` with LogNormal build the *identical* PyMC graph, and only the `*_original_scale` Deterministics differ. Flipping `kwargs["link"]` and rebuilding really does reinterpret an existing posterior correctly rather than requiring a new fit. Worth a sentence in the docstring, because that equivalence is what makes the recipe sound and it is not obvious from reading `LogLinkSpec`.

Three smaller things.

**The log branch widened, and only the PR description says so.** `Censored(LogNormal)` used to hit `TypeError` on the set membership and now passes, which `test_validate_likelihood_compat_log_looks_through_censored` pins. That is probably the behaviour you want, but under log the counterfactual is `exp(mu) - exp(mu - media)`, which is a median-scale delta for the *latent uncensored* variable, not for the observed clipped `y`. For a censored model that is arguably the quantity people want (unconstrained demand), but it is a different quantity from the uncensored case and nothing in the code says so. I would put that distinction in `create_media_contribution_deterministic`'s docstring or in the `LINK_LIKELIHOODS` comment, so it does not have to be rediscovered from the PR description.

**`dist_name is None` reads badly on the log branch.** The identity branch falls back to `type(likelihood).__name__` in the warning, but the log branch interpolates the raw value, so a likelihood with no resolvable name produces `Likelihood 'None' is not compatible with link='log'`. Same fallback would fix it.

**`TruncatedNormal` in `RESPONSE_SCALE_LIKELIHOODS` is correct under the stated criterion but invites a second misreading.** Its `mu` is in the units of the target, so contributions are in target units and the check is right to admit it. But `mu` is not `E[y]` under truncation, so `*_original_scale` will not reconcile against the posterior predictive mean. That is #2834's scope, not this PR's, though a pointer to #2834 next to the frozenset would stop the set reading as "everything here is safe end to end."

On the missing load test: agreed, hand-building an idata to test a state the public API can no longer produce is not worth it. The build-path tests cover the check itself.

anevolbap:
All four applied in d8cd12dc.

I checked the graph equivalence independently before writing it down, since the docstring now asserts it. `inverse_link` has no call site outside `link.py`, and building `link="identity"` and `link="log"` with the same `LogNormal` likelihood and the same `model_config` gives identical free variables (`adstock_alpha`, `intercept_contribution`, `saturation_beta`, `saturation_lam`, `y_sigma`) and an identical observed logp at the initial point (`-107.6040426782519` both). The only difference is the Deterministics: log adds `mu` and `y_original_scale`, and computes `total_media_contribution_original_scale` differently. That is now in the `validate_likelihood_compatibility` docstring next to the recipe.

The censored distinction went into `LogLinkSpec.create_media_contribution_deterministic`, since that is where the counterfactual is built, phrased as a median-scale delta for whatever variable the likelihood puts `mu` on, which under `Censored(LogNormal)` is the latent one.

The log branch now uses the same `type(likelihood).__name__` fallback, with a test.

The `#2834` pointer sits next to the frozenset, saying the set is about units only and that `mu` still need not equal `E[y]`.

One question from the earlier comment is still open: `Laplace`, `Beta` and `InverseGamma` are in `pymc.dims` and take `mu` on the response scale, so they warn today. In or out of this PR?

pymc_marketing/mmm/link.py
anevolbap: On the stated criterion the honest list is five, not three: `Poisson(mu)` and `NegativeBinomial(mu=...)` also take `mu` on the response scale in `pymc.dims`. Three of the five run into #2835, since `Beta` needs the target in `(0, 1)` and the count ones need non-negative integers, and nothing here checks that. Admitting them turns a warning into a silent pass followed by an `-inf` logp, so they stay out of this PR.

juanitorduz:
@anevolbap thank you for working on this one. @PabloRoque pointed out that we have a much better parametrization of the LogNormal Distribution https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/special_priors.py#L262. I guess it makes sense to use this one, right?

anevolbap:
The blocker is `create_likelihood_variable`. `LogNormalPrior` does not have it, and `SpecialPrior` only has `create_variable`, so it cannot go in `model_config["likelihood"]` yet.

Once it can, the check should admit it under identity. `__init__` maps `mu` to `mean`, so `mu=mu_var` lands on the response scale and `E[y] = mu`. Under log it should stay rejected.

Fixed a small bug in e38370b9. `_distribution_name` returned `None` for anything without `.distribution`, so both set lookups compared `None` and the name only reached the warning text. It now falls back to the class name.

I did not add `LogNormalPrior` to `RESPONSE_SCALE_LIKELIHOODS` yet. Today it would pass silently and then fail on `AttributeError`. Add it, or wait for the method?

daimon-pymclabs:
On the open question: leave `LogNormalPrior` out of `RESPONSE_SCALE_LIKELIHOODS`, and when the method lands, admitting it under identity needs more than the units argument.

Two reasons for leaving it out now. `_distribution_name` falls back to `type(likelihood).__name__` for anything without `.distribution`, so today it resolves to the string `LogNormalPrior` and takes the warn branch. That is the correct outcome: it warns, then fails on the missing `create_likelihood_variable` with an `AttributeError` that names the real problem. Adding the string would silence the warning and leave only the `AttributeError`, which is strictly worse for the reader. And the string it would resolve to is a class name, not a distribution name, so the frozenset would hold two namespaces at once while the docstring describes it as distribution names.

The second reason is the one I would want settled before it goes in. `E[y] = mu` holds for the intended positive-scale parameterisation, but not for an unconstrained `mu`. With `mean = m` and `std = s`, `create_variable` sets `mu_log = ln(m**2 / sqrt(m**2 + s**2))` and `sigma_log = sqrt(ln(1 + s**2/m**2))`, so

```
E[y] = exp(mu_log + sigma_log**2 / 2)
     = (m**2 / sqrt(m**2 + s**2)) * sqrt((m**2 + s**2) / m**2)
     = m**2 / |m|
     = |m|
```

Every appearance of `m` is squared, so the sign is folded away rather than rejected. Passing `mean=mu_var` under `link="identity"` means `E[y] = |mu|`, while the contribution Deterministics decompose `mu` itself. Wherever the posterior linear predictor dips below zero the additive decomposition and the likelihood mean disagree in sign, silently, with no NaN to signal it. `LogLinkSpec.validate_target` cannot catch this because the quantity that has to stay positive is the fitted `mu`, not the data.

That is a caveat on the likelihood implementation, not on this PR. It just means the eventual `create_likelihood_variable` should either constrain or check positivity, and the frozenset entry should wait for that rather than land alongside it. Worth its own issue so it does not get rediscovered from this thread.

@juanitorduz on the better parameterisation: agreed it is the one to reach for, and it is a genuinely nicer prior for `mean` and `std` in the units people think in. Note the call site passes `xdist=True` (`mmm.py` L2407), and `LogNormalPrior.create_variable` already requires `xdist=True`, so that part lines up. The likelihood version cannot delegate to `create_variable` though, since that ends in `pm.Deterministic(name, exp(log_phi))` rather than an observed RV. It needs to compute `mu_log`/`sigma_log` from `mu` and `std` and then build the observed `pmd.LogNormal` directly.

juanitorduz:
Ok, I guess we can leave it out for now

daimon-pymclabs:
Agreed. Filed the caveat as #2858 so the positivity issue and the missing `create_likelihood_variable` are tracked separately from this PR. This one is ready from my side.

daimon-pymclabs (approved):
Final pass. Nothing has changed since my last read except `cbadc796`, a merge of `main`, and `main` has not touched `link.py` or the validate call site in the meantime, so the logic I checked before is the logic that would land. The branch is level with `main` and mergeable.

Re-verified the four things from the earlier round, all in:

- The graph-equivalence sentence sits next to the repair recipe in `validate_likelihood_compatibility`, which is what makes "rebuild rather than refit" a safe thing to tell a reader.
- The censoring caveat is in `LogLinkSpec.create_media_contribution_deterministic`, phrased as a median-scale delta for whatever variable the likelihood puts `mu` on, plus the note that `validate_target` rejects non-positive targets so left-censoring at zero cannot be fitted under this link anyway.
- The log branch now runs through `_distribution_name`, so the `'None' is not compatible` message is gone, with `test_validate_likelihood_compat_log_unnamed_uses_class_name` pinning it.
- The `#2834` pointer is next to the frozenset, and the comment says the set is about units only.

Two checks I had not run before, both clean:

- `LogNormalPrior` still takes the warn branch, as intended. `SpecialPrior` does not subclass `Prior`, defines no `distribution` attribute and no `__getattr__`, so `_distribution_name` falls through to `type(likelihood).__name__`. The user gets the warning and then the `AttributeError` naming the missing `create_likelihood_variable`, which is the readable failure. That is the behaviour #2858 assumes.
- No blast radius from the new warn branch. Every explicit `model_config["likelihood"]` in the MMM notebooks is `Normal` or `TruncatedNormal`, both in the set, and the only `LogNormal` occurrences outside `link.py` and its test are `LogNormalPrior` in `tests/test_special_priors.py`, which never reaches this check. The `Poisson` and `NegativeBinomial` likelihoods in `tests/bass` belong to the Bass model, not to the MMM link path. Nothing in docs or tests starts warning or failing because of this.

One small gap, not a blocker. The unwrap was written for `Censored`, and the tests cover the two passing cases (`Censored(Normal)` under identity, `Censored(LogNormal)` under log) plus the unwrap itself in `test_distribution_name_falls_back_to_class_name`. The rejecting case, `Censored(LogNormal)` under identity, is claimed in the PR description but not pinned by a test. That is the case where the unwrap changes a `TypeError` into a real error message, so it is the one most worth a regression test. A single `pytest.raises(ValueError, match="not compatible with link='identity'")` on `Censored(Prior("LogNormal", sigma=1), lower=0)` next to the two existing censored tests would close it.

Approving. CI is still running on the merge commit, so merge on green.

juanitorduz (approved):

juanitorduz:
Thank you @anevolbap !

Issue #60, opened 2022-10-19
Figure out API for incorporating information from lift tests on MMM


ricardoV94:
@juanitorduz mentioned that in the literature this is usually incorporated into the prior information, but I think that in the HelloFresh project @lucianopaz found a way to incorporate such information via observed variables in the same MMM model?

juanitorduz:
I think with the great work by @wd60622 in https://github.com/pymc-labs/pymc-marketing/pull/590, we can close this one 🚀 !

PR #2843, opened 2026-08-08
Add a placebo (negative-control) channel example notebook

ipezygj:
## Why

An MMM is fitted against the same outcome it is later asked to explain, so part of every channel's reported contribution is the share of leftover variation that any plausible-looking spend series would have collected. The package has time-slice CV and lift-test calibration, but nothing answers *"how much credit does this model give to a channel that cannot have caused anything?"*

A placebo channel measures that directly: take a real spend series, destroy its alignment with the outcome, put it back as an ordinary channel, refit. Whatever the model still assigns to it is the floor. This is the negative-control idea from epidemiology (Lipsitch, Tchetgen Tchetgen & Cohen, 2010) applied to attribution.

## What is in the diff

**One example notebook and its gallery entry, no package code.** `docs/source/notebooks/mmm/mmm_placebo_check.ipynb`, built on `MMM` and the data set from the multidimensional example, with the placebo entering `channel_columns` as an ordinary channel. The construction is a dozen lines of NumPy, so the notebook is the whole recipe rather than an API to import.

An earlier revision of this PR did add `pymc_marketing/mmm/placebo.py` with `add_placebo_channel`, `summarize_placebo_contribution` and six tests. That was replaced by the notebook at a maintainer's suggestion, and the description above this line used to describe the module, apologies to anyone who went looking for it.

**That trade is worth naming, because it is a real one.** A gallery entry makes this a supported technique whose copy-paste implementation has no tests behind it. If you would rather have the helper back in the package with its tests and keep the notebook as the narrative, say so and I will reinstate it; the two are not in conflict and the notebook would shrink to a call.

## What the notebook now contains

Four fits, on the simulated multidimensional data set:

1. **The placebo run.** Three channels, one of them a circular shift of `x1` within geo. Reports each channel's posterior contribution share and `P(> placebo)`.
2. **A baseline without the placebo.** Shares are compositional, so a third channel shrinks every share by arithmetic alone. The baseline separates *the placebo revealed that this channel was overstated* from *three channels divide the same pie three ways*, and gives the number a stakeholder actually asks for: how much of this channel's contribution the placebo displaced.
3. **A negative control on the target**, `y` permuted within geo, described as what it is, a check on the plumbing. It cannot show that the statistic discriminates, because permuting the target leaves the three channels exchangeable and almost any share comparison would return 0.5 there.
4. **A positive control on data whose answer is known.** `data/mmm_multidimensional_example_true_parameters.nc` ships alongside the CSV, so `saturation_beta`, `saturation_lam` and `adstock_alpha` can be fixed with `pm.do`: the real channels amplified in their true ratio, the placebo's beta set to exactly zero, one target drawn from the prior predictive. The recipe has to recover the ordering that was put in. This is the run that shows the check discriminates.

The notebook also states, where the reader meets the table, that a placebo built from `x1` competes hardest with `x1`, so `P(> placebo)` for that channel is the least favourable comparison in it; and that multiple placebos belong in separate refits rather than in one model, since five placebos in one fit take a seventh of the pie each by arithmetic.

## Checks

- Notebook executed end to end; stored outputs match the code in the diff.
- Stored outputs no longer carry my local paths, an earlier revision leaked `.venv` warnings onto the public gallery page.
- No new dependencies; no package code touched.

----
📚 Documentation preview 📚: https://pymc-marketing--2843.org.readthedocs.build/en/2843/


juanitorduz:
Hi @ipezygj. Thanks for opening the PR! I think this can be a good example notebook. Meaning, the function itself is a couple of NumPy + pandas code, so I do not think it needs a submodule. That being said, I do think an example notebook doing some placebo analysis should be helpful. Would you be open to it?

ipezygj:
Yes, very much — a notebook is a better home for this than a submodule, and you are right that the construction itself is a few lines of NumPy and pandas.

Here is what I would put in it, so you can tell me early if the shape is wrong:

1. Fit an MMM on one of the datasets already used in the docs, so the notebook does not introduce data of its own.
2. Add a placebo channel: a copy of a real spend series with its time alignment destroyed. Circular shift as the default, because it keeps the flighting and autocorrelation structure intact and is therefore the harder test; a plain permutation shown alongside as the weaker variant, since it keeps only the marginal distribution.
3. Refit and show the placebo channel's posterior contribution share with its HDI.
4. For each real channel, the share of posterior draws in which its contribution exceeds the placebo's. That single number is the part I actually care about: it turns "this channel looks important" into "this channel outperforms a series that cannot have caused anything, in N% of draws".
5. A short read-out section: what a healthy result looks like, what a warning sign looks like, and the limitation stated plainly — a placebo landing near zero is necessary but not sufficient, since it rules out one failure mode rather than confirming the model.

Two questions before I start:

- Which example dataset would you prefer it to build on, so it sits naturally next to the existing notebooks?
- Should it use `MMM`, the multidimensional variant, or both?

I will close this PR and open a fresh one with just the notebook, unless you would rather I convert this one in place.

juanitorduz:
I think you can use https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_multidimensional_example.html and the MMM model form the mmm.py module with is already multidimensional.

ipezygj:
Converted in place, c017bab. The submodule, its tests and its exports are gone; what is left is `docs/source/notebooks/mmm/mmm_placebo_check.ipynb`, built on the `MMM` class from `mmm.py` and the data set from {ref}`mmm_multidimensional_example`, as you suggested. Model spec is that notebook's, unchanged, with the placebo entering `channel_columns` as an ordinary channel. Gallery card added under MMM → Model Evaluation.

What it shows, on your own example data:

| | share | 94% HDI | P(> placebo) |
|---|---:|---|---:|
| x1 | 0.451 | 0.145 – 0.766 | 0.784 |
| x2 | 0.309 | 0.038 – 0.572 | 0.628 |
| x1_placebo | 0.240 | 0.013 – 0.495 | — |

The placebo is a circular shift of `x1` within each geo (correlation with the real `x1`: 0.082), so it carries the shape of a media series and none of its timing. It still collects 24% of the contribution, and neither real channel clears it convincingly. Given that the multidimensional notebook describes this data as deliberately noisy and low-signal, I read that as the honest answer rather than an embarrassing one — and it is the answer a contribution chart alone does not give, because that chart has no scale on it.

The notebook then runs the same analysis with the target scrambled within geo, so that nothing can be found:

| | share | P(> placebo) |
|---|---:|---:|
| x1_placebo | 0.363 | — |
| x1 | 0.354 | 0.491 |
| x2 | 0.283 | 0.385 |

The placebo comes out on top and the real channels sit at a coin flip. I included this because a diagnostic that cannot fail is not a diagnostic; this shows what the check reports when there is genuinely nothing there, and it calibrates the first table.

Two things I would flag for you:

- Sampling uses `nuts_sampler="nutpie"` — my machine has no C compiler, so that is what actually produced the stored outputs. Happy to switch it back to the default and re-run if you would rather the notebook match the others.
- With a strictly positive `saturation_beta`, a placebo can never come back at exactly zero, so the raw share always looks non-trivial. The notebook says this explicitly and points readers at `P(> placebo)` instead. If you would rather the example used a specification where beta can be zero, say so.

Watermark, `nbformat` validation and `gallery.md`-in-sync all pass; `validate-api-docs` and `validate-watermark` fail in my pre-commit only because the hooks call `python3`, which does not exist on Windows — both pass when run directly.

docs/source/notebooks/mmm/mmm_placebo_check.ipynb
```diff
@@ -0,0 +1,844 @@
+{
```
juanitorduz, line 1: <p>Use Marketing Mix Models and not Media Mix Models</p><p> </p><p>Remove all the em dashes, as this is a feature of AI-generated text; use proper punctuation. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2843/discussion/#comment-3757208538'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

juanitorduz, line 1: <p>the legend overlaps with the plot , please fix it</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2843/discussion/#comment-3757208543'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    ax.plot(subset["date"], subset["x1"], label="x1 (real)", color="#1f77b4")
</pre><p>Use "C0" or "C1" (or similar instead of custom color codes</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2843/discussion/#comment-3757208546'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #60.            draws=500,
</pre><p>1000 draws and 1000 tune please</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2843/discussion/#comment-3757208549'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

juanitorduz, line 1: <p>as we did the data generating process, maybe</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2843/discussion/#comment-3757208552'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #10.        ax.hist(
</pre><p>use arviz az.plot_sit</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2843/discussion/#comment-3757208557'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

juanitorduz, line 1: <p>can you more precise about ", but the outcome scrambled within each geo" ... and also what do you expect from it? can you ealborate more?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2843/discussion/#comment-3757208559'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

juanitorduz, line 1: <p>Do you have any recommendation on when to do this and if we need to add more than one placebo?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2843/discussion/#comment-3757208561'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

ipezygj:
Thanks for the review. All of it is in and the notebook is re-executed end to end, so the stored outputs match the new code.

**Wording.** "Marketing Mix Model" throughout, and every em dash is gone; there is now an assertion in my edit script that fails if one comes back.

**Plots.** `C0` / `C1` instead of the hex codes. The legend is out of the axes (`bbox_to_anchor`, constrained layout), so nothing overlaps the series any more. The posterior plot is `az.plot_dist` now. Worth noting for other notebooks: under ArviZ 1.x that is the `arviz-plots` API rather than the old ax-based one, so overlaying the channels in a single panel is

```python
az.plot_dist(
    share.rename("contribution_share").to_dataset(),
    kind="kde",
    aes={"color": ["channel"]},
    cols=None,
    visuals={"credible_interval": False, "point_estimate": False, ...},
)
pc.add_legend("channel")
```

**Sampling.** 1000 draws, 1000 tune, 4 chains, 0 divergences on both fits. The conclusions do not move: the placebo takes 0.239 of the contribution (was 0.240 at 500/500), `x1` 0.455 at P(> placebo) 0.785, `x2` 0.306 at 0.625. On the scrambled target the placebo again comes out on top, 0.359, with the real channels at 0.492 and 0.403.

**"can you be more precise about the outcome scrambled within each geo, and what do you expect from it?"** Rewritten to state the operation and the prediction separately. The permutation is applied to `y` within each geo across dates, so each geo keeps its level and its marginal distribution, the spend series keep their flighting and autocorrelation, and the sample size is unchanged; what is destroyed is the pairing of a date's spend with that same date's sales. The expectation is now explicit: with no effect left to find, the placebo should be indistinguishable from the real channels and every real channel should sit at P(> placebo) near 0.5, and a real channel clearly above 0.5 there would mean the check reports evidence where none exists.

**"any recommendation on when to do this and if we need more than one placebo?"** Both are now their own sections. When: once per specification, before any contribution number leaves the analysis, and again whenever the specification changes, because the floor moves with the priors and controls and not only with the data. How many: one gives a floor, but the offset is random, so three to five from different source channels and offsets turn a point into a range. Read the highest rather than the average, since the question is how much credit the model *can* hand out for nothing. If the placebos disagree widely, that spread is itself the result.

Two open questions back to you.

1. **"as we did the data generating process, maybe"** on the posterior cell. I want to make sure I do what you meant. `data/mmm_multidimensional_example.csv` ships only `date, geo, x1, x2, event_1, event_2, y`, so the true contributions are not recoverable from the file. If you have the generating script with the true betas, comparing the placebo floor against the known truth would make this notebook considerably stronger and I will add it. If you meant something else, say so and I will do that instead.

2. **`nuts_sampler="nutpie"`** is still in the fit. My machine has no C compiler, so that is what produced these outputs. Happy to drop it and re-run if you would rather this notebook match the others.

daimon-pymclabs:
Thanks for this, the writing is clear and the framing of the floor as "credit the model hands out for nothing" is the right one. A few things before this goes in, plus a definite answer to your two open questions.

## Main methodological gap: the placebo is cloned from, and judged against, the same channel

`x1_placebo` is a circular shift of `x1`, and it enters the fit alongside `x1`. It therefore competes hardest with exactly the channel it was cloned from: same marginal, same flighting, same autocorrelation, same scale entering adstock and saturation. Low raw correlation (0.082) does not remove this, because the competition is over the shape of credit the model can allocate, not over the pointwise correlation.

Two consequences worth surfacing in the notebook:

1. **`P(> placebo)` for `x1` is the least favourable comparison in the table**, and the notebook currently reads `x1` at 0.785 and `x2` at 0.625 as if both were measured against a neutral reference. Suggest building the worked placebo from a channel other than the one whose contribution is being interrogated, or building one per channel.
2. **There is no baseline fit without the placebo.** Shares are compositional, so going from two channels to three shrinks every share mechanically, and a reader cannot separate "the placebo revealed that `x1` was overstated" from "adding a third channel divided the same pie three ways". A short two-channel fit shown next to the three-channel one would settle this and costs one extra sample. It would also let you state how much the placebo displaced the real channels, which is itself an informative number.

## "How many placebos" is ambiguous in a way that changes the answer

The section says three to five from different source channels and offsets, and "cost only refits". Five placebos in **one** fit and five **separate** fits with one placebo each give different floors, again because the shares are compositional: with five placebos in one model each takes roughly a seventh of the pie by arithmetic alone, and the floor looks artificially low. Only the separate-refit version produces floors comparable to each other and to the single-placebo table above. Worth making explicit, since the naive reading is to add them all at once.

## Question 1: the true parameters are in the repo

You wrote that the true contributions are not recoverable from the CSV. They are, via a second file that ships alongside it. `scripts/data_generators/mmm_data_generation.py` writes both `data/mmm_multidimensional_example.csv` and `data/mmm_multidimensional_example_true_parameters.nc`, and the latter contains `saturation_beta`, `saturation_lam` and `adstock_alpha` for the generating process. So the comparison against known truth is available today.

That matters more than as a nice-to-have, because it fixes the weakest section of the notebook.

## The scrambled-target null is a weak falsification

Permuting `y` within geo destroys the seasonality, the trend and the level structure, not just the spend-to-sales pairing. Once that is gone, the three channels are exchangeable as far as the model is concerned, so `P(> placebo)` near 0.5 follows from symmetry more or less regardless of whether the statistic is doing anything useful. Almost any share-comparison statistic would return 0.5 there. The section claims the check "discriminates", but this particular test cannot distinguish a working diagnostic from a broken one.

The sharper demonstration is a positive control: a channel with a known-large true beta that clearly beats the placebo, on data whose answer you know. With the `.nc` above, that is now straightforward, and I would rather see it than the scrambled-target run, or alongside it.

## Smaller items

- **Stored outputs leak a local environment.** Cell 2 carries `g++ not available, if using conda: ...`, and the fit cells carry `C:\Users\ipezy.DESKTOP-GD1DJED\oss\pymc-marketing\.venv\...` warnings from `tqdm` and `rich`. These render on the public gallery page. Worth clearing before merge.
- **`np.abs(contribution)`** is a no-op under this specification, since `saturation_beta` is strictly positive, but as written the recipe conflates a large negative contribution with a large positive one if someone reuses it under a specification where beta can be negative. One sentence, or drop the `abs`, would cover it.
- **The PR description is stale.** It still describes `pymc_marketing/mmm/placebo.py`, `summarize_placebo_contribution`, and six tests in `tests/mmm/test_placebo.py`. The diff contains only the notebook and the gallery entries, and the notebook states the opposite design ("the whole recipe rather than an API to import"). Worth rewriting the body so a reviewer is not looking for a module that is not there. Separately, it is a fair question for the maintainers whether the helper should live in the package after all, since a gallery entry makes this a supported technique whose copy-paste implementation has no tests behind it.
- **Gallery title** is long next to its siblings ("Metrics and Model Evaluation", "Causal Identification"). Something like "Placebo Channel Check" would sit better in the grid, with the current sentence as the notebook's H1.

## Question 2: nutpie is fine, no need to re-run

`nutpie>=0.16.7` is a declared dependency in `pyproject.toml`, four other notebooks already set `nuts_sampler`, and `docs/source/conf.py` sets `nb_execution_excludepatterns = ["*.ipynb"]`, so the docs build never re-executes notebooks and uses your stored outputs as-is. Leave it.

ipezygj:
Thank you, this was a better review than the notebook deserved. All of it is in, at `bd034ec`, and two of your points changed the result rather than the wording.

## The baseline, and what it showed

You were right that the table could not separate the two readings, and the answer turns out to be the unflattering one.

```
         share without placebo  share with placebo   lost  dilution alone
x1                       0.598               0.455  0.143           0.143
x2                       0.402               0.306  0.096           0.096

x1 : x2 without the placebo  1.488
x1 : x2 with the placebo     1.489
```

`lost` matches proportional dilution to three decimals and the ratio does not move. The placebo took a proportional cut from both channels and reordered nothing, so it did not reveal that `x1` in particular was overstated. The notebook now says that in those words: what this measures on this data is the whole allocation being generous, which both channels inherit in proportion, not a fact about one line in the media plan. The `lost` / `dilution alone` pair is also the number a stakeholder asks for, so it earns its place regardless.

## The true parameters were there, and the positive control is the run that matters

I was simply wrong that they were not recoverable, and I did not go looking. With `pm.do` fixing `saturation_beta`, `saturation_lam` and `adstock_alpha`, the real channels amplified in their true ratio and the placebo's beta set to exactly zero, one draw from the prior predictive gives a target whose answer is known:

```
generating truth   x1 0.663   x2 0.337   placebo 0.000
recovered          x1 0.643   x2 0.337   placebo 0.019     P(> placebo) = 1.00 for both
```

The residual 0.019 on the placebo is the floor that a strictly positive `saturation_beta` guarantees, which is the same point the notebook makes about why `P(> placebo)` is the readable quantity rather than the raw share.

## The scrambled null

Agreed, and I have stopped claiming it discriminates. It is kept, because a diagnostic that manufactures evidence from a target with none left in it is worth ruling out, but the text now says that permuting `y` leaves the three channels exchangeable and that almost any share comparison would return 0.5 there. It is a check on the plumbing. The positive control is the falsification test.

## The rest

- **Cloned placebo.** Stated where the reader meets the table, not at the end: `P(> placebo)` for `x1` is the least favourable comparison in it, `x2` is measured against a placebo that was not cut from it, and a real analysis should build from another channel or one per channel.
- **How many placebos.** Now explicit that they belong in separate refits, with your arithmetic for why: five in one model take a seventh each and the floor comes out artificially low.
- **Stored outputs.** Fixed in `822e620`, before the rest. They carried my `.venv` paths onto a public page and should never have been in a pushed commit.
- **`np.abs`.** Noted, with the case that breaks it.
- **PR description.** Rewritten. It had described a module that has not existed since `c017bab`, which was my error to leave standing.
- **Gallery title.** "Placebo Channel Check", with the sentence kept as the notebook H1.
- **nutpie.** Thank you for checking, left as is.

On your last point, whether the helper should live in the package after all: I think you are right that a gallery entry makes this a supported technique with no tests behind its implementation, and I would rather that were your call than mine. The module and its six tests are in this branch's history at `16f0bc1` and I am happy to reinstate them, with the notebook kept as the narrative and shortened to a call. Say the word either way.

The notebook is four fits now, about 220s end to end here. Outputs are stored and the docs build does not re-execute, so that cost is mine rather than CI's.

daimon-pymclabs:
The baseline result is the right outcome to have found, and I think the notebook is stronger for it landing the unflattering way. `lost` matching `dilution alone` to three decimals, with the ratio pinned at 1.488 / 1.489, is a cleaner statement of what a placebo does and does not license than any wording change could have been. Keeping the scrambled null as a plumbing check while demoting its claim is also the right call.

On your question, my answer is: reinstate the module, but not as it stands at `16f0bc1`.

The case for the package is the one you already made, and this round strengthened it. The recipe is no longer "shift a column and eyeball a share." It is a two-fit procedure whose correctness depends on details a copy-paste reader will get wrong: the baseline refit, the compositional correction, separate refits rather than five placebos in one model, and the strictly-positive-beta floor that makes `P(> placebo)` the readable quantity instead of the raw share. Every one of those is a thing the notebook now has to say in prose because there is no function enforcing it. That is exactly the shape of logic that belongs behind tests.

But `16f0bc1` predates the finding. `summarize_placebo_contribution` takes a single contribution array, so it cannot express the comparison that turned out to be the point of the exercise. If it comes back, the signature should take both fits, something like `summarize_placebo_contribution(with_placebo, without_placebo, placebo_name)`, and return the `lost` / `dilution alone` pair alongside the shares. A helper that reproduces the table the notebook has now abandoned would be worse than no helper, because it would quietly steer people back to the reading you just disproved.

Two smaller things if you do reinstate:

- `az.hdi` takes `hdi_prob=`, not `prob=`. As written in `summarize_placebo_contribution`, the `hdi_prob` argument either raises or gets swallowed into `**kwargs` and silently ignored, leaving every caller on the 0.94 default. Worth confirming against the pinned arviz before the tests go back in, since a test that only checks the returned columns would not catch it.
- `add_placebo_channel(method="shift")` draws its offset from `rng.integers(1, n)`, which excludes 0 but admits offsets near 1 and near `n-1`. Those are near-copies of the real series with almost the original alignment intact, so on an unlucky seed the placebo is not really a placebo. Constraining the draw to the middle of the range, or documenting the failure mode, would make the default safer for someone who does not pass a seed.

Neither is a reason to keep it out. Happy to review the reinstated version.

ipezygj:
Reinstated at `5a146b1`, with the signature you specified. Both of your smaller points were checked rather than taken, and they came out differently from each other.

## The module

`summarize_placebo_contribution(with_placebo, without_placebo, placebo_name)` returns `lost` and `dilution_alone` beside the shares. Your reason for the new signature is the one that decided the rest of the design: a helper that can still express the old reading would steer people back to it. So the notebook imports it rather than defining its own, and the numbers are identical, shares 0.455 / 0.306 / 0.239 and `lost` 0.143 / 0.096 against `dilution_alone` 0.143 / 0.096, ratio 1.488 to 1.489. A rewrite that moved the result would have been a different function.

That principle cost something and I took the cost. The scrambled null and the positive control each call the summary too, and neither had a baseline. The cheap fix was a single-fit path returning NaN in the comparison columns; that is the door you just asked me to close. They now fit their own baselines, so the notebook is six fits rather than four.

## `az.hdi`

You were right that it was worth confirming, and the confirmation goes the other way. Against the pinned arviz:

```
arviz 1.2.0
az.hdi signature: (data, prob=None, dim=None, group='posterior', ...)
prob=     -> works
hdi_prob= -> TypeError: hdi got an unexpected keyword argument: 'hdi_prob'
```

arviz renamed it in 1.x, so `prob=` is correct here and the suggested change would break the call. I have left it as `prob=`.

Your point about the test still stands and was the useful half. A test that checks only the returned columns would pass while the argument was silently ignored, so the suite now asserts that `hdi_prob=0.99` gives a wider interval than `0.50` on every row. That fails if the keyword ever stops being honoured, whatever it is called.

## The offset draw

This one holds, and it is worse than "documenting the failure mode" would suggest. On the 159-week series of `mmm_multidimensional_example`, four of the 158 available offsets leave the placebo correlated with its source above 0.5, and the worst is 0.76. An unseeded caller draws a near-copy about 2.5 % of the time, per geo.

Rather than document it, `add_placebo_channel` now rejects a draw above `max_abs_corr`, default 0.3, and draws again, raising with a suggestion to use `method="permute"` if no draw qualifies. Under the guard the largest correlation available on that series is 0.32. The notebook already told the reader to check this number by hand every time; enforcing it is the same argument you made for having a module at all.

## Checks

41 tests. One of them is a positive control on the guard itself: with the bound switched off, some seed in 0..59 produces a placebo the default refuses. Without it the guard would pass its own test for the wrong reason, which is the failure mode this notebook is about.

`ruff check` and `ruff format` clean via `uv run`, mypy clean, `validate_api_docs` passes at 21 modules. Two notes from this environment in case they save you time: `validate_api_docs.py` prints an emoji and dies on a Windows cp1252 console unless `PYTHONIOENCODING=utf-8` is set, and the watermark hook fails there with `[WinError 5]` while the script itself passes when run directly.

daimon-pymclabs:
The two-fit signature does what it needed to, and closing the NaN path by giving the scrambled null and the positive control their own baselines was the right call even at two extra fits. Three things below: one where I was wrong, one that still holds, and one small new one.

## arviz

You are right and I was wrong. `pyproject.toml` pins `arviz>=1.2.0,<2.0`, and 1.x moved the keyword to `prob=`. My suggestion would have broken the call. Leaving it as `prob=` is correct.

The test you added is the better version of the point I was reaching for. Asserting `hdi_prob=0.99` gives a wider interval than `0.50` on every row survives the keyword being renamed again, which a signature check would not.

## The guard is pooled, but the offsets are per group

This one holds and is worth another pass. `add_placebo_channel` calls `rng.integers(1, len(values))` inside the per-group loop, so each geo draws its own offset, but the acceptance test is `placebo_correlation(out, channel_column, name)` over the whole frame. Pooling averages a near-copy geo against decorrelated ones, so one market can keep exactly the failure mode you set out to close while the pooled number sits comfortably under 0.3.

Three geos of AR(1) weekly spend, 159 weeks each, calling the function as written on this branch:

```
300 seeds; pooled guard passed but a geo exceeds 0.5: 14
  seed=34   pooled=0.207   geo B = 0.690
  seed=80   pooled=0.119   geo C = 0.688
  seed=82   pooled=0.231   geo B = 0.690
  seed=105  pooled=0.103   geo C = 0.511
```

About 4.7 % of unseeded draws leave one market with a placebo correlated 0.69 against its source, which is worse than the 0.76-at-2.5 % case the guard was written for, because it now passes a check that reads as though it ruled it out. The data is synthetic rather than `mmm_multidimensional_example`, so the exact rate is a property of my series and not yours. The mechanism is not: the statistic is measured at a different level from the one the randomness lives at. Taking the max over groups rather than the pooled correlation fixes it, and that is also what the Notes paragraph in the docstring already describes, since it reasons about one 159-week series rather than a pooled panel.

The suite has the same shape. `test_alignment_is_destroyed`, `test_the_guard_holds_for_every_seed` and `test_a_loose_bound_admits_what_the_default_rejects` all call `placebo_correlation` on the full frame, so the guard's positive control passes while the per-geo case goes unmeasured.

## Smaller: a date-indexed panel raises

`out[name] = pd.concat(parts).sort_index()` needs a unique index to align. A long panel indexed by date with one row per geo per week is a common shape for this data and has duplicate labels:

```python
dates = pd.date_range("2023-01-02", periods=60, freq="W")
X = pd.concat([pd.DataFrame({"geo": g, "x1": rng.gamma(2, 1, 60)}, index=dates)
               for g in ["A", "B"]])
add_placebo_channel(X, "x1", group_column="geo", seed=1)
# ValueError: cannot reindex on an axis with duplicate labels
```

The error does not point at anything a caller would connect to their index. Assigning positionally, or raising early with a message that names the index, would be kinder than pandas' version of it.

ipezygj:
Both fixed at `d168967`, and I have to correct something of mine first, because it is the more useful of the two things I learned this round.

## The claim I made last night does not hold

I wrote that `lost` and `dilution_alone` agree to three decimals and the ratio is pinned at 1.488 to 1.489. That was one seed on one sampler version.

Running `uv run` re-synced the environment mid-branch: pymc 6.0.1 to 6.2.0, pytensor 3.0.7 to 3.2.4, pymc_extras 0.12.2 to 0.14.0. I checked that this was not my code before concluding anything: the placebo column is byte-identical across the two module versions, same sha1, same pooled correlation 0.0818, and no code cell in the notebook differs from the previous commit. Only the sampler moved. The table became 0.137 against 0.141 for `x1`, 0.100 against 0.096 for `x2`, and the ratio 1.467 to 1.500.

Two environments is n = 2, so I measured the spread properly. Repeating both fits at three sampling seeds on the current environment:

```
lost minus dilution_alone
  x1:          -0.0061 to -0.0039
  x2:          +0.0039 to +0.0061
  ratio shift: +0.0317 to +0.0504
```

So the residual is not zero and it is consistently signed: `x1` gives up slightly **less** than proportional dilution would take, not more, and the ratio moves by 2 to 3 %. The qualitative conclusion survives, the placebo did not single `x1` out, and if anything it is the channel it takes least from. But "to three decimals" was a single-run coincidence dressed as a tolerance, in a notebook whose whole subject is numbers that do not survive their own conditions. The notebook now reports the seed spread and says the earlier version did not survive a dependency upgrade.

I would rather you heard that from me than reproduced it and found it yourself.

## The guard was measured at the wrong level

You are right, and it is worse than the rate you measured, because it scales with the group count. Offsets are drawn per group; the acceptance test was pooled; one bad group is divided by the number of groups. On AR(1) panels of 159 weeks, with a pooled bound of 0.3:

| groups | a group exceeds 0.3 | exceeds 0.5 | worst |
|---|---|---|---|
| 2 | 1.3 % | 0 % | 0.394 |
| 3 | 8.7 % | 0 % | 0.411 |
| 6 | 62.7 % | 17.3 % | 0.882 |

Your 4.7 % at three groups I could not reproduce on my series, which is the point you made about the rate being a property of the data. The mechanism is not, and the group-count dependence is a cleaner statement of it than either of our single numbers. Applied per group it is 0 of 300 at every group count tried, worst 0.298.

`placebo_correlation` now takes `group_column` and returns the largest absolute correlation over groups. You were also right that the suite had the same shape: all three guard tests measured the pooled frame, so the guard's positive control passed while the per-group case went unmeasured. They measure per group now, with a six-group regression test for exactly the case the pooled bound let through.

## The duplicated index

Reproduced exactly as you wrote it. The column is assigned positionally now.

Worth recording that my first fix was also wrong: I derived positions with `index.get_indexer_for`, which returns every match under duplicate labels, so a 60-row group produced 120 positions and it failed on a shape mismatch instead. It took running your repro to see it. The positions come from the grouper itself now.

## Checks

44 tests, `ruff check` and `ruff format` clean. The notebook's accepted draw is unchanged under the per-group bound, worst geo 0.255, so the change to the guard moves nothing in the outputs; everything that moved there moved because of the dependency upgrade.

daimon-pymclabs:
Both fixes reproduce as fixed on my side. The duplicated-index panel now returns instead of raising, and re-running the AR(1) sweep against `d168967` with the per-group guard gives 0 of 300 exceeding 0.3 at three groups (worst 0.283) and 0 of 300 at six (worst 0.300, sitting exactly on the inclusive bound). Thank you for the correction on the three-decimal claim as well; a notebook about numbers that do not survive their own conditions is the worst possible place for one that did not, and reporting the seed spread is the right repair.

Two new ones, both in the same place: `groupby` decides which rows exist, and the code downstream of it assumes they all do.

## Rows the grouper drops keep uninitialised memory

`blocks = list(X.groupby(group_column, sort=False).indices.values())` inherits `dropna=True`, so any row with a missing group key contributes no positions. `column = np.empty(len(X))` is then filled only at the positions that do appear, and the rest is whatever was in that allocation.

```python
X = pd.concat([pd.DataFrame({"geo": g, "x1": rng.gamma(2, 1, 200)}) for g in "AB"],
              ignore_index=True)
X.loc[[3, 7, 250], "geo"] = np.nan
out, name = add_placebo_channel(X, "x1", group_column="geo", seed=0)
out[name].iloc[[3, 7, 250]]
```

Across six runs those three rows came back as `[4.68e-310, nan, nan]`, then `[0.0, nan, nan]` five times. Not a fixed sentinel, and a denormal is the shape of value that reaches a sampler rather than stopping at one.

The guard cannot see it, because `placebo_correlation` groups by the same column and drops the same rows: every one of those runs reported a per-group correlation between 0.05 and 0.13 and was accepted. So a panel with one market's label missing for a few weeks, which is an ordinary consequence of a join, produces a placebo channel with junk in it that the acceptance test certifies. `np.full(len(X), np.nan)` at least fails loudly downstream, though the honest version is to count the positions the grouper returned and raise when they do not cover the frame, naming the column and the row count.

## A group whose spend never varies makes the guard order-dependent

A channel that never ran in one market is a constant column there, and its correlation with anything is NaN. `max()` over a generator resolves that by position rather than by rule: `nan > x` is False, so a NaN reached after a finite value is silently discarded, and a NaN reached first is never displaced. Same three geos, same data, only the order of the blocks changed:

```
['A', 'B', 'C'] -> accepted, guard = 0.095
['C', 'A', 'B'] -> ValueError: no draw left abs(corr) <= 0.3 after 50 tries
```

with C the inactive market. Neither branch is the answer you want. The first accepts while reporting a number computed over the two groups that were never in question, and the second spends fifty fits to raise a message about series length and periodicity at a caller whose actual problem is a column of zeros in one geo. A constant group is worth detecting on its own terms and saying so, since a placebo built from a channel with no spend in a market is not measuring anything there in the first place.

## Smaller: a one-row group

`rng.integers(1, len(values))` on a group of length 1 raises `ValueError: low >= high` from numpy, which names nothing the caller controls. A short-panel market, or a stray key from the same join as above, gets there. `len(values) < 2` could raise with the group's label in the message; a one-row group also has no shift that is not the identity, which is worth saying rather than implying.

None of the three has a test, which is consistent with the suite's shape: it builds well-formed panels and then measures the statistic hard. The frames that arrive malformed are the untested edge.

PR #2849, opened 2026-08-10
docs: add MMM campaign credibility disaggregation notebook

cetagostini:
## Summary

Cleanly extracts the MMM campaign credibility disaggregation demo from conflicted PR #2679 without merging or cherry-picking its unrelated history.

- Adds `docs/source/notebooks/mmm/dev/mmm_credibility_disaggregation.ipynb`.
- Adds its six required deterministic synthetic inputs under `docs/source/notebooks/mmm/dev/data/`, including the 394 KB channel-contribution posterior NetCDF.
- Excludes the optional generator script because the notebook does not import it.

## Correctness and robustness

The source notebook described cumulative unduplicated reach as its evidence but calculated `n_reach` from the duplicated sum of daily reach. This PR corrects that mismatch and aligns the entire artifact:

- Effective sample size uses each campaign's final expanding-window, unduplicated reach; daily reach remains the deliberate **normalized daily allocation-score** input.
- Captions and shrinkage prose now match the data: `niche_interest` has the smallest reach and lowest Z; `flash_sale_burst` and `broad_prospecting` begin below the collective rate and pool upward.
- The overlap explanation distinguishes potential targeting-universe overlap from realized delivery-window reach.
- Cumulative-reach values must be finite, non-negative, and non-decreasing; their dates must be unique, chronological, and end with the daily telemetry window. Daily campaign labels must be known and non-blank, preventing stale or silently discarded telemetry before plotting, fitting, or allocating.
- Daily allocation requires exact date-coordinate alignment, uses explicit rate-only fallback shares when a nonzero MMM day has no campaign reach, and raises if the campaign split fails to reconcile within `1e-9`.
- The notebook rejects incomplete/invalid user telemetry before fitting or allocating: every campaign/date reach and raw daily click must be finite and non-negative; terminal cumulative reach, campaign/total clicks, and derived log-rate inputs must be finite and positive. Missing terminal reach cannot inherit a stale value through forward fill, and missing clicks cannot be silently skipped by aggregation.
- Selected MMM and credibility posterior draws must be finite. Stable relative log-rate normalization prevents finite extreme draw values from overflowing while retaining the intended zero-reach fallback.
- The displayed uncertainty is accurately labeled as a seeded, 100-draw daily posterior approximation; saved outputs were regenerated from the repaired notebook.

## Provenance

The package was extracted from source PR #2679 at `89243b6dfb61397b055a7704a63842e564ce4d90`. All six generated inputs remain exact source blobs. The source simulator identifies them as deterministic synthetic demonstration data (`SEED = 42`).

## Validation

After the final repair:

- `uv run pre-commit run --all-files`
- Focused mock smoke: `uv run python scripts/run_notebooks/runner.py --notebooks docs/source/notebooks/mmm/dev/mmm_credibility_disaggregation.ipynb`
- Full saved execution and unmodified execution through `scripts.run_notebooks.runner.actual_run`
- `nbformat` validation with no saved error outputs
- Deterministic guard checks: current telemetry is complete; synthetic interior missing/negative/decreasing reach, missing terminal reach, missing daily click, truncated/reordered/duplicate cumulative-reach timelines, and invalid campaign labels reject; selected non-finite posterior inputs reject; zero-mass fallback sums to one; extreme finite log-rate draws remain finite; reordered dates are rejected by exact xarray alignment; saved reconciliation `1.82e-12 <= 1e-9`

All checks passed. The committed worktree contains no generated checkpoint artifacts.


juanitorduz (approved):
This is a great idea! 

I did a detailed review in https://github.com/pymc-labs/pymc-marketing/pull/2860 (now merged 🎉 !)

At least in this case, it makes a lot of sense. I am in favor of merging this to the dev notebooks from now (and then to the proper docs)

docs/source/notebooks/mmm/dev/mmm_credibility_disaggregation.ipynb
lancechua, line 1: Sharing my thoughts below

## Merits

- Agree the partial pooling is the right tool for disaggregation given the varying amounts of evidence.
- Reach as the evidence measure aligns with Ehrenberg-Bass thinking
- The channel contribution from the MMM is respected as a hard constraint, since this method only reallocates.

## Issues

### Lack of alignment with the main MMM's structure

The disaggregation is built from **click share** and **reach-days**, not from the **media variable** and **response curve** the MMM used to estimate `T` (the channel total). The model result is only used as a target, but the allocation technique isn't connected to the model beyond that.

Say campaigns have different objectives. Two campaigns could deliver identical impressions, which the MMM would credit similarly, but have very different CTRs (e.g. a low-CTR reach/awareness campaign vs. a high-CTR click-optimized campaign). I'd say we should try to inherit the bias of the main MMM since we're anchoring the disaggregation in its results. On a related note, this method also assumes the value of each "click", regardless of campaign source, is the same, which may be too strong. 

We also want to at least incorporate **adstock** (aligned with the main MMM estimated parameter values) so the contribution of a campaign does not abruptly end with its execution, as the main model suggests. 

### Small number of campaigns can make the hyperparameters weakly identified

If we can use time-series data, it would give the hyperparameters more data points to work with, granted it also comes with its own issues (e.g. autocorrelation for unique reach).


### The likelihood ignores that the underlying shares are compositional

Each campaign's naive rate `y_c` is modeled as conditionally independent given `μ` and `σ_s`. But the click shares that `y_c` is built from sum to 1 across campaigns by construction — an error on one campaign's share is mechanically absorbed by the others. That induced negative dependence across campaigns isn't represented anywhere in the likelihood.

I have to admit it was Claude that spotted this. We might want to use a Dirichlet distribution, though I personally don't have too much experience with it.

lancechua (commented):
Added some thoughts on the notebook. It's an interesting idea.

In the end, any disaggregation analysis for campaign (or some other dimension like platform / objective) would have to rely on partial pooling and some assumptions.

docs/source/notebooks/mmm/dev/mmm_credibility_disaggregation.ipynb
cetagostini: > On a related note, this method also assumes the value of each "click"

Click here is a proxy metric, but maybe worth mention in the notebook. The curves, the matrix and so on are metric proxies but if you have other metrics which you believe could represent better the process as noisy realization then good as well to change, it's not something attach to clicks.

If clicks are a biased signal then implementer should evaluate. The notebook says this ("it cures noise, not bias").

> We also want to at least incorporate adstock (aligned with the main MMM estimated parameter values)

Need to think here, but hunch says should be encoded and at least my initial example, show it and I remember doing some comparison, and something like this end-up being just not possible to learn even if you add to your model. In a way, this is the most simple way.

> granted it also comes with its own issues (e.g. autocorrelation for unique reach)

Yes, reason of why going towards the current direction.

juanitorduz:
> Added some thoughts on the notebook. It's an interesting idea.
> 
> In the end, any disaggregation analysis for campaign (or some other dimension like platform / objective) would have to rely on partial pooling and some assumptions.

Thanks @lancechua and @carlosagostini for the interesting discussion. I added these points into a small PR: https://github.com/pymc-labs/pymc-marketing/pull/2863 It would be great to have your thoughts on it :)

docs/source/notebooks/mmm/dev/mmm_credibility_disaggregation.ipynb
lancechua: Thanks for the response.

If I understand correctly, the main ideas of this technique are:
* Use `metric_a` (e.g. clicks) to evaluate campaign impact
* Use `metric_b` (e.g. total unique reach) to judge credibility

On `metric_a`: that makes sense, and I agree it's not inherently tied to clicks. Given that, I'd lean toward aligning `metric_a` with whatever metric the MMM itself was fit on, since we're decomposing contributions that were calculated using that metric. If clicks (or any other proxy) diverges from it, that's exactly the kind of bias the notebook's caveat already flags, so using the same metric as the MMM seems like the more defensible default.

An alternative worth considering is normalizing based on each campaign's declared objective (i.e. evaluating campaigns against their own stated goals rather than one common metric), though with only a handful of campaigns that might be hard to do reliably.

On `metric_b`: unique reach will probably be the right choice for most cases. Just as an aside, budget could in principle serve a similar purpose (higher budget → more credibility).

On adstock: Juan's PR already incorporated this, and I think the revised version arrives at a fairer decomposition. Agreed that the adstock/saturation parameters aren't learnable within this model, so it makes sense to take them directly from the MMM instead. Incorporating decay should be relatively straightforward; saturation seems trickier, since a curve fit at the channel level may not translate cleanly down to an individual campaign's exposure scale.

And agreed on the autocorrelation point. We could consider first-differencing (i.e. newly reached people) as an alternative, but the added complexity likely isn't worth it given the amount of data available.

juanitorduz:
I will merge this one as it is a dev notebook. I a next PR I will promote to the docs :)

PR #2850, opened 2026-08-10
Change default Fourier seasonality prior from Laplace to Normal

matekadlicsko:
## Description

This work is based on [observations in my blog post](https://matekadlicsko.github.io/posts/trigonometric-seasonality-priors/), namely

1. seasonality priors should be invariant under certain rotations (this is to ensure that the prior curve remains invariant under phase shifts),
2. well selected priors can make sure the curve ends up being smooth (regularize Sobolev norm).

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2850.org.readthedocs.build/en/2850/

### DIFF
--- a/pymc_marketing/mmm/fourier.py
+++ b/pymc_marketing/mmm/fourier.py
@@ -119,9 +119,9 @@ Even make it hierarchical...
 
     # "fourier" is the default prefix!
     prior = Prior(
-        "Laplace",
+        "Normal",
         mu=Prior("Normal", dims="fourier"),
-        b=Prior("HalfNormal", sigma=0.1, dims="fourier"),
+        sigma=Prior("HalfNormal", sigma=0.1, dims="fourier"),
         dims=("fourier", "hierarchy"),
     )
     yearly = YearlyFourier(n_order=3, prior=prior)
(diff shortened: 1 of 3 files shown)


matekadlicsko:
The Sobolev-style prior scale decay remains to be added. Perhaps the most flexible way to do it is to scale the feature matrix, since that would support some other rotation invariant distributions other than just Normal. I'm not super familiar with the code base though, so I would appreciate some feedback!

juanitorduz:
> The Sobolev-style prior scale decay remains to be added. Perhaps the most flexible way to do it is to scale the feature matrix, since that would support some other rotation invariant distributions other than just Normal. I'm not super familiar with the code base though, so I would appreciate some feedback!

Thanks @matekadlicsko ! For now, we could work on this one and leave the Sobolev prior for a subsequent PR?

williambdean:
Very interesting. Can we do the Sobolev prior first then change the defaults if the goal is to chagne fourier prior defaults to that.

juanitorduz:
> Very interesting. Can we do the Sobolev prior first then change the defaults if the goal is to chagne fourier prior defaults to that.

yes, that actually makes more sense :)

williambdean:
Well I guess the question is whether we will have this outside of the Prior class. 

I'm fine with this change. Maybe we have a docstring example instead. For example, 

```python
import numpy as np

from pymc_marketing.mmm import YearlyFourier

n_order = 4

harmonics = np.arange(1, n_order + 1, dtype=float)
sobolev_scales = 1.0 / (1.0 + harmonics**2)

before = Prior("Normal", dims="fourier")
after = Prior("Normal", sigma=np.concatenate([sobolev_scales, sobolev_scales]),
dims="fourier")

fourier = YearlyFourier(n_order=n_order, prior=...)
```

I thinking adding this to (module) docstring and referencing the blog post would be good
enough.

Or adding `pymc_marketing.fourier.create_sobolev_scales` function as a helper and using.

```python
def create_sobolev_scales(sigma, n_order: int): 
    harmonics = np.arange(1, n_order + 1, dtype=float)
    sobolev_scales = sigma / (1 + harmonics ** 2)

    return np.concatenate([sobolev_scales, sobolev_scales])
```

Thoughts? @juanitorduz

PR #2853, opened 2026-08-10
refactor(budget-optimizer): allocate_budget returns BudgetOptimizationResult

juanitorduz:
## Description

Re-land of #2801 onto `main`.

#2801 was approved and merged, but its base was the intermediate branch `refactor/budget-optimizer-pm-model` (the stacking base from #2654), so the changes landed as `8fb1753` on that branch and never reached `main`. This PR is a straight cherry-pick of `8fb1753` onto `main` — no code changes on top of what was already reviewed and approved.

Original work and authorship: @PabloRoque.

### What the change does

`allocate_budget` now returns a `BudgetOptimizationResult` instead of a variable-arity tuple (2-tuple, or 3-tuple when `callback=True`):

```python
@dataclass
class BudgetOptimizationResult:
    budgets: DataArray                     # optimized allocation, monetary units
    scipy_result: OptimizeResult           # raw scipy diagnostics
    optimized_vars: dict[str, DataArray]   # {} until #2621
    callback_info: list[OptimizationIterationInfo] | None  # was the 3rd tuple element
```

**Backward compatible:** iterating the result yields `(budgets, scipy_result)`, so every existing `optimal, res = optimizer.allocate_budget(...)` call site works unchanged. `callback=True` no longer changes the return arity — the per-iteration diagnostics land in `result.callback_info`, now typed via `TypedDict`s (`OptimizationIterationInfo`, `ConstraintIterationInfo`).

Also:

* **Remove the `mu_effects` field.** Its only behaviour was raising `NotImplementedError` when an effect had `replace_for_optimization`, and #2621 fills that gap with a different mechanism (`optimizable_vars`) under which the field is dead state. A `DeprecationWarning` is emitted (and the argument ignored) if a caller still passes it. `MMM.budget_optimizer` no longer forwards it.
* **Collapse `_compiled_functions`** — a dict keyed by the fixed `utility_function` field that could only ever hold one entry — into a plain `_objective_and_grad` attribute.

### Motivation

`optimized_vars` gives #2621's co-optimized lever values a typed home instead of attaching attributes onto scipy's `OptimizeResult`, and the fixed return arity removes the `callback=` wart before a third consumer appears. Deleting `mu_effects` now means the placeholder doesn't outlive the design it was placed for.

## Related Issue

* Re-lands #2801 (approved, merged into `refactor/budget-optimizer-pm-model` rather than `main`)
* Follows #2654, precedes #2621

## Checklist

* [x] Cherry-pick applied cleanly, five files, identical to the approved diff in #2801
* [x] `pre-commit` clean (ruff, ruff format, mypy)
* [x] `pytest tests/mmm/test_budget_optimizer.py tests/mmm/test_budget_optimizer_mmm.py tests/mmm/test_cost_per_unit.py` — 107 passed

## Modules affected

* `mmm.budget_optimizer` (public API: return type of `allocate_budget`)
* `mmm.mmm` (`optimize_budget` return type; stops passing `mu_effects`)

## Type of change

* Refactor / breaking-ish change with a compatibility shim (`__iter__`)

----
📚 Documentation preview 📚: https://pymc-marketing--2853.org.readthedocs.build/en/2853/


juanitorduz:
@PabloRoque , because of the change in result type, I will work on fixing the notebooks.

cetagostini (approved):

PR #2859, opened 2026-08-11
Expand mmm.summary for frontend-ready MMM data export (closes #2715)

### LINKED ISSUE #2715: Add return_data parameter to MMM plot methods for JSON-serializable data export
## Problem

In production MMM systems, plotting is often done by a frontend application (React, Dash, Streamlit, custom dashboard) using JavaScript chart libraries — not by the Python backend rendering static matplotlib images. Today, pymc-marketing's plot methods only return rendered figure objects:

- `(matplotlib Figure, NDArray[Axes])` by default
- `PlotCollection` via `return_as_pc=True`

There is no public API to get the underlying plot data (channel contributions, HDI bands, ROAS distributions, saturation curves, etc.) as structured data suitable for JSON serialization and consumption by an arbitrary frontend chart library.

## Why this matters

The plot methods already *compute* the data they render — it's just not exposed. For example, `time_series_contributions()` computes per-channel mean contributions + HDI bands over time, then feeds them into `PlotCollection.wrap()`. That intermediate xarray Dataset is exactly what a frontend needs to render the same chart with Recharts, D3, Chart.js, or any other library.

Without this, the production integrator has two bad options:

1. Duplicate the data-computation logic outside pymc-marketing (brittle, drifts from the library)
2. Render matplotlib images server-side (non-interactive, can't be restyled, wastes compute)

## Proposed API

Add a `return_data: bool = False` parameter to MMM plot methods. When `True`, the method skips rendering and returns the underlying xarray Dataset directly:

```python
ds = mmm.plot.decomposition.time_series_contributions(return_data=True)
# => xr.Dataset with dims (channel, date) and vars: mean, hdi_3, hdi_97, observed

# Ship to frontend:
ds.to_dataframe().reset_index().to_dict(orient="records")
```

**Scope**: all MMM plot methods that currently support `return_as_pc`:

| Namespace | Methods |
|---|---|
| `mmm.plot.decomposition` | `time_series_contributions`, `waterfall` |
| `mmm.plot.diagnostics` | `posterior_predictive`, `prior_predictive` |
| `mmm.plot.transformations` | `spend_vs_contribution`, `saturation_curves` |
| `mmm.plot.budget` | `allocation_roas`, `contribution_over_time` |
| `mmm.plot.cv` | `posterior_predictive`, `param_stability`, `crps` |
| `mmm.plot.sensitivity` | `sweep`, `uplift`, `marginal` |

## Non-goals

- **NOT** asking for Plotly/Vega/any-specific-library JSON spec export. Raw data is backend-agnostic and lets the frontend own presentation.
- **NOT** asking for a new data-computation layer. The data is already computed internally; this just adds a public return path before the rendering step.

## Implementation notes

The data is already prepared before `PlotCollection.wrap()`/`.grid()` is called in every method. No new computation, no breaking changes, backward-compatible (defaults to `False`).

drbenvincent:
## Summary

Implements the reframed approach for #2715: expand the summary layer so frontends can consume pre-aggregated, JSON-serializable DataFrames instead of adding `return_data=True` to plot methods.

- Add **11 new `mmm.summary` methods**: `waterfall`, `channel_share_hdi`, `prior_predictive`, `residuals_over_time`, `residuals_distribution`, `prior_vs_posterior`, `saturation_scatterplot`, and sensitivity trio (`sensitivity_analysis`, `sensitivity_uplift`, `sensitivity_marginal`)
- Add **`BudgetSummaryFactory`** (`allocation_roas`, `contribution_over_time`) for stateless budget-sample summaries
- Add **`cv.summary`** (`MMMCVSummaryFactory`) with `predictions`, `param_stability`, `crps`
- Organize summary code in **`mmm/summary/`** package (mirrors `plotting/`): `factory.py`, `helpers.py`, `budget.py`, `cv.py`, `sensitivity.py`; `__init__.py` re-exports the public API
- Extract shared xarray utilities into **`xarray_utils.py`**; refactor matplotlib plot methods to use summary as single source of truth
- Extend **`plot_interactive`** with `waterfall()` and `channel_share()` backed by summary DataFrames

Target consumer API:

```python
records = mmm.summary.contributions().to_dict(orient="records")
```

Closes #2715.

## Package layout

```
pymc_marketing/mmm/summary/
  __init__.py       # MMMSummaryFactory, BudgetSummaryFactory, MMMCVSummaryFactory, helpers
  factory.py        # MMMSummaryFactory
  helpers.py        # shared computation + dataframe_to_json_records
  budget.py         # BudgetSummaryFactory
  cv.py             # MMMCVSummaryFactory
  sensitivity.py    # sensitivity_* summaries
```

Public imports are unchanged, e.g. `from pymc_marketing.mmm.summary import MMMSummaryFactory`.

### New `mmm.summary` methods
- [x] `waterfall()`
- [x] `channel_share_hdi()`
- [x] `prior_predictive()`
- [x] `residuals_over_time()`
- [x] `residuals_distribution()`
- [x] `prior_vs_posterior()`
- [x] `saturation_scatterplot()`
- [x] `sensitivity_analysis()`
- [x] `sensitivity_uplift()`
- [x] `sensitivity_marginal()`

### Budget / CV
- [x] `BudgetSummaryFactory.allocation_roas()`
- [x] `BudgetSummaryFactory.contribution_over_time()`
- [x] `cv.summary.predictions()`
- [x] `cv.summary.param_stability()`
- [x] `cv.summary.crps()`

### Plot refactors (summary as data source)
- [x] `plot.decomposition` — contributions, waterfall, channel_share_hdi
- [x] `plot.diagnostics` — shared helpers from `summary.helpers`
- [x] `plot.budget` — uses `summary.budget` prepare helpers
- [x] `plot.cv` — shared helpers from `summary.cv`
- [x] `plot.transformation` — `get_channel_x_data` from `summary.helpers`

## Test plan

- [x] `uv run python -m pytest tests/mmm/test_summary.py tests/mmm/test_summary_budget.py tests/mmm/test_summary_cv.py -q --no-cov`
- [x] `uv run python -m pytest tests/mmm/plotting/ -q --no-cov`
- [x] `uv run python -m pytest tests/mmm/test_plot_interactive.py -q --no-cov`
- [ ] CI green on PR

----
📚 Documentation preview 📚: https://pymc-marketing--2859.org.readthedocs.build/en/2859/


pymc_marketing/mmm/summary_helpers.py
```diff
@@ -0,0 +1,346 @@
...
+import pandas as pd
+import xarray as xr
+
+from pymc_marketing.data.idata import MMMIDataWrapper
+from pymc_marketing.data.idata.utils import (
+    get_posterior_predictive,
+    get_prior,
+    get_prior_predictive,
+)
+from pymc_marketing.mmm.plotting._helpers import (
```
drbenvincent: **Layering / circular-import risk.** Summary helpers now depend on `plotting._helpers`, while `plotting/decomposition.py`, `plotting/cv.py`, etc. import back into summary modules. The stated goal is a frontend-facing data layer below plotting — this coupling goes the wrong direction.

Consider moving `_select_dims`, `_apply_aggregation`, `_ensure_chain_draw_dims` to a neutral module (e.g. `mmm/_dim_utils.py` or alongside `idata` utils) that both summary and plotting import, breaking the cycle.

pymc_marketing/mmm/summary.py
```diff
@@ -1050,3 +1066,477 @@ def change_over_time(
...
+            - abs_error_{prob}_lower/upper: HDI bounds for each prob
+            - <custom_dims>: One column per custom dimension when present
+        """
+        effective_hdi_probs = hdi_probs if hdi_probs is not None else self.hdi_probs
+        effective_output_format = (
+            output_format if output_format is not None else self.output_format
+        )
+        self._validate_hdi_probs(effective_hdi_probs)
+
+        channel_contributions = self.data.get_channel_contributions(original_scale=True)
```
drbenvincent: **API inconsistency.** `channel_share_hdi` always calls `get_channel_contributions(original_scale=True)` with no `original_scale` parameter, unlike `waterfall`, `saturation_scatterplot`, and the matplotlib decomposition plots' general pattern.

If a frontend fits in scaled space and exports summaries, channel shares will silently disagree with other scaled summaries. Worth adding `original_scale: bool = True` for parity.

pymc_marketing/mmm/summary.py
```diff
@@ -1050,3 +1066,477 @@ def change_over_time(
...
+        df = self._compute_summary_stats_with_hdi(components, effective_hdi_probs)
+
+        group_cols = [
+            c for c in df.columns if c in self.data.custom_dims or c == "component"
+        ]
+        if group_cols:
+            totals = df.groupby(group_cols, dropna=False)["mean"].transform("sum")
+        else:
+            totals = df["mean"].sum()
+        df["pct_of_total"] = df["mean"] / totals * 100
```
drbenvincent: **`pct_of_total` ignores uncertainty.** This divides each row's posterior *mean* by the sum of means. The returned HDI columns are for component totals, not shares.

Dashboards may treat `pct_of_total` as a point estimate alongside `abs_error_*` and assume they're compatible — they're not. Either document loudly, compute share HDIs (ratio of totals per draw), or drop the column from this method.

pymc_marketing/mmm/summary.py
```diff
@@ -1050,3 +1066,477 @@ def change_over_time(
...
+                posterior_samples = posterior_da.sel(sel).values.flatten()
+
+                if prior_samples.size < 2 or posterior_samples.size < 2:
+                    continue
+
+                all_samples = np.concatenate([prior_samples, posterior_samples])
+                x_grid = np.linspace(
+                    float(all_samples.min()), float(all_samples.max()), num_points
+                )
+                prior_density = gaussian_kde(prior_samples)(x_grid)
```
drbenvincent: **`gaussian_kde` is fragile here.** On constant or near-constant samples (fixed priors, rounded deterministics, very tight posteriors), SciPy raises `LinAlgError` — verified with `np.ones(100)`.

This method is marketed for stable JSON export; it should catch degenerate cases (skip facet with warning, histogram fallback, or return raw quantiles). Also consider bounding `num_points` or downsampling when `var_names=None` on models with dozens of parameters — default 200×vars×facets can explode payload size.

pymc_marketing/mmm/summary/factory.py
```diff
@@ -1050,3 +1066,477 @@ def change_over_time(
...
+                )
+            else:
+                combos = [()]
+
+            for combo in combos:
+                sel = dict(zip(facet_dims, combo, strict=True))
+                prior_samples = prior_da.sel(sel).values.flatten()
+                posterior_samples = posterior_da.sel(sel).values.flatten()
+
+                if prior_samples.size < 2 or posterior_samples.size < 2:
```
drbenvincent, line 1459: **Silent data loss.** Facets with `<2` samples are skipped with no warning. Combined with `var_names=None`, callers can get an empty DataFrame (or partial results) without any signal that filtering removed everything. At minimum warn; ideally raise when the requested scope yields zero rows.

pymc_marketing/mmm/summary/cv.py
```diff
@@ -0,0 +1,422 @@
...
+from pymc_marketing.mmm.plotting._helpers import _select_dims
+from pymc_marketing.mmm.summary_helpers import (
+    DataFrameType,
+    OutputFormat,
+    StatsHelper,
+    compute_summary_stats_with_hdi,
+    convert_output,
+)
+
+__all__ = [
```
drbenvincent, line 48: **Public export of private helpers.** `_build_predictions_arrays`, `_crps_for_split`, etc. are underscore-prefixed but listed in `__all__` and imported by `plotting/cv.py`.

That makes them de-facto public API. If the intent is shared computation, either drop the leading underscore or move these to a dedicated internal module both sides import.

pymc_marketing/mmm/summary/factory.py
```diff
@@ -1050,3 +1066,477 @@ def change_over_time(
...
+        mean_contrib = contributions.mean(dim=["chain", "draw"])
+        x_data = get_channel_x_data(self.data, apply_cost_per_unit)
+
+        scatter_ds = xr.Dataset({"x": x_data, "y": mean_contrib})
+        scatter_ds = _select_dims(scatter_ds, dims)
+        df = scatter_ds.to_dataframe().reset_index()
+
+        return self._convert_output(df, effective_output_format)
+
+    def sensitivity_analysis(
```
drbenvincent, line 1556: **Test gap.** The PR checklist marks `sensitivity_analysis`, `sensitivity_uplift`, and `sensitivity_marginal` as done, but `tests/mmm/test_summary.py` has schema tests for waterfall/residuals/etc. and no coverage for these three.

Given the `prepare_sensitivity_data` / `sweep_x` merge logic (especially `x_sweep_axis="absolute"`), even a single fixture-based schema test per method would catch regressions.

pymc_marketing/mmm/plot_interactive.py
```diff
@@ -1673,3 +1673,60 @@ def _plot_curves(
...
+        nw_df, plotly_kwargs = self._prepare_summaries_for_bar_plot(
+            df=df,
+            required_column="channel",
+            default_title="Channel Share of Contribution",
+            auto_facet=auto_facet,
+            single_dim_facet=single_dim_facet,
+            plotly_kwargs=plotly_kwargs,
+        )
+
+        return self._plot_bar(
```
drbenvincent, line 1731: **Missing tests + implicit x-axis.** `waterfall` passes `x="component"` explicitly; `channel_share` relies on `_plot_bar`'s default `x="channel"`. Works today, but it's asymmetric and untested — `tests/mmm/test_plot_interactive.py` doesn't cover either new method.

Suggest explicit `x="channel"` and at least a smoke test that the figure builds from summary data.

drbenvincent (commented):
## Adversarial review — Composer 2.5 (agent)

Solid direction: summary as the single source of truth for plot data, with a clear frontend export story. Tests pass for the happy path and the refactor removes a lot of duplication.

That said, a few issues look like they could bite real consumers or regress maintainability. Inline comments call out the highest-risk ones; summary below.

### Blockers / high risk
1. **Layering inversion** — `summary_helpers` (and thus `summary`) imports `plotting._helpers`, while plotting now imports summary modules. This creates a circular dependency surface and ties the data layer to matplotlib plotting internals.
2. **`prior_vs_posterior` KDE robustness** — `gaussian_kde` raises `LinAlgError` on degenerate/near-constant samples (common for fixed priors or discrete-like posteriors). No guard; will crash the advertised JSON export path.
3. **Missing test coverage** for the three `sensitivity_*` summary methods and the new `plot_interactive.waterfall` / `channel_share` paths.

### Medium risk
4. **`channel_share_hdi` hardcodes `original_scale=True`** while sibling methods expose `original_scale`; inconsistent with `saturation_scatterplot` and limits scaled-model dashboards.
5. **`waterfall().pct_of_total` is mean-only** — HDI columns describe component totals, but share % ignores posterior uncertainty (can mislead frontend consumers).
6. **Private CV helpers exported in `summary_cv.__all__`** and imported by `plotting/cv.py` — smells like a temporary seam; consider a non-public shared module.
7. **No test that `to_dict(orient="records")` is JSON-serializable** — the PR's stated consumer API; dates/numpy scalars may need explicit conversion.

### Nits
8. New `sensitivity_*` summary methods have stub docstrings vs. the rest of the factory.
9. `plot_interactive.channel_share` omits explicit `x="channel"` (works today via `_plot_bar` default, but asymmetric with `waterfall`).

Happy to re-review after the layering/KDE/test gaps are addressed.

pymc_marketing/mmm/plotting/decomposition.py
```diff
@@ -307,44 +313,19 @@ def waterfall(
             if idata is not None
             else self._data
         )
+        summary = MMMSummaryFactory(data, validate_data=False)
```
drbenvincent: **Heavyweight indirection.** Instantiating `MMMSummaryFactory(data, validate_data=False)` in plot methods just to reach `summary.data` / shared helpers adds an odd dependency from plotting → summary factory → data.

`compute_waterfall_components(data, ...)` (or pass `data` directly) would be thinner and avoids implying plots are summary consumers of the full factory.

pymc_marketing/mmm/mmm.py
```diff
@@ -1617,6 +1617,15 @@ def summary(self) -> Any:  # type: ignore[no-any-return]
             # Get change over time
             df = mmm.summary.change_over_time()
 
+            # Frontend-ready export (library-agnostic JSON)
+            records = mmm.summary.contributions().to_dict(orient="records")
```
drbenvincent, line 1627: **Consumer API not validated.** The docstring shows `mmm.summary.contributions().to_dict(orient="records")` as the target frontend pattern, but no test asserts the result is JSON-serializable (`json.dumps` will fail on `Timestamp`, `numpy.float64`, etc. depending on pandas version).

If JSON export is a first-class goal, a small regression test (or a `summary.to_records()` helper that normalizes types) would lock the contract.

pymc_marketing/mmm/summary_helpers.py
drbenvincent: Fixed in 87287525: moved `_select_dims`, `_apply_aggregation`, `_ensure_chain_draw_dims`, etc. into a new neutral `pymc_marketing/mmm/xarray_utils.py`. `summary_helpers`, `summary`, `summary_budget`, and `summary_cv` now import from there; `plotting/_helpers.py` re-exports for backward compatibility.

pymc_marketing/mmm/summary.py
drbenvincent: Fixed in 87287525: added `original_scale: bool = True` to `channel_share_hdi()` and pass it through to `get_channel_contributions(original_scale=...)`, matching the waterfall API.

pymc_marketing/mmm/summary.py
drbenvincent: Fixed in 87287525: `pct_of_total` is now `share_mean` from per-draw share ratios (`components / total_per_draw * 100`) with full `share_*` HDI columns. Docstring updated to distinguish component-total vs share uncertainty.

pymc_marketing/mmm/summary.py
drbenvincent: Fixed in 87287525: `prior_vs_posterior()` now filters non-finite samples, catches `LinAlgError` from `gaussian_kde`, emits `UserWarning` per skipped facet, and raises `ValueError` if no rows remain.

pymc_marketing/mmm/summary/factory.py
drbenvincent: Fixed in 87287525: skipped facets now log explicit `UserWarning`s (insufficient or degenerate samples). Also warns when summarizing >20 variables without `var_names` to flag large payloads.

pymc_marketing/mmm/summary/cv.py
drbenvincent: Fixed in 87287525: trimmed `summary_cv.__all__` to only `MMMCVSummaryFactory`. Private helpers remain module-internal (still used by `plotting/cv.py`).

pymc_marketing/mmm/summary/factory.py
drbenvincent: Fixed in 87287525: added parametrized schema tests for `sensitivity_analysis`, `sensitivity_uplift`, and `sensitivity_marginal` in `tests/mmm/test_summary.py` (fixture with `sensitivity_analysis` group). Expanded delegate docstrings on `MMMSummaryFactory`.

pymc_marketing/mmm/plot_interactive.py
drbenvincent: Fixed in 87287525: added `TestWaterfallAndChannelShare` smoke tests in `tests/mmm/test_plot_interactive.py`. `channel_share()` now passes `x=plotly_kwargs.pop("x", "channel")` to `_plot_bar` so the channel axis is explicit without conflicting with prepared kwargs (same pattern as waterfall).

pymc_marketing/mmm/plotting/decomposition.py
drbenvincent: Fixed in 87287525: removed `MMMSummaryFactory` wrapper from decomposition plots. `contributions_over_time`, `waterfall`, and `channel_share_hdi` now call `compute_*` helpers and `data` methods directly.

pymc_marketing/mmm/mmm.py
drbenvincent: Fixed in 87287525: added `dataframe_to_json_records()` in `summary_helpers.py` and `test_summary_json_serializable` asserting `json.dumps()` succeeds with ISO date strings.

drbenvincent:
## Review fixes (87287525)

Addressed all inline review threads. Summary:

1. **Layering** — New `xarray_utils.py` module; summary code no longer imports from `plotting._helpers`.
2. **`channel_share_hdi`** — Added `original_scale` parameter.
3. **`waterfall` shares** — `pct_of_total` / `share_*` columns now computed from per-draw ratios with HDIs.
4. **`prior_vs_posterior`** — KDE failures handled with warnings; raises if all facets skipped; payload-size warning for large `var_names`.
5. **`summary_cv.__all__`** — Public API trimmed to `MMMCVSummaryFactory` only.
6. **Tests** — Sensitivity trio schema tests, JSON serializability test, `plot_interactive` waterfall/channel_share smoke tests.
7. **Decomposition plots** — Removed `MMMSummaryFactory` indirection; use `data` + helpers directly.
8. **`plot_interactive.channel_share`** — Explicit `x` axis via `plotly_kwargs.pop("x", "channel")` (avoids duplicate-kwarg conflict with `_prepare_summaries_for_bar_plot`).

329 related tests passing locally.

juanitorduz:
@drbenvincent thanks for workon on this! 

One initial thought: all these new `summary_*.py` files seem repetitive. Can we simply create a sub-module `summary/` and then put `summary/cv.py` inside? and then `summary/__init__.py` takes the main functions for backward compatibility?

drbenvincent:
@juanitorduz Good call — agreed this mirrors `plotting/` better than the flat `summary_*.py` layout.

**What changed (aa81a492):** consolidated into `pymc_marketing/mmm/summary/`:

- `factory.py` — `MMMSummaryFactory` (was `summary.py`)
- `helpers.py` — shared computation + `dataframe_to_json_records`
- `budget.py` — `BudgetSummaryFactory`
- `cv.py` — `MMMCVSummaryFactory`
- `sensitivity.py` — sensitivity summaries
- `__init__.py` — re-exports the public API

**Backward compatibility:** `from pymc_marketing.mmm.summary import MMMSummaryFactory` (and `mmm.summary` property) are unchanged. Internal imports now use `summary.helpers`, `summary.budget`, etc.

Stacked docs PR #2861 was rebased onto this and updated accordingly.

daimon-pymclabs (commented):
Reviewed the full diff (23 files, +3269/-594) against `main`.

**What it does:** adds 11 new `mmm.summary` methods plus `BudgetSummaryFactory` and `cv.summary`, reorganises `summary.py` into a `summary/` package, extracts `xarray_utils.py`, and refactors the matplotlib plot methods to consume the summary layer. Public imports are preserved (`summary/__init__.py` re-exports the old names, `plotting/_helpers.py` keeps a re-export shim for the moved `_select_dims` family). The overall shape is right: one computation, two consumers (tables and plots). Notes below, roughly in priority order.

## Blocking

### 1. The matplotlib waterfall silently changes from mean-over-date to sum-over-date

`plotting/decomposition.py:315-318` now calls `compute_waterfall_components(...)` then `.mean(dim=["chain", "draw"])`. But `helpers.py:149-150` sums over `date`, whereas the code it replaced computed `mean_dims = [d for d in ("chain", "draw", "date") if d in da.dims]` and took a mean. Every bar in the existing plot changes magnitude by a factor of `n_dates`.

That may well be the intended semantics for a waterfall (totals, not per-period averages), but it isn't in the PR description and nothing catches it: the waterfall tests in `tests/mmm/plotting/test_decomposition.py:328-385` assert figure/axes types and ytick labels only, never a bar value.

- [ ] Decide whether the waterfall should sum or average over `date`, and say so in the PR description if it is a deliberate behaviour change.
- [ ] Add a numeric regression test on `waterfall` that would fail if the reduction over `date` changed.
- [ ] Restore the dim filtering at `decomposition.py:318` — it hard-codes `dim=["chain", "draw"]` where the old code filtered to dims actually present, so a component without chain/draw now raises instead of passing through.

### 2. Two live copies of the HDI computation

`MMMSummaryFactory._compute_summary_stats_with_hdi` (`factory.py:208-308`), `_validate_hdi_probs` (147-165) and `_convert_output` (167-206) are byte-for-byte duplicates of `StatsHelper` in `helpers.py:246-330`. The factory never delegates. That undercuts the PR's own "summary as single source of truth" framing, and the two copies will drift.

- [ ] Delete the three duplicated factory methods and have the factory call `StatsHelper`.
- [ ] Collapse the four access paths to `StatsHelper` (the class, the module singleton `_stats_helper`, the module-level wrapper functions, and the per-module `_stats = StatsHelper()` in `budget.py`/`cv.py`/`sensitivity.py`) down to one. It holds no state, so plain module functions would do.

## Worth fixing before merge

### 3. Unit inconsistency across the new API

`waterfall` multiplies shares by 100 (`factory.py:1122`) so `share_mean` is a percentage, while `channel_share_hdi` returns fractions via `compute_channel_shares`. A frontend consuming both gets 0-100 from one and 0-1 from the other, and the `waterfall` docstring says "share point estimates" without saying percent.

- [ ] Pick one convention across all share-returning methods.
- [ ] State the unit explicitly in each docstring.

### 4. `pct_of_total` duplicates `share_mean`

`factory.py:1138` sets them equal, and `test_summary.py:632` asserts the identity.

- [ ] Either drop `pct_of_total` or give it a distinct definition.

### 5. `summary/cv.py` depends on matplotlib

Line 23 imports `matplotlib.dates` solely so `_build_predictions_arrays` can produce `train_end_da` (line 118) as a matplotlib epoch float, which `predictions()` immediately discards on line 245. Only `plotting/cv.py:170` consumes it. For a layer whose selling point is being frontend-ready and serialisable, emitting matplotlib-specific date numbers is backwards.

- [ ] Return a raw timestamp from `_build_predictions_arrays` and call `date2num` in `plotting/cv.py`.
- [ ] Drop the `matplotlib.dates` import from `summary/cv.py`.
- [ ] Give `plotting/cv.py:36` a public entry point instead of reaching into `summary.cv` for the private `_build_predictions_arrays`.

### 6. CRPS is O(rows) xarray `.sel()` calls

`_pred_matrix_for_rows` (`cv.py:147-155`) loops `rows_df.iterrows()` with one `.sel()` per row, and `crps()` calls it twice per fold per extra-dim combo. On daily data with many folds and geos this will be noticeably slow.

- [ ] Vectorise to a single `.sel(date=X_filtered["date"].values)`.
- [ ] Replace the `arr[:n_samples]` truncation on line 155 with an explicit shape check — right now it silently papers over mismatches.

### 7. `_crps_for_split` catches bare `Exception` → NaN + warning

`cv.py:189`. Reasonable for a genuinely flaky per-fold metric, but it will swallow real bugs in the new code paths for the life of the feature.

- [ ] Narrow the except clause to the errors actually expected, or include the traceback in the warning.

### 8. `crps(dims=...)` silently ignores unknown keys

`cv.py:380-383` only consults `dims` for names already in `extra_dims`, so a typo is a no-op. Everywhere else `_select_dims` raises.

- [ ] Route `crps`'s `dims` through the same validation as the rest of the API.

### 9. `prepare_allocation_roas` validates `allocation` and never uses it

`budget.py:54-55`. ROAS is computed as `channel_contribution.sum("date") / total_allocation` (63-66), i.e. per-channel contribution over *total* budget rather than that channel's own spend. This is a faithful extraction of what `plotting/budget.py` does on `main`, so it is pre-existing rather than introduced here. But the PR promotes it from a plot internal to a named `roas` column that frontends will chart.

- [ ] Confirm with the forest plot's author that the denominator is deliberate, then either fix it or document it in the docstring before it hardens into public API.

## Smaller notes

- [ ] `xarray_utils.py` is a public module whose `__all__` contains only underscore-prefixed names — either make the names public or rename the module to `_xarray_utils.py`.
- [ ] HDI column naming `abs_error_{prob}_lower/upper` is pre-existing, but this PR spreads it to ~13 more methods. If it is ever going to be renamed, now is the cheap moment. Separately, `str(int(hdi_prob * 100))` means `hdi_probs=[0.94, 0.945]` silently collide onto one pair of columns — validate against collisions.
- [ ] `saturation_scatterplot` (`factory.py:1532`) builds `xr.Dataset({"x": x_data, "y": mean_contrib})`; mismatched coords inner-join silently. Assert aligned coords first.
- [ ] `dataframe_to_json_records` round-trips through `df.to_json()`, which defaults to 10 decimal places. Fine for display; worth a docstring line for anyone who might sum these values.
- [ ] `plot_interactive.waterfall` / `channel_share`: `hdi_probs = [hdi_prob] if hdi_prob else []` treats `0` as "no HDI" rather than rejecting it as invalid.
- [ ] `factory.py:61-69` has three separate import statements from `summary.sensitivity` — collapse into one.

## Testing

Coverage is broad on the error paths — `test_summary_edge_cases.py` is a good addition (missing groups, degenerate KDE, polars absent, invalid quantiles). It is thin on values: the new schema tests assert column presence and `len(df) > 0` but never a number, which is exactly why finding 1 slipped through.

- [ ] Add hand-computed numeric assertions for `waterfall`.
- [ ] Assert `channel_share_hdi` shares sum to 1 (or 100, per finding 3).
- [ ] Add a known-answer fixture for `crps`.

juanitorduz:
@drbenvincent I think we should not change much about how the waterfall plot is being constructed. We just need to make it consistent.

drbenvincent:
@juanitorduz @daimon-pymclabs Addressed the waterfall consistency feedback in `1309bff2`.

**Juan's point — don't change how the waterfall plot is built, just make it consistent**

The refactor had routed `DecompositionPlots.waterfall` through `compute_waterfall_components`, which **summed** over `date`. The original plot **averaged** over `(chain, draw, date)` (with dim filtering for components like baseline that lack a date dim). That silently scaled every bar by `n_dates`.

- **Plot:** restored the original inline construction from `main` — no longer depends on `compute_waterfall_components`.
- **Summary:** `compute_waterfall_components` now **means** over `date` (keeping `chain`/`draw` for HDIs), so `mmm.summary.waterfall()` `mean` values match the matplotlib bar heights.
- **Docs:** updated `waterfall()` docstring to describe mean per-period contributions, not summed totals.
- **Test:** added `test_waterfall_bar_values_match_summary_mean` — numeric regression that each bar width equals the corresponding summary `mean`.

`channel_share_hdi` was already consistent (sum over date in both plot and summary); unchanged.

Stacked docs PR #2861 does not need changes for this — it only documents the summary API.

daimon-pymclabs:
Checked `1309bff2` against Juan's point and it looks right: `compute_waterfall_components` now means over `date` (keeping `chain`/`draw`), the plot is back to its original inline construction with per-entry `mean_dims` so dimensionless components like baseline are handled, and the `factory.waterfall` docstring says mean per-period contributions rather than totals. So the plot behaviour matches `main` again and the summary agrees with it.

One structural note: with the plot no longer calling `compute_waterfall_components`, the two code paths are duplicated and only `test_waterfall_bar_values_match_summary_mean` keeps them from drifting. That test only covers the `simple_plots` fixture, so a panelled/custom-dims case (where the plot's dim-selection and squeeze logic differs most from the helper's) would be the place a future change could silently diverge. Worth a parametrized variant over the panel fixture if it's cheap.

daimon-pymclabs (commented):
## Review round 2 — at `1309bff2`

Re-checked every item from the first round against current HEAD, plus the new commit.

### Resolved: finding 1 (waterfall reduction)

`1309bff2` fixes this properly and in both directions. Confirmed by diffing `plotting/decomposition.py` against `main`: the `waterfall` body is now byte-identical to `main`, so the plot's bar heights are genuinely unchanged (Juan's ask). And `compute_waterfall_components` (`helpers.py:145-153`) now means over `date`, so `mmm.summary.waterfall()` agrees with the plot rather than being `n_dates` larger. `test_waterfall_bar_values_match_summary_mean` is the right kind of test. Also verified `compute_channel_shares` is a faithful extraction of the inline code it replaced in `channel_share_forest`.

One semantic upside worth calling out in the PR description, because it is a behaviour change to the *summary* that nobody has mentioned: under the old sum-over-date, `baseline` (which has no `date` dim) was left unreduced while channels were summed, so `share_mean` / `pct_of_total` were comparing per-period baseline against `n_dates`-scaled channels. The switch to mean makes those shares meaningful for the first time. Good change, just currently invisible.

### New, introduced by the fix

- [ ] **The waterfall plot no longer consumes the summary layer.** The revert dropped `compute_waterfall_components` from `plotting/decomposition.py`, so the component-assembly loop over `(baseline, channels, controls, seasonality)` with `_select_dims(..., allow_missing=True)` now exists twice: `decomposition.py:314-330` and `helpers.py:135-153`. That is the exact duplication pattern of finding 2, in the one plot the PR's own description holds up as the layering example, and the two copies can now silently drift apart.

  Both blockers to reuse are gone — the helper means over `date`, and the hard-coded `dim=["chain", "draw"]` is no longer in the plot path. So the plot should be able to go back to `compute_waterfall_components(...)` followed by a mean over whichever of `chain`/`draw` are present, and the numeric test then becomes true by construction instead of by coincidence. If the stacked `component` dim genuinely does not work for the panel case (e.g. `xr.concat` broadcasting `baseline` across `extra_dims` changes the `ref_da` selection), please add a comment saying so, so the next refactor does not repeat this.

- [ ] **The numeric regression test only covers the non-panel case.** `test_waterfall_bar_values_match_summary_mean` uses `simple_plots` / `simple_data`, which have no extra dims. The panel branch (`decomposition.py:334-355` `ref_da` search, and the per-panel `sel(...).squeeze()` at 375-381) is the part the revert rewrote most heavily, and it has no numeric assertion — `panel_plots.waterfall()` is only checked for axes count at `test_decomposition.py:340-345`. Extend the test to `panel_plots` with a `dims={"geo": ...}` selection.

- [ ] **It is a consistency test, not a known-answer test.** It asserts plot == summary, so if both sides changed to sum together it would still pass. One hand-computed expected value for a single component would close that.

- [ ] **Nothing asserts shares sum to 100.** Given the semantics change above, `share_mean` summing to 100 across components is now a true invariant worth pinning.

### Still open from round 1 — all re-verified at HEAD, none addressed

Line numbers below are current, and each is unchanged since the last round.

- [ ] **Finding 2 (was blocking): duplicated HDI computation.** `factory.py:147`, `factory.py:167`, `factory.py:208` are still byte-for-byte duplicates of `StatsHelper` in `helpers.py:243-330`, and the factory still calls its own copy at 14 sites. `StatsHelper` is still reachable four ways (class, `helpers.py:333` singleton, module wrappers at 340/347, and `_stats = StatsHelper()` in `budget.py:41`, `cv.py:42`, `sensitivity.py:40`).
- [ ] **Finding 3: share units inconsistent.** `factory.py:1124` still `* 100`; `compute_channel_shares` still returns fractions.
- [ ] **Finding 4: `pct_of_total` duplicates `share_mean`.** `factory.py:1140`.
- [ ] **Finding 5: `summary/cv.py` still imports matplotlib.** `cv.py:23`, `date2num` at `cv.py:118`, and `plotting/cv.py:36` still imports the private `_build_predictions_arrays`.
- [ ] **Finding 6: CRPS still O(rows) `.sel()`.** `cv.py:147-155`, including the silent `arr[:n_samples]` truncation.
- [ ] **Finding 7: bare `except Exception` → NaN.** `cv.py:189`, still no traceback in the warning.
- [ ] **Finding 8: `crps(dims=...)` still silently ignores unknown keys.** `cv.py:375-383`.
- [ ] **Finding 9: `prepare_allocation_roas` validates `allocation` and never uses it.** `budget.py:54-55`, denominator still `total_allocation` at 63-66.
- [ ] `xarray_utils.py:23-29` — public module, `__all__` is all underscore names.
- [ ] `factory.py:269` — `str(int(hdi_prob * 100))` still collides for e.g. `[0.94, 0.945]`.
- [ ] `factory.py:1534` — `xr.Dataset({"x": x_data, "y": mean_contrib})` still inner-joins mismatched coords silently.
- [ ] `helpers.py:357` — `df.to_json()` still rounds to 10 dp with no docstring note.
- [ ] `plot_interactive.py` — `[hdi_prob] if hdi_prob else []` at 801, 927, 1015, 1422, 1539, 1685, 1714 (seven sites, more than I said last time) still treats `0` as "no HDI".
- [ ] `factory.py:61-69` — still three separate imports from `summary.sensitivity`.

Happy for the small notes to be deferred or waved off, but finding 2 was the other blocker and the fix for finding 1 has now made it worse rather than better. CI is green across the board.

juanitorduz:
Thanks @drbenvincent ! I think we are very close! It seems both daimon's still identified some blockers :)

daimon-pymclabs:
@juanitorduz to be clear on my end: nothing outstanding from me is a blocker. My last comment confirmed `1309bff2` does what you asked, the plot is back to its `main` construction and the summary now means over `date` so the two agree. The only thing left was a nit about the two code paths being duplicated and covered by a single fixture, which is fine to defer to a follow-up. Happy for this to go in.

juanitorduz (approved):

juanitorduz (approved):

juanitorduz:
Thanks @drbenvincent !!!

PR #2860, opened 2026-08-11
Feedback:  MMM campaign credibility disaggregation

juanitorduz:
Feedback round on the campaign credibility disaggregation notebook. One substantive correction, a set of new validation checks, and a refactor of the code, figures and prose.

## The substantive fix

The naive rate was divided by **unique reach** while the credit was allocated over **daily reach-days**, so the credibility dial did not interpolate between anything interpretable. The rate is now per reach-day throughout, which makes both ends of the dial exact and asserted in the notebook: at `Z=0` the credit follows delivery (matches the daily-reach split to 0.0020) and at `Z=1` it returns the click split (to 0.0220, the largest gap there being `flash_sale_burst`, whose delivery is the most concentrated in time).

The two quantities are now kept strictly apart, in code and in prose: **unique reach** is the evidence that sets `Z`, **reach-days** are the exposure the credit is spent over.

## New validation

- Baseline scoreboard graded with `pymc_marketing.metrics.crps` alongside MAE and Spearman, plus 94% HDI coverage. The credibility split reaches Spearman 0.893 and CRPS 0.0254, the best probabilistic score of everything tested, against 0.0321 for the closest heuristic.
- Swept a common `Z` against the hidden truth: partial pooling beats both extremes of the dial, and the fitted dials land in the interior region that recovers the ranking.
- Per-campaign daily correlation with the truth, which qualifies the daily-shape claim (0.94 down to 0.28).
- Two sensitivity tables: priors (`sigma_s` scale, `log_k` width) and the notebook's own thesis, swapping the volume measure that sets `Z` from unique reach to reach-days, at three genuinely different seeds.

Where the results do not favour the method they are stated plainly: on point accuracy a plain delivery split is still slightly better (MAE 0.0321 against 0.0347), and seven campaigns cannot settle unique reach versus reach-days as the evidence measure. On the graded scores reach-days are mildly ahead, and the notebook says so before explaining why it overrides them on conceptual grounds.

## Model and sampling

Reparametrized on `log_k` and `sigma_s`, deriving `tau` and `Z`, which samples with **0 divergences** at `target_accept=0.95` over 4 chains. Added `pm.model_to_graphviz` and prior predictive checks that justify the priors on `log_k` and `sigma_s` in units a media analyst can argue with.

Seeds are threaded explicitly: `fit_credibility`, `paired_draws` and `evaluate_variant` all take an integer seed rather than sharing a module-level generator, so the seed level of the sensitivity tables reflects independent fits.

## Numbers cannot drift out of the prose

The recurring scores are interpolated from the tables that produce them, so the scoreboard, prior-sensitivity and evidence-choice paragraphs are generated from `scoreboard`, `prior_sensitivity` and `evidence_choice` rather than typed in. Verdict sentences are computed too, not just the digits. The 16 numbers that have to stay in markdown, in the opening callout, the honest-part callouts and two figure captions, are asserted in a guard cell, so a config change fails the run instead of quietly making the prose wrong.

## Code, figures and prose

- All imports at the top, a `CampaignData` dataclass plus `load_campaign_data` built on `pymc_marketing.paths`, and small typed functions for the direct estimates and the daily allocation.
- New exploratory data analysis section, and a plot contrasting summed daily reach with unique reach that motivates the exposure distinction.
- `arviz-darkgrid` style with `Cn` colors, `sns.heatmap` for the overlap matrix, `azp.plot_dist` for the posterior shares, no clipped axis limits, legends and labels moved off the data, no `plt.show()`, and axis labels set in single `ax.set(...)` calls.
- The blanket `warnings.filterwarnings("ignore")` is replaced by a filter for the one warning the notebook actually raises, and the file is down from 4.99 MB to 2.22 MB after dropping the retina inline format and the widget state.

The six data files move to `data/` with a `credibility_` prefix so the notebook loads them through `pymc_marketing.paths`.

----
📚 Documentation preview 📚: https://pymc-marketing--2860.org.readthedocs.build/en/2860/


daimon-pymclabs (commented):
## Review: campaign credibility disaggregation

Full read of the notebook, `ruff` run against it, and every quoted number diffed against the executed outputs committed in this PR. Inline comments are not possible here (GitHub rejects them: "diff is too large"), so each item below is anchored by cell index and cell id, with a searchable quote.

**The method holds up.** The reach-day unit fix is correct: at `Z=1` the reach-days in the numerator do cancel the denominator, and the allocation reconciles to the MMM draws at 1.8e-12. The `log_k`/`sigma_s` reparametrization is well motivated, `crps`/`per_observation_crps` are called with the sample dim on the left as their signatures require, and the prior predictive checks plus the CRPS-graded scoreboard are real additions. The notebook's willingness to report where the method loses is the best thing about it.

**What blocks merge:** the two sections whose job is to establish trustworthiness describe a run that is not the one committed.

---

### Blocking

- [ ] **1. The seed sweep does not vary the seed.** `cell 60` (`a35faffd`), `for seed in (42, 1, 7)`.
  `evaluate_variant` takes no seed argument and `fit_credibility` unconditionally passes `random_seed=rng`, so these three values only land in the index. The committed output confirms it: all three "unique reach" rows are identical to the last digit (`k_median` 38051.1655, `Z` 0.3339-0.6046, MAE 0.0341, CRPS 0.0252), and likewise all three reach-days rows. Refits at a fixed config are deterministic here, so nothing was reseeded.
  *Fix:* give `fit_credibility` and `evaluate_variant` a `seed: int` parameter, build the rng inside from it, pass `seed` through the comprehension, re-run. If six extra fits are not worth it, drop the `seed` index level and delete the stability claim in cell 61 instead.

- [ ] **2. Cell 61 (`d380746f`) quotes three numbers that disagree with the table above it,** and its last sentence describes the sweep that was not run. Committed reach-days output: MAE **0.0319** (identical across all three rows), coverage **4.0** of 7, CRPS **0.0246**.
  - "slightly better MAE at 0.0326 to 0.0328" -> 0.0319, a single value
  - "worse coverage at three of seven" -> 4 of 7
  - "the same CRPS to within seed noise" -> there is no seed noise to be within
  - "three seeds move MAE and CRPS only in the fourth decimal, and the coverage count by at most one campaign" -> they move by zero, because the seed never changed

- [ ] **3. Cell 59 (`d01dcf42`): four claims do not match the prior sweep output.**
  - "MAE stays between 0.0338 and 0.0354" -> committed range **0.0331 to 0.0348**
  - "CRPS between 0.0245 and 0.0256" -> committed range **0.0242 to 0.0254**
  - "The only divergence anywhere in this sweep appears at the widest pair of priors" -> the `divergences` column is **0 in all six rows**. The 3-divergences-per-fit result is in the reach-days variant one table down, and the prose there never mentions it.
  - "Coverage sits at five of seven in five settings and four in the sixth" -> committed coverage is 5, 5, 4, 5, 5, 4, so **four of seven in two settings**.

  The qualitative conclusions all survive (not a prior artefact, ~6% drift in the threshold, coverage too noisy to tune against), so this is a regeneration rather than a rethink. Worth adding one sentence acknowledging the reach-days variant's 3 divergences: a reader who spots them in the table and finds no comment will discount the rest.

- [ ] **4. Cell 57 (`5efc204b`): "at several seeds so that sampling noise does not get mistaken for a result"** is not true of the committed run. Same fix as item 1.

- [ ] **5. Stale credibility-split scores, repeated in four places.** The scoreboard output says `credibility split  0.8929  0.0341  0.0252`, but MAE **0.0343** / CRPS **0.0250** appear in `cell 47` (`d2e704e2`, items 1, 2 and 3) and again in `cell 62` (`f8625967`, the "MAE is 0.0343 against 0.0321" bullet). The gap is 0.0002 and changes no conclusion, but the PR body says every number was checked against the executed outputs and these are the first four a reader will check. Given how often they recur, consider interpolating them from `scoreboard` rather than hardcoding.

- [ ] **6. `target_accept` mismatch.** `cell 30` (`93a6df0d`) and the PR description both say the reparametrization "produces none" at `target_accept=0.95`; `fit_credibility` in `cell 33` (`c300dca8`) sets **0.92**. Either raise it so the comparison is like-for-like, or restate the claim at 0.92. As written the divergence comparison is against a setting the notebook never runs.

### Substantive

- [ ] **7. State the evidence-measure result as the table states it.** `cell 61` (`d380746f`), "the choice is not paying for itself in accuracy either way". Reach-days beat unique reach on both graded scores (MAE 0.0319 vs 0.0341, CRPS 0.0246 vs 0.0252) at identical Spearman (0.8929). Unique reach wins only on coverage, which cell 59 explicitly calls "far too noisy to tune against". The conceptual argument in the first half of that paragraph is the right one and worth keeping. Suggest saying plainly that the graded scores mildly favour reach-days and you are overriding them because unique reach is the quantity that answers "how much do I know about this campaign?" That is a stronger position than implying the numbers are neutral.

- [ ] **8. The `Z=1` endpoint tolerance is 5x looser than the `Z=0` one** (0.025 vs 0.005 in `cell 43`, `48df3b82`), and the realized `click_residual` is **0.0220** in share units. That is an order of magnitude larger than the MAE differences the notebook argues about (cell 47: "MAE differences of 0.002 are not evidence"). The assert is fine as a guard, but the claim that the dial interpolates cleanly between the two splits is doing more work than 0.0220 supports. Worth naming which campaign carries the residual and why, so the endpoint behaviour is documented rather than just asserted.

- [ ] **9. Add the compositional-likelihood caveat** to `cell 62` (`f8625967`). The seven `y_c` are compositional by construction: click shares sum to one, so a high share mechanically depresses the others, while the likelihood treats campaigns as conditionally independent given `mu` and `tau`. With seven campaigns the induced negative dependence is not negligible, and it plausibly contributes to coverage sitting at 5 of 7 for a 94% interval. It fits alongside the existing "the uncertainty is a lower bound" bullet.

- [ ] **10. One-line comment on `share_at_common_z`** (`cell 43`, `48df3b82`) noting that dropping the `(1 - z) * mu` term is deliberate. It is correct, since a constant added to every `theta` cancels in the normalization, but a reader comparing against the model definition will read the missing term as a bug and stop to re-derive it.

- [ ] **11. `k` is hardcoded at "roughly 39,000" in the opening callout** (`cell 0`, `947c850a`) while the fit returns **38,051**. Quoting a fitted result in the intro drifts silently on any upstream change: the prior sweep alone moves it between 37,320 and 38,537. Either make it explicitly coarse or inject the fitted median.

### Conventions and mechanics

Lower priority. `docs/source/notebooks/*/dev` is excluded from `ruff` (`extend-exclude`), from the docs build (`exclude_patterns`) and from the watermark hook, so none of this fails CI today. It matters when the notebook graduates out of `dev/`.

- [ ] **12. `warnings.filterwarnings("ignore")` at module scope** (`cell 1`, `7b160331`) is unique in this repo: 0 of the other 61 notebooks do it. It will hide real deprecations from anyone running this. Narrow it to the specific message and category.
- [ ] **13. matplotlib floor.** `orientation=` on `boxplot` (`cell 31`, `aa1505d8`) needs matplotlib >= 3.10 and `tick_labels` needs >= 3.9, against a package floor of `matplotlib>=3.5.1`. No other notebook in the repo uses `orientation=`. `vert=False` plus `labels=` is the backwards-compatible spelling.
- [ ] **14. `ruff` items:** four `S101` assert violations, and several `ruff format` diffs (the `fit_credibility` signature fits on one line; the trailing comma in the `paired_draws` call forces an expansion).
- [ ] **15. `paired_draws` builds `np.random.default_rng(seed)` from the module-level global** rather than taking an rng or seed. Same fix as item 1.
- [ ] **16. The committed file is 4.8 MB** (retina inline figures plus widget state), the largest thing in the notebook tree. It passes only because of the `check-added-large-files` exclusion. Stripping widget state and dropping to standard DPI would cut most of it.

---

### Verdict

The unit fix, the reparametrization and the added validation are all real improvements. But the sensitivity narrative currently describes a run that is not the one committed, and the seed sweep does not do what it claims, and both sit in exactly the sections whose purpose is to establish that the results are trustworthy. I would re-execute end to end and regenerate cells 57, 59 and 61 from the actual output before merging.

juanitorduz:
Thanks, this was a useful review: the blocking diagnosis was exactly right. The notebook has been re-executed end to end after fixing the seed plumbing and the sampler setting, so every table and every sentence now comes from the same run. All 16 items are addressed.

**What changed structurally.** `fit_credibility` takes `sample_seed: int`, `paired_draws` takes `pair_seed: int` and `evaluate_variant` takes `run_seed: int`, which the `evidence_choice` comprehension passes through, so the seed level is now real: the unique-reach rows come out at `k_median` 37,061 / 38,104 / 40,298 instead of three identical rows. Thirteen fits total, and the whole notebook still executes in about 30 seconds. The fit runs at `target_accept=0.95`, so the divergence claim in cell 30 is now like-for-like, and it samples with 0 divergences and max R-hat 1.000.

**The recurring scores are interpolated rather than typed.** Cells 47, 59 and 61 are now `display(Markdown(f"..."))` driven by `scoreboard`, `prior_sensitivity` and `evidence_choice`, including the verdicts themselves: whether the credibility split wins on CRPS, and which measure wins on MAE, are computed from the tables rather than asserted in prose. The numbers that cannot be interpolated, because Quarto callouts and figure captions have to stay in markdown, are now covered by a guard cell before "The honest part" that asserts all 16 of them with `# noqa: S101`, so a config change breaks the run instead of quietly making the prose wrong.

### Blocking

1. **Seed sweep.** Fixed as above. The three unique-reach rows now differ, and the stability claim in cell 61 is computed from the actual spread (MAE and CRPS move by at most 0.0008 across seeds, coverage by at most one campaign).
2. **Cell 61.** Fully regenerated from `evidence_choice`, including the ranges rather than single values where the seeds disagree, and the MAE and coverage comparisons in both directions.
3. **Cell 59.** Fully regenerated from `prior_sensitivity`: MAE 0.0338 to 0.0348, CRPS 0.0245 to 0.0254, threshold 36,681 to 39,094 people (about 7%), no divergences anywhere in the sweep, coverage 5 of 7 in four settings and 6 of 7 in two. The divergence and coverage sentences are conditional on the table, so they stay true if the sweep changes. Cell 61 now also states the reach-days divergences explicitly: 1 across the three seeds against 0 for unique reach.
4. **Cell 57.** The "several seeds" sentence is now true of the committed run.
5. **Stale scores.** Interpolated. The current run gives MAE 0.0347 and CRPS 0.0254, and the two callout bullets that keep them in markdown are guarded.
6. **`target_accept`.** Raised to 0.95.

### Substantive

7. **Evidence measure.** Restated as you suggest. The section now opens with "this dataset does not settle the question, and on the graded scores it mildly favours the delivery measure", quotes both scores in both directions, and then makes the conceptual argument explicitly as an override: reach-days measure how hard we pushed, not how much we learned, and an MAE gap of 0.0024 on seven campaigns is not the kind of evidence that should redefine what evidence means. The "not paying for itself either way" line is gone.
8. **`Z=1` endpoint.** The cell now reports which campaign carries the largest gap at that end, `flash_sale_burst`, with the reason (its delivery is the most concentrated in time, so the day-by-day renormalisation does not average back to its window-level click share), and there is a comment above the two asserts explaining why the tolerances differ: `Z=0` is exact up to floating point, while `Z=1` is only exact for a campaign whose reach is spread evenly over the window.
9. **Compositional caveat.** Added next to the "uncertainty is a lower bound" bullet, phrased as you describe: the shares sum to one, the likelihood treats campaigns as conditionally independent, and the induced negative dependence plausibly contributes to 5 of 7.
10. **`share_at_common_z`.** Comment added noting the dropped `(1 - z) * mu` is deliberate because a constant on every `theta` cancels in the normalisation.
11. **Intro `k`.** Now explicitly coarse ("roughly 40,000 reached people, the same order as an average campaign here; the fitted value is reported exactly further down"), with the two dial claims stated qualitatively. The guard cell asserts the coarse values still bracket the fit.

### Conventions and mechanics

12. **Warnings.** The blanket ignore is gone. I ran the notebook under `simplefilter("always")` to see what actually fires: one warning, seaborn passing the deprecated `vert` kwarg to matplotlib's boxplot. The filter is now narrowed to that message and category.
13. **matplotlib floor.** That panel is now `sns.boxplot(..., orient="h", order=z_order)` from a tidy frame of the prior `Z` draws, matching the seaborn boxplots already used in the CTR panel, which avoids the version-specific kwargs entirely.
14. **`ruff`.** `ruff format` is clean and the deliberate asserts carry `# noqa: S101`. One knock-on: `ruff format` strips trailing semicolons, so the cells that ended in a legend call now bind to `_` instead, which keeps the object reprs out of the outputs.
15. **`paired_draws`.** Takes `pair_seed` and no longer reads the module-level generator.
16. **File size.** 4.99 MB to 2.22 MB. Dropping the retina inline format took the 15 embedded figures from 4.81 MB to 2.03 MB, and the notebook-level widget state is stripped.

daimon-pymclabs (commented):
## Re-review at `bf31be8`

I re-fetched the notebook at `bf31be8`, re-ran `ruff check` and `ruff format` against it with the repo's own config, and re-verified every number in the prose against the committed outputs. **All 16 items are addressed.** The blocking ones are genuinely fixed rather than papered over.

### Verified fixed

| # | Item | Evidence at `bf31be8` |
|---|---|---|
| 1 | Seed sweep does not vary the seed | `fit_credibility(*, sample_seed)`, `evaluate_variant(*, run_seed)`, `paired_draws(*, pair_seed)`, all threaded. The output now genuinely differs per seed: `k_median` 37,060 / 38,104 / 40,298 for unique reach. |
| 2 | Cell 61 numbers stale | Whole paragraph now interpolated from `evidence_choice`. Spot-checked the hard cases: max MAE/CRPS seed spread is 0.0008 (reach-days MAE 0.0327/0.0319/0.0324), coverage ranges 3-4 vs 4-5 of 7. Both correct. |
| 3 | Cell 59 numbers stale | Interpolated, including `divergence_claim` and `coverage_claim` branching on the actual column. Output reads 0.0338-0.0348 MAE, 0.0245-0.0254 CRPS, "5 of 7 in 4 settings and 6 of 7 in 2 settings" -- all match the table. |
| 4 | "at several seeds" untrue | True now. |
| 5 | Stale 0.0343 / 0.0250 in four places | Gone. Scoreboard prose interpolated; 0.0347 / 0.0254 consistent everywhere including the closing callouts. |
| 6 | `target_accept` mismatch | Now 0.95 in `fit_credibility`, matching the prose, and the run still reports 0 divergences, so the claim holds as stated. |
| 7 | Evidence-measure result softened | Now says it outright: "on the graded scores it mildly favours the delivery measure", followed by the conceptual override stated as an override. This is the right shape, and the `verdict`/`mae_favours` branching means it stays honest if a refit flips the sign. |
| 8 | `Z=1` endpoint | `click_offender` injected (`flash_sale_burst`, gap 0.0220), plus a comment explaining why the two tolerances legitimately differ (daily renormalisation only recovers the window-level click share under even delivery). That is a better answer than the one I asked for. |
| 9 | Compositional caveat | Added to the "what it cannot do" callout, correctly tied to why the intervals are optimistic. |
| 10 | `share_at_common_z` | Comment added. |
| 11 | Hardcoded `k` | Intro now says "roughly 40,000 ... the fitted value is reported exactly further down", and the exact value is injected in the fit cell. |
| 12 | `filterwarnings("ignore")` | Scoped to message + `PendingDeprecationWarning`, with a comment naming the source. |
| 13 | matplotlib floor | `orientation=` and `tick_labels=` both gone, replaced with seaborn's `orient="h"`. No version floor problem left. |
| 14 | `ruff` | `ruff format --diff`: clean. `S101`: gone (`# noqa`). See below for what the fix introduced. |
| 15 | `paired_draws` global rng | Takes `pair_seed`. |
| 16 | 4.8 MB file | Down to **2.2 MB**, widget state removed. |

The assert-guard cell for the numbers that have to stay in markdown is a better solution than what I suggested -- a config change now breaks the run instead of quietly making prose wrong. I re-derived all 16 guarded quantities from the committed outputs and they hold.

### Three small things, none blocking

- [ ] **The interpolation fix introduced 8 `E501` violations.** Long f-strings cannot be wrapped by `ruff format`, so the three rewritten prose cells now exceed the 120-char limit. Currently invisible (the `dev/` exclusion) but it is the one thing that would fail `ruff check` on graduation:
  ```
  cell 48:40,42,43,45   E501 Line too long (351, 173, 437, 528 > 120)
  cell 60:39,41         E501 Line too long (822, 148 > 120)
  cell 62:54,56         E501 Line too long (1090, 543 > 120)
  ```
  (ruff's 1-indexed cell numbers: the scoreboard prose `ebe73388`, the prior-sweep prose `36993e2f`, the evidence-choice prose `6101b16b`.) Implicit string concatenation across lines fixes it without changing the rendered output.

- [ ] **`I001` import block un-sorted** in cell 1 (`21fe2938`), auto-fixable. Pre-existing, present in both revisions -- flagging only because it is the other thing standing between this notebook and a clean `ruff check`.

- [ ] **The guard covers 16 numbers; a few fit-dependent ones in markdown are still outside it.** All currently correct, so this is completeness rather than an error: the common-`Z` sweep MAEs in `aeef91b0` (0.0310 / 0.0322 / 0.0373 -- cheap to close, since the cell above already interpolates exactly those three), the daily-correlation endpoints in `0714c6e3` (0.94 and 0.77, while the two bad ones 0.56 / 0.28 *are* guarded), the share-error gaps in `64d74d24` ("about 0.05 each", "within 0.02"), and the prior-predictive quantiles restated in `46f9552b`.

### Two observations, no action needed

- The claim in `46f9552b` that sampling $(\tau, s^2)$ directly "produced divergences at `target_accept=0.95`" is now internally consistent, but it is still the one empirical claim in the notebook a reader cannot reproduce from it, since that parametrization is never fitted here. Fine to leave; worth knowing it is load-bearing for the reparametrization argument.
- The coverage guard asserts exactly 5 with zero tolerance, while the scoreboard prose now says the count "moves by a campaign or two across seeds". Both are defensible together (the main fit is seeded), and the observed movement is in fact at most one campaign, so the prose is conservative in the safe direction.

### Verdict

My blockers are cleared. The sensitivity sections now describe the run that is committed, and they are self-maintaining rather than hand-transcribed, which is a stronger position than the notebook was in before the review. Nothing outstanding from my side beyond the `ruff` cleanup, which can wait for the move out of `dev/`.

cetagostini (approved):

Issue #2830, opened 2026-08-06
Incrementality module: first-class support for complex funnel/mediator effects (direct vs mediated decomposition)


juanitorduz:
## Context

#2815 / #2828 introduced an opt-in contract for spend-dependent `mu_effects`: an effect whose contribution depends on `channel_data` must implement `incrementality_spec()` returning an `IncrementalitySpec`, otherwise the incrementality module raises rather than silently reporting the direct path as if it were the total. Once opted in, `compute_incremental_contribution` / `contribution_over_spend` return the **total** (direct + mediated) per-channel increment.

The advanced funnel notebook (#2814, `mmm_funnel_mueffect_advanced.ipynb`) shows that users of complex funnel models also want the **direct-only** path (and implicitly the mediated share). Today that must be assembled by hand from the `channel_contribution` posteriors — fine under the identity link, but not counterfactually meaningful under non-additive links (log, logistic), where contributions and increments no longer coincide.

## Proposal

An explicit estimand option, e.g. `paths="total" | "direct" | "mediated"` on `compute_incremental_contribution` / `contribution_over_spend`:

- `_delta_mu` already computes `channel_delta` and the per-effect deltas separately, so the decomposition is structurally available.
- Under nonlinear links, "direct" should be defined as a **controlled direct effect** (hold the mediator's contribution fixed while intervening on spend). The `pm.do` intervention routing from #2829 provides the machinery for this.

## Also consider

Reporting effect-level quantities users currently hand-roll — e.g. the induced mediator movement (ΔM*) in the mediator's own units. The advanced funnel notebook needs exactly this for its induced-spend ROAS denominator (spend + induced lower-funnel spend) and currently derives it from a manual `sample_posterior_predictive` counterfactual.

## Acceptance sketch

- The advanced funnel notebook's "Funnel (direct only)" row and its induced-spend denominator are computable from the incrementality module alone.
- Guard behavior unchanged: the module never silently reports the direct path as the total; every estimand is explicitly requested and labeled.

## References

- Notebook: `docs/source/notebooks/mmm/mmm_funnel_mueffect_advanced.ipynb` (#2814)
- Opt-in contract: #2815, part of #2828
- Intervention routing via `pm.do`: #2829

🤖 Generated with [Claude Code](https://claude.com/claude-code)

juanitorduz:
I think this one should follow after https://github.com/pymc-labs/pymc-marketing/pull/2813

drbenvincent:
[...]

Issue #2468, opened 2026-04-01
Add a component for the marketing funnel


TeemuSailynoja:
## Goal
Provide a way for users to model lower funnel dependency on upper funnel channels that integrates seamlessly with the library.

## Context
In the past, we have used a custom MuEffect to implement various advanced components into multidimensional MMMs. A structure of marketing funnel keeps coming up and we should provide a component for users to have their first steps into these MMMs with a more complex causal structure.

## Challenges
1. How strong are the assumptions we want to make on the structure?
2. How to integrate the mediated upper funnel contribution into the existing plots and summaries?
3. We need to ensure the optimizer works with the class. (I have a solution with MaskedPrior for this)
4. I think this class should come with a guide on how to expand it.

## Proposed path
1. Make a case study notebook on funnel modeling (I'm already working on one).
2. Agree on an API for the class to include in the library.
3. Update the notebook if needed to use and introduce the implemented class.

PR #2868, opened 2026-08-12
docs: address #2775 review findings in mmm_funnel_mueffect notebook

### LINKED ISSUE #2775: Updates to `mmm_funnel_mueffect` docs page + thoughts on the approach of `FunnelEffect`
Relates to the docs page `mmm_funnel_mueffect.ipynb` as well as the broader approach being taken with `FunnelEffect`.

---

This review has two layers, kept deliberately separate:

1. **The notebook** — is what it shows correct, and are its claims supported by what it runs?
2. **The direction** — does `FunnelEffect`-on-`MMM` scale as the route from "everything is a
   direct effect" to a DAG-first MMM?

The short version: the notebook is good work and its core result is real, with a few claims that
overreach what it computes. The direction concern is separate and unaffected by those fixes.

---

# Part 1 — The notebook

## What it establishes

**The estimation result holds.** On the synthetic DGP, Naive A (mediator as channel) and Naive B
(mediator omitted) are both biased; the funnel model recovers direct, indirect, and total
upper-funnel effects.

**The `FunnelEffect` pattern works.** A `MuEffect` that declares its own likelihood inside
`create_effect` turns "add a term to `mu`" into "fit a second structural equation jointly". That is
a genuinely useful demonstration, and it plugs into `MMM` without touching the class.

**It is the first *joint* funnel notebook.** The repo already ships `mmm_intro_upper_funnel` and
`mmm_upper_funnel_causal_approach`, which tell the same funnel story with a **two-stage** approach
(separate outcome model and mediation model). Joint estimation in one model is the new
contribution and should be credited as such — but see Part 2 on what three overlapping funnel
notebooks with three different mechanics implies.

## Claims that overreach

**"The bias is invisible in predictive fit" is only true for Naive A.** In-sample R² is Naive A
0.9649, Naive B 0.9412, funnel 0.9675. Naive A carries the worst bias (−27.2%) at an R²
indistinguishable from the funnel model — that comparison alone makes the point, emphatically.
Naive B has ~80% more residual variance, so fit *does* partially flag it. The notebook then
concludes "held-out error alone cannot catch it" without ever computing held-out error. Narrow the
claim to Naive A, or compute the holdout.

**The funnel model is the exact generative model.** Data is produced by `MMM(...) +
FunnelEffect(...)` clamped with `do()` and forward-sampled, then refit by the same classes with the
same transform families. "Only the funnel model recovers the truth" therefore holds under *zero*
structural misspecification. That is fine for parameter recovery, but it demonstrates
self-consistency, not robustness. The untested question is what the funnel model does when its DAG
is wrong — when the observed lower spend genuinely drives the target, or when `U` and `C` are
correlated. The notebook's own punchline (fit will not tell you your model is wrong) applies to the
funnel model too, and saying so pre-empts the obvious "but the funnel model works" reply.

**There is no lower-funnel lever in this DAG.** The graph is U→Y, U→D, C→D, D→M, D→Y. Nothing
points *out of* `M`: observed lower-funnel spend is a **measurement** of latent demand, not a cause
of the target, so `do(lower_spend)` has exactly zero effect on `Y`. The consequence is that this
model cannot answer "should I move budget from upper to lower funnel?" — arguably the main business
reason to build a funnel MMM. The notebook is honest that `Y` depends on `D` rather than `M`, but
files the allocation gap under "budget optimization is out of scope", which reads as a tooling
limitation. Even a perfect optimizer would find nothing to allocate here. Worth stating plainly.

**Terminology.** `lower_control` is described as "an exclusion restriction on the mediator
equation". The restriction doing the identification work is `C`'s *absence from the outcome
equation*. Minor, but the identification story leans on that sentence.

## Things that silently give the wrong answer today

None of these are debatable and all are reproducible:

| What | What happens |
|------|--------------|
| Waterfall decomposition | The notebook sets `plot_suite = "new"`; `DecompositionPlots.waterfall` goes through `get_contributions()`, which returns fixed buckets (channels / baseline / controls / seasonality) with no `mu_effects` branch. The funnel contribution is **dropped entirely** and the decomposition no longer sums to the fitted mean. (The *legacy* suite auto-detects `*_contribution_original_scale` and would show it as one mixed bar.) |
| `incrementality` ROAS | Hardcoded to `channel_contribution`. `upper_roas(funnel)` returns a plausible-looking number that omits the mediated path, with **no warning**. The notebook catches this by hand; a user would not. |
| `MMM.save()` / `load()` | `serialize()` succeeds, `deserialize()` raises `ValidationError` ("Input should be an instance of `AdstockTransformation`"): `to_dict` is supplied with no matching `from_dict` (contrast `MediaMuEffect`). The registered type key is `__main__.FunnelEffect`, so it would not resolve in a fresh process anyway. |
| Two decomposition paths disagree | `compute_counterfactual_contributions_dataset` **does** enumerate `mu_effects`; `get_contributions` does not. The same library answers "what are the contributions?" two different ways depending on which door you use. |

## Smaller gaps

- **Data plumbing.** `to_mmm_dataset(DataFrame)` reads only channel/control columns, so the funnel
  model needs an `xr.Dataset`. `make_mmm_dataset` is notebook-local; there is no library helper.
- **Scaling.** Channels are scaled to [0, 1]; the mediator equation runs on raw `lower_spend` /
  `lower_control`, so its `HalfNormal(1)` priors live on a different scale. Self-consistent here
  because the data is generated in-model; a trap on real data. The notebook flags this.
- **Library component.** `FunnelEffect` is notebook-only, untested.

---

# Part 2 — The direction

## `MMM` has no causal DAG in its representation

`MMM` holds channels, controls, transforms, and an ordered list of `mu_effects`. What gets compiled
is:

```text
μ = intercept + Σ channels + Σ controls + seasonality + Σ mu_effects
→ one primary target likelihood on y
```

The optional `dag` argument does **not** define structural equations. It is consumed only in
`__init__`, where it computes a backdoor adjustment set and intersects it with `control_columns`
(and silently nulls `yearly_seasonality` if seasonality is not in that set). It requires
`outcome_node` as well — passing `dag` alone does nothing at all. `BuildModelFromDAG` exists in
`causal.py` but builds linear `slope * parent` edges with no adstock or saturation, and is not used
by `MMM`.

So the PyMC graph contains `funnel_lower_demand` and a second likelihood, while to the `MMM` object
`funnel_effect_contribution` is just another summand. There is no first-class notion of treatment,
mediator, outcome, or valid intervention.

## The wiring makes this concrete

This is not only an architectural framing — it shows up in the code the notebook has to write:

- `create_effect` reads its treatment as `mmm.channel_data_scaled.isel(channel=0)`, a **hardcoded
  positional index**. Add a channel or reorder `channel_columns` and it silently transforms the
  wrong one.
- `channel_data_scaled` is a plain Python attribute carrying `## TODO: Find a better way to save it
  or access it in the pytensor graph.` in `mmm.py`. The only wiring point for a treatment is an
  acknowledged hack.
- `DataVarMuEffect.create_data` registers `pmd.Data(var_name, ...)` **unprefixed**, so two mediator
  effects sharing a control name collide with `ValueError: Variable name lower_control already
  exists.`

An effect cannot ask the model "which variable is my treatment?" because the model does not know.
It can only reach in by position and by global string name.

## Arbitrary causal queries will not generalize

Interventions are possible in PyMC (`pm.do`, graph surgery). They are not generic in `MMM` because
the library cannot represent what is being intervened on.

| Query | Productized `FunnelEffect` | Richer funnel graphs |
|-------|--------------------------|----------------------|
| Total / direct / indirect effects | Could be automated | Needs new components or a graph layer |
| Total ROAS (incl. induced spend) | Feasible with optimizer work | Custom counterfactuals per specification |
| Path-specific effects (one mediator only) | Needs graph semantics | Bespoke analysis |
| Valid `do()` vs input substitution | Needs documented assumptions | Analyst's responsibility |

The notebook's indirect counterfactual (set `upper_spend = 0`, subtract contributions) is correct
for this exogenous DGP — I checked that the scalers stay pinned, so the substitution is sound. It
is not a general intervention operator: under confounded plans or richer graphs, substitution and
`do()` diverge, and the framework cannot warn about it because it does not know the graph.

Identification is a separate axis again. Even with a causal query API, direct vs indirect paths can
be weakly identified (four independent media transforms), confounded plans add sensitivity (#2769),
and fit metrics will not catch misspecification.

## The open question

Productizing `FunnelEffect` — direct/indirect/total estimands, funnel ROAS, eventually funnel-aware
optimization — is a sensible next step and scales to **fixed funnel templates**.

The question to settle early is whether that is the endpoint, or whether funnel models need a
structural representation layer that supports arbitrary path queries. The evidence that it is not
the endpoint: the repo now has three funnel notebooks, three different mechanics, no shared
abstraction, and a decomposition API that disagrees with itself about whether `mu_effects` exist.
That destination is not reachable by accumulating more one-off effect subclasses.

I am not proposing an alternative here — that is a separate piece of work. The ask is only that we
agree on what the current approach can and cannot do before we build more on top of it.

## Suggested follow-ups

1. Decide whether `FunnelEffect`-on-`MMM` is the long-term architecture or a bridge.
2. Fix the silent-wrong-answer cases: waterfall dropping `mu_effects`, direct-only `incrementality`
   ROAS reported without warning, broken `save`/`load` round-trip.
3. Narrow the notebook's predictive-fit claim, and state that lower-funnel spend is a measurement,
   not a lever, in this DAG.
4. Promote `FunnelEffect` to a tested library component with consistent scaling and a working
   `from_dict`.

drbenvincent:
## Summary

Closes #2775.

This PR updates the `mmm_funnel_mueffect` docs notebook to address the notebook-specific review in #2775, without changing the DAG structure or `FunnelEffect` implementation.

- **Narrow predictive-fit claims:** distinguish Naive A (can match funnel R² with large causal bias) from Naive B (biased and worse fit); rename the section to "Predictive fit vs. causal bias".
- **Holdout evaluation:** refit Naive A, Naive B, and the funnel model on the first 80% of weeks and report holdout RMSE/R² on the last 20%.
- **Pedagogy and scope:** add prerequisites linking to the two-stage funnel notebooks; an `{important}` callout on self-consistency vs. robustness; fix exclusion-restriction wording; add a `{note}` that omitting $M \to Y$ is a deliberate teaching simplification.
- **Library footguns:** document `incrementality` direct-only ROAS, waterfall/`get_contributions()` omitting `mu_effects`, the two decomposition APIs, and broken `FunnelEffect` save/load — with links to open follow-up work.
- **ROAS estimand table** and stronger warning on `incrementality` for funnel models.
- **Figure hygiene:** `FIG_WIDTH`/`COLOR_*` constants, technical captions, `hide-input` on pure plotting cells.
- **Dev deps:** add `ipywidgets` to the `test` extra so local notebook runs avoid Rich's Jupyter warning.

Out of scope (left to other issues/PRs): promoting `FunnelEffect` to the library, fixing waterfall/incrementality in code, architecture decisions from Part 2 of #2775.

## Test plan

- [x] `uv run python scripts/run_notebooks/runner.py --notebooks docs/source/notebooks/mmm/mmm_funnel_mueffect.ipynb` (mock runner passes)
- [x] Pre-commit passes on changed files
- [ ] Docs build renders the updated notebook (holdout section adds three MCMC refits — full execute is slow)

----
📚 Documentation preview 📚: https://pymc-marketing--2868.org.readthedocs.build/en/2868/


juanitorduz:
@drbenvincent Looks great! Please take a look at the titles of the plots since they are gone :)

juanitorduz (changes requested):
Minor comments :)

drbenvincent:
@juanitorduz Thanks for the review — addressed in c78decd2:

- **Notation:** `Y_t \sim \mathcal{N}` → `\text{Normal}` to match `\text{TruncatedNormal}`.
- **FunnelEffect caveat:** reframed as a **Notebook-local implementation** warning (instead of "fragility") and linked to #2871.
- **Plot titles:** restored all missing `suptitle` / `title` calls (synthetic data, true decomposition, target posterior predictive, lower-funnel posterior predictive, parameter recovery, and recovered effect decomposition).
- **Colors:** reverted the synthetic data time-series plot to the original `black` / `C0` / `C1` / `C2` palette.

Notebook re-executed locally; mock runner still passes.

juanitorduz:
@drbenvincent, some additional minor comments. The only important one is to soften/remove the last section, which we are aware of via issues, and we are working towards it. It somehow sounds very negative as the final part of the notebook :)

drbenvincent:
@juanitorduz Thanks for the follow-up review — addressed in `413754a8` (includes merge from `main` / #2813).

**Epsilon notation (cell 4)** — We unified notation so structural equations use `$=$` and observed variables use `$\sim$`, matching the PyMC likelihoods. The causal story is unchanged. If explicit shocks help, the equivalent form is:

$$Y_t = \mu_t + \varepsilon^{Y}_t, \quad \varepsilon^{Y}_t \sim \text{Normal}(0, \sigma_Y)$$

with $\mu_t = \alpha + f_{\text{direct}}(U_t) + f_{l\to y}(D_t)$ as in the notebook. For $M_t$, we deliberately avoid a separate $\varepsilon^{M}$: measurement noise is encoded by $M_t \sim \text{TruncatedNormal}(D_t, \sigma_M, \text{lower}=0)$, not additive Gaussian noise. The exclusion story — execution noise in $M$ does not enter $Y$ because only $D_t$ enters $\mu_t$ — is already in the prose below the equations.

**CRPS (cells 67 / 71)** — Added `CRPS` alongside RMSE and $R^2$ in both in-sample and holdout tables via `pymc_marketing.metrics.crps` on the full posterior predictive.

**Stray `#` (cell 68 area)** — Removed an empty code cell that ReviewNB rendered as a lone `#`.

**Ending section** — Removed the “silent wrong answers” block; replaced with a short library-integration sentence linking to #2830, #2871, #2468, and #2804. Also merged #2813 incrementality changes (`incrementality_spec`, mediated ROAS check vs hand-rolled counterfactual).

Mock notebook runner passes locally.

drbenvincent:
Think this could be ready @juanitorduz. Hoping to merge before I address #2892 on the same notebook.

juanitorduz (approved):
Thank you @drbenvincent !

PR #2873, opened 2026-08-12
fix(mmm): support arbitrary VariableFactory in FourierBase.sample_prior

### LINKED ISSUE #1343: Fourier class `sample_prior` support for non-Prior classes
Having `sample_prior` method is not requirement of VariableFactory, only Prior so this needs to be changed:

https://github.com/pymc-labs/pymc-marketing/blob/c989eeba24e11c093161ab7e6294159f3447adfa/pymc_marketing/mmm/fourier.py#L460-L478

Something like in the media transformations will have to be done in order to support non-Prior classes:

https://github.com/pymc-labs/pymc-marketing/blob/c989eeba24e11c093161ab7e6294159f3447adfa/pymc_marketing/mmm/components/base.py#L339-L361

shivamlalakiya:
## Summary

- `FourierBase.sample_prior` called `self.prior.sample_prior(...)` directly — a method that only exists on `pymc_extras.prior.Prior`. Any other `VariableFactory` (e.g. custom priors already supported by `apply()`, which calls `self.prior.create_variable(...)` generically) raised `AttributeError` instead of sampling.
- `Prior.sample_prior` itself is a thin wrapper delegating to the free function `pymc_extras.prior.sample_prior(factory, ...)`, which works for any `VariableFactory`. Routed `sample_prior` through that free function directly, with `xdist=True` to match `apply()`'s variable-creation path.
- Same root cause / fix shape as `mmm/components/base.py`'s `sample_prior`, which already goes through a generic `create_variable` dispatch rather than assuming a `Prior`-specific method.

Closes #1343

## Test plan

- [x] Added `test_fourier_arbitrary_prior_sample_prior` (parametrized over yearly/monthly/weekly) reusing the existing `ArbitraryCode` custom-`VariableFactory` fixture from `test_fourier_arbitrary_prior`
- [x] Confirmed RED before the fix: `AttributeError: 'ArbitraryCode' object has no attribute 'sample_prior'`
- [x] `pytest tests/mmm/test_fourier.py -q` — 126 passed (123 pre-existing + 3 new)
- [x] Blast-radius sweep `pytest tests/mmm -k "fourier or Fourier or seasonality"` — 164 passed, 0 regressions (covers other `sample_prior` call sites this touches, e.g. `test_additive_effect.py`)
- [x] `pre-commit` (ruff, ruff-format, mypy, docs completeness) all pass

----
📚 Documentation preview 📚: https://pymc-marketing--2873.org.readthedocs.build/en/2873/


juanitorduz (approved):
Thank you @shivamlalakiya !

PR #2874, opened 2026-08-12
fix(mmm): enforce theta domain check in delayed_adstock

shivamlalakiya:
## Description

`delayed_adstock`'s docstring states `theta` "Must be between 0 and `l_max` - 1", but out-of-range values silently produced a degenerate all-zero kernel instead of erroring:

```python
delayed_adstock(x, alpha=0.5, theta=-5, l_max=10, dim="time").eval()  # all zeros, no error
delayed_adstock(x, alpha=0.5, theta=20, l_max=10, dim="time").eval()  # all zeros, no error
```

Added a `_check_theta` helper mirroring `_check_alpha` (added in #960 for `geometric_adstock`'s `alpha`), using `pymc.logprob.utils.CheckParameterValue` rather than a numpy check, per the maintainer discussion on #918.

## Related Issue
- [ ] Closes #
- [x] Related to #918

## Test plan

- [x] Confirmed the pre-fix silent-zero behavior for `theta=-5` and `theta=20` with `l_max=10`
- [x] Added `test_delayed_adstock_bad_theta` (parametrized: `-5, -1, 10, 22` against `l_max=10`), asserting `ParameterValueError` on `.eval()`
- [x] Boundary values `theta=0` and `theta=l_max-1` still pass — verified via existing `test_logistic_saturation_delayed_adstock_composition` parametrizations (theta up to 19) and manual boundary check
- [x] `pytest tests/mmm/test_transformers.py` — 148 passed, 1 pre-existing xfail, 0 regressions
- [x] Blast-radius: `pytest tests/mmm/components/test_adstock.py tests/mmm/test_mmm_adstock_curve.py` (covers `DelayedAdstock` component, the only other caller) — 105 passed, 1 pre-existing skip
- [x] `pre-commit` (ruff, ruff-format, mypy, docs completeness) all pass

## Modules affected
- [x] MMM

## Type of change
- [x] Bug fix

----
📚 Documentation preview 📚: https://pymc-marketing--2874.org.readthedocs.build/en/2874/


tests/mmm/test_transformers.py
```diff
@@ -219,6 +219,29 @@ def test_delayed_adstock_x_zero(self):
         )
         np.testing.assert_array_equal(actual=x, desired=y.eval())
 
+    @pytest.mark.parametrize(
```
williambdean, line 237: can we compile the function once as fixture instead of the eval

shivamlalakiya:
Addressed @williambdean's review comment: `test_delayed_adstock_bad_theta` now compiles `delayed_adstock` once into a `pytensor.function` via a class-scoped fixture, and each parametrized `theta` case reuses that compiled function instead of building the graph and calling `.eval()` per case.

Also fixed the CI failure on `all_tests` / `test (numba, 3.12, false, 5)`: `test_full_roundtrip_basic` used `DelayedAdstock(l_max=3)` with the default `theta` prior `HalfNormal(sigma=1)`, which is unbounded above. With `l_max=3` the valid domain is `[0, 2]`, so prior-predictive draws (via `mock_pymc_sample`) landed outside that range often enough (~4.5% of draws) to trip the new `_check_theta` domain check — the old code just silently produced a degenerate all-zero kernel instead. Bumped `l_max` to 10 (matching the docstring's canonical example), pushing the cutoff to 9 (~9 sigma out for `HalfNormal(1)`), so this is no longer flaky.

Pushed in a073bc74..0faf6d56. CI is now green.

ricardoV94:
As someone who doesn't care much about tools like mypy and pydantic... beware these sorts of checks can have big impacts in model speed (sampling, optimization, etc...). I wouldn't push too hard to enforce them.

People usually love JAX for speed, and one of the first compromises JAX does is ignore any forms of checks whatsoever for the sake of speed.

williambdean:
> As someone who doesn't care much about tools like mypy and pydantic... beware these sorts of checks can have big impacts in model speed (sampling, optimization, etc...). I wouldn't push too hard to enforce them.
> 
> People usually love JAX for speed, and one of the first compromises JAX does is ignore any forms of checks whatsoever for the sake of speed.

Maybe a good comment on #918 directly. Should be an easy graph operation to remove, no?

ricardoV94:
> Maybe a good comment on https://github.com/pymc-labs/pymc-marketing/issues/918 directly. Should be an easy graph operation to remove, no?

depends on the constraint, we don't have yet rewrites to understand stuff like [0,1], bounds, we understand positive/negative a bit. And then there is always the concern the values could be `nan`.

That's why PyMC still has the `check_bounds` at the model level

juanitorduz:
Thank you @shivamlalakiya ! Maybe we park this one while we assess the effect in model speed, as pointed out by Reicardo 🙏

Issue #2807, opened 2026-08-03
Budget optimizer: rethink the inert-variable gradient check (report post-hoc rather than probe pre-solve)


PabloRoque:
## Context

`BudgetOptimizer` warns when an optimization variable's gradient is identically zero at the warm start, so a variable the objective cannot see is not silently returned at its seed and reported as optimal. That guard came out of #2621's review, where a discount lever whose event window fell outside the optimization window did exactly that.

The current implementation (see #2802) probes two points — a small step local to `x0` and the furthest reachable bound — and calls a variable inert only when both read a zero gradient.

## The problem with refining this further

The probe is trying to answer *"is this function flat everywhere, or only here?"* from finitely many point evaluations, which no finite probe set can settle: for any set of probe points there is a function flat at exactly those points and responsive elsewhere. The two points cover the two failure modes found so far:

- a **saturating** response, where a distant probe is flatter still and would confirm an "inert" verdict by construction (why a pure midpoint probe was wrong);
- a **dead zone**, where the objective is genuinely flat at `x0` and becomes responsive further out, which a purely local probe cannot escape (why a pure local probe was wrong).

Any third failure mode would motivate a third probe, and so on. This is a heuristic that can be frozen at a chosen level of effort but never completed, so it should not be refined inside a feature PR.

## Proposed alternative: report instead of predict

Rather than predicting flatness before the solve, check the thing users actually care about *after* it:

> did any variable come back exactly at its seed, with a zero gradient there?

That has no probe geometry, no dead zones, and no classification problem — it reports what happened rather than guessing what would happen. It also costs nothing extra: the final gradient is already available from the solver result.

Sketch:

- after `minimize` returns, compare each variable's slice against its `default_x0`/supplied `x0`;
- for any slice that is unchanged to within tolerance, check the gradient at the solution;
- warn naming those variables (per coordinate, as the current warning does).

Trade-off to decide: this warns *after* paying for the optimization rather than before, which is worse for a user who wanted to fail fast, and it cannot distinguish "inert" from "already optimal at the seed" without a probe either — though at the solution that distinction matters much less.

## What should stay regardless

- the **construction-time ancestry check** (`response_variable` must depend on every declared `optimizable_vars` entry): exact, graph-based, no edge cases, catches the common misconfiguration before any compilation;
- the **`check_gradients=False`** escape hatch on `allocate_budget`, for sweeps over many budgets on a large posterior.

## Scope

[...]

PR #2879, opened 2026-08-14
refactor(budget-optimizer): OptimizationVariable abstraction for the decision vector

PabloRoque:
## Description

Replaces #2802, rebuilt from current `main` as a single commit. Same abstraction, **without the inert-variable gradient-check machinery** that made the original hard to follow.

### What this does

The forward map (flat decision vector → model tensors: mask scatter, `channel_scales`, temporal distribution, `cost_per_unit`, adstock zero-padding) and the inverse map (solution → labelled `DataArray`) were written independently, ~150 lines apart, agreeing only by convention. This puts both on one object:

- **`OptimizationVariable`** — the per-variable protocol: `to_model` (forward), `unpack`/`pack` (inverse pair, stated once), `default_x0`, `default_bounds`.
- **`MediaVariable`** — a faithful relocation of the existing media path. No arithmetic changed.
- **`OptimizationVariables`** — owns the flat symbolic input, the contiguous slice layout (tiling `[0, size)` exactly), and the single joint `do()` substitution dict.

`allocate_budget`'s `x0` now also accepts a labelled `DataArray` or `{name: DataArray}` dict, packed through the same layout — warm-starting from a previous solution or a client's current plan was previously impractical for multi-dimensional budgets, since the flat ordering is boolean-mask flattening over transposed dims.

**Scope: media only.** The container can hold several variables, but this PR registers exactly one (`OptimizationVariables([media_variable])`), so behaviour is unchanged. Co-optimizing effect levers alongside media is the follow-up that builds on this.

### Bug fixed along the way

`budgets_to_optimize` was only *transposed*, never *reindexed*, while `cost_per_unit` and `budget_bounds` were already reindexed to model coordinates. A mask whose coordinate order differs from the model's therefore both selected the wrong cells and labelled them with the wrong coordinates — silently, since the result is self-consistent and simply attributes spend to the wrong channel. **This is live in `main` today.** Credit to @daimon-pymclabs, who found the labelling half in the original #2802 review; the selection half turned out to be pre-existing.

Two regression tests cover it: labels must follow the model's coordinates under a shuffled mask, and a partial mask must give the same allocation *and* objective however its coordinates are ordered (the case where the two shifts cancel in the output while the model optimizes the other channel's curve).

### What was dropped versus the original #2802

The zero-gradient probe, its `check_gradients` kwarg, and their tests. That guard tries to answer "is this objective flat everywhere, or only here?" from finitely many point evaluations — undecidable by sampling, and it accreted an edge case per review round without ever affecting results. It's tracked separately in #2807 (with a proposed post-hoc redesign) rather than blocking an abstraction PR.

Net effect: **5 files, +932/−148 in one commit**, against 7 files/+1574/−231 across six commits.

## Verification

- 129 tests green: `test_budget_optimizer.py`, `test_optimization_variables.py`, `test_budget_optimizer_mmm.py`, `test_cost_per_unit.py`.
- Behaviour-neutral by construction; the existing suite passes unchanged apart from one test whose assertion pinned an incidental numpy error message that is now a deliberate shape-validation error.
- New property tests pin the round-trips (`pack(unpack(x)) == x`, `unpack(pack(da)) == da`) over 1-D/2-D shapes with full and partial masks, plus the slice tiling.

## Related

- Replaces #2802 · follow-up: effect levers (#2804) · probe redesign: #2807

## Type of change

- [x] Refactor (behaviour-neutral) + one latent bug fix

----
📚 Documentation preview 📚: https://pymc-marketing--2879.org.readthedocs.build/en/2879/


pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -2059,15 +1984,8 @@ def track_progress(xk):
...
-                self.budgets_to_optimize.values,  # type: ignore
-                dtype=float,
-            )
-            optimal_budgets[self.budgets_to_optimize.values] = result.x  # type: ignore
-
-            optimal_budgets = DataArray(
-                optimal_budgets, dims=self._budget_dims, coords=self._budget_coords
-            )
+            unpacked = self._variables.unpack(result.x)
+            optimal_budgets = unpacked.pop(self.channel_data_var)
```
timbo112711, line 2004: This immediately pops the only key from unpack()'s result. Makes sense with one variable registered, but when the effect-lever work (#2804) lands this will need updating. Would it be worth dropping a `# TODO` here so it's easy to find when that PR comes around?

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -1476,26 +1480,21 @@ def model_post_init(self, context: Any, /) -> None:
...
-            budgets_zeros[bool_mask].set(self._budgets_flat.values),
-            dims=self._budget_dims,
-        )
+        # Align the mask with the model's coordinate order, not just its dim
+        # order: the mask is consumed positionally downstream (scatter into the
+        # model's tensor layout, and the labels the solution is unpacked with),
+        # so a mask ordered differently from the model would otherwise select
+        # the wrong cells and label them with the wrong coordinates.
+        # `cost_per_unit` and `budget_bounds` are reindexed for the same reason.
+        self.budgets_to_optimize = self.budgets_to_optimize.reindex(
```
timbo112711: Nice catch @daimon-pymclabs !

juanitorduz:
Thank you @PabloRoque ! LGTM. Still, I asked a bot for an adversarial review, and these are the findings ;) 

I will let you decide what you wanna tackle in this PR vs opening issues to keep this one scoped.

---

**Verdict.** This is a solid, well-motivated refactor: the new `OptimizationVariable` / `MediaVariable` / `OptimizationVariables` abstraction in `pymc_marketing/mmm/optimization_variables.py` cleanly isolates what the optimizer knows about the decision vector, and the port from the old `BudgetOptimizer` internals was verified faithful in the places that matter (flat-vector ordering is backward compatible for model-ordered inputs, constraint gradients through `_budgets` / `_budgets_flat` stay connected, deleted methods have no remaining callers, no test coverage was lost). However, the review found two behavior regressions introduced by the PR that should be fixed before merge, plus four pre-existing bugs that the refactor surfaces and that these touched code paths are the natural place to fix. Line numbers refer to the PR head.

## Must fix: regressions introduced by this PR

### 1. `budget_distribution_over_period` is not reindexed to model coordinate order, but the mask now is

`pymc_marketing/mmm/budget_optimizer.py:1664-1671` (`_validate_and_process_budget_distribution`)

This PR reindexes `budgets_to_optimize` to the model's coordinate order in `model_post_init` (line 1489), and the comment at line 1488 says `cost_per_unit` and `budget_bounds` are aligned "for the same reason". The third coordinate-sensitive input was missed: `budget_distribution_over_period` is only transposed by dim names (line 1666), then flattened in the user's coordinate order and combined positionally with the now model-ordered boolean mask (line 1671).

Failure scenario: a user passes `budgets_to_optimize` and `budget_distribution_over_period` both with channels ordered `[channel_2, channel_1]` while the model uses `[channel_1, channel_2]`. The mask gets realigned, the distribution does not, so `channel_1` silently receives `channel_2`'s temporal spending profile and vice versa. The optimization completes without error and returns a plausible-looking but wrong allocation. Before this PR both inputs were consumed in the user's order, so a self-consistent shuffled pair still agreed positionally; the PR breaks that.

Suggested fix: reindex the distribution's non-date dims to `self._budget_coords` (with an `isnull` check for missing coords) before the transpose and flatten, exactly mirroring the mask path.

Suggested regression test: build the optimizer twice, once with model-ordered inputs and once with a consistently shuffled coord order for both mask and distribution, and assert the processed distribution tensor (or the resulting allocation) is identical.

### 2. Masks with out-of-model coordinates are now silently truncated

`pymc_marketing/mmm/budget_optimizer.py:1489`

`reindex(self._budget_coords)` drops mask entries whose coordinates are not in the model, without NaN and without error (verified empirically: reindexing a 3-channel mask onto 2 model channels raises nothing). The new `isnull` guard at line 1492 only catches model coords missing from the mask, and the posterior check at line 1477 uses inner-join alignment, so extra labels are invisible to it too. The old code failed loudly at graph build with a boolean-index shape mismatch.

Failure scenario: a user builds `budgets_to_optimize` from a superset channel list (stale config, typo'd channel name, or a channel dropped at fit time) with `True` on a channel the model does not have. Pre-PR: hard error at construction. Post-PR: the cell is silently discarded, the budget is redistributed over the remaining cells, and the user believes the missing channel was considered.

Suggested fix: before reindexing, compare the mask's per-dim coordinate sets against `self._budget_coords` and raise a `ValueError` listing the unknown coordinates.

Suggested regression test: mask covering all model channels plus one extra `True` entry must raise at `BudgetOptimizer` construction.

### 3. `BudgetOptimizationResult.optimized_vars` is never populated

`pymc_marketing/mmm/budget_optimizer.py:1988` (`allocate_budget`)

The PR adds `optimized_vars` to `BudgetOptimizationResult` (line 321) precisely to carry non-media solutions, but `allocate_budget` unpacks the full solution, pops the media entry, and discards the rest; the field stays `{}` unconditionally.

Failure scenario: the day a second `OptimizationVariable` is registered (the #2804 effect-lever work this PR is groundwork for), the optimizer co-optimizes it and then throws its optimal values away; the user cannot recover them short of slicing `result.scipy_result.x` with private slice knowledge, and no error signals the loss.

Suggested fix: after popping the media entry, pass the remainder as `optimized_vars=unpacked`. Testable today with a stub second variable against `OptimizationVariables`, or minimally by asserting the field equals `{}` for the single-variable case via the code path rather than the default.

## Should fix: pre-existing bugs surfaced by the refactor

These are not regressions of this PR (each was verified to exist on `main`), but the PR touches or newly documents each path, so this is the natural moment to fix them, or at least to open tracking issues.

### 4. `channel_scales` is dropped when `budget_distribution_over_period` is active, and the new docstring documents the wrong behavior

`pymc_marketing/mmm/optimization_variables.py:237-253` (`MediaVariable._apply_budget_distribution_over_period`)

The temporal branch builds the distributed budgets from the raw flat `z`; the scaled `budgets` tensor is used only via `zeros_like` for shape and dims, so the division by `channel_scales` performed in `to_model` (line 267) never reaches the tensor substituted into the model. This is exact parity with the old code (which used the raw `self._budgets_flat`), so it is a faithfully preserved pre-existing bug. What is new: the `MediaVariable` docstring now asserts the order "scatter, divide by channel_scales, spread over num_periods", which the temporal branch does not honor.

Failure scenario: `channel_scales != 1` (the normal case via `MMM.budget_optimizer`, which passes max-scaling factors) combined with a temporal distribution substitutes channel data in monetary units instead of scaled units, inflating the modeled response by the scale factor, while the uniform-spread branch is correctly scaled.

Suggested fix: apply the scales division in the temporal branch as well, and make the data flow honest by dropping the misleading `budgets` parameter (the zeros can be built from `self.shape`, `self.dims`, and `num_periods`). If the fix is intentionally deferred, correct the docstring and open a tracking issue.

Suggested regression test: with `channel_scales != 1` and a temporal distribution, evaluate the substituted channel-data tensor and assert it matches the uniform branch's scaling convention.

### 5. `cost_per_unit` NaN guard runs before the reindex

`pymc_marketing/mmm/budget_optimizer.py:1731-1738` (`_validate_and_process_cost_per_unit`)

The `isnull` / positivity check at line 1731 runs before `reindex(budget_coords)` at line 1738, so coordinates present in the model but missing from `cost_per_unit` become NaN after reindexing and flow straight into the graph.

Failure scenario: `cost_per_unit` missing one channel or geo coordinate passes validation, the objective and gradient become NaN, and scipy fails opaquely (or returns garbage with `return_if_fail=True`) with nothing pointing at `cost_per_unit`. This is inconsistent with the strict mask validation the PR adds for exactly this class of input.

Suggested fix: reindex first, then validate (or add a post-reindex `isnull` guard). Test: `cost_per_unit` missing one model coordinate raises an informative error.

### 6. `budget_bounds` DataArray path has no post-reindex NaN guard, and the default-bounds logic is duplicated

`pymc_marketing/mmm/budget_optimizer.py:1873-1915` (`allocate_budget`)

The DataArray branch reindexes to model coords (line 1900) with no NaN check before the mask-select at line 1914, so a bounds array with a missing or typo'd coordinate yields `(nan, nan)` bounds pairs handed to scipy, which treats them as unbounded or fails opaquely depending on the method. Separately, the `budget_bounds=None` branch hand-builds `[0, total_budget]` bounds and always passes them as an override, re-implementing `MediaVariable.default_bounds` and leaving the abstraction's own default-bounds path dead code.

Suggested fix: route bounds construction through the same alignment pipeline as `MediaVariable.pack` (reindex, transpose, NaN check, mask-select), which fixes the NaN hole for free, and let the `None` branch simply call `self._variables.bounds(total_budget)` with no overrides. Test: bounds DataArray missing one model coordinate raises.

### 7. All-False mask crashes with a bare `ZeroDivisionError`

`pymc_marketing/mmm/optimization_variables.py:329` (`MediaVariable.default_x0`)

`total_budget / self.size` has no guard for `size == 0`. An all-False mask (user-supplied, or an auto-detected mask from a degenerate fit where all mean contributions are exactly zero) constructs fine and then crashes in `default_x0` with no indication that the mask is empty.

Suggested fix: raise an informative `ValueError` in `MediaVariable.__init__` when the mask selects zero cells. Test: all-False mask raises at construction with a message naming the mask.

## Non-blocking suggestions

- Build the `scattered()` subgraph once. `model_post_init` (budget_optimizer.py:1534) and `to_model` (optimization_variables.py:242) each build an independent scatter subtree from the same flat input; correctness relies on the two staying textually identical and on PyTensor's merge pass unifying them, and for budget-using utilities the pre-grad rewrite (which does not include the merge pass) backprops through both copies, costing an extra masked gather per scipy iteration. Exposing or caching the node restores the old single-node invariant.
- One `FLAT_DIM` constant. `"budgets_flat"` is defined in three places: `MediaVariable`'s constructor default, `OptimizationVariables`' default, and hardcoded at budget_optimizer.py:1677 for the distribution tensor dims. The container's assign-or-reject `getattr` branch is dead for all shipped variables (MediaVariable always sets a default); a module-level constant used by all three sites would delete roughly ten lines of reconciliation code and survive a rename.
- `channel_scales` hardcodes the dim `('channel',)` (optimization_variables.py:269), the only non-injectable name in the class, and the companion check at budget_optimizer.py:1601 silently skips models without a `channel` coord. Deriving the dims from the mask would match the rest of the design.
- `x0` handling: `np.asarray(x0).astype(dtype)` copies unconditionally (use `copy=False`) and newly coerces string arrays that the old `type.filter` rejected; consider validating the dtype kind.
- `MediaVariable.size` re-sums the mask on every access and is read several times per `allocate_budget`; compute it once in `__init__` (the mask is never mutated). The mask is also bool-cast three times along the construction path.
- Three em-dashes in new docstring prose (optimization_variables.py:17, 20, 146) violate the repository's AGENTS.md no-em-dash rule.

## What was checked and looks good

- Flat-vector ordering is unchanged for model-ordered inputs, so user-supplied flat `x0` and bounds remain backward compatible; `OptimizationVariables.flat` reproduces the old `_budgets_flat` construction exactly.
- Constraint machinery still works: `constraints.py` finds `optimizer._budgets` / `optimizer._budgets_flat`, both assigned in `model_post_init`, and gradients w.r.t. the flat vector stay connected.
- The deleted `_replace_channel_data_by_optimization_variable` and `_apply_budget_distribution_over_period` methods have no remaining callers or subclass overrides anywhere in the repo.
- Pydantic wraps the new `model_post_init` `ValueError` into `ValidationError`, which subclasses `ValueError`, so existing `except ValueError` callers survive; coord-less positional masks still work through the reindex.
- Docs autosummary picks up the new classes in `__all__` automatically; no rst changes needed. New exports are placed correctly in `pymc_marketing/mmm/__init__.py`.
- Tests: no coverage was deleted; the mocked-minimize test's expected error change stays a post-minimize `ValueError`; new tests use fixtures and `parametrize` and mirror the source layout.

PabloRoque:
Thanks @juanitorduz, and thanks to the bot: findings 1, 2 and 4 were genuinely sharp. All seven are fixed in `997d5ff3`, each with a regression test I verified fails without its fix.

Rather than seven separate guards, the four coordinate findings (1, 2, 5, 6) turned out to be one operation done four different ways, so they became one helper:

```python
align_to_model_coords(da, coords, *, label)
```

`reindex` is unsafe in both directions, it drops labels the model does not have and fills missing ones with NaN, so the helper rejects both. `budgets_to_optimize`, `budget_distribution_over_period`, `cost_per_unit` and `budget_bounds` now all enter the model's coordinate frame through it.

**1. Distribution not reindexed.** Confirmed, and confirmed as a regression of this PR: with a self-consistent shuffled pair, model column `channel_1` received `channel_2`'s back-loaded profile. Before the mask fix both inputs were consumed in user order and agreed positionally; realigning only one broke that. Exactly the bug class this PR exists to fix, so it had to be fixed here.

**2. Extra mask coordinates silently truncated.** Confirmed, also a regression: the old code raised on the boolean-index shape mismatch. Now rejected by name.

**3. `optimized_vars` never populated.** Fixed, and it also answers @timbo112711's question about a `# TODO`: the code is now the general form, media is popped as the named output and everything else flows through, so #2804 needs no change at that line.

**4. `channel_scales` dropped in the temporal branch.** Real, and pre-existing (I had preserved it deliberately during the port, and my docstring then asserted an ordering the branch did not honour, which was my error). The fix is structural rather than another branch: the scales division now happens *after* the uniform/temporal split. Scaling is elementwise per channel and spreading elementwise per period, so they commute, which makes the uniform path provably unchanged and the temporal path correct. Verified precisely: with the old ordering the temporal test fails and the uniform one passes.

**5, 6, 7.** `cost_per_unit` now validates after alignment; `budget_bounds` goes through the same alignment, and the `None` branch calls `self._variables.bounds(total_budget)` so `MediaVariable.default_bounds` is no longer dead code; an empty mask raises at construction instead of dividing by zero later.

**`FLAT_DIM`** is now a module constant replacing three independent spellings of the flat dimension's name.

Two corrections for the record, since they may matter if the bot is used again:

- **The em-dash rule does not exist.** I grepped for it: `AGENTS.md` has no such rule, and the file is not even tracked in the repository (it is a local untracked file). I have removed the em-dashes anyway since it is a reasonable convention, but the finding as stated was fabricated.
- **The `flat_dim` "ten lines of reconciliation code"** describes the previous branch, not this one. There is no reconciliation code here, just three independent defaults, which `FLAT_DIM` now collapses.

I have deliberately **not** taken one suggestion: building the `scattered()` subgraph once. The duplicate subtree is real, but every clean fix means caching symbolic nodes or restructuring graph construction for values that already agree, and the extra masked gather is on a tiny tensor. Noting it as a known blemish rather than pretending it is not there.

138 tests green across `test_budget_optimizer.py`, `test_optimization_variables.py`, `test_budget_optimizer_mmm.py` and `test_cost_per_unit.py`.

pymc_marketing/mmm/budget_optimizer.py
PabloRoque: Good catch on the spot, though I went the other way on the remedy: rather than a `# TODO`, `997d5ff3` makes the line not need updating.

The bot review found the same place from a different angle: `optimized_vars` was declared on the result but never populated, so the moment a second variable is registered its solution would be computed and then thrown away. Passing `optimized_vars=unpacked` fixes that and makes this the general form already, media is popped as the named output and everything else flows through, so #2804 changes nothing here.

What your comment did correctly point at is that the comment above the line was stale: it still said "Fill zeros, then place the solution in masked positions", describing scatter logic that now lives in `MediaVariable.unpack`. Replaced with one that explains the pattern instead.

juanitorduz (approved):
Thank you @PabloRoque !

PR #2884, opened 2026-08-17
feat(mmm): co-optimize named pm.Data levers alongside the media budgets

PabloRoque:
## Description

**Lets the budget optimizer solve for things that aren't media budgets, jointly with the media budgets.**

Today the decision vector is exclusively the media grid, so the optimizer can only answer "how do I split this budget across channels and periods." This PR lets you name any one-dimensional `pm.Data` node in the model, give it bounds, and have scipy solve for it and the spend allocation together:

```python
    ...,
    optimizable_vars={"promo_data": [(0.05, 0.45)]},   # native units
    response_variable="total_response_original_scale",  # must reach the lever
)
result = optimizer.allocate_budget(total_budget=100_000)
result.optimized_vars["promo_data"]  # labelled optimum
```

**Why jointly, rather than one and then the other.** The optimal discount depth depends on how much media is pushing the promoted product, and the optimal media split depends on how deep the discount is. Both segments live in one flat vector and go through a single `pymc.do()` substitution, so the Jacobian covers media and levers together and the cross-partials between them are real derivatives rather than an artifact of alternating two separate solves. That is the technical point of the PR; the rest is bookkeeping.

Discounts are the motivating application, but nothing here knows what a promotion is. The optimizer sees "a one-dim node with bounds," which is what keeps the marketing concepts in the follow-up and this PR reviewable as pure optimizer mechanics. An AST guard test enforces that no marketing module is imported.

### What it costs

`LeverVariable` implements the existing protocol in ~90 lines: the forward map is the identity up to relabelling the flat dim, the warm start is the node's current model value clipped to its bounds, and `default_bounds` are the declared native-unit bounds. `LeverVariable.from_model` reads a node's dim, coords and current value straight off the model and rejects anything that cannot serve as a lever, so the optimizer keeps a comprehension rather than an introspection loop. Everything else comes from the abstraction already in `main`, and `result.optimized_vars` was already populated, so the result path needed no change.

The equivalent in #2621 was roughly 400 lines of optimizer diff. This is the abstraction from #2879 paying rent.

Levers are optimized in their own units and **do not** enter the default budget-sum constraint, which sums the media tensor alone: a discount depth is not money drawn from a shared pool.

### Two guards, both exact

- **Construction-time reachability.** A lever the response variable cannot reach has an identically zero gradient, so the solver would return its warm start and report it as optimal. Graph ancestry settles this before anything is compiled. It fires in practice: writing the end-to-end test against the stock media-only objective was rejected, which is how it should behave, so the test effect registers an objective seeing both blocks.
- **Bounds and dim validation** in `LeverVariable.__init__`, so a wrong-length bounds list, a date-varying node or a multi-dim node fails at construction with a message naming the variable, rather than as a shape error from scipy at solve time.

Deliberately *not* included: the inert-variable probe (tracked in #2807) and the `f(x0)` finiteness check. A non-finite objective already fails loudly through scipy, so that check only improved a message, and the case it was written for is caught earlier at effect build time. The "levers scored against the media-contribution objective" warning is also held back: it pattern-matches an MMM variable name, so it belongs at the adaptor layer, in the follow-up.

### `optimization_variables` is now public

Constraining anything other than the media budgets means reaching a variable's segment of the flat vector, which was previously only possible through private attributes (raised by @daimon-pymclabs during #2621: levers optimizable but not constrainable through public API). The accessor closes that, with a docstring example and a test that caps a lever through a custom `Constraint` using only public API while the budget constraint still binds.

The container is exposed for reading, not for injection: it is built inside `model_post_init` and reached through a getter-only property, so "media comes first" stays structural rather than conventional.

### Budget participation seam (added after a spike)

The default sum constraint totalled the media tensor alone, so a second *monetary* variable would have been optimized for free. Variables now declare what they spend via `budget_contribution`, which defaults to `None`:

```python
def budget_contribution(self, z) -> XTensorVariable | None:
    return None            # a lever is not money
    # MediaVariable returns self.scattered(z), the tensor it already builds
```

Behaviour is unchanged, media is still the only contributor, but this is the piece reach-and-frequency budgets (#1968) and funnel spend both need, and including it means neither has to change the protocol later. A spike measured the cost at 30 lines across two files with no regressions, which is why it is folded in here rather than deferred.

A method rather than a flag, because the constraint cannot reconstruct a variable's monetary amount generically: media applies `channel_scales`, R&F would apply `cost_per_unit`. Only the variable knows what its slice costs. The contract follows the media convention of per-period rates rather than horizon totals, since the constraint sums contributions directly.

Tested with a stand-in spend variable: two monetary variables total together, and a lever stays out.

### Also closes a coverage gap from #2879

No test there built `OptimizationVariables` with two variables, so the container's `variable_slice` `isel` branch, which every additional variable depends on, never ran in CI. A lever is the first genuine second variable: the slice layout, the `isel` path, and the concatenation of `x0` and bounds across variables are all now exercised.

## Verification

160 tests green across `test_budget_optimizer.py`, `test_optimization_variables.py`, `test_budget_optimizer_mmm.py`, `test_cost_per_unit.py`. Lint and mypy clean.

## Scope

This generalizes the *decision vector*. It does not change what the objective counts, so mediated effects (funnel) are unaffected here: those need the response variable generalized, which lands with the effect protocol in the follow-up.

## Related

- Follow-up: `OptimizableMuEffect` plus the `MMM`/`link` wiring (second half of the closed #2804), then `DiscountedEventEffect` with its notebook
- Builds on #2879 · supersedes half of #2804 · probe redesign tracked in #2807
- Also unblocks #1968 (frequency/reach), though note R&F budgets are *spend* and so want a variable that participates in the sum constraint, unlike a lever

## Type of change

- [x] New feature (backward compatible: no `optimizable_vars` means behaviour is unchanged)

----
📚 Documentation preview 📚: https://pymc-marketing--2884.org.readthedocs.build/en/2884/

### DIFF
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -1551,30 +1551,12 @@ class BudgetOptimizer(BaseModel):
         # Optimizable vars become lever variables appended after media. They
         # are optimized in their own units and stay out of the budget-sum
         # constraint, which sums the media tensor alone.
-        levers = []
-        for lever_name, lever_bounds in self.optimizable_vars.items():
-            if lever_name not in self.model.named_vars_to_dims:
-                raise ValueError(
-                    f"optimizable_vars entry '{lever_name}' is not a variable "
-                    "with named dims in the model."
-                )
-            lever_dims = list(self.model.named_vars_to_dims[lever_name])
-            if len(lever_dims) != 1 or lever_dims[0] == self.date_dim:
-                raise ValueError(
-                    f"optimizable_vars entry '{lever_name}' must have exactly "
-                    f"one dim, and not the {self.date_dim!r} dim; got "
-                    f"{tuple(lever_dims)}. Date-varying optimizable variables "
-                    "are not supported."
-                )
-            levers.append(
-                LeverVariable(
-                    name=lever_name,
-                    dim=lever_dims[0],
-                    coords=list(self.model.coords[lever_dims[0]]),
-                    bounds=lever_bounds,
-                    initial_value=self.model[lever_name].get_value(),
-                )
+        levers = [
+            LeverVariable.from_model(
+                self.model, lever_name, lever_bounds, date_dim=self.date_dim
             )
+            for lever_name, lever_bounds in self.optimizable_vars.items()
+        ]
 
         self._variables = OptimizationVariables([media_variable, *levers])
         self._budgets_flat = self._variables.flat
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -251,6 +251,7 @@ from pymc.model.transform.optimization import freeze_dims_and_data
 from pytensor import function
 from pytensor.compile.sharedvalue import SharedVariable, shared
 from pytensor.graph import rewrite_graph
+from pytensor.graph.traversal import ancestors
 from pytensor.xtensor.type import XTensorVariable
 from scipy.optimize import OptimizeResult, minimize
 from xarray import DataArray, DataTree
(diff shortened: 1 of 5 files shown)


juanitorduz (approved):
Thank you @PabloRoque 🙌 !

PR #2885, opened 2026-08-17
Docs: build funnel MMMs from YAML

drbenvincent:
## Summary
- Add `extra_vars` to `MMMYamlConfig` and fold funnel mediator columns into `to_mmm_dataset` / `build_mmm_from_yaml`
- Accept `xarray.Dataset` inputs in the YAML builder (needed for `DataVarMuEffect` effects)
- Add `mmm_funnel_from_yml.ipynb`: one naive baseline, then national and geo funnel topologies via DAG → effect class → YAML → PyMC graph
- Rename notebook-local effects to `MeasuredMediatorEffect` and `BudgetDualIndicatorEffect` (topology names per #2892)

## Test plan
- [x] `pytest tests/mmm/builders/test_yaml.py tests/mmm/builders/test_schema.py tests/mmm/test_data_conversion.py`
- [x] Mock-run `mmm_funnel_from_yml.ipynb`
- [x] Gallery sync (`generate_gallery.py`; notebook embeds PNG outputs for thumbnail extraction at docs build)

Partially addresses #2892. More funnel topologies and notebook prose iterations planned in follow-up PRs.

----
📚 Documentation preview 📚: https://pymc-marketing--2885.org.readthedocs.build/en/2885/


drbenvincent:
Tagging @carlosagostini so you can see what we're doing

docs/source/notebooks/mmm/funnel_intro.yml
```diff
@@ -0,0 +1,34 @@
...
+    saturation:
+      class: pymc_marketing.mmm.LogisticSaturation
+      kwargs: {}
+    sampler_config:
+      nuts_sampler: nutpie
+
+extra_vars: [lower_spend, lower_control]
+
+effects:
+  - class: FunnelEffect
```
daimon-pymclabs: **`class: FunnelEffect` is a bare name and will not resolve standalone.**

**Why this matters:** `REGISTRY` in `pymc_marketing/mmm/builders/factories.py` is empty by default, and `locate()` raises `ValueError: Cannot locate 'FunnelEffect'` for a name it cannot import. This YAML only works because the notebook does `REGISTRY["FunnelEffect"] = FunnelEffect` on a module global before calling the builder. Anyone who picks up this config from `data/config_files/` (which is exactly what that directory invites) hits a hard error with no hint that a runtime registration is required. Every other file in that directory uses fully-qualified dotted paths, so this is also an inconsistency in the shipped examples.

**Possible fixes, in rough order of preference:**
1. Promote `FunnelEffect` (and `GeoFunnelEffect`) into the package proper, then use a dotted path here, e.g. `class: pymc_marketing.mmm.effects.FunnelEffect`. Best option if these effects are meant to be reusable.
2. Move these two YAMLs next to the notebook (e.g. under `docs/source/notebooks/mmm/`) so it is unambiguous that they are notebook-scoped and depend on classes defined in that notebook.
3. At minimum, add a header comment in the file stating that the class must be registered in `REGISTRY` first, and show the one-liner.

Same issue in `data/config_files/funnel_advanced.yml` with `GeoFunnelEffect`.

docs/source/notebooks/mmm/funnel_advanced.yml
```diff
@@ -0,0 +1,123 @@
...
+          sigma: 0.5
+          dims: geo
+        dims: [date, geo]
+    sampler_config:
+      nuts_sampler: nutpie
+
+extra_vars: [lower_spend, search_volume, category_demand, lf_budget]
+
+effects:
+  - class: GeoFunnelEffect
```
daimon-pymclabs: Same as the `funnel_intro.yml` comment: `GeoFunnelEffect` is a bare name, `REGISTRY` is empty by default, so `locate()` raises `ValueError: Cannot locate 'GeoFunnelEffect'` unless the notebook has already registered it. Either promote the class and use a dotted path, or move this config next to the notebook that defines the class.

pymc_marketing/mmm/builders/yaml.py
```diff
@@ -144,6 +230,21 @@ def build_mmm_from_yaml(
...
             )
+    elif isinstance(X, xr.Dataset) and date_column and "date" not in X.coords:
+        raise ValueError(
+            f"Date column '{date_column}' specified in config but Dataset "
+            f"input is missing a 'date' coordinate."
+        )
+
+    extra_vars = cfg.extra_vars or []
+    if extra_vars and isinstance(X, pd.DataFrame):
+        X = as_model_dataset(
```
daimon-pymclabs: **Converting `X` here bypasses the flat-`y` reshape, which breaks panel models.**

**Why this matters:** `to_mmm_dataset` only calls `_reshape_flat_target` when `X` is a `DataFrame` (`pymc_marketing/mmm/data_conversion.py:135`). Once this line turns `X` into a `Dataset`, that branch is never taken. A `y` with a plain `RangeIndex` then stays flat: a length-`n_rows` array gets assigned onto a `("date",)` dim whose coordinate has only `n_dates` entries.

Concretely, a geo config plus a flat `y` works today, but adding `extra_vars` to the same config fails with:

```
xarray.structure.alignment.AlignmentError: cannot reindex or align along dimension 'date'
because of conflicting dimension sizes: {8, 4}
```

The new geo test does not catch this because it passes a MultiIndex `y`, which is already in panel shape.

**Possible fix:** reshape `y` at the same point `X` is converted, so the two stay in step. Something like calling `_reshape_flat_target` (or the same logic) on `y` inside this branch when `y` has a non-panel index and `dims` is non-empty. Worth adding a regression test that mirrors `test_build_mmm_from_yaml_extra_vars_geo_dims` but with a flat `RangeIndex` `y`.

pymc_marketing/mmm/builders/yaml.py
```diff
@@ -144,6 +230,21 @@ def build_mmm_from_yaml(
...
+            X,
+            date_column=date_column,
+            channel_columns=model_spec["kwargs"].get("channel_columns", []),
```
daimon-pymclabs: **`control_columns` plus `extra_vars` is currently impossible, and the error message misdirects.**

**Why this matters:** `as_model_dataset` emits `media` plus the extra vars, and never a `_control` variable. So as soon as a config declares both `control_columns` and `extra_vars`, `_validate_mmm_structure` raises something like:

```
control_columns=[...] but dataset is missing '_control' ... Use '_control' instead
```

That advice is unactionable for a user who only wrote a DataFrame and a YAML: they never named a variable `_control` and have no obvious way to create one. This is not a hypothetical combination, it is the natural next step from combining the controls in `funnel_advanced_naive_bplus.yml` with the `extra_vars` in `funnel_advanced.yml`.

**Possible fixes:**
1. Preferred: have `as_model_dataset` also take `control_columns` and emit `_control` the same way the canonical DataFrame path does, so the two paths produce equivalent datasets.
2. If that is out of scope for this PR, fail fast and explicitly: raise at config-parse time with a message like "`extra_vars` is not currently supported together with `control_columns`", so the user gets a clear signal instead of a confusing downstream validation error.

daimon-pymclabs (commented):
Nice piece of work overall: the `extra_vars` schema field and the Dataset-input plumbing are the right shape, and the notebook is executed clean with no error outputs. I have left inline comments with the reasoning and a suggested fix for each. Grouped by priority:

**Blocking-ish (correctness / reproducibility)**
1. The two headline configs (`funnel_intro.yml`, `funnel_advanced.yml`) cannot be built standalone: they use bare class names that only resolve because the notebook mutates `REGISTRY` at runtime.
2. `extra_vars` breaks a flat (RangeIndex) `y` when the model has panel dims, with an `AlignmentError`.
3. `control_columns` combined with `extra_vars` is currently impossible, and the error message misdirects the user.
4. The geo branch of `as_model_dataset` skips the `fillna(0)` the canonical path applies, so ragged panels get NaNs into `pmd.Data`.

**Smaller**
5. The flat branch hand-rolls what `_pandas_to_xarray_dataarray` already does, and diverges from it in three ways.
6. `extra_vars` is silently ignored when `X` is already a Dataset.
7. `tmp_path` is unused in one of the new tests.

**Docs build**
8. All 20 image outputs in the new notebook are `image/svg+xml`, but `scripts/generate_gallery.py` only extracts `image/png`. The new gallery card will fall back to the placeholder thumbnail while the neighbouring funnel cards show real plots. A single matplotlib figure anywhere in the notebook is enough to give the extractor a PNG to pick up. (Not inline-commentable since the notebook is one big added blob.)

Also, the mock-run and gallery-sync boxes in the test plan are still unchecked, worth confirming both before merge.

pymc_marketing/mmm/builders/yaml.py
```diff
@@ -144,6 +230,21 @@ def build_mmm_from_yaml(
...
                 f"in either X or y data."
             )
+    elif isinstance(X, xr.Dataset) and date_column and "date" not in X.coords:
+        raise ValueError(
+            f"Date column '{date_column}' specified in config but Dataset "
+            f"input is missing a 'date' coordinate."
+        )
+
+    extra_vars = cfg.extra_vars or []
+    if extra_vars and isinstance(X, pd.DataFrame):
```
daimon-pymclabs, line 181: **`extra_vars` is silently ignored when `X` is already a `Dataset`.**

**Why this matters:** the `isinstance(X, pd.DataFrame)` guard means that for Dataset input, the declared `extra_vars` are never checked against `X.data_vars`. A typo in the YAML (`serch_volume` instead of `search_volume`) therefore produces no error here and instead surfaces much later, deep inside the effect that tries to read the variable, where the traceback no longer points at the config.

**Possible fix:** in an `else` branch, validate that every name in `extra_vars` is present in `X.data_vars` and raise the same `KeyError` style message `as_model_dataset` already produces, listing what is present. Cheap to add and keeps both input paths behaving the same.

pymc_marketing/mmm/builders/yaml.py
```diff
@@ -23,9 +23,92 @@
...
+        channel_ds = _pandas_to_xarray_dataarray(
+            frame, date_column, dims, channel_columns, "channel"
+        )
+        media = channel_ds["_channel"].rename("media")
+        data: dict[str, xr.DataArray] = {"media": media}
+        for name in extra:
+            pivot = frame[[date_column, *dim_cols, name]].drop_duplicates()
+            pivot = pivot.rename(columns={date_column: "date"})
+            data[name] = pivot.set_index(["date", *dim_cols])[name].to_xarray()
+        return xr.Dataset(data)
```
daimon-pymclabs: **The geo branch skips the `fillna(0)` that the canonical path applies.**

**Why this matters:** the DataFrame path in `data_conversion` finishes with `xr.merge(datasets).fillna(0)`. Here the merged Dataset is returned as-is, so a ragged panel (a geo missing some dates, or a date missing some geos) propagates NaNs straight into `pmd.Data`. The result is that the same input data behaves differently depending on whether `extra_vars` is set, which is a surprising and hard-to-debug divergence: with `extra_vars` you get NaNs in the model, without it you get zeros.

**Possible fix:** `return xr.Dataset(data).fillna(0)`, matching the canonical path. If a deliberate difference is intended (e.g. NaNs are meaningful for some extra vars), that should be documented in the docstring and ideally made an explicit parameter rather than an accident of which branch ran.

pymc_marketing/mmm/builders/yaml.py
```diff
@@ -23,9 +23,92 @@
...
+
+    dates = frame[date_column].to_numpy()
```
daimon-pymclabs: **The flat branch hand-rolls `_pandas_to_xarray_dataarray` and diverges from it in three ways.**

**Why this matters:** `_pandas_to_xarray_dataarray(frame, date_column, (), channel_columns, "channel")` already builds exactly this array, and it is what the geo branch above uses. Because this branch reimplements it, the two branches now differ in ways nobody chose deliberately:

- **dedupe**: the geo branch drops duplicate `(date, dims)` rows, this one does not
- **sort order**: the geo branch sorts via `set_index().to_xarray()`, this one preserves the frame's row order, so an unsorted input frame yields an unsorted `date` coord
- **float coercion**: applied here via `to_numpy(dtype=float)`, but not identically in the geo path

**Possible fix:** collapse both branches onto the shared helper with `dims=()` for the flat case, and build the extra vars through the same mechanism. That removes all three silent inconsistencies at once and shrinks the function.

tests/mmm/builders/test_yaml.py
```diff
@@ -636,3 +637,89 @@ def test_original_scale_vars_ordering_does_not_matter(X_data, y_data):
 
     obs_names = [rv.name for rv in model.model.observed_RVs]
     assert any("lift" in name for name in obs_names)
+
+
+def test_build_mmm_from_yaml_extra_vars_on_dataset(tmp_path):
```
daimon-pymclabs: `tmp_path` is requested but never used in this test: the config is read from the checked-in `tests/mmm/builders/config_files/` fixture and nothing is written to disk. Dropping the fixture argument keeps the signature honest about what the test touches.

docs/source/notebooks/mmm/funnel_intro.yml
drbenvincent: Went with option 2 for this PoC: `funnel_intro.yml` now lives beside the notebook at `docs/source/notebooks/mmm/funnel_intro.yml`, with a header comment that it requires `FunnelEffect` from `mmm_funnel_from_yml`. Promoting `FunnelEffect` into the package is intentionally deferred (the notebook summary already flags that as the next step); `REGISTRY` registration here is demo plumbing, not a supported public API. Naive intro configs remain in `data/config_files/` with only stock dotted class paths.

docs/source/notebooks/mmm/funnel_advanced.yml
drbenvincent: Same resolution as the intro thread: `funnel_advanced.yml` moved to `docs/source/notebooks/mmm/funnel_advanced.yml` with a header noting it needs `GeoFunnelEffect` from this notebook. The notebook loads it via `NOTEBOOK_DIR` after registering the effect on `REGISTRY`.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Fixed by removing `as_model_dataset` and routing DataFrame + `extra_vars` through `to_mmm_dataset(X, y, ..., extra_vars=...)` before `build_model`, then passing `y=None` so `_target` is already panel-shaped and `_reshape_flat_target` runs while `X` is still a DataFrame. Regression: `test_build_mmm_from_yaml_extra_vars_geo_flat_y` (flat `RangeIndex` `y`, geo dims).

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Implemented option 1 on the canonical path: `to_mmm_dataset` now accepts `extra_vars` alongside `control_columns`, so both `_control` and named extra variables are emitted from the same `xr.merge(...).fillna(0)` pipeline. Regression: `test_build_mmm_from_yaml_extra_vars_with_control_columns` + `extra_vars_controls_model.yml`.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: When `X` is already a `Dataset`, `build_mmm_from_yaml` now calls `_validate_extra_vars` and raises the same `KeyError` style message (declared vs present `data_vars`). Regression: `test_build_mmm_from_yaml_extra_vars_dataset_typo_raises`.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: `as_model_dataset` is deleted; extra vars now merge through `to_mmm_dataset`, which always finishes with `xr.merge(datasets).fillna(0)` like the channel/control path. Regression: `TestToMmmDatasetExtraVars.test_extra_vars_panel_fillna_ragged_panel`.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Collapsed by deleting `as_model_dataset` entirely. DataFrame conversion (flat and panel) now goes through `_pandas_to_xarray_dataarray` for channels/controls and `_pandas_columns_to_dataarrays` for extra vars inside `to_mmm_dataset` — one code path, same dedupe/sort/fillna behaviour.

tests/mmm/builders/test_yaml.py
drbenvincent: Removed the unused `tmp_path` fixture from `test_build_mmm_from_yaml_extra_vars_on_dataset`.

drbenvincent:
Thanks for the detailed review — pushed `9e8a5a90` addressing all inline threads plus the gallery note.

**YAML / funnel effects (threads 1–2)**  
Deferred promoting `FunnelEffect` / `GeoFunnelEffect` into the package (this PR is a YAML PoC; promotion is the natural follow-up). Instead, `funnel_intro.yml` and `funnel_advanced.yml` moved to `docs/source/notebooks/mmm/` beside `mmm_funnel_from_yml`, with header comments and notebook callout clarifying they depend on notebook-local `REGISTRY` registration. Naive funnel configs stay in `data/config_files/` as standalone dotted-path examples.

**`extra_vars` plumbing (threads 3–7)**  
Removed `as_model_dataset` and folded `extra_vars` into `to_mmm_dataset` — the sole normalisation seam. `build_mmm_from_yaml` calls it for DataFrame input (with `y` so flat panel targets reshape correctly, then `y=None` into `build_model`) and validates `extra_vars` on Dataset input. Channels, controls, extra vars, and `fillna(0)` now share one path. New regressions cover flat-`y` geo panels, controls + extra vars, Dataset typos, and ragged-panel fillna.

**Test nit (thread 8)**  
Dropped unused `tmp_path` from `test_build_mmm_from_yaml_extra_vars_on_dataset`. Also excluded `funnel_*` from the generic parametrized YAML test — those configs need custom data, not `X_data`/`y_data`.

**Gallery (review summary item 8)**  
`show_causal_dag` now displays Graphviz output as PNG so `generate_gallery.py` can extract a real thumbnail. Notebook re-executed; gallery script run locally.

`pytest tests/mmm/builders/test_schema.py tests/mmm/builders/test_yaml.py` and `TestToMmmDatasetExtraVars` all green locally.

pymc_marketing/mmm/data_conversion.py
```diff
@@ -358,3 +383,27 @@ def _pandas_to_xarray_dataarray(
...
+        raise KeyError(
+            f"Columns {missing} are declared but absent from the data. "
+            f"Present: {sorted(map(str, data.columns))}."
+        )
+
+    dim_cols = [dim for dim in dims if dim in data.columns]
+    index_cols = ["date", *dim_cols]
+    result: dict[str, xr.DataArray] = {}
+    for name in column_names:
+        pivot = data[[date_column, *dim_cols, name]].drop_duplicates()
```
daimon-pymclabs: **`drop_duplicates()` here dedupes on the full row, not on the index keys, which breaks alignment.**

**Why this matters:** the two sibling paths both dedupe on keys and keep the first row:

- `_process_dataframe` (channel and control): `drop_duplicates(subset=[date_column, *valid_dims, metric_coordinate_name])`
- `_reshape_flat_target` (target): `drop_duplicates(subset=index_cols, keep="first")`, and its docstring explicitly documents the supporting case, "when *X* contains extra columns that are not included in *dims* (e.g. a ``market`` column when only ``date`` is used), duplicate dates may appear"

This line dedupes on `[date_column, *dim_cols, name]` including the value column. So in exactly the scenario `_reshape_flat_target` documents as supported, rows that share a key but differ in value are all kept, `set_index` gets a non-unique index, and the resulting DataArray has a duplicated `date` coordinate. The subsequent `xr.merge(...)` in `_to_mmm_xarray` then fails.

**Reproduced end to end** on this branch, identical data and identical config, the only difference being `extra_vars`:

```python
df = pd.DataFrame({
    "date": pd.to_datetime(["2025-01-01", "2025-01-01", "2025-01-08", "2025-01-08"]),
    "market": ["us", "uk", "us", "uk"],       # not in dims
    "upper_spend": [1.0, 2.0, 3.0, 4.0],
    "lower_spend": [10.0, 20.0, 30.0, 40.0],  # varies within a date
})
y = pd.Series([5.0, 6.0, 7.0, 8.0])           # flat RangeIndex

# A: without extra_vars -> ok, sizes {'date': 2, 'channel': 1}
to_mmm_dataset(df, y, date_column="date", dims=(), channel_columns=["upper_spend"])

# B: with extra_vars -> ValueError: cannot reindex or align along dimension 'date'
#    because the (pandas) index has duplicate values
to_mmm_dataset(df, y, date_column="date", dims=(),
               channel_columns=["upper_spend"], extra_vars=["lower_spend"])
```

This is the same class of problem as the flat-`y` misalignment from the last round, and the same "same data behaves differently with and without `extra_vars`" smell. The new geo test does not catch it because a proper `(date, geo)` panel has unique keys by construction.

**Suggested fix**, matching the sibling paths:

```suggestion
        pivot = data[[date_column, *dim_cols, name]].drop_duplicates(
            subset=[date_column, *dim_cols], keep="first"
        )
```

(the `subset` has to name `date_column`, since the rename to `"date"` happens on the next line). I verified this makes case B above return `{'date': 2, 'channel': 1}` with `lower_spend = [10., 30.]`, and leaves the proper geo panel unchanged at `{'date': 2, 'geo': 2, 'channel': 1}`.

Worth a regression test with a long frame whose extra var varies within a date while `dims=()`.

daimon-pymclabs (commented):
Re-reviewed at `efc4e89`. This is a much better shape than the first pass: deleting `as_model_dataset` and teaching the canonical `to_mmm_dataset` about `extra_vars` collapses five of my seven points into one code path, rather than fixing them one at a time. Nice.

**Resolved, verified against the branch:**

1. **Bare class names** — `funnel_intro.yml` and `funnel_advanced.yml` moved to `docs/source/notebooks/mmm/`, next to the notebook that defines and registers the effects, each with a header comment naming that requirement. The notebook now cleanly separates `CONFIG_DIR` (the reusable, dotted-path configs) from `NOTEBOOK_DIR` (the two notebook-scoped ones). The `funnel_*_naive_*.yml` files left in `data/config_files/` all use fully-qualified paths, so they stand alone correctly.
2. **Flat `y` with panel dims** — `y` is now passed into `to_mmm_dataset` while `X` is still a DataFrame, so `_reshape_flat_target` fires as intended, and `y` is set to `None` afterwards (`build_model` explicitly supports that when `X` carries `_target`). `test_build_mmm_from_yaml_extra_vars_geo_flat_y` covers it.
3. **`control_columns` + `extra_vars`** — `control_columns` is now threaded through, so `_control` is emitted. Covered by `test_build_mmm_from_yaml_extra_vars_with_control_columns`.
4. **Missing `fillna(0)`** — moot, the extra vars are merged inside the canonical path that already does `xr.merge(datasets).fillna(0)`.
5. **Hand-rolled flat branch** — gone with `as_model_dataset`.
6. **Unvalidated `extra_vars` on Dataset input** — `_validate_extra_vars` added and called for both input types. Covered by `test_build_mmm_from_yaml_extra_vars_dataset_typo_raises`.
7. **Unused `tmp_path`** — dropped.
8. **SVG-only notebook outputs** — the re-executed notebook now has 11 `image/png` outputs alongside 9 `image/svg+xml`, and no error outputs, so `generate_gallery.py` will find a real thumbnail.

**One new issue**, introduced by the move, left inline on `_pandas_columns_to_dataarrays`: it dedupes on the full row instead of on the index keys, which diverges from both of its sibling paths and makes `extra_vars` fail on data that works without it. I reproduced it end to end.

Minor: `yaml.py` reaches across for the private `_validate_extra_vars`. Since `to_mmm_dataset` already calls it internally, threading `extra_vars` through would keep the private helper private, though it is not worth churn if you prefer the explicit call.

The mock-run and gallery-sync boxes in the test plan are still unchecked.

docs/source/notebooks/mmm/mmm_funnel_from_yml.ipynb
```diff
@@ -0,0 +1,5556 @@
+{
```
juanitorduz, line 1: <p>Give an introduction on why we need these functions :) and maybe how this relates with <a href="https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_build_from_yml_example.html" target="_blank">https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_build_from_yml_example.html</a></p><p> </p><p>Also add </p><p> </p>

<pre class="ql-syntax" spellcheck="false">%config InlineBackend.figure_format = "retina"
</pre><p> </p><p>for better image quality</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2885/discussion/#comment-3798669672'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='1'/>

juanitorduz, line 1: <p>again, do a more gentle introduction to these funnel effects (not too extensive)</p><p> </p><p>Maybe split the funnel effect and geofunneleffect and explain their similarities and differences?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2885/discussion/#comment-3798669677'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='3'/>

juanitorduz:
Thanks @drbenvincent ! There is a small open issue from the Daimon review, and I left some small comments on the notebook. Once we address those, we can merge this one :)

daimon-pymclabs:
@juanitorduz to confirm which item is still open from my review: it's the `drop_duplicates()` on `pymc_marketing/mmm/data_conversion.py` (currently line 406, in `_pandas_columns_to_dataarrays`). Everything else from the first round is addressed by 9e8a5a90.

On the branch as of 0b4bbe8 it still reads:

```python
pivot = data[[date_column, *dim_cols, name]].drop_duplicates()
```

which dedupes on the full row including the value column, unlike the two sibling paths (`_process_dataframe` and `_reshape_flat_target`, which both dedupe on key columns with `keep="first"`). With `dims=()` and a long frame whose extra var varies within a date, `set_index` then gets a non-unique index and the downstream `xr.merge` raises `cannot reindex or align along dimension 'date'`. Repro and the one-line fix are in the inline thread.

@drbenvincent that plus a regression test (long frame, `dims=()`, extra var varying within a date) should be the last blocker on my side. Juan's notebook prose comments are separate.

drbenvincent:
Thanks for review comments so far. I am simplifying the structure... only showing ONE naive/flat model. Then focussing on the 2 more complex funnel examples.

I'm also renaming the in-notebook defined classes, in line with issue #2892.

pymc_marketing/mmm/data_conversion.py
drbenvincent: Fixed in cc6a3958: `_pandas_columns_to_dataarrays` now dedupes with `subset=[date_column, *dim_cols], keep="first"`, matching `_process_dataframe` and `_reshape_flat_target`.

Regression test `test_extra_vars_dedupes_on_index_keys_not_values` covers your repro (long frame, `dims=()`, extra var varying within a date, flat `y`). Mock-run and the targeted pytest suite are green locally.

docs/source/notebooks/mmm/mmm_funnel_from_yml.ipynb
drbenvincent: Addressed in d5f40428 / cc6a3958:

- Opening markdown now points readers to `mmm_build_from_yml_example` first and explains what this notebook adds (`effects:`, `extra_vars`, notebook-local `MuEffect` subclasses).
- `%config InlineBackend.figure_format = 'retina'` is in the setup cell; causal DAGs render as PNG via `show_causal_dag(..., retina=True)`.

The notebook was also simplified to one naive baseline plus two funnel topologies, each following DAG → effect class → YAML → PyMC graph.

drbenvincent: Partly addressed in d5f40428 / cc6a3958: the notebook now has separate Part 1 / Part 2 sections, each with a short intro and the renamed effect classes (`MeasuredMediatorEffect`, `BudgetDualIndicatorEffect` per #2892) defined beside the DAG they implement.

I'm deferring an explicit cross-topology comparison table for now — the plan is to add more graph structures in follow-up notebook iterations, and a comparison section will make more sense once those are in place. Happy to add a lightweight compare/contrast blurb in a later PR if you'd still like one before that.

drbenvincent:
## Review follow-up (d5f40428 + cc6a3958)

Thanks again for the thorough reviews. Quick status after the latest pushes:

### Addressed in this round

**Library / correctness (Daimon)**
- **`extra_vars` dedupe bug** (`data_conversion.py`): `_pandas_columns_to_dataarrays` now dedupes on date/dim keys with `keep="first"`, matching the channel/target paths. Regression test: `test_extra_vars_dedupes_on_index_keys_not_values`.
- **Dataset date validation** (`yaml.py`): added `test_build_mmm_from_yaml_dataset_missing_date_coord_raises` for the missing-`date`-coordinate branch.

**Notebook (Juan + earlier feedback)**
- Simplified to **one naive baseline**, then **two funnel topologies**, each following **DAG → effect class → YAML → PyMC graph**.
- Renamed notebook-local effects to **`MeasuredMediatorEffect`** / **`BudgetDualIndicatorEffect`** (#2892 topology names).
- Opening prose now links to **`mmm_build_from_yml_example`** and explains what this notebook adds (`effects:`, `extra_vars`, custom `MuEffect` subclasses).
- **`%config InlineBackend.figure_format = 'retina'`** plus PNG causal DAGs for gallery thumbnails.
- DataFrame + **`extra_vars`** workflow in `funnel_intro.yml` / `funnel_advanced.yml`.
- Removed unused naive funnel YAML configs; kept `funnel_intro_naive_a.yml` as the single flat baseline.

**Earlier round (9e8a5a90) — still in place**
- Deleted `as_model_dataset`; single `to_mmm_dataset` path for channels, controls, extra vars, and `fillna(0)`.
- Flat `y` + panel dims + `extra_vars`; `control_columns` + `extra_vars`; Dataset `extra_vars` validation.
- Funnel YAMLs moved beside the notebook with header comments; `REGISTRY` registration called out as demo-only.

**Housekeeping**
- Mock-run of `mmm_funnel_from_yml.ipynb` passes.
- `generate_gallery.py` extracts a PNG thumbnail from the notebook at docs build (gallery images are generated, not committed).
- PR test plan checkboxes updated.

### Intentionally deferred (follow-up PRs)

| Item | Why defer |
|------|-----------|
| **Cross-topology comparison** in the notebook | Planning to add more graph structures first; a comparison section is more useful once those exist. Part 1 / Part 2 already each introduce their topology beside the DAG. |
| **Promote effect classes into the package** (#2892) | This PR is the YAML/`extra_vars` PoC; dotted-path YAML configs can come when the classes land in `pymc_marketing`. |
| **Implied causal DAG from the model** (#2886) | Notebook still uses hand-drawn Graphviz DAGs for now. |
| **`yaml.py` importing private `_validate_extra_vars`** | Minor seam; optional cleanup, not blocking. |

Replies posted on the three open inline threads. From my side this should be ready to merge once CI is green — shout if anything else should land in this PR rather than a follow-up.

docs/source/notebooks/mmm/mmm_funnel_from_yml.ipynb
```diff
@@ -0,0 +1,2521 @@
...
```
juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #15.    
</pre><p>Please remove this print statement (can be confusing for new readers)</p><p> </p><p>From the previous round: we need a better introduction and to explain why we need these functions. A more welcoming introduction seems missing </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2885/discussion/#comment-3803640732'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #119.        display(Image(g.pipe(format="png"), retina=retina))
</pre><p>Can we simply use %config InlineBackend.figure_format = "retina" ? as other notebooks? we are introducing unnecessary boilerplate code. Let's keep thinmgs simple</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2885/discussion/#comment-3803640739'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

juanitorduz (changes requested):
Thanks @drbenvincent ! I left some comments on the notebook: just to make a more complete introduction and reduce bolerplate code which I do not think we need for the retina display.

pymc_marketing/mmm/builders/yaml.py
```diff
@@ -144,6 +148,33 @@ def build_mmm_from_yaml(
                 f"Date column '{date_column}' specified in config not found "
                 f"in either X or y data."
             )
+    elif isinstance(X, xr.Dataset) and date_column and "date" not in X.coords:
```
daimon-pymclabs, line 169: **`y` is required even when the `Dataset` already carries `_target`, and passing one silently overwrites it.**

Dataset input to `build_mmm_from_yaml` is new in this PR, but the `y is None` guard at lines 134-137 predates it and still applies unconditionally. The canonical dataset that `to_mmm_dataset` produces contains `_target`, so the natural way to use the new Dataset support is to build it once and pass it in with no separate `y`. That path is blocked, and the obvious workaround is worse than the block.

**Why this matters.** Three symptoms, all reproduced on `cc6a3958`:

**1. The canonical Dataset is rejected, though `MMM` itself accepts it.**

```python
ds = to_mmm_dataset(df, df["y"], date_column="date", dims=(),
                    channel_columns=["c1"], extra_vars=["ev"])
list(ds.data_vars)   # ['_channel', 'ev', '_target']

build_mmm_from_yaml(cfg, X=ds, y=None)
# ValueError: y not provided and no `data.y_path` found in YAML.
```

Yet the underlying model handles exactly this:

```python
MMM(...).build_model(ds, None)      # OK, _target = [0. 1. 2. 3. 4. 5.]
```

So the builder is stricter than the thing it builds, for the one input type this PR was written to support.

**2. The obvious workaround silently corrupts the target.** Supply any `y` to satisfy the guard and `to_mmm_dataset` line 149 (`ds["_target"] = y_da`) clobbers the target already in the Dataset. No warning:

```python
build_mmm_from_yaml(cfg, X=ds, y=pd.Series(np.full(6, -999.0)))
# model.xarray_dataset["_target"] -> [-999. -999. -999. -999. -999. -999.]
# ds["_target"] was              -> [   0.    1.    2.    3.    4.    5.]
```

This is the one I would fix regardless of the rest. A user who assembles a Dataset the documented way, then adds a `y` because the builder demanded one, gets a model fitted to the wrong target with nothing in the logs.

**3. A panel Dataset plus a flat `y` dies unactionably.** `_reshape_flat_target` only runs when `X` is a `DataFrame` (`data_conversion.py:142`), and for a Dataset there are no dim columns to reshape from, so a geo Dataset with a `RangeIndex` `y` surfaces raw xarray internals:

```
AlignmentError: cannot reindex or align along dimension 'date' because of
conflicting dimension sizes: {12, 6} (note: an index is found along that
dimension with size=6)
```

**Suggested fix**, smallest version first:

1. Make `y` optional when `X` is a `Dataset` that already has a target: only raise `y not provided` when `X` is not an `xr.Dataset`, or when it is one and neither `_target` nor `target` is in `X.data_vars`. That alone unblocks symptom 1 and removes the incentive to hit symptom 2.
2. Make the collision explicit rather than silent. When `X` is a `Dataset` containing a target *and* `y` is passed, raise (or at minimum warn) instead of overwriting, e.g. `ValueError: X already contains '_target'; pass either a Dataset with a target or a separate y, not both.`
3. For symptom 3, a targeted error beats the xarray traceback: if `X` is a `Dataset` with panel dims and `y` has no `date` coord, say so, and point at building `y` with a `MultiIndex` or letting `to_mmm_dataset` do the reshape from the DataFrame.

A regression test per symptom would pin all three: Dataset-with-target and `y=None` builds; Dataset-with-target plus `y` raises; panel Dataset plus flat `y` raises something readable.

pymc_marketing/mmm/builders/yaml.py
```diff
@@ -144,6 +148,33 @@ def build_mmm_from_yaml(
...
+    if extra_vars and isinstance(X, pd.DataFrame):
+        X = to_mmm_dataset(
+            X,
+            y,
+            date_column=date_column,
+            dims=dims,
+            channel_columns=channel_columns,
+            control_columns=control_columns,
+            extra_vars=extra_vars,
+            target_column=target_column,
```
daimon-pymclabs, line 190: `to_mmm_dataset` accepts `target_column` but never reads it, so passing it here is misleading.

**Why this matters:** the parameter is declared in the signature and documented in the docstring ("Name of the target / response variable"), but nothing in the function body references it. The target only ever arrives through the `y` argument. Threading `target_column` in from the builder therefore reads as though the target is being extracted from `X` by name, which is exactly the assumption that makes the `y`-handling issue in my other comment hard to spot on a read-through.

**Suggested fix:** either drop the argument at this call site, or, if the intent is for `to_mmm_dataset` to pull the target out of a DataFrame column when `y` is omitted, implement that, and it becomes the clean fix for the sibling issue too. In the meantime the docstring should not describe an argument that has no effect.

daimon-pymclabs (commented):
Re-reviewed at `cc6a3958`. The `_pandas_columns_to_dataarrays` dedupe fix is correct and the regression test covers the exact repro from the last round: `tests/mmm/test_data_conversion.py tests/mmm/builders/test_yaml.py tests/mmm/builders/test_schema.py` is 125 passed / 1 failed on my machine, and the single failure is `test_build_mmm_loads_idata_from_path` erroring on a missing local netCDF backend, unrelated to this branch.

Everything from the previous two rounds is resolved, verified against the branch:

- Dedupe now matches `_process_dataframe` and `_reshape_flat_target` (`subset=[date_column, *dim_cols], keep="first"`).
- Gallery thumbnail: the first image output in the notebook is now `image/png` (cell 4, the naive-baseline DAG), so `generate_gallery.py:213` will find it. Three PNG and three SVG outputs total.
- The notebook now opens by pointing at `mmm_build_from_yml_example` and has separate Part 1 / Part 2 sections per topology.
- Test plan boxes are checked.

One remaining issue, filed inline. It is a single root cause with three symptoms, and it sits squarely on the feature this PR adds, `xr.Dataset` input to `build_mmm_from_yaml`. `y` is still required unconditionally, which was correct when `X` could only be a `DataFrame` but is wrong now that the canonical `Dataset` carries `_target` itself. The bad symptom is the middle one: supplying a redundant `y` to get past the check silently overwrites the target already in the Dataset. Repro and suggested fix are in the inline comment.

Also flagged inline: `to_mmm_dataset` never reads its `target_column` argument, so passing it from the builder reads as load-bearing when it is not.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Fixed in 90c356e4: `y` is optional when `X` is a Dataset with `_target`/`target`; passing both now raises explicitly; panel Dataset + flat `y` raises a readable `ValueError`. Regression tests in `test_yaml.py` and `test_data_conversion.py`.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Fixed in 90c356e4: `to_mmm_dataset` now reads `target_column` from a DataFrame when `y` is omitted; docstring updated to match.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Addressed in `90c356e4` — all three symptoms from the repro:

**1. Dataset with `_target`, `y=None` builds**
`build_mmm_from_yaml` now skips the `y not provided` guard when `X` is an `xr.Dataset` that already carries `_target` or `target`.

**2. Redundant `y` no longer silently overwrites**
Both `yaml.py` and `to_mmm_dataset` raise if a Dataset embeds a target and a separate `y` is also passed.

**3. Panel Dataset + flat `y` → readable error**
`to_mmm_dataset` raises a `ValueError` pointing at MultiIndex / DataFrame reshape instead of an xarray `AlignmentError`.

Regression tests: `test_build_mmm_from_yaml_dataset_with_target_y_none`, `test_build_mmm_from_yaml_dataset_with_target_and_y_raises`, `test_build_mmm_from_yaml_panel_dataset_flat_y_raises`, plus unit tests in `test_data_conversion.py`.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Addressed in `90c356e4`: `to_mmm_dataset` now uses `target_column` when `y` is omitted and `X` is a DataFrame — it extracts that column, applies `_reshape_flat_target` for panel dims, and writes `_target`. The docstring now states that `target_column` is ignored when `y` is supplied or when `X` already embeds a target.

Regression test: `TestToMmmDatasetYInputTypes::test_target_column_extracts_when_y_omitted`.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Thanks for the clear repro — this was a mismatch between the new Dataset input path and guard logic written for DataFrame-only callers.

**Why it broke**

`build_mmm_from_yaml` always required `y`, but the natural Dataset workflow is: call `to_mmm_dataset(...)` once (which embeds `_target`), then pass that Dataset to the builder with `y=None`. `MMM.build_model` already supports that; the builder was stricter than the model it wraps.

The silent overwrite was worse: users who passed a redundant `y` just to satisfy the guard got the wrong target with no warning, because `to_mmm_dataset` always assigned `ds["_target"] = y_da` when `y` was set.

**What we changed (`90c356e4`)**

In `yaml.py`, `y` is optional when `X` is a Dataset that already has `_target`/`target`. If both an embedded target and a separate `y` are supplied, we fail fast in the builder (before `build_model`).

In `to_mmm_dataset` (the shared seam `MMM` also uses), we enforce the same rule and add a targeted error for panel Dataset + flat RangeIndex `y` — pointing callers at MultiIndex `y` or a DataFrame input so `_reshape_flat_target` can run.

**Why both layers**

The builder check catches the common YAML path early. The `to_mmm_dataset` check covers direct Dataset callers and keeps behaviour consistent with `MMM.build_model`.

**Tests**

- `test_build_mmm_from_yaml_dataset_with_target_y_none` — symptom 1
- `test_build_mmm_from_yaml_dataset_with_target_and_y_raises` — symptom 2
- `test_build_mmm_from_yaml_panel_dataset_flat_y_raises` — symptom 3
- `test_dataset_with_target_and_y_raises` / `test_panel_dataset_flat_y_raises` in `test_data_conversion.py`

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Good catch — passing `target_column` from the builder looked load-bearing on read-through, but `to_mmm_dataset` never used it, which made the `y`-handling bug harder to spot.

**Why we implemented it rather than drop the kwarg**

The YAML spec already carries `target_column` (same as the `MMM` constructor). The intended contract is: callers can omit `y` and let the converter pull the response from a named DataFrame column. That aligns the builder with how tabular configs are written, and it is the same field `MMM._generate_and_preprocess_model_data` threads through.

**What we changed (`90c356e4`)**

When `y is None`, `X` is a DataFrame, and `target_column` names a column in `X`, `to_mmm_dataset` now extracts it, runs `_reshape_flat_target` when panel dims require it, and writes `_target`. The docstring now states that `target_column` is ignored when `y` is supplied or when `X` already embeds a target — so the parameter documents real behaviour, not aspirational behaviour.

We kept passing `target_column` from `build_mmm_from_yaml` into the `to_mmm_dataset` call for the DataFrame + `extra_vars` path; that call site is exactly where a YAML-driven user might expect the target column name to matter.

**Test**

`TestToMmmDatasetYInputTypes::test_target_column_extracts_when_y_omitted`

drbenvincent:
## Follow-up on Daimon's latest review (`90c356e4`)

Two inline threads from the Dataset-input review are addressed in **`90c356e4`** (*Fix Dataset target handling in build_mmm_from_yaml*). Replies with what/why are on each thread; summary here for easier review.

### 1. Dataset + `y` handling ([#discussion_r3803690075](https://github.com/pymc-labs/pymc-marketing/pull/2885#discussion_r3803690075))

**Problem:** The builder required `y` even when `X` already carried `_target` from `to_mmm_dataset`. Passing a dummy `y` to satisfy the guard silently overwrote the embedded target. Panel Dataset + flat `y` surfaced an xarray `AlignmentError`.

**Approach:** Align `build_mmm_from_yaml` with `MMM.build_model` — Dataset-with-target is a first-class input. Fail fast when both embedded target and separate `y` are passed. Push panel reshape guidance into `to_mmm_dataset` so the error is actionable.

**Files:** `pymc_marketing/mmm/builders/yaml.py`, `pymc_marketing/mmm/data_conversion.py`

### 2. `target_column` unused ([#discussion_r3803690084](https://github.com/pymc-labs/pymc-marketing/pull/2885#discussion_r3803690084))

**Problem:** `target_column` was threaded from YAML into `to_mmm_dataset` but never read, which misled readers about how targets arrive.

**Approach:** Implement the documented contract — when `y` is omitted and `X` is a DataFrame, extract `target_column` and reshape for panel dims. Update the docstring to spell out when the argument applies.

**Files:** `pymc_marketing/mmm/data_conversion.py` (builder call site unchanged; now functional)

### Tests added

`tests/mmm/builders/test_yaml.py`:
- `test_build_mmm_from_yaml_dataset_with_target_y_none`
- `test_build_mmm_from_yaml_dataset_with_target_and_y_raises`
- `test_build_mmm_from_yaml_panel_dataset_flat_y_raises`

`tests/mmm/test_data_conversion.py`:
- `test_target_column_extracts_when_y_omitted`
- `test_dataset_with_target_and_y_raises`
- `test_panel_dataset_flat_y_raises`

Both inline threads are resolved. Happy to adjust if you prefer `target_column` dropped at the builder call site instead of implemented.

drbenvincent:
Thanks @juanitorduz, I made the intro more introductory. I defer a fuller explanation of the custom effect classes to the existing funnel effect notebooks. I'm due to update those soon

docs/source/notebooks/mmm/mmm_funnel_from_yml.ipynb
```diff
@@ -0,0 +1,2887 @@
...
+    "    for edge in edges:\n",
+    "        if len(edge) == 2:\n",
+    "            g.edge(edge[0], edge[1])\n",
+    "        else:\n",
+    "            g.edge(edge[0], edge[1], **edge[2])\n",
+    "    return g\n",
+    "\n",
+    "\n",
+    "def show_causal_dag(*args, **kwargs) -> None:\n",
+    "    display(draw_causal_dag(*args, **kwargs))"
```
daimon-pymclabs: **Gallery thumbnail regression: the notebook no longer emits any PNG output.**

This line is the cause. At `cc6a3958` the helper was:

```python
def show_causal_dag(*args, retina: bool = True, **kwargs) -> None:
    g = draw_causal_dag(*args, **kwargs)
    display(Image(g.pipe(format="png"), retina=retina))
```

Rasterising through `g.pipe(format="png")` produced `image/png` outputs. The simplification in `68170792` dropped the `Image` import and now displays the `graphviz.Digraph` object directly, so IPython falls back to graphviz's default repr, which is SVG.

I counted the output mimebundles at both heads:

| head | image outputs |
| --- | --- |
| `cc6a3958` | cells 4, 10, 18 `image/png`; cells 8, 16, 26 `image/svg+xml` |
| `90c356e4` | cells 9, 13, 15, 21, 23, 31 all `image/svg+xml`, zero PNG |

`generate_gallery.py:213` only extracts `image/png`:

```python
if "data" in output and "image/png" in output["data"]:
```

With no PNG anywhere in the notebook, that loop falls through to `_use_default_image(f"No image found in {self.notebook_path}")`, so this card renders the placeholder thumbnail while the neighbouring notebook cards show real plots. This is the same issue I raised in round one; it was fixed at the previous head and came back with the DAG simplification.

Worth noting the test plan box now reads `- [x] Gallery sync (generate_gallery.py; notebook embeds PNG outputs for thumbnail extraction at docs build)`. The claim in the box is not true of the committed notebook.

Smallest fix, restoring the one line that mattered while keeping the simplified helper:

```python
def show_causal_dag(*args, **kwargs) -> None:
    display(Image(draw_causal_dag(*args, **kwargs).pipe(format="png")))
```

with `Image` added back to the `IPython.display` import in cell 2. If you would rather keep the DAGs as SVG for crisper docs rendering, then any single matplotlib figure anywhere in the notebook also fixes the thumbnail; the extractor takes the first PNG it finds and does not care which cell it came from. Either way, please re-run `generate_gallery.py` and confirm a real thumbnail lands before ticking that box.

pymc_marketing/mmm/builders/yaml.py
```diff
@@ -128,12 +138,19 @@ def build_mmm_from_yaml(
             raise ValueError("X not provided and no `data.X_path` found in YAML.")
         X = _load_df(data_cfg.X_path)
     if y is None:
```
daimon-pymclabs, line 141: **Minor, consistency: the new `target_column` support is unreachable from the YAML builder.**

This is not a bug, and it is not blocking. It is the same shape as the issue I raised last round, just moved from Dataset input to DataFrame input, so it seems worth naming while the code is fresh.

`to_mmm_dataset` now handles "DataFrame carries the target, `y` omitted", which is exactly what the updated docstring advertises:

> `target_column` -- Name of the target column in a `pd.DataFrame` when *y* is omitted.

I confirmed that works one layer down:

```python
ds = to_mmm_dataset(X, None, date_column="date", channel_columns=["channel_1"],
                    extra_vars=["lower_spend", "lower_control"], target_column="y")
# -> _target present=True, mean=76.5
```

And `MMM.build_model(X, None)` on the same frame builds fine. But this guard only grants the omit-`y` escape hatch to `xr.Dataset` input, so the natural YAML for that case is still refused:

```yaml
model:
  kwargs:
    date_column: date
    channel_columns: [channel_1]
    target_column: y      # column lives inside X_path's CSV
data:
  X_path: X.csv           # no y_path
```

```
ValueError: y not provided and no `data.y_path` found in YAML.
```

One CSV holding dates, channels and the target is the common shape, and the config declares `target_column` right there, so the rejection reads as a bug to the person writing the YAML. The builder is again stricter than the layers beneath it.

Fix mirrors what you already did for Dataset input, extending the escape hatch to the DataFrame case:

```python
if y is None:
    has_embedded_target = (
        isinstance(X, xr.Dataset) and _dataset_has_target(X)
    ) or (
        isinstance(X, pd.DataFrame)
        and (tc := model_spec["kwargs"].get("target_column"))
        and tc in X.columns
    )
    if not has_embedded_target:
        ...
```

That needs `target_column` read before this block rather than below it. If you would rather not widen the scope of a docs PR, a follow-up issue is fine; the asymmetry is just easy to lose track of once `target_column` looks supported in the docstring.

daimon-pymclabs (commented):
Final round at `90c356e4`. The Dataset-target fix is correct and complete: I re-ran all three symptoms from my last review against the branch and each behaves as intended, with no regression on the panel path.

| last round's symptom | at `90c356e4` |
| --- | --- |
| Dataset with embedded target + `y=None` rejected | builds; also works with the unprefixed `target` name |
| supplying `y` silently overwrote the embedded target | raises `... not both` |
| panel Dataset + flat `y` gave a raw `AlignmentError` | raises the readable `no 'date' coordinate` error |
| panel Dataset + MultiIndex `y` | still builds, `_target` shape `(8, 2)` |

Moving the guard into `to_mmm_dataset` was the right call, since `MMM.build_model` routes through the same function and inherits all three fixes for free. `tests/mmm/builders/test_yaml.py` + `tests/mmm/test_data_conversion.py`: 94 passed, 1 failed, and the failure is `test_build_mmm_loads_idata_from_path` erroring on a missing netCDF backend in my sandbox, unrelated to the branch.

One regression to fix before merge, and two notes.

**The gallery thumbnail regressed, and the checked box in the test plan is inaccurate.** Details inline. This one was fixed at the previous head and came back with the notebook simplification, so it is worth re-checking rather than trusting the checkbox.

**Note, scope:** `MMM.build_model(X_df_containing_the_target, y=None)` changed from raising `KeyError: "No variable named '_target'"` to silently building against `X[target_column]`. I verified that flip across `cc6a3958` -> `90c356e4`. It is a sensible improvement, but it is a public API behaviour change landing in a PR titled as a docs change, so it deserves a line in the release notes rather than arriving unannounced.

**Note, asymmetry:** the same capability is now unreachable from the YAML entry point this PR is about. Inline on `yaml.py`.

juanitorduz:
Thanks @drbenvincent ! from my side LGTM :) I just gave a final Daimon review. After those (if relevant), we can merge 🙌

docs/source/notebooks/mmm/mmm_funnel_from_yml.ipynb
drbenvincent: Fixed: `show_causal_dag` again rasterises through `Image(... .pipe(format="png"))` (no `retina=` kwarg — we use `%config InlineBackend.figure_format = "retina"` for matplotlib, per Juan's note). Re-executed the notebook: cells 9, 15, and 23 now emit `image/png`; PyMC graphviz outputs stay SVG. Ran `generate_gallery.py` locally — `docs/source/gallery/images/mmm_funnel_from_yml.png` is a real 275×275 thumbnail, not the placeholder.

pymc_marketing/mmm/builders/yaml.py
drbenvincent: Addressed: extended the `y is None` escape hatch to DataFrames whose `target_column` is present in `X.columns`, mirroring the Dataset path. Regression test: `test_build_mmm_from_yaml_dataframe_target_column_no_y`.

docs/source/notebooks/mmm/mmm_funnel_from_yml.ipynb
drbenvincent: Addressed: removed the stray print; the opening markdown now points at `mmm_build_from_yml_example`, explains why the helper functions exist, and separates funnel background from the YAML wiring focus.

drbenvincent: Addressed: dropped the `retina=` kwarg from `show_causal_dag`; the notebook uses `%config InlineBackend.figure_format = "retina"` like the other gallery notebooks. For gallery thumbnails, DAG cells still rasterise to PNG via `Image(... .pipe(format="png"))` because `generate_gallery.py` only reads `image/png` outputs.

drbenvincent:
Thanks @daimon-pymclabs and @juanitorduz — final review round addressed locally (push incoming).

**Gallery thumbnail (blocker)**  
Restored PNG rasterisation in `show_causal_dag` via `display(Image(draw_causal_dag(...).pipe(format="png")))`, without the old `retina=` kwarg (we use `%config InlineBackend.figure_format = "retina"` for matplotlib, per Juan). Re-executed the notebook: causal DAG cells now emit `image/png` (cells 9, 15, 23); PyMC graphviz outputs stay SVG. Ran `generate_gallery.py` locally and confirmed a real thumbnail lands in `docs/source/gallery/images/mmm_funnel_from_yml.png`.

**YAML `target_column` asymmetry (note from final round)**  
Extended the `y is None` escape hatch in `build_mmm_from_yaml` to DataFrames whose `target_column` is already in `X.columns`, matching the Dataset path added in `90c356e4`. Regression test: `test_build_mmm_from_yaml_dataframe_target_column_no_y`.

**Release-note item (`MMM.build_model(X, y=None)` when target lives in X)**  
Agree this is worth calling out in release notes: `to_mmm_dataset` / `MMM.build_model` now extract the target from a DataFrame column when `y` is omitted and `target_column` is set, instead of raising `KeyError: "_target"`. Suggest a bullet under **Enhancements** in the next release.

**Juan's notebook comments**  
Opening prose now links to `mmm_build_from_yml_example`, separates funnel background from YAML wiring, and drops the stray print / retina boilerplate.

**Tests**  
`pytest tests/mmm/builders/test_yaml.py tests/mmm/test_data_conversion.py` — 96 passed locally (same unrelated netCDF skip as before on my machine).

drbenvincent:
fix for failing codecov incoming

drbenvincent:
Codecov patch coverage fix pushed in `8bf0e698`.

**Problem:** `codecov/patch` was failing at 96.72% with 2 uncovered lines in `pymc_marketing/mmm/builders/yaml.py` (the new Dataset/`target_column` paths from this PR).

**Fix:** Added three focused regression tests:

1. `test_build_mmm_from_yaml_dataframe_missing_date_column_raises` — covers the DataFrame date-column validation branch (`yaml.py` ~165).
2. `test_build_mmm_from_yaml_dataset_with_target_name_y_none` — covers `y=None` when the input Dataset carries unprefixed `target` (not only `_target`).
3. `test_target_column_extracts_panel_target_when_y_omitted` — covers panel `target_column` extraction/reshape in `to_mmm_dataset` when `y` is omitted.

Locally, `pytest tests/mmm/builders/test_yaml.py tests/mmm/test_data_conversion.py` is 99 passed and `yaml.py` patch lines are fully covered (only pre-existing `_load_df` parquet/unknown-extension helpers remain untouched at 42/45, outside this diff).

Waiting on CI for the updated codecov/patch check.

drbenvincent:
hmmm, more code coverage coming I guess

juanitorduz (approved):

drbenvincent:
Failing flaky test, unrelated to this PR @juanitorduz

drbenvincent:
@juanitorduz, I'm just doing a morning review, and this one was approved and it entered the merge queue, but has failing tests which are seemingly unrelated to the work in this PR. Turns out I have ability to re-run failed jobs, so have set that off now. See what happens - if they pass then I think we are good to merge

PR #2887, opened 2026-08-17
feat(mmm): parameterise DelayedAdstock by half-life

JakePiekarski314:
## Description

`DelayedAdstock` can now take a prior on the half-life of the ad effect instead of on `alpha`, extending to the delayed curve the parametrisation that #2839 added for `GeometricAdstock`. Stakeholders reason more readily about "the effect is at half strength two weeks either side of its peak" than about a retention rate raised to a squared lag.

```python
adstock = DelayedAdstock(l_max=12, priors={"halflife": Prior("LogNormal", mu=np.log(3), sigma=0.3)})

# or, to use the default InverseGamma(9, 5.75) prior
adstock = DelayedAdstock(l_max=12, parametrization="halflife")
```

Since the weight at lag $\ell$ is $\alpha^{(\ell-\theta)^2}$, a half-life $h$ gives

$$\alpha = 2^{-1/h^{2}}, \qquad w_\ell = 2^{-\left(\frac{\ell-\theta}{h}\right)^{2}},$$

which maps any positive half-life into $(0, 1)$, so no constraint can be violated. Because the kernel is symmetric about $\theta$, one number carries both readings: $h$ is the time over which effectiveness climbs from half to its peak, and equally the time over which it falls from the peak back to half. The identity is exact for every `l_max` and either setting of `normalize`, since truncation and normalisation rescale the whole kernel and leave the lag at which the weight halves untouched. As in #2839, alpha is computed inline rather than as a `Deterministic`, so the sampled variable is `adstock_halflife` and there is no `adstock_alpha`.

`theta` is untouched and shared by both parametrisations, so only the width of the response around the peak is being re-expressed.

### How the interpretation of half-life differs when using `DelayedAdstock` as opposed to `GeometricAdstock`:

Where GeometricAdstock's half-life is a decay time, the periods needed for the effect to fall to half its initial value under a kernel that declines monotonically from lag zero, DelayedAdstock's squared exponent makes the kernel symmetric about its peak at theta, so its half-life instead measures the width of the response either side of that peak: the time taken to climb from half strength to the peak, and equally the time taken to fall from the peak back to half.

### The default is matched to the default on alpha

Switching parametrisation is a change of coordinates rather than a change of model:

| quantile | half-life | implied alpha | `Beta(1, 3)` alpha |
| --- | --- | --- | --- |
| 5% | 0.398 | 0.013 | 0.017 |
| 25% | 0.532 | 0.087 | 0.091 |
| 50% | 0.663 | 0.207 | 0.206 |
| 75% | 0.841 | 0.375 | 0.370 |
| 95% | 1.225 | 0.630 | 0.632 |

Read as a half-life that default is short, a median of 0.66 periods, which the docstring and the guide both now say out loud. A custom half-life prior needs positive support and should keep its mass away from zero, where the squared exponent flattens the likelihood.

### Changes to the machinery introduced in #2839 

Rather than duplicate the parametrisation plumbing onto a second class, this PR lifts it into a mixin. `GeometricAdstock` comes out behaviourally identical, with the same defaults and byte-identical serialised payloads.

- Everything #2839 put on `GeometricAdstock` that was not specific to geometric decay moved verbatim into a new `_AlphaHalfLifeParametrization`: the `parametrization`-aware `__init__` (including the `default_priors` instance-attribute redirect), the `function_priors` setter that rejects a prior for the inactive parameter, the `_has_defaults_for_all_arguments` override, `to_dict`, and `_alternative_parameters`.
- The exactly-one-of check and the alpha conversion that 2939 had inline in `GeometricAdstock.function` became `_resolve_alpha` on the mixin, with the curve-specific formula supplied by an `_alpha_from_halflife` static method. Both `function` bodies are now a single call to their transformer, with `alpha=self._resolve_alpha(alpha, halflife)`.
- Each class keeps only what distinguishes it: `0.5 ** (1 / h)` for geometric, `2 ** (-1 / h**2)` for delayed, plus its own `default_priors` and `halflife_priors`.
- The mixin is not an `AdstockTransformation` at runtime. Its base is `AdstockTransformation` only under `TYPE_CHECKING` and `object` otherwise, so it stays out of anything that enumerates the transformation subclasses while type checkers can still resolve the cooperative `super().__init__` and the base members it reads. It is listed first in both class definitions so its setter and prior checks take precedence in the MRO.
- `_has_defaults_for_all_arguments` gets its first real workout here. `DelayedAdstock.halflife_priors` has to carry `theta` alongside `halflife`, so the override matters: it subtracts both alternatives from the required set and adds back only the active one, leaving every other argument of `function` still needing a prior. With geometric's single parameter that path was untested.

`pymc_marketing/mmm/transformers.py` is untouched: `delayed_adstock(...)` still takes `alpha` only, and the conversion lives in the component wrapper, as it does for geometric.

### Note for reviewers

`halflife` is inserted before `theta` in `DelayedAdstock.function`, which changes the positional order from `(x, alpha, theta)` to `(x, alpha, halflife, theta)`. `Transformation.apply` calls `self.function(x, dim=core_dim, **kwargs)`, so everything inside the library passes these by keyword and is unaffected, but a caller reaching for `adstock.function(x, 0.5, 2)` directly would now bind `2` to `halflife`. The alternative was appending `halflife` last, which would have read oddly against the geometric signature and against the docstring order. Flagging it as the one visible difference from the purely additive change in `418dea2`, where `alpha` was already the final argument.

As in #2839 , half-life models have no `adstock_alpha` in the trace, so helpers that name it explicitly need `adstock_halflife` instead. This is called out in the class docstring.

The guide gained a "Parametrizing by Half-Life" section for the delayed curve, mirroring the geometric one: the mapping, decay curves for a range of half-lives at fixed `theta`, and how to choose a prior. The formula for the delayed kernel in the existing text said $\alpha^{|i-\theta|}$ where the implementation squares the lag, so that is corrected in the same cell.

### Test plan

- `tests/mmm/components/test_adstock.py` and `tests/mmm/test_mmm.py::TestDelayedAdstockHalfLife` (187 passed, 1 skipped): parametrisation inference from the priors, the class attributes staying untouched, priors rejected in both directions and for both at once, rejection on assignment after construction, `theta` accepted under both parametrisations, the default matching `Beta(1, 3)` across five quantiles via scipy, numerical equivalence against the corresponding `alpha` across 5 alphas by 3 thetas by 3 `l_max` values by `normalize` on and off, the half-peak reading at $\theta \pm h$, symmetry about `theta`, monotone widening in `h`, `function` rejecting neither-or-both, channel-specific priors broadcasting over named dims, `sample_prior`/`sample_curve`, and serialisation round-trips including that `parametrization` is omitted for `alpha`
- `tests/mmm/test_mmm.py`, `tests/mmm/components`, `tests/mmm/builders`, `tests/mmm/test_transformers.py` (821 passed, 1 skipped): `adstock_halflife` and `adstock_theta` both reach the built model, `model_config` overrides the half-life, a stale `adstock_alpha` warns, and an `MMM.save`/`load` round-trip keeps the parametrisation

## Checklist

- [x] pre-commit.
- [x] Included tests that prove the fix is effective or that the new feature works.
- [x] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format].

## Feedback (relevant to this PR)

1. Problem: Adding halflife ahead of theta in DelayedAdstock.function meant an existing positional call like function(x, 0.5, 2) silently bound 2 to halflife and let theta fall back to its theta=0 default, fitting an undelayed model with no error.

- [x] Solution: Making theta keyword-only and required turns that into an immediate TypeError, at no cost, since apply already passes every parameter by keyword and the prior check reads keyword-only names straight out of the signature. Commit: `5ad12d3d`.

2. Problem: halflife means different things in the two classes. A decay time from lag zero for GeometricAdstock, but a half-width either side of the peak for DelayedAdstock. A reader arriving from geometric could carry the decay-time reading across and misjudge how wide a response their prior implies.

- [x] Solution: Keeping the shared keyword for cross-class consistency and naming the difference explicitly, in a docstring paragraph placed before the matched-defaults discussion and in the guide's half-life section, so the rise-to-peak half of the interpretation cannot be missed. Commit: `b7a91dc3`.

3. Problem: _has_defaults_for_all_arguments was only ever tested on its passing path, so the mixin's override never actually raised in a test — including the missing_parameters branch that the default_priors shadowing newly makes reachable.

- [x] Solution: Two DelayedAdstock subclasses, one with halflife_priors missing theta and one carrying an argument function doesn't take, assert on the message and on an empty priors set so they fail if the base implementation runs instead. Commit: `2dc5c5ce`


daimon-pymclabs:
Follow-up to the review, with concrete patches for each point.

### 1. `theta=0` is a silent default — make it keyword-only and required

`function(self, x, alpha=None, halflife=None, theta=0, *, dim)` was forced by Python's ordering rule, but it converts the positional hazard flagged in the "Note for reviewers" section from loud into quiet. `adstock.function(x, 0.5, 2)` used to bind `2` to `theta`; it now binds to `halflife` and `theta` silently falls back to `0`, which is a different model with no error.

Every internal caller already passes by keyword (`apply` calls `self.function(x, dim=core_dim, **kwargs)`), so this costs nothing:

```diff
-    def function(self, x, alpha=None, halflife=None, theta=0, *, dim: str):
+    def function(self, x, alpha=None, halflife=None, *, theta, dim: str):
```

`_has_defaults_for_all_arguments` reads `signature().parameters`, which includes keyword-only names, so the prior check is unaffected. The exact call the PR body worries about then fails loudly instead of silently fitting `theta=0`.

### 2. Half-life near zero yields NaN, not a flat likelihood

`2 ** (-1/h²)` underflows to exactly `0.0` once `h < 0.0305` (against `h < 0.00093` for the geometric `0.5 ** (1/h)` — the square is what brings it into reach). `_check_alpha` tests `alpha >= 0`, so zero passes despite the `"0 < alpha <= 1"` message. With `normalize=True` (the default) and a non-integer `theta`, every weight is then `0 ** positive == 0`, the sum is `0`, and the kernel is `0/0`. Since `theta ~ HalfNormal(1)` is continuous, it is never an integer in practice.

The shipped default prior puts no meaningful mass there, but a user-supplied diffuse or heavy-left-tailed half-life prior lets the sampler wander in, and the failure mode is a NaN logp rather than a rejected proposal. Cheapest fix is to floor the mapping:

```diff
     @staticmethod
     def _alpha_from_halflife(halflife):
         """Map a half-life onto the retention rate of the decay."""
-        return 2.0 ** (-1.0 / halflife**2)
+        # Floored at the smallest positive normal: below h ~ 0.03 the exact
+        # expression underflows to 0, and with normalize=True a non-integer
+        # theta then makes every weight 0 and the normalised kernel 0/0.
+        return pt.maximum(2.0 ** (-1.0 / halflife**2), np.finfo(float).tiny)
```

If you would rather not clamp, then at minimum sharpen the docstring line, which currently understates it:

```diff
-    half-life prior must have positive support and should keep its mass away
-    from zero, where the likelihood goes numerically flat. The
+    half-life prior must have positive support and should keep its mass away
+    from zero: below a half-life of roughly 0.03 periods the implied ``alpha``
+    underflows to zero and, with ``normalize=True``, the kernel evaluates to
+    NaN rather than merely going flat. The
```

### 3. `parametrization="halflife"` now means two different quantities

For `GeometricAdstock` the half-life is a decay time. Here it is the half-width at half maximum of a Gaussian-shaped kernel (`h = σ√(2 ln 2)`), symmetric about the peak. The PR is candid about this and the docs explain it well, so this is not a blocker — but it is worth deciding on purpose rather than by default. Two coherent options:

- **Keep `halflife`.** Cross-class consistency of the keyword is a real argument, and the docstring already spells out the symmetry. If so, add one line to the class docstring naming the difference explicitly, so a reader coming from `GeometricAdstock` cannot carry the decay-time reading over.
- **Rename to `halfwidth`** for `DelayedAdstock` only, with `halflife` accepted as a deprecated alias for one release. More precise, at the cost of the shared keyword.

My preference is the first, with the extra docstring line. Flagging so the call is recorded either way.

### 4. Default half-life reads oddly, but keep it

`InverseGamma(9, 5.75)` matches `Beta(1,3)` well (alpha quantiles 0.013 / 0.087 / 0.207 / 0.375 / 0.630 against 0.017 / 0.091 / 0.206 / 0.370 / 0.632 — reproduced independently). But it implies a median half-life of 0.66 periods: the effect is at half strength two-thirds of a period either side of the peak, essentially a spike. Someone reaching for `parametrization="halflife"` for interpretability is unlikely to expect that.

The backward-compatibility argument is strong, so I would keep the default. Suggestion is only to surface the number where users will hit it — the class docstring says "median half-life of 0.66 periods" already; adding the same sentence to the guide notebook's half-life section, next to the matched-priors plot, would put it in front of the reader who is actually choosing a prior.

### 5. Missing test: `ParameterPriorException` through the mixin override

The PR body says `_has_defaults_for_all_arguments` "gets its first real workout here", but the tests only exercise the passing path. Nothing raises `ParameterPriorException` through the mixin's override; the only tests for it are in `test_base.py`, against the base implementation. Suggested addition to `tests/mmm/components/test_adstock.py`:

```python
def test_halflife_priors_missing_argument_raises() -> None:
    class IncompleteDelayedAdstock(DelayedAdstock):
        lookup_name = "incomplete_delayed"
        # theta is an argument of function but has no prior here
        halflife_priors = {"halflife": Prior("InverseGamma", alpha=9, beta=5.75)}

    with pytest.raises(ParameterPriorException):
        IncompleteDelayedAdstock(l_max=10, parametrization="halflife")


def test_halflife_priors_extra_argument_raises() -> None:
    class ExtraDelayedAdstock(DelayedAdstock):
        lookup_name = "extra_delayed"
        halflife_priors = {
            "halflife": Prior("InverseGamma", alpha=9, beta=5.75),
            "theta": Prior("HalfNormal", sigma=1),
            "not_an_argument": Prior("HalfNormal", sigma=1),
        }

    with pytest.raises(ParameterPriorException):
        ExtraDelayedAdstock(l_max=10, parametrization="halflife")
```

The second one covers the `missing_parameters` branch, which is the half of the check the shadowing makes newly reachable.

Everything else in the diff checks out: the MRO linearizes cleanly with `_MixinBase = object`, the `function_priors` property override wins over the base because the mixin is listed first, `AdstockTransformation.function_priors.fset` still reaches the original setter, `variable_mapping` and `model_config` key off `default_priors` and so follow the shadowing, serialization round-trips both directions, and `GeometricAdstock` comes out behaviourally identical.

juanitorduz:
I think 

```python
+        # Floored at the smallest positive normal: below h ~ 0.03 the exact
+        # expression underflows to 0, and with normalize=True a non-integer
+        # theta then makes every weight 0 and the normalised kernel 0/0.
+        return pt.maximum(2.0 ** (-1.0 / halflife**2), np.finfo(float).tiny)
``` 

should be using pytensor not numpy

daimon-pymclabs:
Correction to the clamp patch in my earlier comment — @juanitorduz is right that it should not use numpy for the operation, and it was wrong in two further ways.

**Namespace.** `pt` is not imported in `adstock.py` at all, and even with the import `pt.maximum` would be wrong here: `delayed_adstock` builds the kernel with `ptx.math.power` on `as_xtensor` inputs, so these are `XTensorVariable`s and the clamp has to go through `pytensor.xtensor`. There is no `ptx.maximum`; it is `ptx.math.maximum`.

**Dtype.** Numpy is still the right way to *obtain* the constant (pytensor has no `finfo` equivalent), but keying it to `float` hardcodes float64. Under `floatX=float32` the float64 constant silently upcasts the whole expression back to float64; `np.finfo(pytensor.config.floatX).tiny` stays float32 at `1.1754944e-38`. Corrected patch:

```python
import pytensor

    @staticmethod
    def _alpha_from_halflife(halflife):
        """Map a half-life onto the retention rate of the decay."""
        # Floored at the smallest positive normal: below h ~ 0.03 the exact
        # expression underflows to 0, and with normalize=True a non-integer
        # theta then makes every weight 0 and the normalised kernel 0/0.
        return ptx.math.maximum(
            2.0 ** (-1.0 / halflife**2), np.finfo(pytensor.config.floatX).tiny
        )
```

**But I would reconsider the layer, and you may want to apply nothing here.** The NaN is not introduced by this PR. `delayed_adstock` has `alpha: XTensorLike = 0.0` as its own default, and `delayed_adstock(x, alpha=0.0, theta=0.5)` produces an all-NaN kernel on `main` today with no half-life involved. `_check_alpha` carries the message `"0 < alpha <= 1"` but tests `(alpha >= 0) & (alpha <= 1)`, so zero passes the check that its own message excludes. Patching `_alpha_from_halflife` fixes only the path this PR opens and leaves the direct call NaN-ing.

Filed as #2889 for a follow-up PR against `transformers.py`. If you would rather keep this PR to the parametrization, the alternative is to change nothing in the code here and only sharpen the docstring line, which currently understates the failure:

```diff
-    half-life prior must have positive support and should keep its mass away
-    from zero, where the likelihood goes numerically flat. The
+    half-life prior must have positive support and should keep its mass away
+    from zero: below a half-life of roughly 0.03 periods the implied ``alpha``
+    underflows to zero and, with ``normalize=True``, the kernel evaluates to
+    NaN rather than merely going flat. The
```

The other four items in my earlier comment stand as written.

juanitorduz:
so @JakePiekarski314 , please ignore 2 for now ;)

juanitorduz (approved):
Thank you @JakePiekarski314 !

PR #2890, opened 2026-08-17
docs(mmm): funnel-aware budget optimization in the advanced funnel notebook

juanitorduz:
## Description

Adds a **Funnel-aware budget optimization** section to `mmm_funnel_mueffect_advanced.ipynb` (two upper-funnel channels x three geos, both feeding a latent demand pool through a custom `DataVarMuEffect`), plus the pieces around it listed under "Also in this PR". The section puts the fitted funnel model through `BudgetOptimizer` in-sample: optimize the weekly allocation across the six geo x channel cells at a fixed total, +/-50% per cell, the historical spend pattern kept fixed (`budget_distribution_over_period`), then re-weight the historical spend by the optimal multipliers and score the plans with the posterior and against the known truth.

What the section shows:

- **The three requirements for optimizing a funnel model**, each stated and verified at run time. (1) The effect must read `mmm.channel_data_scaled`, so the optimizer's `pm.do` intervention on `channel_data` propagates into the mediated path. (2) The objective must be `total_response_original_scale`, the variable a model with `mu_effects` registers (#2895); the default `total_media_contribution_original_scale` scores the direct path only and runs as a like-for-like control on the same posterior. (3) The effect's exogenous data must cover the window: `create_optimization_model(start_date, end_date)` builds the window dataset with every variable the effects read (observed values on the training dates, zeros beyond) and sizes the carry-over by `effective_carryover_lags()`, the model's `l_max` plus the effect's declared `additional_carryover_lags` (#2898); the in-sample window is chosen so that window plus carry-over is exactly the training range.
- Verification through the public API (`optimization_variables`, `extract_response_distribution`), all asserted at run time: at the historical plan the direct, mediated and total response equal the in-sample posterior means; the exact gradient of the mediated response with respect to every cell's weekly spend is positive and the gradient of the total is the direct plus the mediated gradient (the xtensor graph is lowered before `pytensor.grad`, as the optimizer does for its own objective); each optimizer wins on its own objective; the loop closes at both optima against the posterior-predictive recomputation of the re-weighted plans and `scipy_result.fun`.
- A per-cell marginal table (posterior direct, mediated and total response per unit of weekly spend, and the share the default objective sees) next to the same marginals on the generative truth (central differences on the clamped generative model): the default objective sees 56% to 72% of each cell's marginal, and least exactly where the two objectives disagree (`south, tv_spend` flips sign). At the level it sees 68% of the media-driven response (61% in truth).
- A subsection "Why the plans differ where they do": each optimizer equalises the marginal it sees (the direct marginal at 26.9 per unit of weekly spend at the direct-only optimum, the total marginal at 41.3 at the funnel-aware one; asserted), a figure of the per-cell marginals at the historical plan against those two levels, and a split of each mediated marginal into demand created per unit of spend and demand converted, posterior next to truth. `south, tv_spend` is the one cell below the first level and above the second: its direct marginal is the twin of `north` TV's while its mediated marginal is twice as large, because a unit of its spend creates half again as much demand and `south`'s lower funnel converts it at a steeper point of the shared conversion curve than `north`'s, which runs at the highest structural spend level.
- Uplifts are small (about 1%) because the synthetic plan was already close to the posterior optimum within the box; on this dataset the funnel-aware plan is worth more on the generative truth (+0.51% vs +0.36%), the true uplifts lie inside the posterior 94% intervals, and the lower-funnel spend each plan induces is reported next to the uplift (gross, not priced).
- Figures mirror the case study: original vs optimized spend per channel/geo, media-driven response over time with HDIs, uplift densities with the truth line, plus allocation, marginal, uplift and level tables.

Also in this PR:

- `TestFunnelEffectPropagation` in `tests/mmm/test_budget_optimizer_mmm.py` pins the same facts on the conftest `FunnelEffect` fixture (ancestry of the decision vector, in-sample identity, gradient vs finite differences, the two objectives differing by exactly the mediated gradient), in both the default and the `cvm` compile modes.
- `LinearTrend` gains a `prefix` option that prefixes its changepoint coordinate and parameter names, so several trends can coexist in one model without collisions, with tests (#2903).
- The basic funnel notebook's implementation note describes what the optimizer does with a funnel effect and points at the advanced section (markdown only).
- The `BudgetOptimizer` docstring's parameter list documents `optimizable_vars`.

**Execution note.** The advanced notebook runs end to end under pymc 6.2 / pytensor 3.2.4 (`jupyter execute` with the project kernel, about 3.5 minutes); mock runner and pre-commit pass. Every quoted number is tied to the committed execution: a re-run in the same environment reproduces the tables, while dependency upgrades move posterior means at the second or third significant digit, so the prose needs re-checking against the tables after any environment change.

### Known limitations / scope notes

1. **Out-of-sample windows are a scenario choice.** Beyond the training dates the window dataset zero-fills the effect's variables ("no committed activity"); to plan against committed lower-funnel activity, `pm.set_data` the effect's variables on the optimization model before handing it to `BudgetOptimizer`. The notebook's caveats spell this out.
2. **Levers**: `lf_budget` is a `(date, geo)` series, so it cannot be an `optimizable_vars` lever (levers are one-dimensional, non-date `pm.Data`, #2884), and a lever stays outside the budget-sum constraint; a joint upper/lower-funnel decision needs a per-geo lever inside the effect and a money constraint. The induced lower-funnel spend is therefore uncounted cost in the current decision problem (reported next to the uplifts, gross).
3. `BudgetOptimizerWrapper.sample_response_distribution` and `MMMPlotSuite.budget_allocation` report direct channel contributions only.
4. **Numba**: the default backend evaluates this model's objective and gradient (unit-tested in both compile modes); the minimal reproducer of pymc-devs/pytensor#2360 still fails, so the caveats keep `compile_kwargs={"mode": Mode(linker="cvm")}` as the documented fallback.

## Related Issue

- [ ] Closes #
- [x] Related to #2654, #2879, #2813, #2884, #2895, #2898, #2900, #2903

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--2890.org.readthedocs.build/en/2890/


juanitorduz:
This feature depends on https://github.com/pymc-labs/pymc-marketing/pull/2884 and https://github.com/pymc-labs/pymc-marketing/pull/2895

daimon-pymclabs:
## Review action items

Follow-up to the detailed review, restated as actionable items with concrete fixes. Cell references use the committed `execution_count` (`In [79]` etc.) in `mmm_funnel_mueffect_advanced.ipynb`.

Nothing here is a bug in the current run: every number in the new prose reconciles with the committed outputs, and the library semantics the section relies on check out in source. Items 1, 3 and 8 are latent correctness issues (wrong on a different model, or wrong after an exception); the rest are robustness and cost.

### Should fix

- [ ] **1. Make the posterior bumps labelled, so both halves of the marginal table index the same cell** (`In [79]`)

  The posterior half bumps positionally, `x[i] *= 1.01` over `itertools.product(mask["geo"], mask["channel"])`, while the truth half bumps by label, `multiplier.loc[{"geo": geo, "channel": channel}]`. The flat order is actually `OptimizationVariables.dims == tuple(mask.dims)`, i.e. the mask's own dim order, not the `("geo", "channel")` order the labels are built in. It matches here, so the table is right; on a model with another budget dim, or a mask whose dims come through in another order, the two halves of the table would compare different cells and no assertion would fire. That table is the section's funnel-slope check, so it is the one place worth making order-proof.

  `pack` already reindexes and transposes to `self.dims`, so building the plan by label costs one extra line and makes both halves symmetric:

  ```python
  bumps = {}
  for label, (geo, channel) in zip(cell_labels, cell_index, strict=True):
      plan = mean_plan.copy()
      plan.loc[{"geo": geo, "channel": channel}] *= 1.01
      bumps[label] = response_at(optimizer.optimization_variables.pack(plan)) - at_history
  ```

  `per_unit` in `In [80]` inherits the same positional assumption (`mean_plan.transpose("geo", "channel").to_numpy().ravel()` lines up with the flat vector only under the same coincidence). Since it is now consumed alongside label-indexed rows, derive it from `cell_index` too:

  ```python
  per_unit = 0.01 * np.array(
      [float(mean_plan.sel(geo=geo, channel=channel)) for geo, channel in cell_index]
  )
  ```

- [ ] **3. Scale the optimality-guard slack to `ftol`, and derive the interior margin from `budget_bounds`** (`In [81]`)

  `at_optimum[funnel]["total"] < at_optimum[direct]["total"] - 1e-9` puts an absolute `1e-9` slack on an objective of magnitude ~536, while the run's own `ftol` is `1e-8` absolute. The inequality holds mathematically, but a legitimately converged run can land inside `ftol` of the other plan's value and trip the guard, so this can raise on a correct result. Tie the slack to the tolerance actually in force:

  ```python
  ftol = minimize_kwargs["options"]["ftol"]
  slack = 10 * ftol * max(abs(at_optimum[funnel_label]["total"]), 1.0)
  if at_optimum[funnel_label]["total"] < at_optimum[direct_label]["total"] - slack:
      raise RuntimeError("the funnel-aware optimum loses on its own objective")
  ```

  Same cell, `if float(multipliers.min()) < 0.52 or float(multipliers.max()) > 1.48` hardcodes the interior margin against the `0.5x / 1.5x` box set in `In [76]`. Change the box and the check silently stops testing the claim it is named after. Deriving it keeps the two coupled and reproduces today's numbers exactly:

  ```python
  ratio = budget_bounds / mean_plan
  lo, hi = float(ratio.sel(bound="lower").max()), float(ratio.sel(bound="upper").min())
  margin = 0.02 * (hi - lo)
  if float(multipliers.min()) < lo + margin or float(multipliers.max()) > hi - margin:
      raise RuntimeError("an optimal multiplier sits at the budget bounds")
  ```

- [ ] **8. Wrap the generative-model `set_data` round trip in `try` / `finally`** (`In [80]`, and the same pattern in `In [20]`)

  `true_response_parts` sets `channel_data` to the perturbed media, draws, and sets it back, with the restoration asserted afterwards. If anything between the two `set_data` calls raises, `gen.model` is left on the perturbed media and every later truth number in the notebook is computed against the wrong DGP, silently. Two lines:

  ```python
  with gen.model:
      try:
          pm.set_data({"channel_data": media})
          values = pm.draw([gen.model[name] for name in names])
      finally:
          pm.set_data({"channel_data": media_raw})
  ```

  `forward_with_zeroed` in `In [20]` has the same exposure: the restore is a separate statement after the `cf_gen` loop, so a raise inside the loop leaves the model perturbed.

### Worth considering

- [ ] **2. Use the exact gradient instead of a one-sided secant for the marginal table** (`In [79]`, `In [80]`)

  The prose calls the per-unit column "the quantity the optimizer equalises across cells at an interior optimum", but it is a forward 1% difference across 130 weeks of logistic saturation, so it sits below the derivative it is standing in for. The exact object is free, it is the same graph the new `test_gradient_matches_finite_differences` validates, and it is one compiled call rather than seven evaluations:

  ```python
  gradients = pytensor.function(
      [flat],
      [pytensor.grad(mean_over_posterior(name), flat) for name in response_vars.values()],
  )
  exact = {
      part: next(iter(optimizer.optimization_variables.unpack(g).values()))
      for part, g in zip(response_vars, gradients(x0), strict=True)
  }
  marginals = pd.DataFrame(
      {
          part: [float(da.sel(geo=geo, channel=channel)) for geo, channel in cell_index]
          for part, da in exact.items()
      },
      index=cell_labels,
  )
  ```

  These are already per unit of weekly spend, so the `per_unit` division drops out of the posterior half, and the `mediated > 0` and `total == direct + mediated` assertions carry over unchanged. It would make the 56%-72% marginal shares exact rather than approximately right. The truth half still needs finite differences (no gradient through the clamped generative model), which is fine as long as the prose says which side is which.

- [ ] **5. Consider `compile_kwargs` on `make_optimizer`, and watch the numba notebook job** (`In [78]`)

  `test_notebook.yml` runs the matrix `linker: [cvm, numba]` with `PYTENSOR_FLAGS: linker=${{ matrix.linker }}`, and both funnel notebooks sort into the `--notebooks mmm --start-idx 10` split, so `notebooks (numba, --notebooks mmm --start-idx 10)` compiles this objective and its gradient under numba with no `compile_kwargs`. That is the configuration caveat 4 and pytensor#2360 are about. `TestFunnelEffectPropagation` handles it properly by parametrizing default against `Mode(linker="cvm")`; the notebook has no equivalent. As of this comment both `--start-idx 10` notebook jobs are still running, so this may resolve itself; if the numba one goes red, `compile_kwargs={"mode": Mode(linker="cvm")}` in `make_optimizer` is the fix, and it is worth a one-line comment either way.

- [ ] **4. Trim the ~1.6 MB of new PNG in the two allocation figures** (`In [85]`, `In [86]`)

  The notebook goes from 7.9 MB to 9.9 MB in this PR, of which 9.2 MB is embedded images, and it is a git-tracked artifact re-executed on every stack bump. `In [85]` alone is 1.06 MB: six panels of original against optimized weekly spend, where the optimized line differs from the original by a constant multiplier that is already printed in the legend, in the allocation table and in the bar chart. `In [86]` is another 0.53 MB. Either halve the dpi on both, or drop `In [85]` and keep `In [86]`, which is the one that shows something the tables do not (the overlapping HDI bands behind the "modest tilt" claim).

- [ ] **6. Say why the tests reach for `_objective_and_grad`, or route them through the public API** (`tests/mmm/test_budget_optimizer_mmm.py`)

  Three of the four new tests in `TestFunnelEffectPropagation` call `BudgetOptimizer._objective_and_grad`, and they are the only place in the suite that touches it. That is pragmatic and it is also the seam that will break silently on refactor, while the notebook next door makes a point of proving the same facts through `optimization_variables` and `extract_response_distribution`. Either add a class-docstring line saying the private hook is deliberately the right level (it is the exact callable handed to SLSQP, which is the thing under test), or build the objective and its gradient from `extract_response_distribution` plus `pytensor.grad` the way `In [79]` does.

- [ ] **7. Two wording tweaks where the epistemics run slightly ahead of the evidence** (md cell before `In [86]` output, and the executive summary)

  "The truth lines fall inside both 94% intervals but in their lower halves" reads as coverage validation, but both plans were selected using the same posterior that produced those intervals, so they are not calibrated for the selected plan. The paragraph already names the optimizer's curse and the extrapolation, so this needs a clause, not a rewrite, e.g. "fall inside both 94% intervals, which are not coverage claims for a plan chosen on this posterior, but in their lower halves".

  In the executive summary, "its plan is worth more on the generative truth (+0.51% against +0.36%...)" is one dataset and one fit, and a 0.15pp margin at n=1 does not license a general ordering. The "on one synthetic dataset it is not separable from" caution is currently attached to the 2x overstatement only. Adding "on this dataset" to the summary sentence would carry it forward.

### Not action items, listed so they are not lost in review

The loop-closure assertions are the strongest part of the section and I would not touch them: `at_optimum` against the posterior-predictive recomputation against `-scipy_result.fun` at `rtol=1e-6` pins the optimizer, the re-weighted plan and scipy's own value to one number; `mediated_zero` cross-checked against `response_at(zeros)` ties the both-off yardstick to the optimizer's own graph rather than to a parallel implementation; and `total == direct + mediated` plus the strictly-positive mediated bump is the right minimal pair to prove the intervention reaches the effect. Caveat 4's three requirements and two non-capabilities are accurate against the source, including the `optimizable_vars` one-dimensional non-date restriction. Dropping the #2804 pointer from the basic notebook is right, that issue is closed.

juanitorduz:
Thanks for the itemised review. Everything below is on the branch (commits 028e3540 and 0ea144f4), the notebook re-executed end to end (the fits reproduce the committed numbers exactly, so nothing outside the edited cells moved).

**Should fix**

- [x] **1. Label-indexed marginals.** The posterior half no longer bumps positionally: it is now the exact gradient (item 2), unpacked with `optimization_variables.unpack(...)[optimizer.channel_data_var]` and read by `.sel(geo=..., channel=...)` for each `(geo, channel)` in `cell_index`, so both halves of the table index the same cell by label. `per_unit` is derived from `cell_index` (`0.01 * float(mean_plan.sel(geo=geo, channel=channel))`) and only used on the truth half; the posterior half is already per unit.
- [x] **3. Guard slack and interior margin.** Both taken as written: `slack = 10 * ftol * max(abs(value), 1.0)` with `ftol = minimize_kwargs["options"]["ftol"]` for the two "wins on its own objective" guards, and the interior margin from `budget_bounds / mean_plan` (`margin = 0.02 * (hi - lo)`), which reproduces today's [0.52, 1.48].
- [x] **8. `try` / `finally`.** Both round trips (`true_response_parts` and `forward_with_zeroed`) restore `channel_data` in a `finally`; the separate restore after the `cf_gen` loop is replaced by an assertion that the media are back.

**Worth considering**

- [x] **2. Exact gradient.** Adopted, with one wrinkle worth recording: `pytensor.grad` does not go through an xtensor graph (`XTensorVariable` has no `ones_like`; on `.values` the pullback of `XReduce` is not implemented), so the graph is lowered first, exactly as `BudgetOptimizer._compile_objective_and_grad` does: `rewrite_graph(mean_over_posterior(name).values, include=("lower_xtensor", "canonicalize", "stabilize"))`, then `pytensor.grad(..., flat)`, one compiled function for the direct, mediated, total and `funnel_lf_spend` gradients. The `mediated > 0` and `total == direct + mediated` assertions carry over on the gradients. The truth half uses central differences of +/-0.5% (two forward passes per cell); on this model they agree with an exact `pytensor.grad` through the clamped generative model to four significant digits (e.g. `north, tv_spend` total 34.257 both ways, where the forward 1% secant gave 34.100), and the prose says which side is which. The shares the default objective sees are unchanged at the integer percent (56% to 72%).
- [x] **4. PNG size.** Kept as they are on purpose: the two figures mirror the case study's presentation (spend per channel/geo, response over time with HDIs), which is what the section is written against.
- [x] **5. Compile mode.** The `notebooks (numba, --notebooks mmm --start-idx 10)` job went green on the previous push (run 32137544126), so the notebook's objective and gradient compile under numba in CI as they do locally, where the default backend is numba as well; the new gradient function is exercised the same way. No `compile_kwargs`. A one-line comment in `make_optimizer` names the default mode and points at the caveats' `Mode(linker="cvm")` fallback.
- [x] **6. `_objective_and_grad` in the tests.** Documented in the class docstring: it is the exact callable handed to SLSQP (objective and gradient in one compiled function), so the tests pin what the solver sees; the notebook proves the same facts through the public API.
- [x] **7. Wording.** Both clauses added ("which are not coverage claims for a plan chosen on this posterior"; "on this dataset" in the executive summary and in "Reading the results").

**Also new**, prompted by a question from @juanitorduz on the allocation bar chart (why the two objectives split on `north, tv_spend` and `south, tv_spend`): a subsection "Why the plans differ where they do" that reads the exact marginals at both optima (each optimizer equalises the marginal it sees: the direct one at 26.9 per unit of weekly spend, the total one at 41.3, asserted to 1e-3), a figure of the per-cell marginals at the historical plan against those two levels, and a table splitting each mediated marginal into demand created per unit of spend (the `funnel_lf_spend` gradient) and demand converted, posterior next to truth. `south, tv_spend` is the one cell that sits below the direct level and above the total one: its direct marginal is the twin of `north` TV's (24.9 vs 24.5) while its mediated marginal is twice as large (19.8 vs 9.6; 16.0 vs 11.1 in truth), because a unit of its spend creates half again as much demand (a smaller cell on the same pooled TV curves) and `south`'s lower funnel converts it at a steeper point of the shared conversion curve than `north`'s, which runs at the highest structural spend level.

juanitorduz:
We will get additional improvements form https://github.com/pymc-labs/pymc-marketing/pull/2898 :)

cetagostini (approved):
Hey all good, great notebook! a few things:
1. Feel its too verbose and could be split at least in two notebooks.
2. There are many phrases which I feel don't belong on a notebook of funnel effects, e.g: "optimizer returns any model variable as a graph of the decision vector (optimizer.optimization_variables.flat), conditioned on the posterior draws, and optimization_variables.pack turns a labelled plan into that vector." - Sure, okay but why that's even relevant here? is a notebook about deep-dive optimizer stuff?

Let's merge and open an issue to polish a bit, I feel it's great but as soon as you see the density becomes hard to follow. Many words and phrases are structure in a way a bit heavy to read, I needed two or three reads to grasp what was saying.

I'm happy with an issue and review before next release!

juanitorduz:
We are waiting for refinement from https://github.com/pymc-labs/pymc-marketing/pull/2908

Issue #2886, opened 2026-08-17
Formalise the MMM's implicit causal DAG as a composable NetworkX graph


drbenvincent:
## Summary

Every `MMM` already encodes a causal assumption: every media channel and every control is a **direct cause of the outcome**. That assumption is implicit. It lives in how `build_model` wires terms into `μ`, not in a graph the library can plot, inspect, or identify against.

This issue proposes that **building an MMM also builds a NetworkX causal DAG** that makes that assumption explicit, and that the DAG **grows with the model**. The first slice attaches the graph to the `MMM` itself (the default star: media + controls → outcome). The next slice gives `MuEffect` a causal fragment and has `MMM.add_mu_effect` compose it. The third slice does the same for funnel effects, which is where the graph stops being a star and starts representing mediation, latents, and measurement.

The graph is for **graphviz plotting** and, later, **causal identification** (d-separation, backdoor, path enumeration). It is a representation of what the objects actually assume, not a second independently-authored artefact.

## Problem

A vanilla `MMM` with channels `{tv, social}` and controls `{price, holiday}` is, causally, this graph:

```
tv      →  y
social  →  y
price   →  y
holiday →  y
```

Nothing in the object says so. A user who wants that picture today either draws it by hand or passes a DOT string into `MMM(dag=…)` — and that string is **not** derived from the model. It is an optional overlay used only to prune `control_columns` via `CausalGraphModel` / dowhy. If you omit it, the implicit star still holds; if you pass a DAG that disagrees with the regression, the two representations drift.

The gap gets worse as the model is composed:

- `add_mu_effect(ControlMuEffect(…))` or `MediaMuEffect(…)` adds more direct causes. The implicit graph should grow. It does not.
- A funnel `MuEffect` (notebooks `mmm_funnel_mueffect`, `mmm_funnel_mueffect_advanced`; test fixture in `tests/mmm/conftest.py`) adds **mediated** structure: upper spend → latent demand → lower-funnel spend → outcome, plus measurement / indicator edges. The advanced notebook already maintains a hand-written `DAG_EDGES` list, builds a `networkx.DiGraph` from it, and runs `nx.is_d_separator` / backdoor surgery **beside** the model. The model does not own that graph.
[...]

PR #2891, opened 2026-08-17
Add implied causal DAG on MMM

drbenvincent:
Towards #2886

## Summary
- Add `MMM.causal_graph` (NetworkX) rebuilt on access as the MMM star union `MuEffect.causal_graph_fragment` outputs (empty default).
- Add `MMM.plot_causal_graph()` for graphviz rendering before `build_model` / `fit`.
- Promote `networkx` to core dependencies.
- Document in `mmm_quickstart` with an inline causal DAG figure.

Part of #2886 Phase 1. Does not touch `MMM(dag=...)` / `CausalGraphModel`. Built-in effect fragment overrides and funnel notebooks are follow-ups.

## Test plan
- [x] `uv run pytest tests/mmm/test_causal_graph.py`
- [x] `uv run pre-commit` on changed files
- [ ] CI

----
📚 Documentation preview 📚: https://pymc-marketing--2891.org.readthedocs.build/en/2891/


drbenvincent (commented):
## Critical review (Bugbot + manual)

**Verdict:** Request changes — one must-fix (see follow-up commit).

### Must fix

**Season node collision** (`pymc_marketing/mmm/causal_graph.py`)

When `yearly_seasonality` is set, the star builder always appends the fixed `season` node. If a channel or control column is also named `season`, the DAG shows a single `season → target` edge while the fitted MMM includes both that column and the Fourier seasonality term in μ. Same class of bug as `target_column` matching a predictor — should reject at graph build time.

### Should fix (non-blocking)

- `plot_causal_graph` / `causal_graph_to_graphviz` return `object` — prefer `graphviz.Digraph` under `TYPE_CHECKING`.
- `ImportError` for missing graphviz should mention system Graphviz binaries (`dot`), not only `pip install graphviz`.

### Deferred (Phase 2/3, per #2886 scope)

- Built-in `MuEffect` subclasses still return empty fragments — `add_mu_effect(ControlMuEffect(...))` does not add edges yet.
- No test that `causal_graph` reflects `dag=` control pruning (intentionally separate from `MMM(dag=...)` / DoWhy).

### Looks good

- Rebuild-on-access + `compose_causal_graph` is the right seam for Phase 2 fragments.
- Tests cover geo dim, cycles, snapshot semantics, and plot smoke.
- Notebooks separate implied causal DAG from PyMC computation graph; multidimensional example shows `geo` is not a causal node.

drbenvincent:
## Fixes applied (commit efd82ea5)

Addressing the review above:

### Must fix — done
- **Season node collision:** `build_mmm_star_graph` now raises `ValueError` when `yearly_seasonality` is set and a channel or control column is named `season` (mirrors the existing `target_column` clash guard). Added `test_build_star_graph_raises_when_season_column_conflicts_with_fourier`.

### Should fix — done
- **`plot_causal_graph` / `causal_graph_to_graphviz`** return `graphviz.Digraph` via `TYPE_CHECKING` imports instead of `object`.
- **`ImportError`** message now mentions both the Python `graphviz` package and system Graphviz binaries.

### Not changed (deferred)
- Built-in `MuEffect` fragment overrides — Phase 2.
- `dag=` / DoWhy interaction tests — out of scope for this PR.

All 13 tests in `tests/mmm/test_causal_graph.py` pass; pre-commit green on changed files.

drbenvincent:
The issue #2886 outlines the desired functionality in terms of building a network DAG representation of the MMM as it is being built, programmatically.

Currently this PR implements this just for when you build an `MMM` object. The functionality is shown in 2 of the docs notebooks.

For example in `mmm_quickstart.ipynb`:

Screenshot 2026-08-18 at 09 33 03

and in `mmm_multidimensional_example.ipynb`:

Screenshot 2026-08-18 at 11 10 42

---

My question is: Do we want to review and merge this PR now to bank the progress for this simple case? Or should I push forward and add for `MuEffects` and subclasses? As in, small baby steps, or flesh out the full capability before we push code to main? @juanitorduz @timbo112711 @PabloRoque

drbenvincent:
The main reason why it could be a bad idea to merge into main now is that we are only offering causal graphs of flat `MMM` objects. So if a user then adds more complexity, then the causal graph will no longer be correct

drbenvincent:
@juanitorduz - just doing some morning review. I think this is a useful and important feature to have, but am I right in thinking I should hold off on this until there is a little more clarity in terms of `MuEffects` subclasses (including the proposal #2901)?

juanitorduz:
> @juanitorduz - just doing some morning review. I think this is a useful and important feature to have, but am I right in thinking I should hold off on this until there is a little more clarity in terms of `MuEffects` subclasses (including the proposal #2901)?

yes, let's hold on on this one and lets think about the design integration :) This will be very helpful 💪

PR #2894, opened 2026-08-17
fix(mmm): correct the identity-link mean scale for TruncatedNormal

### LINKED ISSUE #2834: IdentityLinkSpec.mean_correction is unconditionally 1.0, so TruncatedNormal contributions are silently off by the truncation factor

anevolbap:
Fixes #2834

## Problem

`IdentityLinkSpec.mean_correction` returned `1.0` for every likelihood. That is right for `Normal` and wrong for `TruncatedNormal`, where clipping shifts `E[y]` off `mu`. Mean-scale contributions were off by the truncation offset, with no warning.

## Changes

- `LinkSpec.to_mean_scale` applies the correction to a contribution dataset. Identity link: adds the truncation offset to the `intercept` term. Log link: multiplies by `exp(sigma**2 / 2)`, as before.
- `LinkSpec.mean_scale_factor` returns the correction as a factor, for `incrementality.py`, which folds it into a scale. It raises for `TruncatedNormal`, where the correction is an offset.
- Both go through `_validate_mean_defined`, so both reject the same likelihoods: `Censored` and `Scaled` wrappers, `StudentT` with `nu <= 1`, and `LogNormal` under the identity link. An unknown likelihood warns and stays uncorrected.
- `mean_correction` is deprecated. Log link: it still returns `exp(sigma**2 / 2)`. Identity link: it raises, since the right correction depends on the likelihood, which this signature cannot see.
- `mu` is registered as a `Deterministic` under the identity link too. The correction is pointwise in `mu`, and summing the contribution `Deterministic`s is not safe, because `MuEffect.create_effect` only has to return its term.

## Behaviour changes

- Every identity-link posterior now carries `mu`, and identity is the default link. The identity branch does not transpose, unlike the log branch: on a panel model the dims are `(chain, draw, country, date)`.
- A `MuEffect` that registers a `Deterministic` named `mu` used to build. It now fails in `build_model` with `Variable name mu already exists` and has to be renamed.
- Posterior predictive sampling on the fitting data now uses the `mu` stored in the posterior. It does not rebuild `mu` from the contributions.
  - A hand edit to a contribution in the posterior, for example to `intercept_contribution`, no longer changes the predictive draws.
  - Two `LogNormalPrior` tests in `tests/mmm/test_mmm.py` used such an edit to force the sign of `mu`. They now set `mu` directly.
- Models fitted before this change have no `mu` in the posterior. Under `TruncatedNormal` with `central_tendency="mean"` they raise, with a message to refit or use the median.
- `to_mean_scale` is now the abstract method on `LinkSpec`, and `mean_correction` is not. A `LinkSpec` subclass outside this repo that implements only `mean_correction` no longer instantiates.

## Differences from the issue

- The issue lists `NegativeBinomial` as needing no correction. #2841 left it out of `RESPONSE_SCALE_LIKELIHOODS` (discrete likelihood, non-integer scaled target), so this follows the merged set.
- `StudentT` raises when `nu <= 1`, because no mean exists. The issue does not mention this. One draw at or below 1 is enough, and the message reports how many draws are affected.


pymc_marketing/mmm/link.py
```diff
@@ -322,13 +414,135 @@ def create_media_contribution_deterministic(
...
+
+        # The textbook ratio cancels to zero once the truncation point is about
+        # ten sigma from mu, which the identity link permits, and returns nan
+        # there.  erfcx is the scaled complementary error function, which keeps
+        # the one-sided cases exact and is a ufunc, so it stays vectorised.
+        # scipy.stats.truncnorm is exact too but roughly 2000x slower, which
+        # matters on a full posterior.
+        root_two = np.sqrt(2.0)
+        if np.isposinf(bounds["upper"]):
+            return sigma * np.sqrt(2 / np.pi) / erfcx(alpha / root_two)
```
anevolbap, line 1045: Not the textbook `phi/Phi` ratio, deliberately. That form cancels to zero once the truncation point is about ten sigma from `mu`, which the identity link permits: at `mu = -10, sigma = 1, lower = 0` it is already `inf` while the true value is `10.098093`. `erfcx` agrees with `scipy.stats.truncnorm` to ten decimals and is about 2000x faster, 0.12s against 166s on a 2M-point posterior, which matters since this runs on the full posterior.

tests/mmm/test_mmm.py
```diff
@@ -1010,7 +1010,10 @@ def create_data(self, mmm):
         pass
 
     def create_effect(self, mmm):
-        return as_xtensor(pt.zeros(1), dims=["date"])
+        # Date-length, like the sibling effect below. A length-1 tensor claiming
+        # the date dim only survives while nothing forces the shape, which stops
+        # being true once "mu" is a registered Deterministic.
+        return as_xtensor(pt.zeros(mmm.model.dim_lengths["date"]), dims=["date"])
```
anevolbap: Unrelated to the fix, but registering `mu` forced it. This returned `as_xtensor(pt.zeros(1), dims=["date"])`, a length-1 tensor claiming a dim of length 14, which survives only while nothing forces the shape. Once `mu` is a registered Deterministic it fails in `_sample_with_deterministics`. The sibling effect below already uses `mmm.model.dim_lengths["date"]`, so I matched it. Worth knowing that user-defined effects could have the same latent problem.

pymc_marketing/mmm/link.py
```diff
@@ -222,34 +264,118 @@ def create_total_response_deterministic(
...
+            f"{type(self).__name__}.mean_correction is deprecated, use "
+            f"to_mean_scale instead. A single factor cannot express the "
+            f"identity-link correction, which depends on the likelihood and "
+            f"is additive rather than proportional.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self._mean_ratio(posterior, output_var)
+
+    def mean_scale_factor(
```
anevolbap, line 367: Separate from `to_mean_scale` because `incrementality.py` needs a factor to fold into a scale, not a dataset to apply a correction to. Raises where the correction is additive so the two entry points can't silently disagree.

pymc_marketing/mmm/link.py
```diff
@@ -383,13 +509,150 @@ def create_media_contribution_deterministic(
...
+            offset = self._truncation_offset(posterior, likelihood, output_var)
+            corrected = dataset.copy()
+            corrected[BASELINE_PART] = corrected[BASELINE_PART] + offset * target_scale
+            return corrected
+
+        if dist_name == "StudentT":
+            _check_studentt_mean_exists(posterior, likelihood, output_var)
+            return dataset
+
+        if dist_name in RESPONSE_SCALE_LIKELIHOODS:
```
anevolbap, line 710: Issue lists `NegativeBinomial` as needing no correction, but #2841 already excludes it here, so this follows the merged set instead of the issue text.

pymc_marketing/mmm/link.py
```diff
@@ -72,6 +74,46 @@ class LinkFunction(StrEnum):
...
+#: factor, so it cannot be folded into a scale.
+ADDITIVE_CORRECTION_LIKELIHOODS = frozenset({"TruncatedNormal"})
+
+#: Key of the baseline term in a counterfactual contribution dataset.  Under
+#: the identity link a correction that belongs to the noise distribution is
+#: added here rather than spread across the component contributions.
+BASELINE_PART = "intercept"
+
+
+def _check_studentt_mean_exists(
```
anevolbap, line 87: Not in the issue: `StudentT` has no mean when `nu <= 1`. `nu` can be a fitted prior, so this raises instead of silently returning the median.

tests/mmm/test_linear_regression.py
```diff
@@ -86,6 +86,7 @@ def test_fancy_linear_regression(
         "draw",
         "gamma_control",
         "intercept_contribution",
+        "mu",
```
anevolbap, line 89: `mu` now shows up in the posterior under the identity link too. Flags this for anyone else asserting an exact variable set.

juanitorduz (approved):

daimon-pymclabs:
Review of the math and test coverage. The `erfcx` one-sided identities check out algebraically (`sqrt(2/pi)/erfcx(alpha/sqrt2)` is exactly `phi/(1-Phi)`), the two-sided scipy fallback is correctly parameterized, and `offset * target_scale` matches the fact that contributions come back in original units. Test coverage here is unusually good.

Four findings, one of them a real bug.

**1. Bug: fixed non-scalar `sigma` bypasses `_label_bound` (`pymc_marketing/mmm/link.py:1053`)**

A fixed non-scalar `sigma` is used raw, while the bounds go through `_label_bound`. Something like `Prior("TruncatedNormal", lower=0, sigma=np.array([...]), dims="date")` then hits exactly the positional-broadcast misalignment that `_label_bound` exists to prevent: it either raises an opaque shape error or aligns against the wrong axis of `mu`, which the identity link deliberately leaves untransposed. Suggest running `sigma` through `_label_bound` too, and adding the array-sigma case to `test_array_valued_bounds_are_handled`.

**2. Brittle wrapper detection (`link.py:561`)**

`getattr(likelihood, "parameters", None) is None` is duck-typing the `Censored`/`Scaled` detection. If pymc-extras ever gives those wrappers a `parameters` attribute, the hard error silently turns into wrong mean-scale numbers. An `isinstance` check against the wrapper types fails loudly instead.

**3. Missing regression test for out-of-sample `mu`**

Posterior predictive now reads `mu` from the trace, so out-of-sample correctness rests on PyMC's volatility detection marking `mu` volatile when the data changes. That behavior is load-bearing and currently untested. A test that OOS prediction with new `X` of the same length differs from the in-sample draws would pin it.

**4. Minor: draw count in `_check_studentt_mean_exists`**

`nu_values.size` is reported as a draw count, which overstates it when `nu` carries extra dims.

<sub>Posted by Daimon on behalf of @juanitorduz.</sub>

juanitorduz:
All four validated and fixed in 3915387.

**1. Fixed non-scalar `sigma` — confirmed, real bug.** Reproduced on a panel posterior with `mu` dims `(chain, draw, country, date)`:

```
per-country sigma: ValueError operands could not be broadcast together with shapes (1,1,2,3) (2,)
per-date sigma:    silently aligned to the trailing axis
```

Same two failure modes as the bounds, for the same reason. `_label_bound` is now `_label_parameter` and `sigma` goes through it. New tests: `test_a_fixed_array_sigma_follows_its_dim_too` (per-country scale on an untransposed `mu`, checked per country against `scipy.stats.truncnorm`) and `test_an_unnameable_fixed_array_sigma_is_refused`.

Kept as two separate tests rather than a row in `test_array_valued_bounds_are_handled`: that test's posterior is `(chain, date)`, where `date` is trailing and positional broadcasting succeeds by luck — adding a row there would not have caught this.

**2. Wrapper detection — agreed.** Now type-first, duck-type second:

```python
WRAPPER_LIKELIHOODS: tuple[type, ...] = (Censored, Scaled)
...
if (
    isinstance(likelihood, WRAPPER_LIKELIHOODS)
    or getattr(likelihood, "parameters", None) is None
):
```

`isinstance` alone would drop the catch-all for a wrapper this module has not been told about, so both arms stay. Verified today that neither `Censored` nor `Scaled` has `parameters`, which is exactly the assumption that should not be load-bearing.

**3. OOS regression test — added.** `TestRegisteredMuAndPosteriorPredictive::test_new_spend_still_moves_the_predictive_draws`. Counterfactual `X` of the *same* length, all media zeroed, so a shape change cannot do the volatility detection's job. Measured discriminating power: the predictive mean moves 58% (2.934 → 1.239), so a frozen `mu` fails loudly rather than marginally.

**4. Draw count — agreed.** `_check_studentt_mean_exists` now reduces over the non-`(chain, draw)` dims with `.any()` before counting: a draw is unusable as soon as one of its cells is at or below 1, and the raw cell count overstated it. The existing test was also using `date` as a stand-in for `draw`, which is what let the bug hide — it now uses `draw`, and `test_a_hierarchical_nu_is_counted_in_draws_not_in_cells` covers `nu` with an extra `country` dim (3 of 6 cells, 2 of 3 draws, reports "2 of 3").

`702 passed, 15 skipped` across `test_link`, `test_spend_reach`, `test_incrementality`, `test_mmm`, `test_counterfactual`, `test_linear_regression`, `tests/data`; `368 passed, 8 skipped` re-run after rebasing onto the `main` merge. `pre-commit` clean.

daimon-pymclabs (approved):
Approving on behalf of @juanitorduz.

Re-reviewed at `39153873`. All four findings from my earlier comment are fixed, and the fixes go beyond the ask: `_label_bound` became `_label_parameter` and now covers `sigma`, the wrapper check keeps the duck-type as a catch-all behind the `isinstance`, the hierarchical `nu` count reduces over non-sample dims before counting, and the OOS test uses a same-length counterfactual `X` (the only version of that test that actually pins volatility detection, since a shape change would pass for the wrong reason). I checked that the one-sided `erfcx` path stays label-aligned and the two-sided path goes through `xr.apply_ufunc`, so scipy never sees bare arrays.

Three small non-blocking notes for a follow-up:

1. `link.py:1273` asymmetry: the log link's `_mean_ratio` still hard-refuses a fixed `sigma` ("give sigma a prior"), even though the identity link now handles one via `_label_parameter`. The same machinery would work there.
2. Test gap: the fixed array `sigma` is only exercised in the one-sided branch. The two-sided `apply_ufunc` path, with `sigma` and a bound on different dims, is where a dim-order surprise would actually bite.
3. `link.py:225`: the error message lost its noun, "cannot place the 1-dimensional 'sigma': the likelihood declares dims ()".

Review is from reading the diff, not running the suite.

daimon-pymclabs:
Read `3915387`. All four are addressed, and two of the fixes are better than what I asked for.

**1.** `_label_parameter` on `sigma` is the right shape for the fix, and the reasoning for keeping the two new tests separate from `test_array_valued_bounds_are_handled` holds: that fixture's `(chain, date)` posterior has `date` trailing, so positional broadcasting succeeds there for the wrong reason and a new row would have been a passing test over a live bug. `test_an_unnameable_fixed_array_sigma_is_refused` also pins the `ndim > len(dims)` arm, which the bounds path had and `sigma` did not.

**2.** Keeping both arms is correct and I was wrong to suggest replacing rather than prepending. `isinstance` alone narrows the check to wrappers this module already knows, which is the opposite of what the catch-all is for.

**3.** Same-length counterfactual `X` is the detail that makes this test actually test volatility detection rather than shape inference, and quantifying the discriminating power (2.934 -> 1.239) means a future regression fails by a margin instead of by a tolerance.

**4.** Reducing with `.any()` over the non-sample dims matches the semantics of the message. One small thing, not blocking: when `nu` has no `chain`/`draw` dims at all the branch reports "posterior values" against the raw cell count, which is the honest wording for that case, so this reads as deliberate rather than as the bug I flagged.

Nothing further from me.

<sub>Posted by Daimon on behalf of @juanitorduz.</sub>

PR #2895, opened 2026-08-17
feat(mmm): an objective that sees mu-effect contributions (total_response_original_scale)

PabloRoque:
## Description

**An objective the budget optimizer can point at when part of the response travels through a `MuEffect`.**

`total_media_contribution_original_scale` is built from the `channel_contribution` tensor alone (`link.py`), so any response routed through an effect — a funnel mediator, a promotional lever — is invisible to it. A budget optimized against it undervalues whatever drives that effect, and does so **silently**: you get a plausible allocation that is wrong in a direction nothing reports.

This registers `total_response_original_scale` when the model has mu effects:

```python
    ...,
    response_variable="total_response_original_scale",
)
```

Two things already in the repo were waiting on it:

- **#2884** shipped `LeverVariable` with a docstring and PR description pointing users at `response_variable="total_response_original_scale"` — a variable that did not exist. A lever scored against the media-only objective has an identically zero gradient, so #2884's reachability check refuses it outright. The capability was unusable end to end.
- **#2890** measured the funnel case on a real two-channel × three-geo model: the default objective sees **68%** of the media-driven response (61% against the generative truth), the two objectives move the mix differently in two of six cells, and the sign flips for `south, tv_spend`. That notebook currently hand-registers a scalar `Deterministic` to work around it.

### How it is built

A concrete helper on `LinkSpec`, alongside `create_media_contribution_deterministic`, so the definition lives with the other link-dependent objective:

```python
pmd.Deterministic(
    "total_response_original_scale",
    self.original_scale_transform(mu_var, target_scale).sum(dim="date").sum(),
)
```

It reuses the existing `original_scale_transform` (`variable * target_scale` for identity, `exp(variable) * target_scale` for log), so it is correct under both links **by construction** rather than by a per-link formula that could drift. Because `mu_var` already includes every additive effect, no rule is needed for which effects count — promos, funnel and reach/frequency are all covered by the same line.

This also closes the gap #2890 hit directly: `y_original_scale` is unusable as an objective (observed RV in the graph) and `mu` is **not a registered variable** under the identity link, so there was no scalar to point at.

### Scope, stated honestly

This is an *objective*, not an attribution. It includes the baseline, which is an additive constant for the default mean utility — so the `argmax` over budgets is unchanged — but **not** for a risk-adjusted utility, which is called out in the docstring along with the fact that in an optimization model the date sum spans the `adstock_periods` carry-over tail. For the direct-versus-mediated decomposition, `Incrementality` already computes proper counterfactuals and the docstring points there.

### Two deliberate decisions

**Gated on the model having mu effects.** #2621 gated the equivalent deterministic on `isinstance(effect, OptimizableMuEffect)`. That gate needs a protocol that does not exist yet and, more importantly, would **not** fire for a funnel effect — which is a plain `DataVarMuEffect`, and precisely why #2890 still had to hand-roll it. The broader gate serves every case with one rule while leaving plain media models' posteriors untouched.

**The default `response_variable` is unchanged.** Switching it is a behaviour change that deserves #2890's numbers to validate it, so it belongs with the funnel work. Meanwhile incorrect usage already fails loudly: the optimizer's construction-time reachability check errors when a declared lever cannot reach the chosen objective.

### Backward compatibility — one narrow note

Under the identity link, `mu` was previously **never computed as a deterministic**; it is only named and feeds the likelihood. This objective pulls it into `compute_deterministics` for the first time, so a contribution carrying a `date` dim of the wrong length is now evaluated rather than ignored. That surfaced a test fixture declaring `pt.zeros(1)` with `dims=["date"]` in a 14-date model — harmless before, an assertion error now, and the one-line fixture change here. Any user effect relying on length-1 `date` broadcasting is affected the same way, and fails loudly at fit time.

## Verification

452 tests green across `test_link.py`, `test_mmm.py`, `test_budget_optimizer.py`, `test_budget_optimizer_mmm.py`, `test_optimization_variables.py`. Lint and mypy clean.

Tests at four levels:

1. The gate — registered with mu effects, absent without, under both links.
2. Log link cross-checked against the independently built `y_original_scale`.
3. Identity link: moving an effect changes the response while leaving `total_media_contribution_original_scale` **bit-identical** — the silent bug encoded directly as a test.
4. End to end: a fitted model with a mediated effect optimized under both objectives produces different allocations, and the effect-aware one funds the channel driving the mediator. This is #2890's finding reproduced as a unit test; it fails if the feature breaks in either direction.

## Related

- Unblocks #2890 (its hand-rolled `Deterministic` becomes this variable; the published numbers are the oracle for it)
- Completes #2884, whose `LeverVariable` documented this variable before it existed
- Next: the `OptimizableMuEffect` wiring (promos) and monetary spend variables (funnel), which both depend on this

## Type of change

- [x] New feature (registered only when the model has mu effects; the default response variable is unchanged)

----
📚 Documentation preview 📚: https://pymc-marketing--2895.org.readthedocs.build/en/2895/


pymc_marketing/mmm/link.py
```diff
@@ -160,6 +165,56 @@ def create_media_contribution_deterministic(
...
+        intended plan.
+
+        This is a response total, not a media attribution: for the
+        direct-versus-mediated decomposition see
+        :class:`~pymc_marketing.mmm.incrementality.Incrementality`, which
+        computes proper counterfactuals.
+        """
+        pmd.Deterministic(
+            "total_response_original_scale",
+            self.original_scale_transform(mu_var, target_scale).sum(dim="date").sum(),
```
juanitorduz: do we need to make this "date" ? what about date_col from the MMM class?

daimon-pymclabs (commented):
Reviewed at `036606e` (the explicit-`response_variable=None` commit). The core change is right and well-argued, so most of this is about scoping the warning and tidying two loose ends.

`total_response_original_scale` built as `original_scale_transform(mu_var, target_scale)` is correct for both links by construction rather than by a per-link formula, and putting it on `LinkSpec` next to `create_media_contribution_deterministic` is the correct home. The identity-link test that asserts the media objective stays bit-identical while the response moves is the best test here: it encodes the bug, not the fix. The `.get` vs `not in` fix in the latest commit is also the right call, and the docstring rationale on it is worth keeping.

Five things I'd raise, roughly in order of how much they matter.

## 1. The warning fires for mu effects that are not budget-dependent

This is the one I'd block on.

`self.mu_effects` is not a proxy for "response travels through an effect the media objective misses". `add_events` appends an `EventAdditiveEffect`, `add_mu_effect(LinearTrendEffect(...))` appends a trend, and `FourierEffect` / `ControlMuEffect` are all `MuEffect` subclasses. None of them are downstream of `channel_data`. For those models `total_media_contribution_original_scale` is exactly correct, and the suggested replacement only adds a budget-independent baseline to the objective, which does not move the argmax.

The consequence is that every existing user with an events or trend model now gets a `UserWarning` on every `budget_optimizer()` call telling them their objective undercounts, and offering them the choice of switching objectives or passing "the default explicitly to silence this warning". That is a behavior change to workflows that were already correct. The PR description's "the default `response_variable` is unchanged" is true of the value but undersells it, since the observable behavior did change for those users.

**Suggested fix.** The predicate you want is graph reachability, not effect-list non-emptiness: does the effect's contribution have the optimization variable among its ancestors? The machinery is already imported in this file and used for exactly this shape of question in step 8b of `_set_optimization_model`:

```python
reachable = set(ancestors([self._pymc_model[self.response_variable]]))
```

The same check applied per effect fires correctly for a `DataVarMuEffect` funnel (its contribution is downstream of `channel_data`) while staying quiet for events, trends, Fourier, and controls. Roughly:

```python
from pytensor.graph.traversal import ancestors

def _effects_carry_media_response(self, pymc_model) -> bool:
    """True when some mu-effect contribution is downstream of the media data."""
    media = pymc_model["channel_data"]
    return any(
        media in set(ancestors([pymc_model[name]]))
        for name in self._mu_effect_contribution_names(pymc_model)
    )
```

then gate the warning on that instead of on `if self.mu_effects:`. The one wrinkle is that this needs a built model, so it has to run where `pymc_model` exists rather than in `_resolve_response_variable` as currently placed.

If that is more surgery than you want here, the clean split is to land the deterministic plus the docs in this PR and land the warning separately, where the reachability check and its own tests can live together. Registering the deterministic broadly is cheap and harmless; warning broadly is not.

## 2. `.sum(dim="date").sum()` in `create_total_response_deterministic`

Two separate parts to the `"date"` question raised in review.

The dim name is safe today. `_generate_and_preprocess_model_data` builds `coord_dict = {"date": dates}`, so `date_column` names a dataframe column and never renames the model dim. No bug.

But the double reduction is redundant. The second `.sum()` with no `dim` already reduces over every dimension, so `.sum(dim="date")` buys nothing except an implicit assertion that a `date` dim exists. And `MMM.budget_optimizer` already carries a configurable `date_dim` ("Honour a caller-supplied `date_dim` rather than assuming 'date'"), so hardcoding `"date"` one layer down is inconsistent with a convention this codebase already adopted.

**Suggested fix.** Drop the first reduction:

```python
pmd.Deterministic(
    "total_response_original_scale",
    self.original_scale_transform(mu_var, target_scale).sum(),
)
```

That resolves the review comment and removes the question. If the assertion is deliberate, keep it but say so in a comment and thread the same `date_dim` through rather than hardcoding. Worth contrasting with the `sum(dim="date")` in `create_media_contribution_deterministic`, where it is load-bearing because of the `* target_scale` broadcast that follows: that one should stay exactly as it is.

## 3. The `optimize_budget` warning path is untested

`stacklevel=3` is correct for both entry points, since both call `_resolve_response_variable` one frame below the user. But every test in `TestDefaultObjectiveWarning` goes through `MMM.budget_optimizer`. `BudgetOptimizerWrapper.optimize_budget` also calls `self.model_class._resolve_response_variable(...)` and has no coverage for either the warning or the blame frame, so a future delegation layer around it would silently break the guarantee `test_warning_points_at_the_caller` exists to protect.

**Suggested fix.** Parametrize over the two entry points rather than adding a parallel test:

```python
@pytest.fixture(params=["budget_optimizer", "optimize_budget"])
def warn_entry_point(self, request, dummy_df, fitted_mmm_mediated):
    """Both public paths resolve the objective, so both must warn and both must
    blame the caller."""
    ...
```

and run `test_warns_when_the_objective_is_left_unset`, `test_silent_when_the_objective_is_chosen_explicitly`, and `test_warning_points_at_the_caller` through it. If point 1 lands, this is also where the events/trend negative case belongs.

## 4. Direct `BudgetOptimizer(...)` construction stays silent

That path is documented in the class docstring and cannot see `mu_effects`, so the guard genuinely only exists at the MMM layer. Not fixable at that layer, and I do not think it should be. But it is worth one line in the `response_variable` docs noting that the warning is an MMM-layer convenience, so nobody reads it as a general safety net and concludes silence means correctness.

## 5. `DEFAULT_RESPONSE_VARIABLE` is a half-introduced constant

`BudgetOptimizer.response_variable` still hardcodes `default="total_media_contribution_original_scale"` in its `Field(...)`, so the new constant covers only the MMM side. The two can now drift in opposite directions, which is strictly worse than one hardcoded string in each place.

**Suggested fix.** Pick one. Either wire it through, which means moving the constant somewhere both modules can import from without a cycle and referencing it in the `Field` default, or drop the constant and keep the literal at both sites. Separately, its current placement mid-class between `create_optimization_model` and `_resolve_response_variable` is unusual; class constants sit at the top of the class body elsewhere in this file.

## Smaller note on the docstring

Under the log link this sums `exp(mu) * target_scale`, which is the conditional median, not the mean. `LogLinkSpec.mean_correction` exists precisely for that gap. The argmax is safe, since `exp(sigma**2 / 2)` is per-draw and budget-independent, exactly like the baseline. But the Warnings block is already doing careful scope-honesty work, and a reader taking the value as "total predicted response" is off by that factor. One sentence pointing at `mean_correction` would close it.

---

The backward-compat note about identity-link `mu` entering `compute_deterministics` for the first time is correct and honestly disclosed. The fixture change in `test_mmm.py` is the right fix rather than a workaround, and module-scoped `fitted_mmm_mediated` matches the existing `dummy_df` / `fitted_mmm` scoping.

PabloRoque:
Thanks — point 1 was right and I had it wrong. All five addressed in `79756b57`.

## 1. The warning fired for effects that carry no media response — fixed

You're right that `self.mu_effects` is not a proxy for the property I care about, and the consequence you describe is the real cost: an events or trend user, whose objective was already exactly correct, got told it undercounts and was offered the choice of switching or restating the default to silence it.

Gated on graph reachability now, using the same `ancestors` traversal the optimizer already applies to the response variable:

```python
def _effects_carry_media_response(self) -> bool:
    media = self.model["channel_data"]
    for effect in self.mu_effects:
        name = effect.contribution_var_name
        if media in set(ancestors([self.model[name]])):
            return True
    return False
```

It runs on `self.model`, which exists after `build_model`, so it stays in `_resolve_response_variable` and both entry points keep sharing it.

Two judgement calls worth flagging. An effect whose `contribution_var_name` raises `NotImplementedError`, or whose contribution isn't in `named_vars`, can't be checked — those are treated as though they might carry response, on the grounds that a spurious warning is recoverable and a silently undercounted objective is not. And I kept *registration* gated on `mu_effects` rather than reachability: a promotional lever effect is downstream of `promo_data`, not `channel_data`, so a reachability gate there would fail to register the variable for the promos case. Registering broadly is cheap; warning broadly is not — which was your point exactly.

Pinned with a `LinearTrendEffect` model asserting silence. I mutation-tested it: restore `if self.mu_effects:` and it fails with `unexpected objective warning`.

## 2. `.sum(dim="date").sum()` — dropped

Agreed on both halves, and this is also what @juanitorduz asked inline. The double reduction bought nothing but an implicit assertion that a `date` dim exists, while hardcoding a name one layer below a configurable `date_dim`. Now a bare `.sum()`. I left `create_media_contribution_deterministic` alone for the reason you give — there the `sum(dim=)` is load-bearing for the `* target_scale` broadcast.

## 3. The `optimize_budget` path — parametrized

Done as suggested, one fixture over both entry points, with the three warning tests running through each. One wrinkle: `optimize_budget` emits its own unrelated "No budget bounds provided" `UserWarning`, so the silence assertions now check that no warning matching *this* guard was raised, rather than `simplefilter("error")`. That's a more precise assertion anyway.

## 4. Direct construction stays silent — documented

Added to the `response_variable` docs: the warning is an MMM-layer convenience, constructing `BudgetOptimizer` directly has no view of the model's effects, so silence there is not evidence the default is right.

## 5. `DEFAULT_RESPONSE_VARIABLE` — wired through

`DEFAULT_RESPONSE_VARIABLE` now lives in `budget_optimizer.py` after the imports, referenced both by the `Field` default and by `MMM`. No cycle, since the dependency already runs that direction.

## Log link median vs mean

Correct, and the Warnings block now says so and points at `mean_correction`, noting the argmax is safe because the factor is per-draw and budget-independent.

426 tests green across `test_budget_optimizer_mmm`, `test_link`, `test_budget_optimizer` and `test_mmm`.

pymc_marketing/mmm/link.py
PabloRoque: Good catch — dropped it entirely in `79756b57`.

Short answer to the `date_col` question: the model dim is always literally `"date"`. `_generate_and_preprocess_model_data` builds `coord_dict = {"date": dates}`, so `date_column` names a column in the incoming dataframe and never renames the dim. So there was no bug here.

But the line shouldn't have existed anyway: the trailing bare `.sum()` already reduces over every dimension, so `.sum(dim="date")` bought nothing except an implicit assertion that a `date` dim exists — while hardcoding a name one layer below `MMM.budget_optimizer`'s configurable `date_dim`, which is the inconsistency you were picking up on. It's now just:

```python
self.original_scale_transform(mu_var, target_scale).sum()
```

Note this is the opposite call from `create_media_contribution_deterministic` a few lines up, where the `sum(dim="date")` is load-bearing for the `* target_scale` broadcast that follows — that one stays as it is.

daimon-pymclabs (commented):
Round 2, at `79756b57`. All five points from the last round are addressed, and the reachability predicate is the right shape. One confirmed false positive survives it, for a reason outside this PR, plus three smaller things.

**What's resolved:** `_effects_carry_media_response` replaces the `mu_effects` non-emptiness test with ancestor reachability from `channel_data`, and `test_silent_for_effects_that_do_not_carry_media_response` pins the trend case. `DEFAULT_RESPONSE_VARIABLE` is module-level and wired into the `Field` default, so the two sites can no longer drift, and it introduces no new import cycle since `mmm.py` already imported `OptimizerCompatibleModelWrapper` from `budget_optimizer`. The redundant `.sum(dim="date")` is gone. The median-vs-mean note and the "silence here is not evidence" note both landed. The `call_entry_point` fixture covers both entry points, which was point 3.

I also want to confirm the scoping decision in the `TestDefaultObjectiveWarning` docstring, because it reads as an omission and is not one: levers acting through a `MuEffect` are deliberately *not* this warning's job, since step 8b of `_set_optimization_model` already raises `ValueError` for a lever the objective cannot reach. An error beats a warning, and the docstring's "only fires for declared levers, so a model with a mediating effect and no levers would otherwise misconfigure silently" is the correct statement of the remaining gap. Nothing to change; worth not re-litigating in a later round.

## 1. `add_events` models still get the spurious warning

This is the same regression point 1 was about. It now arrives through the conservative fallback rather than through the predicate, so the fix moved it rather than closing it.

`EventAdditiveEffect.create_effect` registers `f"{self.prefix}_total_effect"`. It does not override `contribution_var_name`, so the base property returns `f"{prefix}_effect_contribution"`, which is never registered anywhere: nothing in `mmm.py` or `additive_effect.py` wraps an event effect under that name. `_effects_carry_media_response` therefore takes the `if name not in self.model.named_vars: return True` branch and warns.

Existing tests confirm the registered name independently. `test_mmm.py` asserts `"holiday_total_effect"` in `named_vars` and in `mmm.posterior`, and there is no `holiday_effect_contribution` anywhere in the suite.

So every `mmm.add_events(...)` model still gets, on every `budget_optimizer()` call, a warning saying its objective undercounts a media path, when the events contribution is not downstream of `channel_data` and the default objective is exactly right for it. CI is green because the new negative test uses `LinearTrendEffect`, which does register `f"{prefix}_effect_contribution"`, and nothing covers `add_events`.

The base property's docstring is what led here: it says the default convention is "used by `LinearTrendEffect` and `EventEffect`". That is true of `LinearTrendEffect` and false of `EventAdditiveEffect`.

**Suggested fix**, one line on `EventAdditiveEffect`:

```python
@property
def contribution_var_name(self) -> str:
    """Event effects register ``f"{prefix}_total_effect"``."""
    return f"{self.prefix}_total_effect"
```

plus a negative test alongside the trend one, parametrized over the two families rather than duplicated, so the next effect that names its contribution differently gets caught:

```python
@pytest.mark.parametrize("fixture_name", ["fitted_mmm_trend", "fitted_mmm_events"])
def test_silent_for_effects_that_do_not_carry_media_response(...)
```

Worth flagging that the mismatch is not created by this PR. `contribution_var_name` is documented as the lookup `compute_counterfactual_contributions_dataset` uses to locate an effect's contribution, so that method looks like it has been unable to find the events contribution all along, and this PR is the first thing to depend on the lookup being right. If that turns out to be a live bug rather than a path events models never reach, the one-line override fixes both and probably deserves its own note in the PR description. Worth confirming before assuming it is only cosmetic.

## 2. `self.model is None` cannot fire; it raises instead

`ModelBuilder.__init__` has `self.model: pm.Model` as a bare annotation, which declares a type without creating the attribute, and `model: pm.Model` at class level is likewise annotation-only. `self.model` does not exist until `build_model` assigns it, so on an unbuilt model the guard raises `AttributeError` rather than returning `False`.

Short-circuiting hides half of it: `not self.mu_effects` is evaluated first, so an unbuilt model with no effects is fine. An unbuilt model *with* effects raises. Neither entry point can reach that today, since `budget_optimizer` calls `create_optimization_model` first and `optimize_budget` wraps a fitted model, so this is latent rather than live. But the guard reads as though it handles the case and it does not.

**Suggested fix:**

```python
if not self.mu_effects or getattr(self, "model", None) is None:
    return False
```

## 3. `record[0]` assumes our warning is the first one

`test_warning_points_at_the_caller` asserts on `record[0].filename`. On the newly-covered `optimize_budget` path, `optimize_budget` emits a `cost_per_unit` `UserWarning` about ten lines *before* it calls `_resolve_response_variable`, so any model with `data.cost_per_unit` set puts that warning at index 0 and the assertion checks the wrong frame. The current fixture does not set it, so the test passes; adding the second entry point is what introduced the ordering assumption.

**Suggested fix** — select by message rather than by position, matching what `_assert_no_objective_warning` already does:

```python
with pytest.warns(UserWarning, match="undercounts that path") as record:
    call_entry_point(fitted_mmm_mediated)

ours = next(w for w in record if "undercounts that path" in str(w.message))
assert ours.filename == __file__, f"warning blamed {ours.filename}, not the caller"
```

## 4. The three fallback branches have no test

`codecov/patch` is red, and reading the tests, the uncovered lines look like the conservative branches of `_effects_carry_media_response`: the `NotImplementedError` path, the missing-name path, and the `"channel_data" not in named_vars` path. Nothing in `TestDefaultObjectiveWarning` exercises any of them. Point 1 is a bug sitting on the second of those, which is the argument for covering them rather than waiving the patch check: an effect with no `prefix`, and an effect whose declared name is absent, are both cheap to construct as local `MuEffect` subclasses next to `_MediatedEffect`.

## 5. Small inconsistency in the conservative default

The docstring commits to a policy: "a spurious warning is recoverable, a silently undercounted objective is not." Two branches follow it (unnamed contribution warns, missing name warns) and one contradicts it: `if "channel_data" not in self.model.named_vars: return False` goes silent on a model it cannot analyze. Defensive rather than reachable for `MMM`, so this is a consistency note, not a bug. Either flip it or say in a comment why absent channel data means there is nothing to undercount.

## 6. One traversal instead of N

`set(ancestors([...]))` is rebuilt per effect, so an n-effect model walks the graph n times. Since only the boolean matters, one call over all named contributions is equivalent:

```python
named = [self.model[n] for n in names]          # after the unnamed/missing checks
return media in set(ancestors(named))
```

Negligible for realistic effect counts, so purely a readability call: it does cost the early exit, and the current loop reads more like the docstring. Take it or leave it.

juanitorduz (approved):
thanks @PabloRoque 🎉

PR #2897, opened 2026-08-17
docs(mmm): promote the credibility disaggregation notebook to the docs gallery

juanitorduz:
## Description

Follow-up to #2849: promotes the MMM campaign credibility disaggregation notebook from `docs/source/notebooks/mmm/dev/` to the documentation gallery, and rewrites it for its real audience, a marketing analyst who knows MMMs and campaign optimization but is not a statistician.

**Notebook** (`docs/source/notebooks/mmm/mmm_credibility_disaggregation.ipynb`, moved out of `dev/`; the diff renders as delete plus add because of the rewrite, so please read the file rather than the diff)

- Same method, same model, same data, same seeds; every result reproduces the dev version. This is a prose and structure pass, not a modelling change.
- Restructured as a storyline: problem and the credibility idea, data and EDA (reordered so the expanding-window reach report is explained before unique reach is used), Step 1 (each campaign's noisy own story), Step 2 (fit the trust dial), Step 3 (hand the credit back day by day, with a "what you bring to the meeting" table), grading against the hidden truth, the adstock extension, three robustness checks (prior sensitivity, unique reach vs reach-days as evidence, compositional variants), and a closing "what it can and cannot do".
- Three new figures that carry the concept: the credibility curve $Z = n/(n+k)$ before any data (also the gallery thumbnail), the fitted dial with the posterior band on $k$ and the seven campaigns on it, and a share bar chart with HDI whiskers against the truth and two heuristics (replacing the tall per-campaign `plot_dist` figure).
- Every term is defined the first time it appears in marketing language (HDI, partial pooling, shrinkage, log rate, adstock, Spearman, MAE, CRPS, coverage), and share metrics are also stated in percentage points of the channel's credit.
- Quarto-style callouts converted to MyST admonitions and dropdowns (the statistical detail lives in dropdowns); numbered section references replaced by named ones; headings normalised.
- The ten code cells that built Markdown paragraphs from f-strings with `if/else` verdicts are now static markdown written to match the executed outputs; code cells only compute and show tables and plots. All numbers quoted in prose were checked against the executed notebook with a script (a replacement for the removed guard cell) that re-executes the notebook source and asserts each quoted value.
- Verification-only cells removed: the guard cell that asserted sixteen posterior numbers and every `assert` inside analysis cells (reconciliation is now printed). This is also what makes the notebook pass the CI mock runner, which it now falls under (`docs/source/notebooks/*/*.ipynb`), together with ruff.
- Watermark now lists `pymc_marketing` (required by the `validate-watermark` hook); warning filters moved below the imports (ruff `E402`).

**Gallery** (`docs/source/gallery/gallery.yaml`, `gallery.md` regenerated with `scripts/generate_gallery.py`)

- The top-level section previously titled "Predicted Incrementality by Experimentation (PIE)" is renamed **"Campaign Optimization"**; it keeps the PIE card and gains the new card. Renaming rather than replacing the PIE example was my reading of the request; happy to adjust.

**Checks run locally**

- `uv run jupyter execute --inplace --kernel_name <repo venv kernel>` end to end, twice (final outputs come from the final source).
- Prose checker: all quoted numbers match the executed outputs; no em-dashes.
- CI mock runner: `uv run python scripts/run_notebooks/runner.py --notebooks mmm/mmm_credibility_disaggregation.ipynb --kernel <repo venv kernel>` passes. Note that `test_notebook.yml` excludes `docs/source/notebooks/mmm/**.ipynb` from its path filter, so this PR only triggers the docs build.
- `pre-commit run --files` on the notebook and the two gallery files (ruff, watermark, nbformat, gallery-in-sync).
- Local `sphinx-build -W --keep-going`: the notebook page and the gallery card render (admonitions, dropdowns, math, links); the only warnings in my local build are unrelated autosummary signature warnings from Python 3.14, so the CI docs job (Python 3.12) is the real check.

## Related Issue

- [ ] Closes #
- [x] Related to #2849, #2860, #2863

## Checklist

- [ ] Included tests that prove the fix is effective or that the new feature works (documentation only)

----
📚 Documentation preview 📚: https://pymc-marketing--2897.org.readthedocs.build/en/2897/


timbo112711 (approved):
Great stuff here!! Excited to try this out in practice

lancechua (approved):
The notebook is honest with its limitations and using a measure that sets credibility (unique people reached in this case) is an interesting idea.

The decay fix is already incorporated, which helps give a more accurate decomposition across time.

juanitorduz:
Thanks @jamespooley and @timbo112711 for the careful reads. All points are addressed in 521ea21, and the notebook was fully re-executed end to end afterwards. Point by point:

1. **Fay-Herriot named but not explained** (cell 0): kept the reference and added a one-sentence explainer in the "For the statistically minded" dropdown (shrink each area's noisy direct estimate toward a model-based mean, more strongly the larger its known sampling variance; here the variances come from reach instead of being supplied by hand).

2. **Mixed delivery/exposure terminology**: added a short paragraph at first use in the EDA pinning down the vocabulary (delivery = the volume the platform serves; exposure = the actuarial base a rate is charged over, always reach-days here) and aligned the one usage that contradicted it.

3. **"283to37k" rendering error** (cell 11): the dollar signs were being parsed as inline math; escaped them so the sentence renders as written.

4. **Overlap measured, not corrected for** (cell 23): now stated explicitly right under the overlap figure (the credibility weights treat each campaign's unique reach as independent evidence, so contested audiences are trusted as if disjoint) and echoed as its own bullet in the closing "What it cannot do" list.

5. **Z=1 and the click split** (cell 44): you are right, and this was a real imprecision, not just a wording one. The cancellation holds for the flight-total score, but the per-day renormalisation means Z=1 gives the click split spread over each campaign's own daily delivery profile; the two coincide exactly only when every campaign has the same normalised daily delivery profile. Step 3 now says this precisely, the sweep intro and caption label the Z=1 endpoint as the daily-renormalised click split, and the caption explains why its 0.0373 differs from the scoreboard's 0.0432 for the naive click split.

6. **Non-mathy "don't use it for" guidance** (cell 78): added a dedicated warning bullet: the split hands out average credit for spend already delivered, not the marginal return on the next dollar; no per-campaign response curve is fitted and saturation is estimated at channel grain, so the ranking is not reallocation advice, and this would hold even if the proxy measured incrementality perfectly. "It never breaks the budget" is now qualified as a property of the construction (a floor, not a merit badge), the intro's "defend in the budget meeting" phrasing is gone, and the bullet points to the budget allocation notebook for the forward-looking question.

7. **Daily vs weekly** (@timbo112711): added a "Shown daily, applies weekly" note before "Run it on your own data": same machinery one level coarser (reach-weeks, per-week renormalisation, unchanged unique-reach query), with the honest cost being resolution (no within-week shape, bursts blur into their week) rather than validity.

Two consistency fixes came out of the full re-run: the Dirichlet-multinomial paragraph was still quoting numbers from an earlier environment and now matches the committed outputs (MAE 0.0360, CRPS 0.0299, 10 divergences, click-split gap 0.0304, kappa posterior IQR 3,642), and the Related-notebooks cross-references had broken MyST role syntax (missing backticks) and now resolve. All 139 numbers quoted in the prose are verified against the executed code.

PR #1652, opened 2025-04-25
Create wrapper to make budget optimizer compatible with multidimensional class

cetagostini:
## Description

The new PR creates a wrapper which allows the optimizer to work on top of the multi-dimensional class. The only reason to not replace directly the code on the `optimizer.py` is to give us time to make the transition between the classes MMM and multidimensional MMM, and not break users current behavior which can be potentially an issue with the needed changes for the class.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--1652.org.readthedocs.build/en/1652/

PR #2898, opened 2026-08-18
fix(mmm): a correct optimization window — effect data, observed inputs, carry-in, effective carryover

PabloRoque:
## Description

**The optimization window was wrong in its contents and at both of its bounds.** Each defect was independently survivable, which is why all four lasted: they only bite together, on a model with a `DataVarMuEffect` optimized over a window that is not the training one — a combination that until #2890 existed in no notebook and no test.

None of this is a refactor regression. `create_zero_dataset` dates from 2025-05 (#1652); `DataVarMuEffect` from 2026-07 (#2677). The zero-dataset builder was written fourteen months before effect-referenced data variables existed, and adding them never taught it they existed. The pre-refactor optimization path is structurally identical to today's — same four steps, same `include_last_observations=False`, same `num_periods = len(dates) - l_max`, same `[budgets, zeros]` concat.

```python
# now just works, in or out of sample
optimizer = mmm.budget_optimizer(start_date=..., end_date=...)
```

### The contents

**Effect variables were absent, so out-of-sample optimization failed on shape.** `create_zero_dataset` built only `_channel` and `_control`, and `DataVarMuEffect.set_data` sets only what the dataset carries, so the effect's `pm.Data` silently kept its fit-time length. In-sample the lengths happen to agree and nothing looks wrong. Every other window raises.

**Non-decision inputs were zeroed, so in-sample optimization scored a model the posterior was never fitted to.** Re-optimizing a window that overlaps training is a counterfactual on the *spend*; every other input should still hold its fitted value. Zeroed, an exogenous series an effect reads — a category-demand index, a committed budget — came back as zero. `create_optimization_model` now restores observed values wherever the training data reaches. This is not a flag: making it unconditional broke exactly one test out of 400, the one written to pin the old behaviour.

### The bounds

**The window opened on a cold adstock**, as if the business had gone dark before the plan started. Because adstock feeds saturation this is not an additive constant — it moves the operating point over the first `l_max` periods and so changes the marginal return the optimizer equalises. The data was never missing: it is the tail of the training set.

**Both ends were sized by `adstock.l_max` alone**, while `spend_reach.py` already honours an effect's declared `additional_carryover_lags` when it sizes the incrementality evaluation windows. Same declaration, two answers, and the optimization window was the one truncating the tail.

## The date axis

Three blocks, and only the middle one is decided:

```
[ spend already made ][ decisions ][ zero-spend carry-over ]
        L_eff              N               L_eff
```

`pm.do` replaces the channel tensor wholesale, so the leading block cannot simply sit in the model's data and be left alone — `MediaVariable` re-emits it as a constant, read by `BudgetOptimizer` off the model it was handed rather than from anything that knows what an MMM is. The AST no-marketing-imports guard stays green.

`create_zero_dataset` takes its leading dates from the training *index* rather than generating them, so they are real observed dates by construction and the count clips itself when the window starts near the beginning of training.

`num_periods` now subtracts **both** flanking blocks. Subtracting only the tail turns history into decisions and spreads budget over the past.

`MMM.effective_carryover_lags()` is the single source for `L_eff` — `adstock.l_max` plus the widest additional carryover any registered effect declares — and all three sites read it. **Declarations only, never a probe:** this runs at build time, where measuring reach is neither available nor affordable, so an effect that declares nothing contributes nothing rather than being guessed at.

### A vocabulary fix that comes with it

Two mechanisms had come to share one set of words. `include_last_observations` prepends real observations and slices them off (`isel(date=slice(l_max, None))`, "Remove extra observations used for adstock continuity"); `include_carryover` / `adstock_periods` appends zero-spend periods. #2654 introduced "prepended warm-up" wording for the *second* of these in five places, including `BudgetOptimizer.adstock_periods` itself, and `create_zero_dataset` claimed its extension gave "leading observations" while extending `end_date`. Corrected here for the sites this PR touches.

## Behaviour changes

- **In-sample optimization** now restores observed values for controls and effect variables instead of zeroing them. This changes results for anyone optimizing a window that overlaps training — in the direction of the posterior they fitted.
- **Window sizes change** for any model whose effects declare `additional_carryover_lags`. On the conftest funnel fixture `l_max=3` and the mediator declares 3 more, so the axis goes from `3+8+3` to `6+8+6`.
- **`sample_response_distribution` is deliberately unchanged.** It scores an allocation against a zeroed baseline, so `create_zero_dataset` keeps `preserve_observed=False` as its default, now pinned by its own test.
- Models with no `mu_effects`, and effects with no `data_vars`, keep today's behaviour apart from carry-in and `L_eff`.

## Measurements, and their limits

Both bound fixes are **small on the available fixture**, and that is worth stating plainly rather than leaving a reader to assume otherwise:

| | allocation shift |
|---|---|
| carry-in vs cold start | 0.036% |
| `L_eff` vs `l_max` | 0.011% |

The fixture's window is 8 periods against `l_max=3` and sits on a near-linear stretch of the saturation curve; the correction should grow as the window shortens relative to adstock, which is the regime still unmeasured.

The objective value also changes a great deal under `L_eff` (−766 vs −609), but **that is not evidence of anything**: a longer window sums over more dates, baseline included, so the two values are not comparable. The case for these two changes is that the leading spend and the mediated tail belong in the window — not that they measurably moved this fixture's plan.

## Verification

692 passed, 6 skipped across `test_utils.py`, `test_budget_optimizer.py`, `test_budget_optimizer_mmm.py`, `test_optimization_variables.py`, `test_mmm.py`, `test_additive_effect.py`, `test_cost_per_unit.py` and `test_incrementality.py` — the last included because it shares the `additional_carryover_lags` concept. Lint and mypy clean.

Every claim is mutation-verified — each mutation below fails exactly the test that asserts the corresponding guarantee, and no others:

| mutation | caught by |
|---|---|
| effect fill disabled | `ValueError: Input dimension mismatch: (input[1].shape[0] = 12, input[3].shape[0] = 24)` — the real bug, in the end-to-end optimize |
| unknown-variable guard removed | bare `KeyError` from xarray instead of a message naming the effect |
| non-date variable zero-filled | a per-channel constant silently replaced by zeros |
| `_channel` overwritten | the decision variable erased |
| multi-dim variable collapsed to 1-D | channel dim and coords lost |
| `preserve_observed` forced on / off | in-sample restoration, and the default |
| `to_model` drops the carry-in block | `(6,3) == (8,3)` — the `do()` clobber |
| `num_periods` keeps the carry-in rows | two tests, including `shape 18 vs 8` in the end-to-end optimize |
| declared lags ignored | the window stops widening |

The end-to-end test allocates a real budget on a funnel model over a genuine future window, pinned to `total_response_original_scale` (#2895) and compiled with the `cvm` linker — the numba backend cannot compile the gradient of this effect's second, sample-batched adstock (pytensor#2360, upstream and unrelated).

## Deliberately out of scope

- **`TemporalReach.requires_full_axis`.** An effect declaring `evaluation_mode="full"` cannot be reproduced by *any* window, however wide. `effective_carryover_lags()` ignores that flag; it deserves at least a warning, but it is a new failure mode rather than a sizing fix.
- **Excluding carry-in dates from the objective.** Their response is a genuine constant — adstock is backward-looking and no decision reaches them — so the argmax is unaffected and only the reported value is inflated. Worth doing when the objective gains a window notion.
- **A monetary lower-funnel spend variable**, so `lf_budget` can be a *decision*. It is date × geo money, which `LeverVariable` (#2884) does not cover.

## Related

- Retires limitation (ii) of #2890 and removes both of its workarounds: the `pm.set_data` shape fix-up, and the `end_date=date_range[-(l_max + 1)]` gymnastics that pinned its window to the training range. Its caveat (3) — "a `DataVarMuEffect` keeps its training values, which is right in-sample" — documents the bug as correct behaviour and needs rewriting.
- Follows #2895 (the objective that sees those effects). Independent, but the two together are what make funnel optimization work out of sample.

## Type of change

- [x] Bug fix
- [x] Behaviour change (see **Behaviour changes**)


daimon-pymclabs:
## Points to fix

Follow-up on the review above, narrowed to the items I think should change before merge, each with the reason and a concrete fix.

---

### 1. Effect variables with a `channel` or `control` dim can silently become NaN

**Reason.** `_posterior_predictive_data_transformation` ends in `_reindex_dataset_to_user_order` (`mmm.py:1734-1736`), which reindexes the **whole dataset** on `channel` (and `control`). That was safe when the dataset only held `_channel` / `_control`, whose coords are `channel_columns` by construction. Section 6 now adds arbitrary effect variables to the same dataset, and any of them carrying a dim named `channel` whose coord set is not exactly `channel_columns` gets its missing labels filled with NaN by xarray. Reproduced mechanically on a zero-filled var with `channel = ['c2']` against `channel_columns = ['c1','c2']`:

```
zero-filled:               [[0.0], [0.0]]
after user-order reindex:  [[nan, 0.0], [nan, 0.0]]
```

Those NaNs go straight into `pm.set_data` for the effect's variable with no error at the boundary, so the optimizer runs on NaN. Reachable by a funnel mediator that reads only the lower-funnel channels, or by any effect variable that reuses the dim name `channel` for a different entity list. Same story for a `control`-dim variable when `control_columns` is set.

To be clear on reachability: this is plausible rather than demonstrated on a shipped model. The `MediaMuEffect` examples in `additive_effect.py:155-187` all carry the full channel set, and the test fixture's `lf_budget` is `("date",)` only. But nothing in the new code prevents it, and the failure mode is silent.

**Suggested fix (preferred).** Keep the section-6 variables out of the user-order reindex. Restrict `_reindex_dataset_to_user_order` to the variables it was written for (`_channel`, `_control`), or pass it an explicit var list, and reindex effect variables along `date` only.

**Suggested fix (minimum).** After section 6 and after the reindex, assert that no NaN was introduced in the variables section 6 added, and raise naming the effect and the variable name. That converts a silent numeric corruption into a loud error.

---

### 2. The multidimensional-effect test stops short of the path that breaks

**Reason.** `test_a_multidimensional_effect_variable_keeps_its_other_dims` asserts on the output of `create_zero_dataset`, which is the one point where the channel coords are still intact. The NaN in point 1 appears later, inside `create_optimization_model`. So the test that looks like it covers this case cannot catch it.

**Suggested fix.** Extend that test (or add a sibling) to build a model whose effect reads a variable with a `channel` dim over a **subset** of `channel_columns`, call `create_optimization_model` for an out-of-sample window, and assert `np.isfinite(model["<var_name>"].get_value()).all()` plus the expected shape. That exercises the `set_data` boundary rather than the dataset constructor.

---

### 3. In-sample numbers change silently

**Reason.** For an in-sample window, `set_data` previously skipped the effect's variables, so the **fitted** values stayed in place. They are now zeroed. Any funnel-style model with a nonzero exogenous input therefore gets different results from `budget_optimizer`, the deprecated `BudgetOptimizerWrapper`, and `sample_response_distribution` over a training-window date range. The docstring `Notes` block says this plainly, which is good, but the PR is labelled bug fix only and the description frames in-sample as the case where "nothing looks wrong", so a user on that path sees a numeric change with no signal.

**Suggested fix.** Two things:
- Add a changelog / release-note entry stating that in-sample optimization results change for models with `mu_effects` reading date-varying data, and that zero-filling is now the documented default.
- Emit a `UserWarning` when the requested window overlaps the training dates *and* section 6 zero-filled at least one variable, naming the variables and pointing at the `pm.set_data` escape hatch. This matches the pattern `_resolve_response_variable` already uses for its analogous silent-undercount trap.

---

### 4. `reads` loses attribution when two effects share a variable

**Reason.** The mapping is keyed by variable name, so two effects reading the same variable collapse to one entry and the error message names whichever effect was registered last. Cosmetic, but the message claims a specific attribution it does not actually have.

**Suggested fix.** Key on `(var_name, effect_name)` in a `set`, or make the insert first-wins and mention that the variable may be read by more than one effect. Either is a couple of lines.

---

### 5. Comment volume

**Reason.** Roughly 40 lines of prose comment for about 20 lines of code, and the trailing block at `utils.py:450-455` restates the docstring `Notes` and the PR body verbatim. Duplicated prose drifts out of sync with the code it describes, and the docstring is the copy users actually read.

**Suggested fix.** Delete the trailing block and let the docstring `Notes` carry it. Compress the `getattr`-vs-`isinstance` import-cycle rationale to one line. Trim the test docstrings that run longer than their assertions.

---

Not on this list, deliberately: the new `ValueError` is near-unreachable in practice (`DataVarMuEffect.create_data` at `additive_effect.py:552-554` already indexes `mmm.xarray_dataset[var_name]`, so a missing variable raises `KeyError` at `build_model` first) but it is cheap defense and worth keeping as-is. The `include_carryover` docstring correction is right and needs nothing further.

daimon-pymclabs:
Second pass on `1426c33`, short version.

`carry_in` is the right shape and I prefer it to the `UserWarning` I suggested earlier. Two things I would still hold merge on:

1. **The reindex NaN interaction.** `_reindex_dataset_to_user_order` reindexes the whole dataset on `channel`/`control`, so an effect variable whose coords are a subset of `channel_columns` gets NaN-filled after the zero/carry fill. With `carry_in=True` this now overwrites the observed history that `carry_in` exists to preserve, so it went from cosmetic to data-destroying. Fix: keep the section-6 variables out of the user-order reindex (reindex them on `date` only), or assert no NaN was introduced and raise naming the effect and variable.

2. **The compatibility claim reads backwards.** `test_the_default_still_zero_fills_an_in_sample_window`'s docstring says "without it the old behaviour stands", but pre-PR an in-sample window kept the *fitted* values (`set_data` skipped absent variables); the default now writes zeros over them. So the default is the new behaviour and `carry_in=True` restores the old one. The PR body still has a "No new parameter for planned activity" section and a bug-fix-only `Type of change`. Between the two, a reviewer cannot see that in-sample numbers move at the default. Fix: retitle the test and say it is a change, and refresh the body plus `Type of change` (and consider `feat(mmm):` in the title, since squash-merge makes it the release note).

Two additions I would push for: a `UserWarning` when `carry_in=True` and the window only partially overlaps training (naming the uncovered dates, since a control stepping to zero mid-window is the broken identity `carry_in` was added to avoid), and a test that checks the claim in the commit subject, i.e. `carry_in=True` plus historical `channel_data` reproduces the in-sample posterior predictive mean, not just that the data carried through.

Rest is polish.

daimon-pymclabs:
Another round on this, and I want to lead with the honest headline: this PR has gotten noticeably better each time I've read it. `carry_in_periods` for adstock warm-start is a stronger idea than the `UserWarning` I proposed two rounds ago, splitting it from `preserve_observed` was the right call (they really are two concepts), and sizing both the window and `adstock_periods` from a single `effective_carryover_lags()` is the kind of thing that prevents a whole class of future bugs. A few things I flagged earlier are cleanly resolved: the compatibility docstring now states the default plainly, and the `num_periods` arithmetic correctly subtracts the leading block before the tail, pinned by `test_the_leading_dates_are_not_decisions`.

I also went looking for the obvious scale trap on the new carry-in and did not find one: `MediaVariable` reads the block from `self.model[channel_data_var].get_value()` (model space) and divides only the decision budgets by `channel_scales`, so both blocks land in the same units. `test_the_leading_dates_hold_the_spend_that_already_happened` confirms the round trip.

Below are the things I'd like to see before merge, each with a concrete suggestion. Three are substantive (A, B, C); the rest are quick.

---

### A. `_reindex_dataset_to_user_order` can NaN out an effect variable

**Why.** `_reindex_dataset_to_user_order` (`mmm.py:1726-1736`) reindexes the **whole** dataset on `channel` and `control` with no `fill_value`, and `_posterior_predictive_data_transformation` ends on it (`mmm.py:2529`). That was harmless when the dataset only held `_channel`/`_control`, whose coords are `channel_columns` by construction. Now section 6 adds effect variables, and any one whose `channel` coords are a strict subset gets NaN-filled for the missing labels. Probed it:

```
effect_var coords: ['channel_2']   model channel: ['channel_1','channel_2']
rows out: [NaN, 1.623], [NaN, 1.649], [NaN, 0.563]     nan fraction 0.50
```

The NaN fraction is identical before and after the transform, so the hole is created at reindex and then propagated untouched into `pm.set_data`. No error at the boundary. And because `create_optimization_model` now hardcodes `preserve_observed=True`, the NaN overwrites real observed history rather than a zero. A funnel mediator reading only lower-funnel channels hits this, as does any effect variable using a dim named `channel` for a different entity list.

**Suggested fix.** Scope the reindex to the variables it was written for:

```python
def _reindex_dataset_to_user_order(self, dataset: xr.Dataset) -> xr.Dataset:
    owned = [name for name in ("_channel", "_control") if name in dataset.data_vars]
    indexers = {...}  # as today
    reindexed = dataset[owned].reindex(indexers)
    return xr.merge([dataset.drop_vars(owned), reindexed], join="left")
```

If that's more surgery than you want here, the minimum is a guard right after the transform: assert no NaN was introduced in the variables section 6 added, and raise naming the effect and the variable. A loud failure is a big improvement over silent NaN.

---

### B. Carry-in never checks that the history is adjacent to the window

**Why.** Section 2a in `utils.py` takes `training_dates[training_dates < new_dates[0]][-carry_in_periods:]`, which is the last training dates however far in the past they are. The clipping guard handles *not enough* history; nothing handles *non-contiguous* history. Probed with a window opening about a year after training ends:

```
training ends 2023-06-12   window starts 2024-06-10   lags 6
date gaps present: {7 days, 364 days}    max gap 364 days
carry-in block nonzero: True
```

So six weeks of year-old spend get spliced directly onto the window, and the adstock convolves them as if they were the immediately preceding weeks. The window comes up warm on activity that decayed to nothing twelve months earlier. This mirrors `include_last_observations`, which grabs `isel(date=slice(-l_max, None))` just as blindly, but that one is opt-in and this is on for every call to `create_optimization_model`.

**Suggested fix.** Require contiguity, and fall back to the existing zero path with the same kind of message the short-history case already emits:

```python
before = training_dates[training_dates < new_dates[0]][-carry_in_periods:]
if len(before):
    step = new_dates[1] - new_dates[0]
    contiguous = (
        before[-1] == new_dates[0] - step
        and (np.diff(before.values) == step).all()
    )
    if not contiguous:
        warnings.warn(
            f"Requested {carry_in_periods} carry-in periods, but the training data "
            f"ending {before[-1].date()} is not contiguous with the window starting "
            f"{new_dates[0].date()}. Falling back to a cold start (zeros).",
            UserWarning,
            stacklevel=2,
        )
        before = before[:0]
```

A test with a far-future window asserting the cold-start fallback would pin it.

---

### C. The documented direct-`BudgetOptimizer` path doesn't account for the leading block

**Why.** `create_zero_dataset`'s docstring and `test_a_planned_calendar_can_be_set_on_the_built_model` both point users at: build with `create_optimization_model`, `pm.set_data` the variable, hand the model to `BudgetOptimizer` directly. But `create_optimization_model` now always prepends `effective_carryover_lags()` dates, `BudgetOptimizer.carry_in_periods` defaults to `0`, and nothing validates that `carry_in_periods + num_periods + adstock_periods` matches `len(model.coords[date_dim])`. I ran the documented path end to end. With `lags=6`, an 8-period window and a 20-date model:

```
naive call (num_periods=8):
  ValueError: Input dimension mismatch: (input[1].shape[0] = 14, input[3].shape[0] = 20)
  raised inside the pytensor function, under scipy's _differentiable_functions

workaround (num_periods=14):
  converges, budgets [5.0598, 4.9402]
  but 14 = carry_in + window, so the budget is spread over history
```

The optimizer constructs without complaint and emits no length message, so the mismatch only surfaces at optimize time as that pytensor shape error, several frames deep in scipy. The wrapper path is fine (it resolves 8 correctly); it's the documented direct path that breaks. The existing test stops at `pm.set_data` and never constructs the optimizer, which is why it can't catch this.

**Suggested fix.** Validate in `model_post_init` with a message that names the three blocks:

```python
expected = self.carry_in_periods + self.num_periods + self.adstock_periods
actual = len(self.model.coords[self.date_dim])
if expected != actual:
    raise ValueError(
        f"Date length mismatch: the model has {actual} dates but "
        f"carry_in_periods ({self.carry_in_periods}) + num_periods "
        f"({self.num_periods}) + adstock_periods ({self.adstock_periods}) "
        f"= {expected}. If this model came from create_optimization_model, "
        f"pass carry_in_periods=mmm.effective_carryover_lags()."
    )
```

Then update the docstring example to pass `carry_in_periods=mmm.effective_carryover_lags()`, and extend `test_a_planned_calendar_can_be_set_on_the_built_model` to actually construct the optimizer and assert the allocation lands on the window dates, not the leading block.

---

### D. `MMM.sample_response_distribution` is a dead Sphinx target

**Why.** `utils.py:287` justifies the `preserve_observed=False` default by cross-referencing `` :meth:`~pymc_marketing.mmm.mmm.MMM.sample_response_distribution` ``. That method doesn't exist on `MMM`. The only definition is at `mmm.py:4283`, inside `BudgetOptimizerWrapper` (class opens at `3939`), which raises `DeprecationWarning` on construction. Both call sites relying on the default (`mmm.py:3959`, `mmm.py:4321`) are inside that same wrapper. So the default is defended by a deprecated path while the live path hardcodes the opposite.

**Suggested fix.** Either flip the default to `True` and have the wrapper pass `preserve_observed=False` explicitly at its two call sites, which makes the live path's intent the default and the deprecated behaviour the exception; or keep the default and change the reference to name `BudgetOptimizerWrapper.sample_response_distribution` with a note that it's deprecated. The first reads better to me.

---

### E. Release note, and the objective value moves

**Why.** `preserve_observed` isn't reachable from the live path at all now, since `create_optimization_model` hardcodes it and doesn't expose it. That's defensible, but it means a public method returns different in-sample numbers than before with no opt-out, under a `fix(mmm):` title that squash-merge turns into the release line and that `pr-auto-label` labels off.

One number worth calling out. Comparing carry-in against zeroed on the same model, allocations barely move but the achieved objective moves a lot:

| objective | allocation shift | objective (zeroed → carry-in) |
| --- | --- | --- |
| `average_response` | 0.035% | -540 → -766 |
| `value_at_risk` | 1.20% | -38.8 → -44.8 |

All four optimizations converged. So plans are robust across the change, which is reassuring, but the reported objective is not comparable to previous releases, and people do track that number.

**Suggested fix.** Retitle to `feat(mmm):`, add the in-sample behaviour change to `Type of change`, and put a line in the release note along the lines of: *optimization now conditions on observed history for non-decision variables; allocations are largely unchanged but reported objective values are not comparable to prior releases.*

---

### Smaller polish

- The trailing comment block at the end of section 6 in `utils.py` is now orphaned: it sits after the loop it describes, immediately above `return xr.Dataset(data_vars)`, so it reads as commenting the return. It also restates the docstring `Notes` verbatim. Suggest cutting it and letting the docstring carry it.
- `reads` still collapses duplicate variables to one dict entry, so the error names whichever effect came last. Keying on `(var, effect)` (or first-wins) would attribute correctly.
- `create_zero_dataset`'s `Returns` section still says "plus one zero-filled variable per date-varying name", which is only the `preserve_observed=False` case now.
- `effective_carryover_lags` handles a `None` spec incidentally, via `getattr(None, ...)` returning `None`, rather than explicitly as its docstring claims. One `if spec is None: return 0` would match the documented contract.
- `TestEffectiveCarryover.test_the_declaration_widens_the_window` never checks a window: it asserts a scalar, and the assertion it ends on is the value *after* the declaration is removed, which reads backwards against the name. Either rename it or add the widened-window assertion.

---

Genuinely nice work on this one. A, B and C are the three I'd hold for; D and E are a docstring and a release note, and the rest is polish you can take or leave. Happy to dig into any of them further, and I'm glad to re-run the probes against a new head once you've pushed.

juanitorduz:
thanks @PabloRoque ! I think we are very close 🚀

juanitorduz:
@daimon-pymclabs thanks for the thorough round. I pushed two commits onto this branch (`03ed10c3`, `284d883f`) to address it so we can get this merged; @PabloRoque please `git pull` before pushing again so nothing gets dropped. Summary per item:

**A. NaN in effect variables.** Confirmed the symptom, but the hole is not born at `_reindex_dataset_to_user_order`: an `xr.Dataset` holds one index per dimension, so an effect variable whose `channel` labels are a strict subset is NaN-filled the moment it is aligned (already by `xr.Dataset(data_vars)` in `create_zero_dataset`, or earlier in the training dataset itself). Scoping the reindex cannot help: the `xr.merge(join="left")` variant still carries the NaN, and a Dataset cannot hold two `channel` orders anyway (and positional `pm.set_data` needs effect variables in model order too). So I took your stated minimum and put it at the source: `create_zero_dataset` now raises a `ValueError` naming the effect(s) and the variable when an effect variable contains NaN after alignment. Covers the live path and the deprecated wrapper path. Test: `test_an_effect_variable_with_holes_is_a_loud_error` (your probe, reproduced via `Dataset.assign` of a single-channel variable).

**B. Non-contiguous carry-in.** Done as suggested, frequency-based: the history must equal `pd.date_range(end=window_start, periods=n+1, freq=inferred_freq)[:-1]` (elementwise, since `DatetimeIndex.equals` is False across datetime64 units); otherwise a `UserWarning` and a cold start, without the misleading "clipped" debug record. Tests: far-future window falls back with the warning; the week-after-training window stays silent. Note: the test `WINDOW` constant opened three weeks after training ended, which is now a warned cold start, so I moved it to the week after training (same length).

**C. Direct `BudgetOptimizer` path.** `model_post_init` now validates `carry_in_periods + num_periods + adstock_periods` against the date axis and raises your message naming the three blocks. Two refinements: the length is measured on the `channel_data` tensor (what `do()` replaces) rather than `model.coords[date_dim]`, which is `None` for a dim declared by length only; and a legacy wrapper (`BudgetOptimizerWrapper`, `BuildMergedModel` over one) whose own `num_periods` disagrees with the caller's gets its own message ("pass `num_periods=wrapper.num_periods`"), because those users have no carry-in and the wrapper's `optimization_model()` ignores the argument. Docstring examples (class, `merge_inference_data`, `merge_models_and_idata`, `create_zero_dataset` Notes) now pass `carry_in_periods=mmm.effective_carryover_lags()`. `test_a_planned_calendar_can_be_set_on_the_built_model` now constructs the optimizer, allocates, and evaluates the substituted channel tensor: leading rows equal the training tail, the middle rows carry the allocation, the tail is zero. The check surfaced one docs notebook, `mmm_budget_allocation_example.ipynb` cell 45, which passed `num_periods=campaign_period` (12) to a wrapper whose window is 13 dates; it now passes `optimizable_model.num_periods` (3-line diff, outputs not regenerated, so that cell's printed allocation is one period stale; worth a docs re-run in a follow-up). Both optimizer notebooks pass the mocked CI runner.

**D. `preserve_observed` default.** Kept `False` and fixed the reference: it now points at `BudgetOptimizerWrapper.sample_response_distribution` (deprecated) and justifies the default by the function's contract (a window with no committed activity, `_control` unchanged for existing callers) with `create_optimization_model` opting in explicitly. Happy to flip it in a follow-up if you feel strongly.

**E. Release note / title.** Not editing the PR metadata from my side; suggested title `feat(mmm): a correct optimization window: effect data, observed inputs, carry-in, effective carryover`, and a release line: "optimization now conditions on observed history for non-decision variables; allocations are largely unchanged but reported objective values are not comparable to prior releases. `BudgetOptimizer` now raises when `carry_in_periods + num_periods + adstock_periods` does not match the model's date axis (legacy wrapper users: pass `num_periods=wrapper.num_periods`); carry-in falls back to a cold start with a `UserWarning` when training history is not contiguous with the window." Release-note sections are driven by labels, so an `enhancement` label would file it under New Features.

**Polish.** Orphaned comment block removed; `reads` keeps every effect per variable and the error names them all (tested); `Returns` section corrected; `effective_carryover_lags` handles `spec is None` explicitly (as `continue`, not `return 0`, so one effect opting out does not silence another's declaration); `test_the_declaration_widens_the_window` now asserts the window widths with and without the declaration.

Verification: 745 passed, 6 skipped across `test_utils`, `test_optimization_variables`, `test_budget_optimizer`, `test_budget_optimizer_mmm`, `test_budget_optimizer_merge_helpers`, `test_multidimensional_shim`, `test_mmm`, `test_additive_effect`, `test_cost_per_unit`, `test_incrementality`, `test_pytensor_utils`; pre-commit clean; both optimizer notebooks green under `scripts/run_notebooks/runner.py`. Happy for you to re-run the probes against `284d883f`.

daimon-pymclabs:
Read `03ed10c3` and `284d883f` end to end. To be straight about the basis: this is a read of the diff and of the surrounding call sites at `284d883f`, not a re-run of the probes in a live env, so anything below that I could not settle statically I have said so.

Three of the four substantive items land, and one of them lands better than what I asked for.

**A.** You are right and I was wrong about where the hole is born. An `xr.Dataset` carries one index per dimension, so the NaN is created at alignment, and scoping `_reindex_dataset_to_user_order` would only move it. Refusing at the source in `create_zero_dataset` is the correct place, and naming every reader rather than the last one is the better version of what I filed. One consequence worth a release-note line rather than a change: the guard also fires on a genuine gap in observed history, which only reaches it now that `create_optimization_model` sets `preserve_observed=True`. A model whose effect variable has missing dates in training used to optimize (zero-filled through `zeros_like`) and will now raise. That is the right call, but it is a behaviour change for someone, and your error text already covers it ("or the training data itself has gaps").

**B.** The contiguity test is sound, including the part I would have got wrong: `new_dates[0]` is produced by `pd.date_range(start=..., freq=inferred_freq)`, so it is always on the offset's anchor, which is what stops `pd.date_range(end=new_dates[0], ...)` from rolling backwards and shifting `expected` by one period. Elementwise rather than `.equals` is right for the same reason you give. One residual, and I do not think it needs fixing: when `pd.infer_freq` returns `None` the fallback is `"W"` (W-SUN), so training dates on any other weekday are judged non-contiguous and get a cold start. That is arguably the honest answer for an irregular index, and the frequency warning fires alongside the carry-in warning, so the pair is diagnosable.

**C.** Measuring on the channel tensor instead of `model.coords[date_dim]` is the right fix for the length, and the separate wrapper message is a good call. I checked the interaction I was most worried about, which is B and C disagreeing: if the carry-in falls back to a cold start, the model's date axis loses the leading block, and a `carry_in_periods` computed independently would then fail the new length check. On the supported path it cannot happen, because `MMM.budget_optimizer` derives `carry_in_periods = int((model_dates < start_date).sum())` from the built model's own coords, so both sides read the same axis.

The direct path is where I would still change something. The docstrings now tell users to pass `carry_in_periods=mmm.effective_carryover_lags()`, and that is exactly wrong in the two cases `create_zero_dataset` handles: a window near the start of training (leading block clipped, `logger.debug` only) and a window not contiguous with training (block dropped, `UserWarning`). In both, the model has fewer leading dates than `effective_carryover_lags()` and the user gets `Date length mismatch` telling them to pass the value that just failed. The error text does hedge with "or the number of model dates before your start_date", but the examples above it do not. Suggestion, small: make the examples show the derived form, the same one-liner `budget_optimizer` uses,

```python
model_dates = pd.DatetimeIndex(list(pymc_model.coords["date"]))
carry_in_periods = int((model_dates < pd.Timestamp(start_date)).sum())
```

so copy-pasting the example is correct in every case rather than the common one. Alternatively record the realised carry-in on the model returned by `create_optimization_model` and read it back, which would also let the mismatch error say what the axis actually holds.

**D and E.** Both fine by me. `preserve_observed=False` justified by the contract rather than by the deprecated caller is the right reasoning, and I do not feel strongly enough to ask for the flip. The release line reads well; agreed that the notebook's stale printed allocation wants a docs re-run rather than a rerun in this PR.

Nothing here blocks. A is a release-note line, C is a docstring change, and I would merge with or without them.

juanitorduz (approved):

PR #2899, opened 2026-08-18
fix(mmm): floor alpha away from exact zero so adstock normalization never divides 0 / 0

### LINKED ISSUE #2889: delayed_adstock returns NaN for alpha=0 with non-integer theta
### Description

`delayed_adstock` returns an all-NaN kernel for `alpha=0`, which is its own default argument, whenever `theta` is not an integer and `normalize=True`.

```python
import numpy as np
import pytensor.tensor as pt
from pymc_marketing.mmm.transformers import delayed_adstock

x = ptx.as_xtensor(pt.as_tensor_variable(np.ones(20)), dims=("time",))
print(delayed_adstock(x, alpha=0.0, theta=0.5, l_max=12, dim="time", normalize=True).eval())
# array([nan, nan, nan, ...])
```

The weights are `w = alpha ** ((lag - theta) ** 2)`. With `alpha = 0` and non-integer `theta` every exponent is strictly positive, so every weight is `0`, `w.sum(dim=kernel_dim)` is `0`, and the normalisation is `0 / 0`. With integer `theta` one exponent is exactly `0`, that weight is `1`, and the result is finite, which is why this is not caught by the current tests.

### Why it passes the guard

`_check_alpha` uses the message `"0 < alpha <= 1"` but tests `(alpha_tensor >= 0) & (alpha_tensor <= 1)`. Zero therefore passes a check whose own message excludes it.

```python
checked_alpha_tensor = alpha_check_op(
    alpha_tensor, ((alpha_tensor >= 0) & (alpha_tensor <= 1)).all()
)
```

### Why it matters beyond a literal `alpha=0`

Values that are not exactly zero underflow to it. For `geometric_adstock` the relevant mapping is far from the underflow boundary, but `DelayedAdstock`'s new half-life parametrisation (#2887) maps `alpha = 2 ** (-1 / h**2)`, and the square brings the boundary up to `h ~ 0.0305` in float64 (`h ~ 0.082` in float32). A user-specified diffuse or heavy-left-tailed half-life prior lets a sampler reach it, and `theta ~ HalfNormal(1)` is continuous so it is never an integer in practice. The failure is a NaN logp rather than a rejected proposal, which is much harder to diagnose.

### Suggested fix

Fix it in `transformers.py` so both the direct call and every parametrisation on top of it are covered. Either:

1. Make `_check_alpha` match its message, so `alpha=0` raises instead of silently producing NaN. This is the stricter reading and matches the documented contract, but it is a behaviour change for anyone currently relying on `alpha=0` with integer `theta`.
2. Floor `alpha` at the smallest positive normal before the `power`, keeping the graph finite:

```python
alpha = ptx.math.maximum(alpha, np.finfo(pytensor.config.floatX).tiny)
```

Note the dtype must come from `pytensor.config.floatX`: a hardcoded `np.finfo(float).tiny` is float64 and silently upcasts a float32 graph.

Option 1 is cleaner if the default argument `alpha: XTensorLike = 0.0` is also changed, since that default is itself the degenerate case. My preference is 1 plus a sensible default, but 2 is the non-breaking choice.

### Tests to add

- `delayed_adstock` with `alpha=0.0` and non-integer `theta`, `normalize=True`, asserting the chosen behaviour (raises, or finite).
- The same with integer `theta`, which is finite today and must stay that way under option 2.
- A `float32` case, so the floor is not silently a no-op or a silent upcast.

Context: surfaced while reviewing #2887 (https://github.com/pymc-labs/pymc-marketing/pull/2887#issuecomment-5316108104). The bug predates that PR and is independent of it.

vaibhav8a:
Closes #2889.

## Problem

`delayed_adstock` returned an all-NaN kernel for `alpha=0.0` (its own default) whenever `theta` was non-integer and `normalize=True`:

```python
x = ptx.as_xtensor(pt.as_tensor_variable(np.ones(20)), dims=("time",))
delayed_adstock(x, alpha=0.0, theta=0.5, l_max=12, dim="time", normalize=True).eval()
# array([nan, nan, nan, ...])
```

Every weight `alpha ** ((lag - theta) ** 2)` has a strictly positive exponent when `theta` is non-integer, so every weight is `0`, the sum is `0`, and normalization is `0 / 0`. Integer `theta` was accidentally spared — the weight at `lag == theta` has exponent exactly `0`, so `0 ** 0 = 1` keeps the sum non-zero, which is why the existing test suite never caught it.

## Fix

Took the non-breaking option the issue lists (flooring) over the breaking one (raising on `alpha=0` and changing the `alpha: XTensorLike = 0.0` default that is itself the degenerate case):

```python
tiny = np.finfo(pytensor.config.floatX).tiny
checked_alpha_tensor = pt.maximum(checked_alpha_tensor, tiny)
```

Fixed in `_check_alpha` itself, not in `delayed_adstock` alone, per the issues "Fix it in transformers.py so both the direct call and every parametrisation on top of it are covered": `geometric_adstock` was already safe by construction (its exponent is always the integer `lags` array, never derived from a continuous parameter), but `binomial_adstock` had its own unguarded `alpha=0` failure mode — `1 / alpha - 1` is `1 / 0`, an `inf` rather than a `NaN`, a different symptom of the same root cause. Fixing the shared check covers both without touching either caller.

`np.finfo(pytensor.config.floatX).tiny`, not `np.finfo(float).tiny`, per the issues own note: the latter is always float64 and would silently upcast a float32 graph. Verified directly rather than assumed: with `floatX=float32`, `_check_alpha(float32 alpha)` returns a float32 result equal to `np.finfo(np.float32).tiny`, not float64s.

## What changes for a caller

Nothing observable for any `alpha` that isnt exactly zero. For `alpha=0` with integer `theta`, the previously-exact spike (`[0,0,0,0,0,1,0,0,0,0,0,0]`) becomes numerically indistinguishable from one (values ~1e-308 next to the spike, still summing to `1.0` within tolerance) — pinned as a regression guard rather than silently assumed unchanged. For `alpha=0` with non-integer `theta`, the previously-NaN result becomes the intuitively correct split between the two nearest integer lags (e.g. `theta=0.5` → `[0.5, 0.5, 0, ...]`).

## Tests

Added to `TestsAdstockTransformers` in `tests/mmm/test_transformers.py`:

- `test_delayed_adstock_alpha_zero_noninteger_theta_is_finite` — the issues own repro.
- `test_delayed_adstock_alpha_zero_integer_theta_still_spikes` — regression guard for the case that was already finite.
- `test_binomial_adstock_alpha_zero_is_finite` — the second `_check_alpha` callers own failure mode (`inf`, not `NaN`).
- `test_check_alpha_floor_respects_floatx_no_upcast` — isolated from `delayed_adstock`/`theta` (thetas own default dtype would otherwise confound the result), and explicitly sets `floatX=float32` rather than relying on the ambient default (`float64`) — under the default, a hardcoded float64 floor would pass this assertion for the wrong reason.

## Verification

```
uv run --extra test pytest tests/mmm/test_transformers.py::TestsAdstockTransformers -v
53 passed, 1 xfailed (pre-existing, unrelated)
```

`ruff check` clean on both touched files; `ruff format` applied to the test file (whitespace only). `mypy` reports zero errors in `transformers.py` specifically (1072 pre-existing errors elsewhere in the tree, none in this file).

----
📚 Documentation preview 📚: https://pymc-marketing--2899.org.readthedocs.build/en/2899/


pymc_marketing/mmm/transformers.py
```diff
@@ -55,6 +56,20 @@ def _check_alpha(
...
+    # alpha to a non-negative power and, when normalize=True, divides by the
+    # sum of those powers: at alpha == 0 every term with a strictly positive
+    # exponent is 0, the sum is 0, and normalization is 0 / 0 -> NaN (not
+    # caught here because delayed_adstock's exponent (lag - theta) ** 2 is
+    # zero, and the whole array finite, only when theta happens to be an
+    # integer). Flooring away from exact zero keeps every adstock's weights
+    # finite without changing the accepted range or raising for any input
+    # that already worked. `pytensor.config.floatX`, not a hardcoded
+    # `np.finfo(float)`: alpha may be float32, and float64's tiny would
+    # silently upcast the graph.
```
juanitorduz: remove this comment please

juanitorduz (changes requested):
thanks @vaibhav8a . some initial comment

daimon-pymclabs (commented):
Thanks for the clear write-up and for adding tests alongside the fix. The core diagnosis is right, the shared-helper placement is the right call, and I verified the `theta=0.5` limit you describe: floored weights are `tiny**0.25 = 1.22e-77` at lags 0 and 1, normalizing to exactly `[0.5, 0.5, 0, ...]`, which is the correct `alpha -> 0+` limit.

Two changes I'd like before merge.

### 1. `test_binomial_adstock_alpha_zero_is_finite` passes on `main`, so it isn't guarding this fix

The PR body says `binomial_adstock` had its own `alpha=0` failure mode producing `inf`. `1 / alpha - 1` is indeed `inf`, but that value is only an intermediate *exponent*, and the observable output was already finite before this change, because `1.0 ** inf == 1.0` and every later base is `< 1`:

```
pre-fix   w = (1 - lags/13) ** inf       -> [1.0, 0.0, 0.0, ...]   sum = 1.0
post-fix  w = (1 - lags/13) ** 4.49e307  -> [1.0, 0.0, 0.0, ...]   sum = 1.0
```

The normalized kernel is bit-identical either way, and `pyproject.toml` sets no `-W error`, so the `RuntimeWarning` from the division doesn't fail the test either. Could you confirm by reverting just the `transformers.py` hunk and running that one test? If it passes, please pick one:

- **Keep it, relabel it.** Reword the docstring so it reads as a pin on already-correct behavior ("`binomial_adstock` was already finite at `alpha=0`; assert the floor doesn't change that") rather than as a regression guard, and drop the `inf` failure-mode claim from the PR description.
- **Or make it guard the intermediate**, if that's what you actually want protected. Assert on `rescaled_alpha` itself rather than on the convolution output:
  ```python
  alpha = _check_alpha(ptx.as_xtensor(pt.as_tensor_variable(0.0)))
  assert np.isfinite((1 / alpha - 1).eval())
  ```

Either is fine. The thing to avoid is a test whose name implies it would have caught the bug when it wouldn't.

### 2. Derive the floor from `alpha`'s dtype, not from `floatX`

The comment says `pytensor.config.floatX` is used so that "float64's tiny would [not] silently upcast the graph", but reading `floatX` doesn't achieve that when the two disagree. With the default `floatX="float64"` and a `float32` `alpha` (a user passing an explicit `np.float32` array), `tiny` is an `np.float64` scalar and `pt.maximum` promotes the whole downstream graph to float64 -- exactly the upcast the comment is guarding against. `test_check_alpha_floor_respects_floatx_no_upcast` only covers `floatX="float32"` with `float32` alpha, the combination that happens to work.

Suggested fix, keying off the tensor's own dtype with a fallback for the integer case (`alpha=1` as a Python int arrives as `int64`, where `np.finfo` raises):

```python
alpha_dtype = checked_alpha_tensor.dtype
tiny = np.finfo(
    alpha_dtype if alpha_dtype.startswith("float") else pytensor.config.floatX
).tiny
checked_alpha_tensor = pt.maximum(checked_alpha_tensor, tiny)
```

And parametrize the test over the mismatched combination so the regression is actually covered:

```python
@pytest.mark.parametrize(
    "floatX, alpha_dtype", [("float32", "float32"), ("float64", "float32"), ("float64", "float64")]
)
def test_check_alpha_floor_does_not_upcast(self, floatX, alpha_dtype):
    with pytensor.config.change_flags(floatX=floatX):
        alpha = pt.as_tensor_variable(np.zeros((), dtype=alpha_dtype))
        checked = _check_alpha(ptx.as_xtensor(alpha))
        assert checked.dtype == alpha_dtype
        assert checked.eval() == np.finfo(alpha_dtype).tiny
```

The middle case is the one that fails today.

### Non-blocking

One thing worth noting rather than fixing here: flooring at the smallest *normal* value is the weakest floor possible for a base raised to a power, since `tiny ** e` underflows back to zero once `e` exceeds roughly 1.05 (float64) or 1.2 (float32). So the all-zero kernel and the NaN still occur when the smallest `(lag - theta) ** 2` over `lag in [0, l_max - 1]` clears that threshold, i.e. when `theta` sits more than ~1 outside the lag range:

```
theta=11.5, l_max=12: sum = 1.22e-77   ok
theta=12.5, l_max=12: sum = 0.0        still NaN
theta=13.7, l_max=12: sum = 0.0        still NaN
```

Reachability is low: the docstring bounds `theta` to `[0, l_max - 1]`, and `alpha` is never exactly `0` under the `Beta(1, 3)` prior during sampling, so this needs a manual out-of-contract call. I wouldn't hold the PR for it. But the "never divides 0 / 0" in the title claims more than the fix delivers, so either soften that phrasing, or, if you want the guarantee to actually hold, guard the denominator (`w / pt.switch(pt.gt(s, 0), s, 1.0)`) or build the kernel in log space (`exponent * log(alpha)`, normalized by softmax), which reproduces the exact `alpha -> 0+` limit for any `theta` with no underflow at all.

Smaller things, take or leave: the non-integer-`theta` test asserts only finiteness, so it's worth an `assert_allclose` pinning the `[0.5, 0.5, 0, ...]` kernel your description highlights; `_check_alpha` now modifies its input, which its name doesn't suggest, so a one-line docstring would help; the `CheckParameterValue` message still reads `0 < alpha <= 1` while the predicate accepts `alpha >= 0`, so that mismatch from the issue survives (fine as a deliberate choice, just call it out); `np.finfo(...).smallest_normal` is the canonical name for `.tiny`; and the 14-line comment could compress to two or three lines plus a reference to the issue, which is where `transformers.py` usually leaves the detail.

vaibhav8a:
Thanks both for the detailed review — addressed both required points.

**1. `test_binomial_adstock_alpha_zero_is_finite` didn't guard the fix**

Confirmed by reverting just the `transformers.py` hunk and computing the pre-fix path directly: `1 / alpha - 1` at `alpha=0` is `inf`, but `1.0 ** inf == 1.0` and every other base is `< 1`, so the kernel `[1.0, 0.0, 0.0, ...]` was already finite either way. Replaced it with `test_check_alpha_floor_keeps_binomial_rescaled_alpha_finite`, which asserts on `1 / alpha - 1` directly — the actual quantity the floor protects.

**2. Floor should key off alpha's dtype, not `floatX`**

Fixed as suggested: `_check_alpha` now reads `checked_alpha_tensor.dtype` and falls back to `pytensor.config.floatX` only for non-float (integer) alpha, where `np.finfo` would raise. Parametrized the floatX test over all three combinations (the `float64`/`float32` mismatch is the one that fails without this fix), and added a separate test for the integer-alpha fallback path.

Also applied the smaller items: pinned the exact `[0.5, 0.5, 0, ...]` kernel in the theta=0.5 test, added a docstring to `_check_alpha` noting it mutates its input, and switched to `np.finfo(...).smallest_normal`.

Left the `CheckParameterValue` message mismatch and the log-space rebuild for the underflow-at-large-theta case as-is, per the review calling both out as optional/pre-existing.

One honest caveat: this machine hit a severe, unrelated environmental stall while I was verifying locally (iCloud Drive's sync daemons pinning heavy CPU/IO against this directory — every `pytensor`-importing command and even plain `git` operations here were hanging for many minutes at a time). I verified the logic by other means instead — reproduced the binomial-adstock finiteness claim directly in plain numpy, and derived the pinned kernel values from `batched_convolution`'s own already-passing `ConvMode.After` test semantics rather than a fresh live run. Flagging this rather than claiming a local test run I didn't actually get a clean one of — happy to re-verify and follow up if CI turns up anything my reasoning missed.

vaibhav8a:
Follow-up: the local environment cleared up enough to get a clean run after all.

\`\`\`
tests/mmm/test_transformers.py::TestsAdstockTransformers::test_delayed_adstock_alpha_zero_noninteger_theta_is_finite PASSED
tests/mmm/test_transformers.py::TestsAdstockTransformers::test_delayed_adstock_alpha_zero_integer_theta_still_spikes PASSED
tests/mmm/test_transformers.py::TestsAdstockTransformers::test_check_alpha_floor_keeps_binomial_rescaled_alpha_finite PASSED
tests/mmm/test_transformers.py::TestsAdstockTransformers::test_check_alpha_floor_respects_alpha_dtype_no_upcast[floatX=32,alpha=32] PASSED
tests/mmm/test_transformers.py::TestsAdstockTransformers::test_check_alpha_floor_respects_alpha_dtype_no_upcast[floatX=64,alpha=32] PASSED
tests/mmm/test_transformers.py::TestsAdstockTransformers::test_check_alpha_floor_respects_alpha_dtype_no_upcast[floatX=64,alpha=64] PASSED
tests/mmm/test_transformers.py::TestsAdstockTransformers::test_check_alpha_floor_falls_back_to_floatx_for_integer_alpha PASSED
7 passed in 195.31s
\`\`\`

Confirms the `[0.5, 0.5, ...]` kernel values I'd derived from `test_batched_convolution`'s semantics rather than run live in my last update — they check out exactly. All green.

pymc_marketing/mmm/transformers.py
```diff
@@ -50,11 +51,27 @@ def _check_alpha(
         msg="0 < alpha <= 1", can_be_replaced_by_ninf=False
     ),
 ):
+    """Validate alpha's range and floor it away from exact zero.
+
+    The floor avoids the 0 / 0 that `delayed_adstock` hits when `theta` is
+    non-integer (see #2889); returned value differs from the input at
+    alpha == 0 even though the check itself accepts it.
+    """
```
juanitorduz: plese remove these redundant comments

juanitorduz (changes requested):

vaibhav8a:
Removed — both the docstring and the dtype comment are gone, so `_check_alpha` reads as it did before this PR plus the four lines that do the flooring. The reasoning stays here in the thread if it's ever needed.

The only production change left is:

```python
alpha_dtype = checked_alpha_tensor.dtype
tiny = np.finfo(
    alpha_dtype if alpha_dtype.startswith("float") else pytensor.config.floatX
).smallest_normal
checked_alpha_tensor = pt.maximum(checked_alpha_tensor, tiny)
```

This commit only deletes comment lines, so I checked it with `ruff check`/`ruff format --check` (clean) rather than re-running the suite that already passed on the previous commit.

vaibhav8a:
@juanitorduz both comment blocks are gone (dropped in `fix(mmm): drop the explanatory comments from _check_alpha`), and the branch is up to date with `main`. Ready for another look whenever you have a moment.

PR #1031, opened 2024-09-13
Add `BetaGeoBetaBinomModel`

ColtAllen:
## Description

Reopening https://github.com/pymc-labs/pymc-marketing/pull/922 due to the force push.

The BG/BB model is for non-contractual purchase opportunities across discrete time periods; a good example would be sporting events.

Sampling is rather slow because NUTS defaults to compound sampling due to the discrete distributions used in this model. I recommend that https://github.com/pymc-labs/pymc-marketing/pull/707 be merged to speed up performance. 

A small addition to `CLVBaseModel` was required for the input validations, and the dev notebook is rather minimalist right now but will be expanded into a full how-to in a future PR. An inefficiency in `distribution_new_customers` that is shared by all CLV models was also identified and will be fixed in a separate PR.

## Related Issue

- [ ] Closes https://github.com/pymc-labs/pymc-marketing/issues/176
- [ ] Related to https://github.com/pymc-labs/pymc-marketing/pull/707

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [x] Included tests that prove the fix is effective or that the new feature works
- [x] Added necessary documentation (docstrings and/or example notebooks)
- [x] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Modules affected

- [ ] MMM
- [x] CLV

## Type of change

- [x] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1031.org.readthedocs.build/en/1031/

PR #2902, opened 2026-08-18
fix(clv): make BetaGeoBetaBinom.logp natively broadcastable and cache the compiled logp

### LINKED ISSUE #1033: Create Class Properties for CLV `_logp` Methods
`ParetoNBDModel` and `BetaGeoBetaBinomModel` use their respective `logp` functions for some of the predictive methods. Right now`logp` is being compiled every time a predictive method is called, which is inefficient. Caching this a class property would be an easy way to speed up predictions for these models.

### LINKED ISSUE #2919: BUG: BetaGeoBetaBinom.logp is silently wrong for a length-1 value (collapses BetaGeoBetaBinomModel posterior draws)
## Summary

`BetaGeoBetaBinom.logp` silently returns a wrong result when `value` has length 1 against a longer parameter batch. `BetaGeoBetaBinomModel._logp` is exactly that call, so it does not produce a per-draw log-likelihood: the data-dependent term is computed once, effectively from draw 0, and reused for every other draw with only the `-betaln(alpha, beta) - betaln(gamma, delta)` normalizer varying across draws.

Everything that flows through `_logp` is affected: `expected_purchases`, `expected_probability_alive` and `expected_purchase_probability`. Posterior means stay close to correct, but posterior *variation* is collapsed, so credible intervals from these methods are not meaningful.

Direct users of the distribution are affected too, independently of the CLV models.

ParetoNBD is not affected.

## Root cause

One line in `pymc_marketing/clv/distributions.py` (L313-315). `x` is missing from the `broadcast_arrays` call:

```python
t_x, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
    t_x, alpha, beta, gamma, delta, T
)
...
unnorm_logp = scan(
    fn=logp_customer_died,
    outputs_info=[None],
    sequences=[t_x, x, alpha, beta, gamma, delta, T],
    return_updates=False,
)
```

Six of the seven scan sequences get broadcast to the common batch length; `x` keeps whatever length `value` came in with. `scan` truncates `n_steps` to the shortest sequence, so it follows `len(value)`, not the parameter batch. The length-1 result then broadcasts back out against the full-length normalizer, so the output has the right shape and nothing warns.

`t_x` also comes out of `value` and *is* in the broadcast list, so `len(value)` must be either 1 or the full batch length. That gives two reachable cases and the length-1 one is silent. From @shivamlalakiya's isolation, 6 parameter draws, one customer (`x=2`, `t_x=3`, `T=6`), against a scipy transcription of Fader, Hardie & Shang (2010) eq. (5), no `vectorize_graph` anywhere:

| `len(T)` | `len(value)` | max abs error | outcome |
|---|---|---|---|
| 1 | 1 | 0.513 | silently wrong |
| 1 | 3 | n/a | raises `Incompatible Elemwise input shapes [(6,), (3,)]` |
| 1 | 6 | 2.7e-15 | correct |
| 6 | 1 | 0.513 | silently wrong |
| 6 | 3 | n/a | raises |
| 6 | 6 | 2.7e-15 | correct |

Adding `x` to the broadcast turns every row that returns a number into 2.7e-15.

`BetaGeoBetaBinomModel._logp` builds `dummy_values = pt.tensor(shape=(1, 2), dtype=int)`, which is exactly the silent case. The dummy construction is not itself broken: it feeds the distribution a shape-`(1,)` value against shape-`(n_samples,)` parameters, which the signature permits, and the distribution returns garbage.

## Evidence

Checked against an independent scipy transcription of the BG/BB log-likelihood, 6 customers x 4 draws:

| branch | `max abs(branch - reference)` |
|---|---|
| `main` (`e9fc42e5`) | 14.56 (log space) |
| #2902 | 3.6e-14 |

Two checks that do not depend on the reference implementation being correct:

1. On `main`, the difference between draw 0 and each other draw is identical for every customer (e.g. `[0, 2.142, 1.556, -1.809]` repeated across all rows). Real per-draw likelihoods do not behave that way.
2. On `main`, runtime is flat in the number of draws. 300 customers at 25 / 100 / 400 draws: 0.77s / 1.12s / 0.98s. The scan is not running per draw.

@shivamlalakiya reproduced the `x` diagnosis independently on #2902, and @ricardoV94 pushed back on the original `vectorize_graph` explanation, which is what led to it.

## Why CI did not catch it

`test_expected_purchases` compares `est_num_purchases.mean(("chain", "draw"))` against the Fader/Hardie reference at `rtol=0.01`. Averaging over draws is exactly the operation that hides a draw-collapse, and the `mock_fit` draws are tight enough to stay inside 1%.

At the distribution level, `test_logp_matches_excel` uses scalar parameters with a 22-row `value`, so it lands on the correct row of the table above by accident. There is no test covering a length-1 `value` against a longer parameter batch.

## Affected versions

Two different ranges, because the distribution bug and the caller that triggers it arrived separately:

- **Distribution** (`BetaGeoBetaBinom.logp`, direct users): `x` has been outside the broadcast since cc6ee50c (`Add BetaGeoBetaBinom Distribution Block`, #431, 2024-05-27), first released in **0.7.0**. Affected `0.7.0` through `1.0.0` inclusive.
- **Model** (`BetaGeoBetaBinomModel`): affected from **0.9.0**, when 660686b (#1031, 2024-09-13) added the caller that passes a length-1 `value`. Affected `0.9.0` through `1.0.0` inclusive.

Verified with `git tag --contains` on both commits.

## Fix

The one-line fix is to add `x` to the broadcast:

```python
t_x, x, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
    t_x, x, alpha, beta, gamma, delta, T
)
```

plus a `TestBetaGeoBetaBinom` case covering a length-1 `value` against a longer parameter batch. With that in place, `_logp` can keep `main`'s structure (`dummy_T` / `dummy_values` / `vectorize_graph` all retained) and only needs the four parameters lifted to symbolic inputs so the compiled function can be cached for #1033. Scope is being settled on #2902.

shivamlalakiya:
## Summary

Two things, coupled:

1. `BetaGeoBetaBinom.logp` silently returned a wrong result for a length-1 `value` against a longer parameter batch. `BetaGeoBetaBinomModel._logp` was exactly that call, so BG/BB predictive methods collapsed all posterior variation to one draw plus a constant shift. Fixed by making the logp natively broadcastable (#2919).
2. `ParetoNBDModel._logp` and `BetaGeoBetaBinomModel._logp` rebuilt and compiled a fresh pytensor graph on every call, as flagged by the `# TODO: cache this as a property` comments on both. Both are now instance methods backed by a `@cached_property _logp_fn` compiled once per model instance and reused (#1033).

## The bug

`x` was missing from the `broadcast_arrays` call in `BetaGeoBetaBinom.logp`:

```python
t_x, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
    t_x, alpha, beta, gamma, delta, T
)
```

Six of the seven `scan` sequences got broadcast to the common batch length; `x` kept the length `value` came in with. `scan` truncates `n_steps` to the shortest sequence, so it followed `len(value)`. The length-1 result then broadcast back out against the full-length `- betaln(alpha, beta) - betaln(gamma, delta)` normalizer, so the output had the right shape and nothing warned.

`t_x` also comes out of `value` and *was* in the broadcast list, so `len(value)` had to be either 1 or the full batch length. Two reachable cases, and the length-1 one was the silent one. 6 parameter draws, one customer (`x=2`, `t_x=3`, `T=6`), against a scipy transcription of Fader, Hardie & Shang (2010) eq. (5):

| `len(T)` | `len(value)` | max abs error | outcome |
|---|---|---|---|
| 1 | 1 | 0.513 | silently wrong |
| 1 | 3 | n/a | raises `Incompatible Elemwise input shapes [(6,), (3,)]` |
| 1 | 6 | 2.7e-15 | correct |
| 6 | 1 | 0.513 | silently wrong |
| 6 | 3 | n/a | raises |
| 6 | 6 | 2.7e-15 | correct |

`_logp` on `main` builds `dummy_values = pt.tensor(shape=(1, 2), dtype=int)`, which is the silent case. So on releases 0.9.0 through 1.0.0, `expected_purchases`, `expected_probability_alive` and `expected_purchase_probability` for BG/BB have posterior means roughly right and credible intervals that are not meaningful. The distribution itself has been affected since 0.7.0 for any direct caller passing a length-1 `value`. Full write-up in #2919.

Why CI missed it: `test_expected_purchases` compares `.mean(("chain", "draw"))` at `rtol=0.01`, and averaging over draws is exactly the operation that hides a draw collapse. At the distribution level, `test_logp_matches_excel` uses scalar parameters with a 22-row `value`, so it lands on a correct row of the table above by accident.

## The fix

Per @ricardoV94's suggestion, the `scan` is gone. It only ever existed to hide the data-dependent `i_vec = pt.arange(pt.maximum(T - t_x, 0) + 1)`, which is not rectangular and so cannot be vectorized naively. That is what `OpFromGraph` is for. The core scalar computation is now wrapped in an `OpFromGraph` and `pt.vectorize`d, which makes the logp broadcast natively and removes:

- the `broadcast_arrays` call, and the bug with it (structurally, not patched)
- the `ndim > 1` NotImplementedError limitation
- the `atleast_1d` / `specify_shape(logp, 1).squeeze(0)` shape juggling
- the outer `vectorize_graph` in `BetaGeoBetaBinomModel._logp`, which now passes its natural shapes instead of tiling every argument into a flat `(customer * sample)` vector

Verified: matches the scipy reference to 3.1e-15; a length-1 `value` now equals a materialized one exactly (`0`, not `2.7e-15`); `ndim=2` parameters work; gradients through the `OpFromGraph` match finite differences to 3.2e-9, which matters because BG/BB is sampled with NUTS.

## Timings

BG/BB, end-to-end `_logp` including the numpy input prep, OpFromGraph versus the tile/repeat version this PR previously carried:

| customers | samples | OpFromGraph | tile/repeat + scan | ratio |
|---|---|---|---|---|
| 300 | 100 | 0.015s | 0.017s | 1.17x |
| 2,357 | 100 | 0.119s | 0.125s | 1.05x |
| 11,104 | 100 | 0.548s | 0.606s | 1.11x |
| 2,357 | 400 | 0.461s | 0.507s | 1.10x |

Results agree to 1.8e-15. The compiled call itself is about 14% slower than the flat scan, and that is more than offset by never materializing the `n_customers * n_samples` input arrays. That also removes the memory concern that motivated the customer chunking, which is out of this PR.

BG/BB **cannot** be compared against `main`: `main` was running roughly `1 / n_draws` of the work, so any speedup measured against it is partly just the missing computation.

ParetoNBD is a like-for-like comparison, since its maths is unchanged:

| | 5 calls to `expected_purchases`, 2,357 customers, 2x50 draws |
|---|---|
| `e9fc42e5` | 0.78s |
| this PR | 0.12s (~6x, results agree to `rtol=1e-6`) |

## Tests

- `test_logp_broadcasts_value_against_parameters`: a length-1 `value` must equal a materialized one. Asserts the invariant rather than comparing to a reference, per @daimon-pymclabs, so it holds even if a reference transcription drifts. Fails on the previous code with 5/6 elements mismatched, max abs difference 0.513.
- `test_logp_supports_batched_parameters`: `ndim=2` parameters, which previously raised. Replaces `test_notimplemented_logp`, which guarded the removed limitation.
- `test_logp_varies_across_draws` (BG/BB): compares each draw's offset from draw 0 *per customer*, so a draw collapse cannot hide behind the average. Verified to fail on `e9fc42e5` (`assert 7.7e-16 > 1e-08`) and pass here (4.1e-3).
- `test_logp_fn_is_compiled_once` (both models): the cached function is compiled once and reused.

- [x] `uv run pytest tests/clv/ -q` — 339 passed, 16 skipped (pre-existing skips), 0 failed
- [x] `uv run pre-commit run` on the changed files (ruff, ruff-format, mypy) — clean

## Notes

- All logp inputs are now cast to `floatX`. Under the default `floatX=float64` this is a no-op for the existing call sites; it is a real change from the previous integer `dummy_T` / `dummy_values` path.
- Not done here: tightening the input validation to reject inconsistent lengths, which @daimon-pymclabs raised. With native broadcasting there is no longer a guard to tighten, and nothing has reported a need for one. Happy to open a follow-up if wanted.
- Left non-blocking: BG/BB's `_logp_fn` doesn't reference `self`, so a module-level cache would amortize the compile across model instances. Kept per-instance to match ParetoNBD, whose graph genuinely depends on the covariate columns.

Closes #1033
Closes #2919

----
📚 Documentation preview 📚: https://pymc-marketing--2902.org.readthedocs.build/en/2902/


daimon-pymclabs:
@shivamlalakiya this is a really nice piece of work, and it turns out to be considerably more valuable than the issue it closes. While benchmarking it I went looking for why the two branches disagreed numerically, and the answer is that **`main` is wrong and your rewrite fixes it**. Thank you for pulling on this thread.

Below: the correctness finding, then four concrete asks. Three of them are small.

---

## 1. This PR is a bug fix, not just a perf win

`BetaGeoBetaBinomModel._logp` on `main` (`e9fc42e5`) does not produce a per-draw log-likelihood. It computes the data-dependent term once, from draw 0, and reuses it for every other draw with only the `-betaln(alpha, beta) - betaln(gamma, delta)` normalizer varying.

Checked against an independent scipy transcription of the BG/BB log-likelihood, 6 customers, 4 draws:

| branch | `max abs(branch - reference)` |
|---|---|
| this PR | 3.6e-14 |
| `main` | 14.56 (log space) |

Two checks that do not depend on my reference implementation being correct:

1. On `main`, the difference between draw 0 and each other draw is **identical for every customer**: `[0, 2.142, 1.556, -1.809]` repeated across all 6 rows. Real per-draw likelihoods do not behave that way.
2. On `main`, runtime is flat in the number of draws. 300 customers at 25 / 100 / 400 draws: 0.77s / 1.12s / 0.98s. The scan is simply not running per draw.

Root cause is exactly the construction you replaced. `dummy_T` and `dummy_values` have shape `(1,)`, so inside `BetaGeoBetaBinom.logp` the scan sequences get truncated to length 1 instead of length `n_samples`, and the length-1 result broadcasts back out through `vectorize_graph`.

Consequence: on released versions, `expected_purchases`, `expected_probability_alive` and `expected_purchase_probability` for BG/BB collapse all posterior variation to draw 0 plus a constant shift. Point estimates are close, but credible intervals are not meaningful.

This also explains the timing difference some might read as a regression: `main` at 11,104 customers × 4,000 draws takes 78.0s versus 168.4s here, because `main` was doing roughly `1 / n_draws` of the scan work. Your branch is the one doing the full computation.

Why CI never caught it: `test_expected_purchases` compares `est_num_purchases.mean(("chain", "draw"))` against the Fader/Hardie reference at `rtol=0.01`. Averaging over draws is precisely the operation that hides a draw-collapse, and the `mock_fit` draws are tight enough to stay inside 1%.

**Ask:** retitle to something like `fix(clv): correct BG/BB per-draw logp and cache compiled logp function`, swap the `enhancement` / `priority: low` labels for `bug`, and add a release note. Worth opening a companion bug issue describing the `main` defect so it is searchable for anyone who sees odd BG/BB intervals on a released version. Happy to open that issue if you prefer.

## 2. Please add customer chunking to BG/BB (the one blocker)

Building `alpha_rep` and friends materializes seven arrays of `n_customers × n_samples` elements up front (`values_rep` is 2×). Measured peak RSS at 4 chains × 1,000 draws: **25,000 customers needs about 12 GB**, which will OOM on most laptops. The benchmark in the description used `draws=50`, which is 80× smaller, so it could not surface this.

Fix is a loop over customer chunks. The function is already compiled once, so this costs nothing:

```python
# module level
_MAX_CHUNK_ELEMENTS = 2_000_000  # ~16 MB per float64 intermediate


def _logp(self, alpha, beta, gamma, delta, x, t_x, T):
    n_chain, n_draw = alpha.sizes["chain"], alpha.sizes["draw"]
    n_samples = n_chain * n_draw

    a = alpha.stack(sample=("chain", "draw")).values
    b = beta.stack(sample=("chain", "draw")).values
    g = gamma.stack(sample=("chain", "draw")).values
    d = delta.stack(sample=("chain", "draw")).values
    T_vals = T.values
    values = np.stack((t_x.values, x.values), axis=-1)

    chunk = max(1, _MAX_CHUNK_ELEMENTS // n_samples)
    loglike = np.concatenate(
        [
            self._logp_fn(a, b, g, d, T_vals[i : i + chunk], values[i : i + chunk])
            for i in range(0, len(T_vals), chunk)
        ],
        axis=0,
    )  # (n_customers, n_samples)

    loglike = np.moveaxis(loglike.reshape((-1, n_chain, n_draw)), 0, -1)
    return xarray.DataArray(data=loglike, dims=("chain", "draw", "customer_id"))
```

Note this also lets the `np.tile` / `np.repeat` block go away entirely: the tiling moves inside `_logp_fn`, which needs the same broadcast it already performs. Keep whichever shape handling you find clearer, the point is only that peak memory becomes `O(chunk × n_samples)` rather than `O(n_customers × n_samples)`.

I measured the chunked version as bit-identical to this PR at every size I tried, and marginally faster. The knob is not sensitive: at 11,104 customers, `500_000` gives 437 MB / 158.8s and `8_000_000` gives 1260 MB / 153.4s. `2_000_000` is a good default.

ParetoNBD does not need this. Its inputs stay at their natural sizes.

## 3. Add a test that would have caught the draw-collapse

The current reference test cannot see this class of bug because it averages over draws. Something like this would have failed on `main` and passes here:

```python
def test_logp_is_not_collapsed_across_draws(self, model_with_fit, data):
    """Each posterior draw must have its own data-dependent log-likelihood.

    Before this fix the scan ran only for draw 0 and the result was broadcast, so
    the per-draw offsets were identical for every customer.
    """
    posterior = model_with_fit.idata.posterior
    loglike = model_with_fit._logp(
        alpha=posterior["alpha"],
        beta=posterior["beta"],
        gamma=posterior["gamma"],
        delta=posterior["delta"],
        x=data["x"].to_xarray(),
        t_x=data["t_x"].to_xarray(),
        T=data["T"].to_xarray(),
    )  # dims: (chain, draw, customer_id)

    # offset of each draw relative to draw 0, per customer
    offsets = loglike - loglike.isel(chain=0, draw=0)
    # a pure normalizer shift would be constant across customers
    assert float(offsets.std("customer_id").max()) > 1e-8
```

Adapt the fixture names to whatever `tests/clv/models/test_beta_geo_beta_binom.py` already provides.

## 4. Add a cheap assertion that the caching cannot silently regress

```python
def test_logp_fn_is_compiled_once(self, model_with_fit):
    assert "_logp_fn" not in model_with_fit.__dict__  # not compiled yet
    model_with_fit.expected_purchases()
    compiled = model_with_fit.__dict__["_logp_fn"]
    model_with_fit.expected_purchases()
    assert model_with_fit.__dict__["_logp_fn"] is compiled
```

---

### Minor, non-blocking

- `allow_input_downcast=True` combined with casting to `floatX` is a real behavioural change from the previous integer `dummy_T` / `dummy_values`. One docstring line saying inputs are cast to `floatX` would help the next reader.
- BG/BB's `_logp_fn` does not reference `self` at all, so a module-level cache would amortize the roughly 1.9s compile across model instances rather than paying it per instance. Fine to leave as is for now.
- `_logp_fn` has no return type annotation. `pytensor.compile.function.types.Function` works.

Nice work on this. Once chunking and the two tests land, plus the relabel, I think this is good to go.

shivamlalakiya:
Thanks, this was a genuinely useful review. I reproduced the draw-collapse independently before changing anything, and you're right on all three counts.

**1. It's a bug fix.** Confirmed against my own scipy transcription of eq. (5), 6 customers x 4 draws: this branch is 5.3e-15 from the reference, `e9fc42e5` is 4.54 off in log space. And the tell you pointed at reproduces exactly: on `e9fc42e5` the per-draw offset from draw 0 is `[0, -0.456, -1.346, +1.989]` repeated identically across every customer.

Retitled to `fix(clv): ...`. There's no `CHANGELOG` in the repo, so the title is the release note; I can't change labels myself, so `enhancement` / `priority: low` -> `bug` needs someone with write access. I've also dropped the `~17x` BG/BB number from the description rather than keep a speedup that was partly just the missing scan work, and noted explicitly that this branch is slower than `e9fc42e5` at high draw counts for that reason. On the companion issue: happy to open it, but I'd rather not file against the repo unprompted, so say the word (or open it yourself) and I'll write it up.

**2. Chunking is in.** Went with keeping the `np.tile` / `np.repeat` inside the chunk loop rather than pushing the tiling into the graph, since it's the smaller diff and the peak-memory property is the same. Module-level `MAX_LOGP_CHUNK_ELEMENTS = 2_000_000`, as you suggested. Measured at 20,000 customers x (2 chains, 500 draws): 2173 MB unchunked vs 823 MB chunked, identical checksum. Bit-identical at every chunk size I tried, down to `chunk_size == 1`.

**3 and 4. Both tests added,** plus one for the chunking itself:

- `test_logp_varies_across_draws` — I checked it actually fails on `e9fc42e5` rather than assuming: `assert 7.7e-16 > 1e-08`. On this branch the same quantity is 4.1e-3, so there's about five orders of magnitude of margin either side of the threshold.
- `test_logp_is_chunked_over_customers` — bounded inputs to the compiled function, unchanged result.
- `test_logp_fn_is_compiled_once` — added for ParetoNBD too, since the caching change lands in both models.

**Minor points.** `floatX` cast is now documented in both `_logp_fn` docstrings. Return type annotated, though note `pytensor.compile.function.types.Function` no longer resolves on pytensor 3.2.4; `pytensor.compile.Function` is the path that works. Left the module-level cache alone for now: BG/BB's `_logp_fn` doesn't touch `self`, but keeping it per-instance matches ParetoNBD, whose graph does depend on the covariate columns, and the per-instance version already removes the per-call compile.

`uv run pytest tests/clv/ -q`: 339 passed, 16 skipped, 0 failed.

daimon-pymclabs:
All four handled on my side, and this now looks good to me.

**Labels.** Done: `enhancement` and `priority: low` off, `bug` and `priority: high` on.

**Companion issue.** Opened as #2919 with the reproduction, the root cause, the reason CI missed it, and the affected range. On that last point: the construction came in with 660686b (#1031, 2024-09-13), so every release from `0.9.0` through `1.0.0` inclusive is affected, which is worth knowing for anyone who has published BG/BB intervals from a released version.

**Chunking.** Read through the implementation. Keeping `tile`/`repeat` in the loop is the right call, and the index arithmetic is consistent end to end: `tile` varies the sample fastest, `repeat` varies the customer slowest, so the flat position is `customer * n_samples + sample`, which is exactly what `reshape((n_customers, n_chain, n_draw))` then unpacks given `stack(sample=("chain", "draw"))` orders samples as `chain * n_draw + draw`. Concatenating chunk-wise preserves that because chunks are contiguous in customer order. Peak memory is now `O(chunk_size * n_samples)` as intended.

**Tests.** `test_logp_varies_across_draws` is the one that matters and I like that you measured both sides of the threshold rather than trusting it. Five orders of magnitude of margin is plenty.

Two small notes, neither blocking:

- `pytensor.compile.Function` is the correct import, agreed. `pytensor.compile.function.types.Function` is a private path that has moved before, so the public one is the safer annotation regardless of version.
- `cached_property` is fine here. I checked that neither `CLVModel` nor `ModelBuilder` pickles or deep-copies the instance, so stashing a compiled function in the instance `__dict__` has nothing to trip over. Worth remembering if serialization ever changes.

Agreed on leaving the cache per-instance. ParetoNBD's graph genuinely depends on the covariate columns, and symmetry between the two models is worth more than the marginal saving of sharing BG/BB's graph across instances.

Nothing further from me. The remaining step is a maintainer merge.

Labeeb2339 (commented):
I reviewed the cached-function construction and the BG/BB chunking path. The compiled inputs use dynamic lengths, the flattened parameter/data ordering matches `np.repeat` over each customer and `np.tile` over posterior samples, and the reshape restores `(chain, draw, customer_id)` in the documented order. The chunk size is derived from the sample count and the test replaces the cache with a spy to prove bounded calls while comparing the complete result. The per-draw regression test checks the failure mode that averaging reference values would hide, and the ParetoNBD cache test covers reuse across different horizons. The full CLV test matrix and lint checks are green; I found no additional correctness issue.

ricardoV94:
> Root cause is exactly the construction you replaced. dummy_T and dummy_values have shape (1,), so inside BetaGeoBetaBinom.logp the scan sequences get truncated to length 1 instead of length n_samples, and the length-1 result broadcasts back out through vectorize_graph.

This doesn't make sense to me. The distribution was defined as having scalar parameters (see the signature). The scan was just used to go over batch dimensions of the arguments, we couldn't do a naive vectorization because the inner graph is not rectangular (there's a data/parameter dependent variable size expansion inside the logsumexp). The outer vectorize_graph with shape=(1,) on the core scan should create a Blockwise where every step is the valid scan? The only thing the Scan does here is to work like an OpFromGraph to hide the intermediate variable-size vector, but otherwise it never needed to iterate over "a sequence". There may be a problem, maybe scan eager n_steps==1 inlined output. The explanation and solution is very unconvincing.

CC @daimon-pymclabs

pymc_marketing/clv/models/beta_geo_beta_binom.py
```diff
@@ -258,31 +288,43 @@ def _logp(
...
+        gamma_samples = gamma.stack(sample=("chain", "draw")).values
+        delta_samples = delta.stack(sample=("chain", "draw")).values
+        T_values = T.values
+        values = np.stack((t_x.values, x.values), axis=-1)
+
+        # The BetaGeoBetaBinom distribution only works with vector parameters, so the
+        # per-sample params and per-customer data are each repeated to cover every
+        # (customer, sample) pair as a single flat vector. Customers are evaluated in
+        # chunks so that peak memory stays bounded instead of growing with
+        # `n_customers * n_samples`.
```
ricardoV94: anybody asked for this?

daimon-pymclabs:
@ricardoV94 You're right, and my explanation was wrong. I went and reproduced it rather than argue: the failure has nothing to do with `vectorize_graph` and nothing to do with scan iterating over batch dimensions. It is a one-line bug inside `BetaGeoBetaBinom.logp`.

`x` is missing from the `broadcast_arrays` call (`distributions.py:313-315`):

```python
t_x, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
    t_x, alpha, beta, gamma, delta, T
)
...
unnorm_logp = scan(
    fn=logp_customer_died,
    outputs_info=[None],
    sequences=[t_x, x, alpha, beta, gamma, delta, T],
    return_updates=False,
)
```

Six of the seven sequences get broadcast to the common batch length; `x` keeps whatever length `value` came in with. `scan` truncates to the shortest sequence, so `n_steps` follows `len(x)`, not the broadcast batch shape. The length-1 scan output then broadcasts back out against the full-length `- betaln(alpha, beta) - betaln(gamma, delta)` correction, so the result has the right shape and no warning is raised.

Isolation, 6 parameter draws, one customer (`x=2`, `t_x=3`, `T=6`), against a scipy transcription of Fader/Hardie eq. (5). No `vectorize_graph` anywhere:

| `T` shape | `value` shape | scan `n_steps` | max abs error |
|---|---|---|---|
| `(1,)` | `(1, 2)` | 1 | 2.16 |
| `(1,)` | `(6, 2)` | 6 | 8.9e-16 |
| `(6,)` | `(1, 2)` | 1 | 2.16 |
| `(6,)` | `(6, 2)` | 6 | 8.9e-16 |

It tracks `value`'s length exactly, in both directions, which is what pins it on `x` rather than on `T`. Adding `x` to the broadcast:

```python
t_x, x, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
    t_x, x, alpha, beta, gamma, delta, T
)
```

gives `n_steps=6` and 8.9e-16 on the first row too. (pytensor 3.0.7, pymc 6.0.1, pymc-marketing at `fabba92`.)

So the correct framing is the opposite of what I wrote. `_logp`'s dummy construction was not itself broken; it fed the distribution a shape-`(1,)` value against shape-`(n_samples,)` parameters, which the signature permits, and the distribution silently returned garbage. Materializing every argument to full length, which is what this PR does, sidesteps the bug but does not fix it: `BetaGeoBetaBinom.logp` is still wrong for any broadcasting call, including a direct user call outside the CLV models.

That changes what should happen here:

1. The one-line broadcast fix belongs in `distributions.py`, with a test that a broadcast `value` matches a materialized one. I'd rather that landed than the workaround alone.
2. #2919 needs correcting: the root cause and the reproduction I gave there are both wrong. The affected range claim survives only if `x` has been outside that broadcast since 660686b, which I have not checked yet.
3. Whether `_logp` still wants the flat-vector rewrite once the distribution is fixed is a separate question, and honestly a weaker one, since the original construction would then be correct.

@shivamlalakiya sorry, this lands on your PR. Happy to do (1) and (2) myself if you'd rather keep this branch focused.

On the chunking, `beta_geo_beta_binom.py:307`: nobody asked for it. I suggested it in review off my own concern about `n_customers * n_samples` intermediates, not off a reported problem, and Shivam implemented what I asked for. That's on me, and it doesn't belong in a bug fix. My vote is to strip the chunking and `MAX_LOGP_CHUNK_ELEMENTS` out of this PR entirely, and for it to come back only if someone actually hits the memory ceiling.

shivamlalakiya:
@ricardoV94 @daimon-pymclabs

I reproduced the `x` diagnosis independently and it holds. One detail worth adding, because it pins the blast radius exactly.

`t_x` and `x` both come out of `value`, but only `t_x` is in the `broadcast_arrays` call, so `len(value)` already has to be broadcast-compatible with the parameter batch. That leaves two reachable shapes, not a spectrum:

| `len(T)` | `len(value)` | max abs error vs scipy | outcome |
|---|---|---|---|
| 1 | 1 | 0.513 | silently wrong |
| 1 | 3 | n/a | raises `Incompatible Elemwise input shapes [(6,), (3,)]` |
| 1 | 6 | 2.7e-15 | correct |
| 6 | 1 | 0.513 | silently wrong |
| 6 | 3 | n/a | raises |
| 6 | 6 | 2.7e-15 | correct |

6 parameter draws, one customer (`x=2`, `t_x=3`, `T=6`), reference is a scipy transcription of Fader/Hardie eq. (5), no `vectorize_graph` anywhere. Adding `x` to the broadcast turns every row that currently returns a number into 2.7e-15.

So there is no quiet middle case: `len(value)` is either 1 or the full batch length, and the length-1 case is the silent one. `_logp`'s `dummy_values = pt.tensor(shape=(1, 2), dtype=int)` is exactly that case.

I also confirmed the one-liner is sufficient for the model path: with `x` broadcast, `n_steps` becomes `n_samples`, `vectorize_graph` Blockwises that core scan over customers, and the result comes out `(n_customers, n_samples)`. So my `_logp` rewrite is not a correctness fix and I should not have framed it as one.

**Chunking: removed.** `MAX_LOGP_CHUNK_ELEMENTS`, the chunk loop and `test_logp_is_chunked_over_customers` are all gone. `uv run pytest tests/clv/ -q`: 338 passed, 16 skipped. Agreed it did not belong in a bug fix.

**On the remaining scope**, there is a third option between "keep the flat-vector rewrite" and "revert to main" that I do not think has been named yet. What blocks caching on `main` is that the graph bakes the posterior in as numpy constants (`alpha=alpha.stack(sample=("chain", "draw")).values` passed to `.dist()`), so the four parameters have to become symbolic inputs either way. But `dummy_T`, `dummy_values` and `vectorize_graph` can all stay exactly as they are, and the compiled function still caches. That gets #1033 with a much smaller diff than this branch currently carries, and no tile/repeat at all.

@ricardoV94 your call on which you would rather have:

1. distribution one-liner lands here, plus caching only via the `vectorize_graph`-preserving version above
2. distribution one-liner lands separately (@daimon-pymclabs offered to take it), this PR narrows to caching only
3. something else on the distribution side, for example extending the existing `ndim > 1` guard to reject inconsistent lengths rather than leaning on broadcast

Fine with any of them. Happy to write the distribution fix plus a `TestBetaGeoBetaBinom` case covering the length-1 `value` if you would rather it came from here. I would just rather not pick the scope myself twice in one PR.

**For #2919:** the missing-`x` line predates #1031. It came in with cc6ee50c (#431, 2024-05-27), first released in 0.7.0. So the distribution has been affected since 0.7.0, and the BG/BB model since 0.9.0, because that is when #1031 added the caller that passes a length-1 `value`. The model-level affected range in the issue is right, the root-cause commit is not.

shivamlalakiya:
Follow-up on option 1, because I had a claim of my own to retract.

This PR's description said that making the placeholders symbolic while keeping `vectorize_graph` "hits a pytensor vectorize/Blockwise limitation (`Runtime broadcasting not allowed`)". That was wrong. It is the same missing `x` broadcast, one layer down.

With `x` absent from the broadcast, `vectorize_graph` over the symbolic-parameter graph fails at build time:

```
ValueError: Incompatible vectorized shapes for input 2 and axis 1.
Input 2 has shape 1, but is not statically known to have shape 1.
```

Baking the parameters in as constants let pytensor constant-fold that shape away, which is why the original code compiled at all. It compiled and returned the collapsed result. Symbolic inputs simply removed the constant-folding that was hiding the bug, so pytensor was right to complain and I misread it as a library limitation.

Adding `x` to the broadcast makes the symbolic version compile and agree with the flat tile/repeat approach to `2.66e-15` (6 samples x 4 customers, `dummy_T` / `dummy_values` / `vectorize_graph` all retained exactly as on `main`).

So option 1 is verified, not just plausible: with the distribution one-liner in place, `_logp` keeps `main`'s structure and only lifts the four parameters to symbolic inputs so the compiled function can be cached. That is a much smaller diff than this branch, and it drops the `np.tile` / `np.repeat` block entirely.

It also means the "avoided a pytensor bug" framing was backwards: there was no pytensor bug, and the flat-vector rewrite was working around our own. I have corrected the description accordingly.

Still yours to call, @ricardoV94, but I would now actively prefer option 1 over what is on the branch.

daimon-pymclabs:
@shivamlalakiya Both of your corrections check out, and I've verified them rather than just taking them.

**The commit range.** I owed this one, since I filed the affected range on #2919 and then said I hadn't checked whether `x` predated 660686b. `git show cc6ee50c:pymc_marketing/clv/distributions.py` has:

```python
_, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
    t_x, alpha, beta, gamma, delta, T
)
```

so `x` has been outside the broadcast since the distribution landed. (`t_x` was discarded outright at that point and only later bound to the broadcast result, which changes nothing about `x`.) `git tag --contains` gives 0.7.0 for cc6ee50c and 0.9.0 for 660686b, so your split is right: distribution affected from 0.7.0 for any direct caller passing a length-1 `value`, model affected from 0.9.0 once #1031 added such a caller.

**#2919 is corrected.** New root cause, split affected ranges, retitled to name the distribution rather than the model, and a comment on the issue saying what changed and why, so the wrong version isn't the one people find. Your two-reachable-shapes table is in there, credited.

**Option 1 is my preference too**, now that you've shown the symbolic-parameter version compiles with `dummy_T` / `dummy_values` / `vectorize_graph` all retained. My offer to take the distribution fix separately was made when option 1 looked blocked by a pytensor limitation; it isn't, so splitting it now would only mean the fix and its regression test land in different PRs, with this branch carrying a workaround in the meantime. Keep it here. Ricardo's call on the merge, but I don't think option 2 has anything left to recommend it.

**On option 3** (tightening the `ndim > 1` guard at L304-308 to reject inconsistent lengths instead of leaning on broadcast): worth doing, but not instead of the one-liner. Length-1-against-batch is legitimate broadcasting that the signature permits, and the callers rely on it. A guard would convert today's silent wrong answer into an exception on a call that ought to work. The one-liner makes it correct; a guard only makes the already-raising case raise earlier with a better message.

**Scope check on the rest of the module**, since the natural next question is how wide this goes: `BetaGeoBetaBinom` is the only distribution in `distributions.py` that uses `scan` over sequences, and the only one that calls `broadcast_arrays`. `ParetoNBD`, `BetaGeoNBD`, `ModifiedBetaGeoNBD` and `BetaGeoBetaBinom`'s neighbours are pure elemwise, so numpy broadcasting handles mixed lengths natively and there is no `n_steps` to truncate. The bug is confined to BG/BB.

One request for the test you offered to write: alongside the length-1 `value` case, assert that the length-1 result *equals* the materialized full-length one, not just that it's close to a reference. That's the invariant the broadcast is supposed to guarantee, and it's the thing that would catch a future regression here regardless of whether the reference transcription drifts.

ricardoV94:
I suggest dropping the scan altogether and using OpFromGraph + pt.vectorize for the base distribution logp. Then can remove the ndim>1 limitation and the need for the outer vectorize_graph for the response function as the logp will be natively broadcastable

shivamlalakiya:
@ricardoV94 Done, pushed in 68fe841. You were right twice: the scan was never doing what my explanation claimed, and `OpFromGraph` + `pt.vectorize` is a better fix than the one-line broadcast I was about to write.

What it removed, beyond the bug:

- the `broadcast_arrays` call, so the missing-`x` bug is gone structurally rather than patched
- the `ndim > 1` NotImplementedError, as you said it would
- the `atleast_1d` / `specify_shape(logp, 1).squeeze(0)` shape juggling
- the outer `vectorize_graph` in `BetaGeoBetaBinomModel._logp`, which now just passes its natural shapes

Verification before I committed to the approach:

| check | result |
|---|---|
| scipy transcription of eq. (5), 22 customers | 3.1e-15 |
| length-1 `value` vs materialized | `0` |
| `ndim=2` parameters vs flat equivalent | `0` |
| `pt.grad` wrt alpha vs finite differences | 3.2e-9 |

The gradient one is why I checked before rewriting rather than after: BG/BB is sampled with NUTS, so the logp has to differentiate through the variable-size `arange` inside the `OpFromGraph`. It does.

On perf, since #1033 is a perf issue and I did not want to hand you a regression. The compiled call is about 14% slower than the flat scan, but end-to-end `_logp`, including the numpy prep the flat path needs, it comes out ahead because the `n_customers * n_samples` arrays are never built:

| customers | samples | OpFromGraph | tile/repeat + scan | ratio |
|---|---|---|---|---|
| 300 | 100 | 0.015s | 0.017s | 1.17x |
| 2,357 | 100 | 0.119s | 0.125s | 1.05x |
| 11,104 | 100 | 0.548s | 0.606s | 1.11x |
| 2,357 | 400 | 0.461s | 0.507s | 1.10x |

Agreeing to 1.8e-15. That also disposes of the memory concern the chunking was for.

`test_notimplemented_logp` is gone, since it guarded the limitation you removed. In its place: a test that a length-1 `value` equals a materialized one (fails on the old code, 5/6 elements mismatched, max abs difference 0.513) and a test that batched parameters work.

`uv run pytest tests/clv/ -q`: 339 passed, 16 skipped, 0 failed. Description rewritten to match. @daimon-pymclabs the invariant-not-reference framing you asked for is in `test_logp_broadcasts_value_against_parameters`.

One thing I did not do: tighten the input validation to reject inconsistent lengths. With native broadcasting there is no guard left to tighten, and nothing has reported needing one, so it seemed like the wrong thing to add to a bug fix. Happy to open a follow-up issue instead if either of you wants it tracked.

daimon-pymclabs:
@shivamlalakiya I reproduced the `OpFromGraph` version rather than reading the diff, since I got the diagnosis wrong once in this thread already and would rather not do it twice. It holds, and it holds for the reason you gave.

pytensor 3.0.7, pymc 6.0.1, PR at `68fe841`, `main` at `fabba92`, `floatX=float64`. Reference is an independent scipy transcription of Fader, Hardie & Shang (2010) eq. (5), written from the paper rather than from either branch.

| check | `68fe841` | `fabba92` (main) |
|---|---|---|
| length-1 `value`, 6 parameter draws, vs scipy | 1.8e-15 | 2.001 |
| length-1 `value` vs materialized | exactly `0` | 2.001 |
| 22 customers, scalar params, ragged `T - t_x`, vs scipy | 2.7e-15 | n/a |
| `ndim=2` parameters (3x4) vs scipy | 2.7e-15 | `NotImplementedError` |
| `pt.grad` wrt alpha vs central differences | 1.9e-9 | n/a |
| `value.ndim == 1` | scalar out, 8.9e-16 | |
| `T < t_x` | `-inf`, no crash | |

Three things I would flag as load-bearing. The length-1 case is now *exactly* equal to the materialized one, not merely close, which is what you would expect if the batch is a real broadcast rather than a re-materialization, and it is the cleanest evidence that the bug is gone structurally. The `ndim > 1` limitation is removed rather than relocated. And the gradient survives the variable-size `arange` inside the `OpFromGraph`, which is the part I would have worried about most given BG/BB is sampled with NUTS.

Two notes, neither blocking:

**Non-integer `T`.** It is accepted silently and returns a different number than the rounded value (`T=6.5` gives -4.6436, `T=6.0` gives -4.5159). I checked main and it produces the identical pair, so this is not a regression. I raise it only because the description now says all inputs are cast to `floatX`, which reads like it might be new behaviour. It isn't, and rejecting it belongs in its own issue if anyone wants it.

**The ParetoNBD covariate branch.** `_logp_fn` moves it from a runtime check on `alpha.dims` to a compile-time check on `self.purchase_covariate_cols`. Those are equivalent: `build_model` gives `alpha` `dims="customer_id"` exactly when `purchase_covariate_cols` is truthy, and `_extract_predictive_variables` mirrors that. Worth saying out loud because the compiled input now carries a fixed ndim, so if the two ever drifted it would surface as a pytensor type error rather than a silent broadcast, which is the failure mode you want.

On your open question: I do not think the input-length validation needs a follow-up issue. It would have guarded a shape combination that now broadcasts correctly, so there is nothing left to reject. Agreed it does not belong here either way.

`test_logp_broadcasts_value_against_parameters` is the right shape for that test, thanks for taking the invariant-over-reference point.

One thing I did not do: re-run the full `tests/clv/` suite locally. CI is green on `68fe841` (55 passing, 2 skipped publish jobs), so I am relying on that rather than claiming it myself.

PR #2903, opened 2026-08-18
feat(mmm): add prefix option to LinearTrend changepoint coords/variables

### LINKED ISSUE #1568: Add prefix to all change point coords / variables
Unlikely to have multiple within a model. However, names like "delta" or "changepoint" become unusable without

https://github.com/pymc-labs/pymc-marketing/blob/9942c75cd8119a6c71f76d1882ead82b724db2aa/pymc_marketing/mmm/linear_trend.py?plain=1#L210-L226

shivamlalakiya:
## Summary
- `LinearTrend.apply()` always registers the `"changepoint"` coordinate and `"delta"`/`"k"` variables with fixed names. Two `LinearTrend` instances in the same `pm.Model` (e.g. a national + a regional trend) collide on these names.
- Adds an optional `prefix: str = ""` field to `LinearTrend`. When set, the changepoint coordinate and parameter variable names are namespaced as `f"{prefix}_changepoint"` / `f"{prefix}_delta"` / `f"{prefix}_k"`. Default is `""`, which reproduces the exact current unprefixed names — fully backward compatible, no behavior change for existing users.
- `sample_prior()` / `sample_curve()` updated consistently so the prefixed variable names round-trip correctly between the two.
- `_check_dims_are_subsets` validator updated so priors with `dims="<prefix>_changepoint"` validate correctly when a prefix is set.

## Design note (judgment call, flagging for discussion)
`LinearTrendEffect` in `additive_effect.py` already carries its own `prefix` (used for `{prefix}_t` / `{prefix}_effect_contribution`) and internally calls `self.trend.apply(t)`. I deliberately did **not** wire `LinearTrendEffect`'s prefix into `self.trend.prefix` automatically (unlike the analogous `EventAdditiveEffect.model_post_init` -> `self.effect.basis.prefix = self.prefix` pattern), because doing so would silently rename `"delta"`/`"k"`/`"changepoint"` to `f"{prefix}_..."` for every existing `LinearTrendEffect` user, breaking idata/trace compatibility.

Left as opt-in: users who want multiple `LinearTrend`s in one model pass `prefix=...` to `LinearTrend` directly today. Happy to add the auto-wiring in `LinearTrendEffect` as a follow-up if maintainers are fine with it being a documented behavior change (with a release note), per the issue's own "prefix spelling is a judgment call" ambiguity.

## Test plan
- [x] `uv run pytest tests/mmm/test_linear_trend.py -q` — 23 passed (19 existing + 4 new: `test_apply_with_prefix`, `test_two_prefixed_trends_in_same_model_do_not_collide`, `test_prefix_sample_prior_and_curve_roundtrip`)
- [x] `uv run pytest tests/mmm/test_additive_effect.py tests/mmm/test_linear_trend.py -q` — 61 passed, confirming `LinearTrendEffect` is unaffected
- [x] `pre-commit run` on both changed files (ruff, ruff-format, mypy) — clean

Closes #1568

----
📚 Documentation preview 📚: https://pymc-marketing--2903.org.readthedocs.build/en/2903/


juanitorduz (approved):
Thanks @shivamlalakiya

PR #2905, opened 2026-08-18
Add ECDF plot_type to plot_expected_purchases_ppc

### LINKED ISSUE #1226: Add `arviz` ECDF plots to `clv.plotting.plot_expected_purchases_ppc`
The ECDF plots from the [ParetoNBD notebook](https://github.com/pymc-labs/pymc-marketing/blob/main/docs/source/notebooks/clv/pareto_nbd.ipynb) could be a good addition to [`plot_expected_purchases_ppc`](https://github.com/pymc-labs/pymc-marketing/pull/1222):

image

shivamlalakiya:
## Summary

Adds `plot_type="ecdf"` to `plot_expected_purchases_ppc`. The default stays `"hist"` and is unchanged. The new option draws the observed ECDF of purchase counts against a 95% simultaneous confidence band, plus a companion difference plot, which is the figure asked for in #1226. Closes #1226.

- **The band is not reimplemented here.** `arviz` no longer ships the top-level `plot_ecdf` (removed in the 1.x split into `arviz_base`/`arviz_stats`/`arviz_plots`), but the statistics survive in `arviz_stats.ecdf_utils`, so the band comes from `compute_ecdf` + `simulate_confidence_bands` + `get_pointwise_confidence_band` (Säilynoja et al. 2021). Only the two-panel assembly is local. There is no public arviz path that returns this band: `arviz_stats.ecdf` computes points only, and `arviz_plots.plot_ppc_pit` draws a different chart (PIT difference ECDF on a [0, 1] axis) and returns a `PlotCollection` rather than Axes.
- **`arviz-stats` is now declared explicitly** rather than inherited from the `arviz` meta-package, since we import it directly. No version cap: capping one arviz sub-package would freeze the whole stack, because `arviz` requires them to share a minor version.
- **Shown in the notebooks**, next to the existing histogram in `bg_nbd`, `mbg_nbd` and `pareto_nbd`.

## Behaviour worth knowing

- `plot_type` is the **last** named parameter, so existing positional calls keep their meaning.
- On a point-estimate fit the reference CDF is estimated from a fixed 100 predictive draws per customer (`_REFERENCE_DRAWS`). This matters more than it sounds: `ParetoNBDModel.fit` defaults to `method="map"`, so that is the default path for that model. The band assumes the reference CDF is *known*, and with one draw per customer it came out roughly √2 too narrow — simulated coverage of the nominal 95% band was 0.28–0.41, so a well-fitting model failed the check most of the time. Coverage saturates by a ratio of about 50, and cost is linear in the draws (at 20k customers a ratio of 1000 costs 646 MB and 171 s against 38 MB and 9 s at 50), hence a bounded constant rather than `samples`. `samples` keeps its original meaning: prior predictive checks only.
- Raises `ValueError` for an unknown `plot_type`, for an `ax` that is not the right shape for the chosen plot, and for observed data with a single distinct purchase count.
- Warns when the predictive samples behind the reference CDF fall below 20 per observation, and when `max_purchases` is passed alongside `plot_type="ecdf"`, where it is ignored.
- The `ecdf` branch forwards `**kwargs` to the band, so `confidence_level` and `num_trials` are reachable.

## Test plan

- [x] `uv run pytest tests/clv` — 352 passed, 16 skipped
- [x] `uv run pytest tests/clv/test_plotting.py` — 43 passed (26 pre-existing + the ECDF option, the `ax` and `plot_type` rejections, the degenerate-data guard, the two warnings, and the point-estimate path)
- [x] **Mutation-checked the new assertions.** Plotting the predictive CDF instead of the observed one, shifting the band one step off the grid, reverting the grid to a 100-point linspace, dropping the `**kwargs` forwarding, and reverting the reference draws each fail the suite. The earlier versions of these assertions passed four of the five, which is why they were rewritten.
- [x] `uv run pytest tests/clv/test_plotting.py` on the arviz **1.3** stack (`--with arviz==1.3.0 --with arviz-base==1.3.0 --with arviz-plots==1.3.0 --with arviz-stats==1.3.0`) — 43 passed, and the band values are bit-identical to 1.2.0 for a fixed seed
- [x] Simulated coverage of the band, negative-binomial counts, n = 2357, nominal 0.95: 0.28–0.41 at a predictive-to-observed ratio of 1, 0.83 at 10, 0.89 at 20, and indistinguishable from nominal from 50 up. An exact reference CDF gives 0.93, so the discreteness of the counts costs little on its own.
- [x] `uv run python scripts/run_notebooks/runner.py --notebooks <the three CLV notebooks>` — all pass
- [x] All three notebooks re-executed with real sampling; only the new cells' outputs are committed, so the rest of each notebook is byte-identical to `main`
- [x] `uv run pre-commit run --files <changed files>` — all hooks pass
- [x] `uv lock --check` — clean

----
📚 Documentation preview 📚: https://pymc-marketing--2905.org.readthedocs.build/en/2905/


juanitorduz:
Is this something we could get from ArviZ without doing it by hand 🤔 ?

shivamlalakiya:
Good question, I went and checked, and yes for the part that matters. **Done in 61cd72a** — the band math now comes from ArviZ. Only the finished figure stays local, because nothing in ArviZ 1.x draws it.

### What ArviZ 1.x does ship

`arviz-stats` 1.2.0 (previously pulled in transitively by our `arviz>=1.2.0,<2.0` pin) has `arviz_stats.ecdf_utils`, which contains both primitives I had hand-rolled:

| was hand-rolled | now used instead |
|---|---|
| `_ecdf` | `arviz_stats.ecdf_utils.compute_ecdf` |
| `_ecdf_confidence_band` | `simulate_confidence_bands` + `get_pointwise_confidence_band` |

They are not merely similar, they are the same algorithm. Same seed, same eval points, the bands come out **bit-for-bit identical** (`np.array_equal`, not just `allclose`):

```
max |lower diff| = 0.0
max |upper diff| = 0.0
```

(ArviZ parameterizes gamma as `1 - 2*min(...)` and takes the `prob` quantile where I took `2*min(...)` at the `1 - prob` quantile, and it uses `bdtr`/`bdtrik` instead of `scipy.stats.binom.cdf`/`.ppf`, but those are the same quantities.) So the swap was a pure deletion: `_ecdf`, `_ecdf_confidence_band` and the `scipy.stats.binom` import are gone, net -38/+20 lines.

One thing worth your call: `ecdf_utils` is a submodule, not part of the documented top-level `arviz_stats` API, so we are leaning on a semi-private surface. We already import `arviz_plots` and `arviz_base` directly in `pymc_marketing/mmm/plotting/*`, so importing an arviz sub-package isn't unprecedented, and I've added an explicit `arviz-stats>=1.2.0,<2.0` pin rather than letting it ride the `arviz` meta-package. If you'd rather not depend on an undocumented submodule at all, say so and I'll restore the frozen local copy with a comment explaining why.

### What ArviZ doesn't give us

None of the 1.x plotting functions produce the two-panel figure from #1226:

- `azp.plot_ecdf_pit` plots a Δ-ECDF of **PIT values** and expects them already transformed, in a DataTree under an SBC / `loo_pit` group. Our input is raw discrete purchase counts (PIT for discrete observations needs randomization), and it renders the difference panel only, not the ECDF panel.
- `azp.plot_ppc_dist(kind="ecdf")` overlays `num_samples` individual predictive ECDFs instead of one simultaneous confidence band, so it's a visibly different chart from the reference image in the issue.
- `azp.plot_ppc_rootogram` / `plot_ppc_pava` are ArviZ's recommended discrete-data PPC plots, but again, a different chart than the one requested.

So: statistics in ArviZ, two-panel assembly here. If you'd rather we not carry a bespoke ECDF plot at all and instead point CLV users at `plot_ppc_rootogram`, that's a reasonable outcome too, just tell me and I'll close this.

shivamlalakiya:
Pushed in 61cd72a (and updated my comment above, since it now describes something that's already done rather than proposed).

- `_ecdf` and `_ecdf_confidence_band` deleted, along with the `scipy.stats.binom` import. The band now comes from `arviz_stats.ecdf_utils` (`compute_ecdf`, `simulate_confidence_bands`, `get_pointwise_confidence_band`). Bit-for-bit identical output to what it replaced, so no behaviour change.
- Added an explicit `arviz-stats>=1.2.0,<2.0` pin, since we now import it directly instead of relying on the `arviz` meta-package. `uv lock --check` is clean.
- `tests/clv/test_plotting.py`: 32 passed. Pre-commit green on all four touched files. No test changes were needed, the tests only ever exercised the public `plot_expected_purchases_ppc`.

One separate question before this closes #1226. The ECDF plots in the issue screenshot are no longer in `docs/source/notebooks/clv/pareto_nbd.ipynb` — as far as I can tell they were dropped during the arviz 1.x migration when `az.plot_ecdf` disappeared, and there are now zero `ecdf` references in any CLV notebook source. Do you want the notebook rewired to call `plot_expected_purchases_ppc(..., plot_type="ecdf")` as part of this PR, so the thing the issue points at actually exists again? Happy either way, it's a notebook re-run so I'd rather ask than surprise you with the output diff.

pyproject.toml
```diff
@@ -26,6 +26,7 @@ classifiers = [
 
 dependencies = [
     "arviz-plots[matplotlib]>=1.2.0,<2.0",
+    "arviz-stats>=1.2.0,<2.0",
```
juanitorduz, line 31: why do we need this?

juanitorduz (commented):
Can nyou also show these plots in the relevant notebooks please?

pymc_marketing/clv/plotting.py
```diff
@@ -19,6 +19,11 @@
 import numpy as np
 import pandas as pd
 import pymc as pm
+from arviz_stats.ecdf_utils import (
+    compute_ecdf,
+    get_pointwise_confidence_band,
+    simulate_confidence_bands,
+)
```
daimon-pymclabs, line 27: `arviz_stats.ecdf_utils` is not part of arviz-stats' public API. It is not re-exported from `arviz_stats/__init__.py` (which exposes `ecdf` via `arviz_stats.visualization`) and the module has no `__all__`. With the pin at `>=1.2.0,<2.0`, any 1.x minor can rename or relocate these three functions and break `import pymc_marketing.clv.plotting` at import time, not just at call time.

Two options:
1. Use the public path instead. `arviz_plots.plot_ppc_pit` already exists and does predictive-vs-observed PIT ECDF with the same simultaneous band, and `arviz-plots[matplotlib]>=1.2.0` is *already* a dependency here, so this would also make the new `arviz-stats` pin unnecessary.
2. If you keep the private import, tighten the pin and add an explicit import smoke test so a dependency bump fails loudly in CI rather than at user import time.

Also worth a look: `arviz_plots.plot_ppc_rootogram` is the conventional PPC display for count data and may be a better fit than an ECDF for purchase frequencies.

daimon-pymclabs (commented):
Reviewed the code, the dependency change, and the arviz-stats API this builds on. The core approach is right: delegating the band to `arviz_stats.ecdf_utils` rather than reimplementing Säilynoja et al. is the correct call, and I verified the call convention matches how arviz uses these functions internally (`eval_points` is the reference CDF evaluated on the grid, passed to both `simulate_confidence_bands` and `get_pointwise_confidence_band`). The `hist` default is genuinely unchanged and the notebooks do demonstrate the new option.

Main things I would like resolved before merge:

1. **The import is from a non-public module.** `arviz_stats.ecdf_utils` is not exported from `arviz_stats/__init__.py`, so the `>=1.2.0,<2.0` pin lets a minor release break `import pymc_marketing.clv.plotting`. `arviz_plots.plot_ppc_pit` covers this use case through a public API, and `arviz-plots[matplotlib]` is already a dependency, which would also make the new `arviz-stats` pin unnecessary.
2. **The band assumes a continuous reference CDF, but purchase frequency is integer-valued.** arviz-stats corrects for this itself in `difference_ecdf_pit` via a randomized PIT for integer dtypes; this code does not. My quick simulation suggests the effect here is conservative rather than anti-conservative, so it costs power rather than producing false alarms, but it should be either corrected or documented.
3. **The `95% HDI` legend label is wrong.** This is a simultaneous confidence band, not a highest-density interval, and the label is now baked into three notebooks.
4. **The new test does not exercise the path it claims to.** `(plt.subplot(), plt.subplot())` is the same Axes twice.

The rest are smaller: no validation on user-supplied `ax`, percent-labelled axes carrying proportions, a one-point misalignment between the curve and the band, and no way for an ECDF caller to reach `confidence_level`. Details inline.

One process note rather than a code comment: the three notebooks account for ~2,200 of the 2,332 added lines, almost all re-executed output. That is expected for this repo, but it does make the substantive 90-line change hard to find. Not a blocker.

pymc_marketing/clv/plotting.py
```diff
@@ -558,6 +584,62 @@ def plot_expected_purchases_ppc(
...
+    confidence_level: float = 0.95,
+) -> tuple[plt.Axes, plt.Axes]:
+    """Plot an ECDF plot and companion difference plot against a 95% confidence band."""
+    if ax is None:
+        _, (ax_ecdf, ax_diff) = plt.subplots(2, 1)
+    else:
+        ax_ecdf, ax_diff = ax
+
+    rng = np.random.default_rng(random_seed)
+    x = np.linspace(observed.min(), observed.max(), npoints)
```
daimon-pymclabs: Two things on the eval grid.

**Discrete support.** `frequency` is integer-valued, so a 100-point `linspace` over `[min, max]` mostly evaluates the same step repeatedly (for a typical CDNOW-style dataset there are ~30 distinct counts), and the plotted step edges land on non-integer x positions. `np.arange(observed.min(), observed.max() + 1)` would give one eval point per attainable count, cheaper and with risers on the lattice.

**Degenerate case.** If every customer has the same frequency, `observed.min() == observed.max()` and `x` becomes 100 identical points. `compute_ecdf` still returns something, but the plot is meaningless and the band simulation runs on a constant `eval_points` vector. Worth an explicit guard.

Separately, and more substantively: the Säilynoja et al. band assumes the reference CDF is continuous, so that PIT values are Uniform(0, 1). That does not hold for counts. arviz-stats handles this itself in `difference_ecdf_pit`, which detects integer dtype and applies a randomized PIT (`urvs * vals_less + (1 - urvs) * vals`) before calling `ecdf_pit`. This code skips that correction. I ran a quick coverage simulation (negative-binomial counts vs a gamma control, 60 replicates) and the discrete band came out conservative rather than anti-conservative, so the practical consequence looks like lost power rather than false alarms. Still worth either applying the randomization or documenting the caveat, since a conservative band makes the check less likely to flag real misfit.

pymc_marketing/clv/plotting.py
```diff
@@ -558,6 +584,62 @@ def plot_expected_purchases_ppc(
...
+    ax: plt.Axes | Sequence[plt.Axes] | None = None,
+    npoints: int = 100,
+    num_trials: int = 500,
+    confidence_level: float = 0.95,
+) -> tuple[plt.Axes, plt.Axes]:
+    """Plot an ECDF plot and companion difference plot against a 95% confidence band."""
+    if ax is None:
+        _, (ax_ecdf, ax_diff) = plt.subplots(2, 1)
+    else:
+        ax_ecdf, ax_diff = ax
```
daimon-pymclabs, line 669: No validation on user-supplied `ax`. The public docstring invites callers to pass axes, but:

- `plot_expected_purchases_ppc(m, plot_type="ecdf", ax=some_single_axes)` fails here with `TypeError: cannot unpack non-iterable Axes object`.
- A sequence of the wrong length gives a bare `ValueError: too many values to unpack`.
- The mirror case, `plot_type="hist"` with a 2-tuple, falls through to `pandas.DataFrame.plot(ax=(ax1, ax2))` and fails somewhere inside pandas.

All three are cryptic relative to the clear `NameError` you raise for a bad `plot_type`. Suggest validating up front, e.g. that `ax` is a sequence of exactly two Axes when `plot_type="ecdf"` and a single Axes when `plot_type="hist"`.

pymc_marketing/clv/plotting.py
```diff
@@ -558,6 +584,62 @@ def plot_expected_purchases_ppc(
...
+        n_draws=n,
+        n_chains=1,
+        eval_points=z,
+        prob=confidence_level,
+        n_simulations=num_trials,
+        rng=rng,
+    )
+    lower, upper = get_pointwise_confidence_band(gamma, n, z)
+
+    band_label = f"{int(confidence_level * 100)}% HDI"
```
daimon-pymclabs: `"95% HDI"` is the wrong name for this object. What `simulate_confidence_bands` + `get_pointwise_confidence_band` return is a simultaneous confidence band for the ECDF under the reference distribution. It is not a highest-density interval, and it is not derived from the posterior at all. Calling it an HDI in the legend will mislead readers, and the label is now baked into the three notebook outputs in this PR. Suggest `"95% simultaneous band"` or similar.

pymc_marketing/clv/plotting.py
```diff
@@ -558,6 +584,62 @@ def plot_expected_purchases_ppc(
...
+    observed_ecdf = compute_ecdf(observed, x)
+
+    ecdf_x = np.insert(x, 0, x[0])
+    ecdf_y = np.insert(observed_ecdf, 0, 0)
+    ax_ecdf.step(ecdf_x, ecdf_y, where="post", label="Observed")
+    ax_ecdf.fill_between(x, lower, upper, step="post", alpha=0.2, label=band_label)
+    ax_ecdf.set_title(f"{title_prefix} ECDF Plot")
+    ax_ecdf.set_xlabel("Purchases per Customer")
+    ax_ecdf.set_ylabel("Percent of Total")
```
daimon-pymclabs: The y-values here are proportions on [0, 1], not percentages, so `"Percent of Total"` (and `"Percent Deviation from Expectation"` on the difference panel below) mislabels the axis. Either rename to "Proportion of customers" / "Deviation from expected proportion", or apply `matplotlib.ticker.PercentFormatter(1.0)` so the ticks match the label.

pymc_marketing/clv/plotting.py
```diff
@@ -558,6 +584,62 @@ def plot_expected_purchases_ppc(
...
+        n_simulations=num_trials,
+        rng=rng,
+    )
+    lower, upper = get_pointwise_confidence_band(gamma, n, z)
+
+    band_label = f"{int(confidence_level * 100)}% HDI"
+    observed_ecdf = compute_ecdf(observed, x)
+
+    ecdf_x = np.insert(x, 0, x[0])
+    ecdf_y = np.insert(observed_ecdf, 0, 0)
```
daimon-pymclabs: The ECDF line is drawn on `ecdf_x` (101 points, with a duplicated leading x to draw the riser from 0) while the band on the next line is drawn on `x` (100 points). The band therefore does not extend to the start of the curve. The difference panel below also gets no equivalent leading-riser treatment, so the two panels are drawn inconsistently. Prepending the same point to `lower`/`upper` (and to both series on the difference panel) would make them line up.

pymc_marketing/clv/plotting.py
```diff
@@ -558,6 +584,62 @@ def plot_expected_purchases_ppc(
...
 
+def _plot_expected_purchases_ecdf(
+    observed: np.ndarray,
+    ppc: np.ndarray,
+    title_prefix: str,
+    random_seed: int,
+    ax: plt.Axes | Sequence[plt.Axes] | None = None,
+    npoints: int = 100,
+    num_trials: int = 500,
+    confidence_level: float = 0.95,
```
daimon-pymclabs, line 637: `npoints`, `num_trials` and `confidence_level` are only reachable on this private helper. The public `plot_expected_purchases_ppc` documents `**kwargs` as hist-only, so an ECDF caller cannot change the confidence level, the grid resolution, or the number of simulation trials, even though the legend advertises a specific level. Consider forwarding `**kwargs` to this helper on the `ecdf` branch.

Related: on the `ecdf` branch, `max_purchases` and any `**kwargs` the caller passes are silently discarded. It is documented, but silent is easy to miss. A `warnings.warn` when they are non-default would be friendlier than a doc note.

pymc_marketing/clv/plotting.py
```diff
@@ -488,22 +494,31 @@ def plot_expected_purchases_ppc(
...
+        ``plot_type`` is 'hist'.
 
     Returns
     -------
-    axes : matplotlib.AxesSubplot
+    axes : matplotlib.AxesSubplot, or tuple of two matplotlib.AxesSubplot when ``plot_type`` is 'ecdf'
     """
-    if ax is None:
+    if plot_type not in ("hist", "ecdf"):
+        raise NameError("Specify 'hist' or 'ecdf' for 'plot_type' parameter.")
```
daimon-pymclabs: Nit: `NameError` is the wrong builtin for an invalid argument value; `ValueError` is the conventional one. I see this matches the existing `ppc` check two blocks down, so consistency is a fair argument for leaving it. Flagging in case you would rather introduce the correct type here and deprecate the other one separately.

tests/clv/test_plotting.py
```diff
@@ -208,3 +213,23 @@ def test_plot_expected_purchases_ppc(
...
+@pytest.mark.parametrize(
+    "ppc, subplot",
+    [("prior", None), ("posterior", (plt.subplot(), plt.subplot()))],
+)
+def test_plot_expected_purchases_ppc_ecdf(fitted_model, ppc, subplot):
+    ax_ecdf, ax_diff = plot_expected_purchases_ppc(
+        model=fitted_model,
+        ppc=ppc,
+        plot_type="ecdf",
+        samples=100,
```
daimon-pymclabs, line 271: This does not test what it looks like it tests. `plt.subplot()` called twice with identical arguments returns the *same* Axes instance (confirmed on matplotlib 3.11: `plt.subplot() is plt.subplot()` is `True`). So `subplot` is `(A, A)`, both panels get drawn onto one Axes, and the two `isinstance` assertions pass anyway. The user-supplied-axes path is effectively untested.

Use `plt.subplots(2, 1)[1]`, and build it inside the test rather than in the `parametrize` decorator, since decorator arguments are evaluated at collection time and leak a stray global figure into every other test in the module. (The existing `test_plot_expected_purchases_ppc` above has the same collection-time issue, though not the aliasing bug.)

While here, the assertions are quite thin. Worth adding at least one check on the actual output, e.g. that the ECDF panel is monotone non-decreasing and bounded in [0, 1], that the band satisfies `lower <= upper` everywhere, and that `plot_type="hist"` still returns a single Axes so the default-unchanged claim is pinned by a test.

pyproject.toml
shivamlalakiya: `pymc_marketing/clv/plotting.py` imports `arviz_stats` directly now (`compute_ecdf`, `simulate_confidence_bands`, `get_pointwise_confidence_band` for the ECDF confidence band), so it is a direct dependency rather than something we inherit from the `arviz` meta-package. Declaring it keeps a future `arviz` repackaging from silently dropping it out of the environment.

In 066bde2 I tightened it to a single minor, `>=1.2.0,<1.3.0`, because those functions live in `arviz_stats.ecdf_utils`, which is not part of the documented public API. With the pin narrowed, a 1.3 release that moves them fails at lock time instead of at user import time.

If you would rather not carry the pin at all, the alternative is dropping back to a frozen local copy of the band math, which is what this replaced. I think the pin is the better trade, but it is your call.

pymc_marketing/clv/plotting.py
shivamlalakiya: Agreed that it is a private surface. I looked for a public path and did not find one that produces this plot:

- `arviz_stats` 1.2.0 exposes `ecdf` (via `arviz_stats.visualization`), but it only computes the ECDF points. There is no public entry point for the simultaneous band; `ecdf_pit` and the two band helpers all live in `ecdf_utils`.
- `arviz_plots.plot_ppc_pit` does compute the band, but it plots the difference ECDF of PIT values on a [0, 1] axis, with no ECDF panel, and returns a `PlotCollection` rather than Axes. It is a different chart from the one in #1226 and would not fit this function's `-> plt.Axes` contract.
- `plot_ppc_rootogram` is a good discrete PPC display, but again a different chart. Worth adding to the CLV notebooks on its own merits; I would rather not fold it into this PR.

So I took your option 2 in 066bde2: the pin is now `arviz-stats>=1.2.0,<1.3.0` with a comment at the pin explaining why it is narrower than the rest of the arviz pins.

On the smoke test: `tests/clv/test_plotting.py` imports `pymc_marketing.clv.plotting` at module scope, so a rename in `ecdf_utils` fails at collection time in CI, before any test body runs. I did not add a second test that only asserts the import, since it would fail in exactly the same place.

pymc_marketing/clv/plotting.py
shivamlalakiya: All three addressed in 066bde2.

**Discrete support.** The grid is now `np.arange(observed.min(), observed.max() + 1)`, so there is one eval point per attainable count and the risers land on the lattice. `npoints` is gone, since the data now determines the resolution.

**Degenerate case.** `_plot_expected_purchases_ecdf` raises `ValueError("An ECDF plot requires more than one distinct observed purchase count.")` when `observed.min() == observed.max()`, covered by a test.

**Randomized PIT.** Documented rather than applied. Applying the randomization turns this into a PIT plot on a [0, 1] axis, which is `plot_ppc_pit`'s chart, not the one #1226 asks for. Since the effect is conservative, the caveat is now in the public docstring under Notes ("less likely to flag a real misfit, rather than more likely to raise a false alarm") and repeated on the helper. Thanks for running the coverage simulation; that is what made the direction of the error decidable.

pymc_marketing/clv/plotting.py
shivamlalakiya: Fixed in 066bde2. Validation happens up front, before any sampling, so it fails fast:

- `plot_type="ecdf"` with a single Axes or a wrong-length sequence: `ValueError("Specify a sequence of two matplotlib Axes for 'ax' when 'plot_type' is 'ecdf'.")`
- `plot_type="hist"` with a sequence: `ValueError("Specify a single matplotlib Axes for 'ax' when 'plot_type' is 'hist'.")`

Both paths have tests.

pymc_marketing/clv/plotting.py
shivamlalakiya: You are right, it is a simultaneous confidence band under the reference distribution, with no posterior involved. Renamed to `"95% simultaneous band"` in 066bde2, and the three notebook figures were re-executed so the images match.

pymc_marketing/clv/plotting.py
shivamlalakiya: Renamed in 066bde2: `"Proportion of Customers"` on the ECDF panel and `"Deviation from Expected Proportion"` on the difference panel. I went with renaming rather than `PercentFormatter(1.0)` since the deviations are small enough (order 0.01) that percent ticks would round to a wall of zeros.

pymc_marketing/clv/plotting.py
shivamlalakiya: Fixed in 066bde2 by dropping the synthetic leading riser rather than extending it. Both panels are now drawn on the same `x`, so the curve and the band line up and the two panels are consistent. The ECDF simply starts at `F(min)` instead of rising from a duplicated first point. A test asserts the two panels share the same x data.

pymc_marketing/clv/plotting.py
shivamlalakiya: `**kwargs` is now forwarded to the helper on the `ecdf` branch in 066bde2, so `num_trials` and `confidence_level` are reachable from the public function. `npoints` no longer exists; the grid is determined by the observed counts. The docstring lists what the `ecdf` branch accepts.

`max_purchases` now raises a `UserWarning` when it is non-default and `plot_type="ecdf"`, rather than being silently dropped.

pymc_marketing/clv/plotting.py
shivamlalakiya: Changed the `plot_type` check to `ValueError` in 066bde2. `plot_type` is new in this PR, so there is no caller to break, and it seemed better to introduce the right type than to propagate the wrong one for symmetry.

I left the `ppc` check as `NameError`. Changing it would break any existing `except NameError`, so it belongs in its own PR with a note in the release notes. Happy to open that separately if you want it.

tests/clv/test_plotting.py
shivamlalakiya: Good catch on the aliasing, `plt.subplot()` twice really is one Axes. Fixed in 066bde2:

- The axes are built inside the test with `plt.subplots(2, 1)[1]`, not in the decorator, so no figure leaks at collection time. Same fix applied to `test_plot_expected_purchases_ppc` above, which had the collection-time issue you noted.
- The test now asserts the returned axes are the ones passed in, so the user-supplied path is actually pinned.
- Added assertions on the output: the ECDF is monotone non-decreasing and bounded in [0, 1], the band polygon stays within [0, 1], and both panels are drawn on the same x grid.
- `test_plot_expected_purchases_ppc` now asserts the default `plot_type="hist"` returns a single Axes.
- Also switched the teardown from `plt.clf()` to `plt.close("all")`, since `clf()` leaves the figure open.

I did not add a `lower <= upper` assertion. That ordering comes from `get_pointwise_confidence_band`'s two quantiles inside arviz-stats, not from anything this PR does, so the assertion would be testing their code.

shivamlalakiya:
@juanitorduz the notebooks were done in 289ca79 (bg_nbd, mbg_nbd, pareto_nbd now show the `plot_type="ecdf"` figure next to the existing histogram). @daimon-pymclabs everything from your review is addressed in 066bde2, with a reply on each thread. Summary:

| Finding | Resolution |
|---|---|
| Private `arviz_stats.ecdf_utils` import | Pin narrowed to `>=1.2.0,<1.3.0` with a comment. No public path computes this band (details on the thread). |
| Continuous-CDF band on integer counts | Documented in the public docstring Notes. Applying the randomized PIT would turn this into `plot_ppc_pit`'s chart. |
| Discrete eval grid | Now `np.arange(min, max + 1)`; `npoints` removed |
| Degenerate all-equal counts | Explicit `ValueError` |
| `95% HDI` label | `95% simultaneous band` |
| Percent-labelled proportion axes | Renamed to "Proportion of Customers" / "Deviation from Expected Proportion" |
| Curve/band one-point misalignment | Leading riser dropped, both panels on the same grid |
| No `ax` validation | Validated up front, before sampling, on both branches |
| `confidence_level` unreachable | `**kwargs` forwarded on the `ecdf` branch; `max_purchases` now warns |
| `NameError` for `plot_type` | `ValueError`. Left the pre-existing `ppc` check alone. |
| Test aliased one Axes into both panels | `plt.subplots(2, 1)[1]`, built inside the test, plus assertions on the plotted output |

On the notebook-diff volume: point taken, and this round is only 2 changed lines per notebook. The three notebooks were re-executed end to end with real sampling, but only the ECDF figure outputs were written back, so the rest of each notebook is byte-identical.

Verification:
- `uv run pytest tests/clv` — 350 passed, 16 skipped
- `uv run pytest tests/clv/test_plotting.py` — 41 passed (was 32)
- `uv run pre-commit run --files <the six changed files>` — all hooks pass
- `uv lock --check` — clean
- All three notebooks re-executed with `nbclient`, exit code 0, no error outputs

daimon-pymclabs:
Thanks, this round addresses the review. I re-checked each item against `066bde2`: the pin comment, the `ValueError` for `plot_type`, the `95% simultaneous band` label, the integer eval grid, the axis labels, the shared grid across both panels, the up-front `ax` validation, the `**kwargs` forwarding, and the rebuilt test with real assertions on the plotted output all look right. Agreed on the `arviz_stats.ecdf_utils` tradeoff given there is no public path to this band; the narrowed pin plus the comment is a reasonable place to land.

One item I want to revisit, because my original framing of it was wrong and the new docstring now encodes that framing.

**The band is anti-conservative when the reference CDF comes from a predictive sample the same size as the data, which is what a MAP fit produces here.**

`_plot_expected_purchases_ecdf` builds the reference from `z = compute_ecdf(ppc, x)`, and `simulate_confidence_bands(n_draws=n, ...)` assumes `z` is a *known* CDF. That holds for an MCMC fit, where `ppc_freq.values.ravel()` is `chain × draw × customer` and `z` is effectively exact. It does not hold for a MAP fit: `sample_posterior_predictive` takes the `dataset.sizes["draw"] == 1` branch, expands to `n_samples`, and the plotting call hardcodes `n_samples=1`, so `ppc` has exactly `n` values. Then both curves carry sampling error of the same order, the band is roughly a factor of √2 too narrow, and a correctly specified model fails it most of the time.

Simultaneous coverage over 300 replicates, nominal 0.95, `n = 2357` (the notebook size), reference and data drawn from the same distribution:

| reference CDF | Poisson(1) | NegBinom(2, 0.3) |
|---|---|---|
| exact | 0.927 | 0.947 |
| `ppc` size = 200·n (MCMC path) | 0.940 | 0.947 |
| `ppc` size = n (MAP path) | 0.517 | 0.323 |

```python
import numpy as np
from scipy import stats
from arviz_stats.ecdf_utils import (
    compute_ecdf, simulate_confidence_bands, get_pointwise_confidence_band,
)

rng = np.random.default_rng(1)

def coverage(dist, n, ppc_size, reps=300):
    hits = 0
    for _ in range(reps):
        obs = dist.rvs(size=n, random_state=rng)
        x = np.arange(obs.min(), obs.max() + 1)
        z = dist.cdf(x) if ppc_size is None else compute_ecdf(
            dist.rvs(size=ppc_size, random_state=rng), x
        )
        gamma = simulate_confidence_bands(
            n_draws=n, n_chains=1, eval_points=z, prob=0.95,
            n_simulations=500, rng=rng,
        )
        lower, upper = get_pointwise_confidence_band(gamma, n, z)
        ecdf = compute_ecdf(obs, x)
        hits += bool(np.all((ecdf >= lower) & (ecdf <= upper)))
    return hits / reps

n, d = 2357, stats.nbinom(2, 0.3)
print(coverage(d, n, None), coverage(d, n, 200 * n), coverage(d, n, n))
```

Two consequences for the current diff:

1. The failure mode is silent. `model.fit(fit_method="map")` is documented in the `BetaGeoModel`, `ModifiedBetaGeoModel` and `BetaGeoBetaBinomModel` docstrings, so a user can reach this by following the docs, and the plot will show the observed ECDF outside the band on a model that fits fine. The fix is small: either draw many predictive samples for the reference (pass the function's `samples` through instead of the hardcoded `n_samples=1`), or raise when `len(ppc)` is not comfortably larger than `len(observed)`. I would prefer the first, with the second as a guard.

2. The new Notes paragraph says the band is conservative and points at the continuous-CDF assumption as the reason. Row 1 of the table is my earlier point measured in isolation: with an exact reference, discreteness costs 0.927 to 0.947 against a nominal 0.95, so it is close to nominal and marginally *liberal*, not conservative. My original comment asserted a direction I had only spot-checked, which was my error, and the docstring inherited it. Once the reference-sample issue is fixed I would either drop that sentence or reword it to state the assumption without claiming a direction.

The existing tests would not catch this. `fitted_bg` and friends set a fake fit with `draw=50`, so they never take the `draw == 1` branch. A test that fits with `fit_method="map"` and asserts the observed ECDF lies inside the band would cover it.

Nothing else outstanding from my side.

shivamlalakiya:
Reproduced both claims, and both are right. Thanks, this one mattered.

**The MAP path.** Confirmed at the source: `distribution_new_customer` only expands the draw dimension when `dataset.sizes["chain"] == 1 and dataset.sizes["draw"] == 1`, so the hardcoded `n_samples=1` was doing nothing for a sampled posterior and everything for a point estimate. On `data/clv_quickstart.csv` (n = 2357) with a `method="map"` fit:

```
idata dims {'chain': 1, 'draw': 1}
n_samples=1:    ppc dims {'chain': 1, 'draw': 1,    'customer_id': 2357} ->    2357 values, n=2357
n_samples=1000: ppc dims {'chain': 1, 'draw': 1000, 'customer_id': 2357} -> 2357000 values, n=2357
```

**Coverage.** Your script, my seed, `nbinom(2, 0.3)`, n = 2357, nominal 0.95:

| reference | coverage |
|---|---|
| exact CDF | 0.927 (0.933 for Poisson(1)) |
| ratio 1 (the MAP path) | 0.28 |
| ratio 10 | 0.83 |
| ratio 20 | 0.89 |
| ratio 50 | 0.91 |
| ratio 1000 | 0.92 |

**Fix.** The posterior branch now passes `n_samples=samples` (default 1000) instead of 1, so a point-estimate fit gets a reference CDF built from `samples * n_customers` draws. The sampled-posterior path is untouched: `n_samples` is only read inside that expansion branch, and I checked the plotted ECDF is bit-identical (`np.array_equal`) for `samples=1` vs `samples=1000` on a fixture with `draw=50`. The `samples` docstring now says when it applies.

**Guard.** Implemented at ratio < 20, but as a `UserWarning` rather than a raise:

> The confidence band treats the predictive CDF as known, which needs at least 20 predictive samples per observation; got 1. The band is too narrow, so a well-fitting model can fall outside it. Increase 'samples', or fit the model with more draws.

I went with warning over raising for two reasons. A short fit is a legitimate thing to want to look at, and the plot is still readable once you know the band is optimistic. And our mocked-notebook CI substitutes `pymc.testing.mock_sample`, whose default is 10 draws on a single chain, so a hard raise would turn every ECDF cell in the CLV notebooks into a notebook-CI failure for a reason that has nothing to do with the user's model. Happy to make it a raise if you would rather have the hard stop.

The threshold comment in the source carries the coverage numbers above, so the 20 is not a bare magic number.

**Notes wording.** Dropped the direction claim. It now reads "assumes a continuous reference CDF, which does not hold for integer purchase counts, so its coverage is only approximately the nominal level", with the known-reference assumption documented separately on the helper. Your 0.927 against nominal 0.95 matches what I measured, so "conservative" was wrong in both of our accounts of it.

**Test.** Added `map_fitted_bg` to `tests/clv/conftest.py`, a model whose idata has the single `(chain, draw)` shape a MAP fit produces, and `test_plot_expected_purchases_ppc_ecdf_map_fit` asserts no small-sample warning fires on that path. It fails on the pre-fix code with "got 1" and passes after, so it pins the regression rather than just exercising the line. `test_plot_expected_purchases_ecdf_warns_on_small_ppc` covers the guard directly.

**Notebooks.** No re-run needed. bg_nbd and mbg_nbd fit with 4 chains x 2000 draws and pareto_nbd with the MCMC default, so all three are on the sampled-posterior path where `n_samples` is ignored, and the committed figures are unchanged.

Fixed in c0b38bb. `tests/clv`: 352 passed, 16 skipped. Pre-commit clean.

pyproject.toml
```diff
@@ -26,6 +26,9 @@ classifiers = [
 
 dependencies = [
     "arviz-plots[matplotlib]>=1.2.0,<2.0",
+    # clv.plotting imports arviz_stats.ecdf_utils, which is not part of the public API,
+    # so this is pinned to a single minor rather than the whole 1.x range.
+    "arviz-stats>=1.2.0,<1.3.0",
```
juanitorduz: Do we really need this? can we test thet this works in https://github.com/arviz-devs/arviz-stats/releases 1.3.0 ?

juanitorduz (commented):
samll question

pyproject.toml
shivamlalakiya: Cap removed, it is now back to `arviz-stats>=1.2.0,<2.0`. You were right to push on it, and for a stronger reason than I had: capping one arviz package on its own would have silently frozen the whole stack. `arviz` enforces that its sub-packages share a minor version, so installing 1.3 of just this one fails at import:

```
ImportError: Incompatible ArviZ packages versions detected:
- arviz_base: 1.2
- arviz_plots: 1.2
- arviz_stats: 1.3
All ArviZ packages must share the same minor version.
```

With `arviz-stats<1.3.0` in place, a future `arviz` 1.3 would have been held back to 1.2 by the resolver rather than failing loudly. That is worse than the problem the cap was meant to guard against.

Tested against 1.3.0, three ways:

1. The three functions are still there, with identical signatures:
   ```
   arviz_stats 1.3.0
   compute_ecdf (sample, eval_points)
   get_pointwise_confidence_band (prob, ndraws, eval_points)
   simulate_confidence_bands (n_draws, n_chains, eval_points, prob, n_simulations, rng)
   ```
2. The band is unchanged. Same seed and inputs under 1.2.0 and 1.3.0, comparing with `==` rather than a tolerance:
   ```
   versions: 1.2.0 vs 1.3.0
     ecdf:  identical = True
     z:     identical = True
     lower: identical = True
     upper: identical = True
     gamma: 0.9905074245840846 vs 0.9905074245840846, identical = True
   ```
3. `uv run --with 'arviz==1.3.0' --with 'arviz-base==1.3.0' --with 'arviz-plots==1.3.0' --with 'arviz-stats==1.3.0' pytest tests/clv/test_plotting.py` — 43 passed.

The dependency line itself stays, since `pymc_marketing/clv/plotting.py` imports `arviz_stats` directly and should say so rather than inherit it from the `arviz` meta-package. The lockfile still resolves to 1.2.0, matching the rest of the arviz pins, and `uv lock --check` is clean. A stack-wide bump to 1.3 looks safe from the CLV side whenever you want to do it, but that belongs in its own PR.

shivamlalakiya:
Went back over the whole diff change by change before asking for another look. Three things came out of it, pushed as separate commits.

**adf76ad — the `n_samples` fix was too broad.** Passing `samples` through applied to the histogram as well, which does not need a precise reference CDF. For a point-estimate fit the predictive array it allocates is `n_samples * n_customers * 2`:

| customers | `n_samples=1` | `n_samples=1000` |
|---|---|---|
| 2,357 | 0.04 MB | 38 MB |
| 100,000 | 1.6 MB | 1,600 MB |
| 1,000,000 | 16 MB | 16,000 MB |

A 1.6 GB allocation on `plot_expected_purchases_ppc(model)`, the default call, is not a trade I want to make for a plot that was fine with the single draw. It is now `n_samples=samples if plot_type == "ecdf" else 1`, so only the band pays. The histogram path is byte-identical to what it did before this PR.

Same commit finishes the `ax` validation @daimon-pymclabs asked for. A length-two sequence whose elements are not Axes, `plt.subplots(2, 2)[1]` being the realistic mistake, slipped through the length check and failed later with `AttributeError: 'numpy.ndarray' object has no attribute 'step'`, which is the same cryptic-error class the original comment was about. Test added.

**3d3af48 — arviz-stats pin widened**, with the evidence on @juanitorduz's thread.

**b4a0727 — the notebook diffs.** @daimon-pymclabs flagged that ~2,200 of the added lines were re-executed output. Only two cells per notebook are actually new, so every pre-existing cell now comes from the base commit and only the new cells come from the executed run, with execution counts renumbered so each notebook still reads as one sequential run. Verified per notebook that the cell sources match the executed version exactly, that the new cells carry the real figures, and that no pre-existing cell differs from base beyond its execution count. This also removes my machine's `language_info` from the committed metadata.

The notebook part of the diff goes from +2,332/-819 to +375/-45, and the PR as a whole to +678/-62.

Also checked and deliberately left alone:

- `NameError` on the `ppc` parameter: still out of scope, still happy to change it separately.
- The `layout="constrained"` on the ECDF figure: measured rather than assumed, it widens the gap between the two panels from 34 px to 74 px, so it is doing real work against the title/xlabel collision.
- `x = np.arange(observed.min(), observed.max() + 1)` grows with the observed maximum, so a dataset with extreme purchase counts pays more in the band simulation than the old fixed 100-point grid did. The per-simulation cost is dominated by the `n`-sized uniform draw either way, and repeat-purchase counts do not realistically reach the thousands, so I left it rather than add a cap.
- No `lower <= upper` assertion: that ordering is arviz's, not ours.

Verification: `tests/clv` 352 passed, 16 skipped. `tests/clv/test_plotting.py` 43 passed, and 43 passed again on the arviz 1.3 stack. Mock notebook runner green on all three notebooks after the rebuild ("Notebooks run successfully!"). Pre-commit clean on all six changed files. `uv lock --check` clean.

daimon-pymclabs:
Re-reviewed at `b4a0727`. The reference-CDF fix is right, and the guard is better placed than what I asked for.

I re-ran the coverage simulation against the constant in the source, `arviz-stats 1.3`, `nbinom(2, 0.3)`, `n = 2357`, nominal 0.95, 150 replicates per cell:

| reference | coverage |
|---|---|
| exact CDF | 0.933 |
| ratio 1 (seeds 1 / 13 / 45) | 0.393 / 0.387 / 0.360 |
| ratio 20 | 0.900 |

So the numbers in the `_MIN_PPC_RATIO` comment hold, including the 0.41 at ratio 1, which is the seed-to-seed range I get rather than the 0.28 you quoted in the thread. Nothing to change there; the comment is the version that reproduces.

Worth stating explicitly, because it is the part that is easy to get wrong and this diff gets it right: the guard compares `len(ppc)` to `len(observed)`, which for a sampled posterior reduces to `chain * draw` and not to the raw sample count. That is the correct denominator. Customers within one posterior draw share parameters, so they are not independent contributions to the reference CDF, and the Monte Carlo error of `z` is governed by the number of draws, not by `chain * draw * customer`. A ratio built on the total count would have read ~2000x too optimistic on a 10-draw fit. Warning instead of raising is also the right call given `mock_sample`, agreed.

Two things left, neither blocking.

**The ECDF path still carries the allocation you removed from the histogram.** `n_samples=samples if plot_type == "ecdf" else 1` means a point-estimate fit calling the default `plot_expected_purchases_ppc(model, plot_type="ecdf")` draws `1000 * n_customers * 2`, which is the 1,600 MB row of your own table at 1M customers. MAP is exactly the fit method someone reaches for at that size, so this is a reachable default rather than a corner. The guard needs a ratio of 20, and your table puts ratio 20 at 0.89 against 0.92 at ratio 1000, so the last 50x of allocation buys about three points of coverage. I do not think it needs to become a cap in this PR, but the `samples` docstring currently describes only when the parameter applies, not that it multiplies the predictive array by `n_customers` on that path. One sentence there would let someone size it down deliberately.

**`len(ax)` can still raise the cryptic error the validation exists to prevent.** `ax=42` or any non-sized object reaches `len(ax)` before anything else rejects it and comes out as `TypeError: object of type 'int' has no len()`. Adding a `Sequence`/`ndarray` check ahead of the length test keeps every bad input on the same `ValueError`.

The rest I checked and have nothing on: the `n_samples` narrowing is bit-identical for the histogram, `test_plot_expected_purchases_ppc_ecdf_map_fit` fails on pre-fix code for the right reason, the notebook diff rebuild is a real reduction rather than a squash, and dropping the direction claim from the Notes matches what both of us measured.

I could not run `tests/clv` here, so I am taking the 352 passed and the arviz 1.3 run on your report. Nothing blocking from my side; leaving the formal approve to @ColtAllen and @juanitorduz.

shivamlalakiya:
Before asking for another pass I re-reviewed the whole diff adversarially rather than only the two items left open. That turned up a correction to something I told you earlier, so that first.

## Correction: one committed figure was stale, and I said otherwise

I wrote above that "bg_nbd and mbg_nbd fit with 4 chains x 2000 draws and pareto_nbd with the MCMC default, so all three are on the sampled-posterior path where `n_samples` is ignored, and the committed figures are unchanged." The last clause was wrong. **`ParetoNBDModel.fit` defaults to `method="map"`**, unlike `BetaGeoModel` and `ModifiedBetaGeoModel` which default to `mcmc`:

```
ParetoNBDModel      fit method default = map
BetaGeoModel        fit method default = mcmc
ModifiedBetaGeoModel fit method default = mcmc
```

So `pnbd_map.fit(data=rfm_data)` in pareto_nbd cell 28 is a point-estimate fit, and its posterior ECDF cell was on exactly the path the reference-CDF fix changed. The docs were shipping the miscalibrated band this PR exists to remove, with the observed curve stepping outside it one cell above the markdown that says the model simulates customer behaviour reliably.

Fixed in bc3f5d81 by re-executing. To be sure of the blast radius I re-executed all three notebooks on HEAD and diffed every new figure payload rather than reasoning about it:

| notebook | cell | figure changed on HEAD |
|---|---|---|
| bg_nbd | 32 prior, 34 posterior | no, no |
| mbg_nbd | 33 prior, 35 posterior | no, no |
| pareto_nbd | 26 prior | no |
| pareto_nbd | **29 posterior** | **yes** |

Only the changed payload is committed, so that commit is a one-line diff. With the corrected band the observed ECDF stays inside it, so the figure and the surrounding text now agree.

Also worth correcting: a `15 of 44 points outside the band` number I derived while investigating described my own standalone reproduction, which used the default time unit and skipped the notebook's `spent > 0` filter. It does not describe the notebook. What that reproduction did establish is the mechanism: on the MAP path the bands from one draw per customer and from many are demonstrably different.

## The other thing that matters: item (A) needed a code change, not a docstring line

You asked for a sentence rather than a cap, and I agreed. Then I measured it, and with `ParetoNBDModel` defaulting to MAP this is the default path for that model rather than a corner:

| customers | `samples=50` | `samples=1000` (the default) |
|---|---|---|
| 5,000 | 39 MB / 4.6 s | 164 MB / 44.7 s |
| 20,000 | 38 MB / 9.5 s | 646 MB / 171.5 s |
| 100,000 (linear) | ~40 MB / ~45 s | ~3.2 GB / ~14 min |

Wall time is the worse half, and you only mentioned memory: the CLV random variables fall back to numba object mode, so the draws are expensive. Meanwhile coverage saturates long before the default. Measured, nominal 0.95, negative-binomial counts, n = 2357:

| ratio | 1 | 10 | 20 | 50 | 100 | 1000 |
|---|---|---|---|---|---|---|
| coverage | 0.28-0.41 | 0.83 | 0.89 | 0.91 | 0.955 | 0.92 |

Everything from 50 up is one bundle of Monte Carlo noise around nominal. So a 14-minute default buys nothing over a 45-second one.

In e95d787f the ecdf branch asks for a fixed 100 draws per customer (`_REFERENCE_DRAWS`), and `samples` goes back to meaning exactly what it meant before this PR: prior predictive checks only. That is a smaller surface than what I proposed to you, not a larger one: it deletes the docstring caveat, and it removes the "increase `samples`" advice being a no-op on the posterior path. Happy to switch to a tunable knob if you would rather have one, but I could not measure a reason to want it.

## The rest

- **`plot_type` was inserted as the 3rd positional parameter** of a public function in a 1.0.0 release, so `plot_expected_purchases_ppc(model, "posterior", 20)` meant `max_purchases=20` before this PR and raises now. Moved to last named parameter, so every pre-existing positional slot keeps its meaning. Nothing in-tree used positional arguments, so this is a downstream-only break that is now avoided rather than documented.
- **Item (B), the `ax` guard.** Done, but not with the `Sequence`/`ndarray` check we discussed: that still raises `TypeError` on a 0-d object array. `np.shape(ax) == (2,)` handles every input I could think of, and additionally rejects a `set` of two Axes, which passed the old check and then unpacked in arbitrary order. Verified over 13 input shapes: `ndarray (2,)`, list, tuple accepted; set, single Axes, `Figure`, 2x2 array, 0-d array, `42`, `"ab"`, `[1, 2]`, generator, dict all rejected with the ValueError, none raising.
- **The narrow-band warning told users to do something that cannot work.** "Increase 'samples'" is a no-op for a posterior check, which is the case that most often triggers it. Now scoped per path.
- **The band legend truncated instead of rounding**: `int(confidence_level * 100)` labelled a 99.9% band "99%". Now `:.4g`.
- **The test assertions could not fail.** This is the one I am least happy about, since you raised it once already. Monotonicity and [0, 1] bounds are properties of `compute_ecdf` and `get_pointwise_confidence_band` regardless of what gets plotted, and the "same grid" check compared two views of one array. 64b7d5f6 replaces them with value-level checks and I verified them by mutation:

| mutation | old assertions | new assertions |
|---|---|---|
| plot the predictive CDF instead of the observed one | pass | **fail** |
| shift the band one step off the curve grid | pass | **fail** |
| revert the grid to a 100-point linspace | pass | **fail** |
| stop forwarding `**kwargs` to the band | pass | **fail** |
| revert the reference draws to one per customer | fail | fail |

## Checked and deliberately left alone

No `Raises` section (this file has none in 710 lines); the `Sequence[plt.Axes]` annotation versus the ndarray `plt.subplots` returns (checked with a type checker, matplotlib's own overloads make it fine); empty observed data (`_validate_rfm_data` accepts empty frames, pre-existing and not this PR); `num_trials=0` raising `IndexError` (reachable only by hand-feeding the private helper); the numba stderr line and `metadata.execution` in the new notebook cells (32 of 69 committed notebooks already carry absolute paths, so this is repo practice rather than a regression); and the pre-existing `NameError` on the `ppc` parameter.

I also left the Notes wording as it is. The claim that the discreteness costs nothing does not survive the numbers: an exact reference CDF measures 0.93 against nominal 0.95, and at these replicate counts that gap is about 1.5 standard errors, so "only approximately the nominal level" claims exactly as much as the evidence supports.

## Verification

- `uv run pytest tests/clv` — 352 passed, 16 skipped
- `uv run pytest tests/clv/test_plotting.py` — 43 passed
- Same file on the arviz **1.3** stack — 43 passed
- Mutation table above
- All three notebooks re-executed with real sampling; mock notebook runner green
- `uv run pre-commit run --files <the changed files>` — all hooks pass
- `uv lock --check` — clean
- Merged `upstream/main`, so this is no longer behind

PR description refreshed; it had been describing the first commit. The notebook part of the diff is now +375/-45 rather than the +2,332/-819 you flagged, and the whole PR is +702/-61.

shivamlalakiya:
Heads up on the red check, since it is not from this PR: `notebooks (numba, --notebooks clv)` failed on a network error while fetching data, in a notebook this branch does not touch.

```
Error running notebook: clv_quickstart.ipynb
Exception encountered at "In [4]":
  raw_trans = pd.read_csv(
      "https://raw.githubusercontent.com/pymc-labs/pymc-marketing/main/data/cdnow_transactions.csv"
  )
...
  File /usr/lib/python3.12/ssl.py:1320, in SSLSocket.do_handshake
ConnectionResetError: [Errno 104] Connection reset by peer
URLError: <urlopen error [Errno 104] Connection reset by peer>
```

The reset happens inside `ssl.do_handshake`, so the TLS connection to raw.githubusercontent.com was dropped before any HTTP response arrived. Nothing in `pymc_marketing` appears in the stack.

Why it is not this PR:

- `docs/source/notebooks/clv/clv_quickstart.ipynb` has an empty diff here, and does not call `plot_expected_purchases_ppc`.
- The same workflow passed twice earlier today on this same branch, at 15:25 and 19:03, with that identical notebook.
- It passes locally on the current HEAD: `uv run python scripts/run_notebooks/runner.py --notebooks docs/source/notebooks/clv/clv_quickstart.ipynb` reports "Notebooks run successfully!".
- Every other check on this SHA is green.

I cannot re-run it myself: `gh run rerun --failed` returns "Must have admin rights to Repository". If either of you clicks **Re-run failed jobs** it will re-run just that one job. Otherwise I will push an empty retrigger commit once the remaining checks settle, which re-runs the whole matrix.

Not fixing it here, but flagging it as a recurring CI fragility worth its own change: `clv_quickstart.ipynb` and `pareto_nbd.ipynb` both fetch the CDNOW CSV over HTTPS even though `data/cdnow_transactions.csv` is in the repo, so any CLV notebook job can fail this way. The two candidate fixes pull in opposite directions, which is why I would rather you pick: reading the local file removes the flake but breaks the notebooks' standalone-runnable property, while a retry around the fetch in `scripts/run_notebooks/runner.py` keeps the docs as they are but hides genuine outages. Happy to open an issue or a small PR for whichever you prefer.

daimon-pymclabs:
Re-reviewed at `6992f1e`. Both open items are resolved, and the correction is the right call.

**On the `ParetoNBDModel` figure.** Confirmed, and it is the more important half of this round. `ParetoNBDModel.fit` does default to `method="map"` while `BetaGeoModel` and `ModifiedBetaGeoModel` default to `mcmc`, so `pareto_nbd` cell 28 was a point-estimate fit and its posterior ECDF was on the miscalibrated path. Re-executing all three notebooks and diffing figure payloads rather than reasoning about which ones changed is the correct way to bound that, and the one-cell diff matches what I would expect from it.

**On `_REFERENCE_DRAWS`.** I checked the thing the constant depends on, since it is the failure mode that would be expensive and silent: if any supported model applied `n_samples` unconditionally, `_REFERENCE_DRAWS = 100` would multiply a sampled-posterior ECDF by 100x with no warning. It does not. All four models that reach this function gate the expansion on the same MAP condition:

| model | file | gate |
|---|---|---|
| `ParetoNBDModel` | `pareto_nbd.py:937` | `chain == 1 and draw == 1` |
| `BetaGeoModel` | `beta_geo.py:725` | same |
| `ModifiedBetaGeoModel` | `modified_beta_geo.py:484` | same |
| `BetaGeoBetaBinomModel` | `beta_geo_beta_binom.py:572` | same |

So on a sampled posterior the constant is a no-op and the array is unchanged from `main`, and on a point-estimate fit it fixes the ratio at exactly 100 against a `_MIN_PPC_RATIO` of 20. That also makes the `samples` docstring true again, which is a better outcome than the sentence I asked you for. Fixing the draws rather than exposing a knob is the right trade given you could not measure a use for the knob.

Worth noting for anyone reading the constant later: the ratio is conservative on the MAP path in a way it is not on the sampled path. With one parameter draw the 100 predictive draws per customer are conditionally i.i.d. given theta, so the effective sample count behind `z` is `100 * n_customers`, not 100. The guard measures the sampled-posterior quantity in both cases, which errs the safe direction.

**On the `ax` guard.** `np.shape(ax) == (2,)` is better than the `Sequence`/`ndarray` check I proposed. I had missed that a 0-d object array satisfies the isinstance test and then fails at unpack, and rejecting a two-element `set` is a real fix rather than a bonus: it passed the old length check and then unpacked in arbitrary order, so the ECDF and difference panels could swap silently.

Two leftovers, neither blocking and neither new to this revision.

`reference` in `test_plot_expected_purchases_ppc_ecdf` is recovered as `observed_ecdf - (observed_ecdf - z)`, so `np.all(np.diff(reference) >= 0)` and the `[0, 1]` bounds hold for any `z` that `compute_ecdf` can return. They cannot fail. The value-level assertions above them carry the test, so this is dead weight rather than a gap, but the mutation table would not change if they were deleted.

Nothing asserts that the reference CDF is built from more than one draw per customer at a specific ratio. `test_plot_expected_purchases_ppc_ecdf_map_fit` asserts the warning is absent, which pins the ratio at 20 or above rather than at `_REFERENCE_DRAWS`. Dropping the constant to 25 would keep every test green. Not worth a test on its own, but if you touch the file again, asserting the predictive length on the MAP path would close it.

The red `notebooks (numba, --notebooks clv)` was a data-fetch network failure rather than anything in the diff, agreed, and the current run is in flight.

I could not run the suite here, so the 352 passed and the arviz-stats 1.3 run are on your report as before. Nothing blocking from me. Approval still belongs with @ColtAllen and @juanitorduz.

juanitorduz (approved):
Thanks @shivamlalakiya !

daimon-pymclabs (approved):
Re-reviewed the full diff. Every thread from my earlier pass is resolved, and the follow-up work goes further than what I asked for:

- `n_samples=_REFERENCE_DRAWS if plot_type == "ecdf" else 1` is the right call. Scaling the reference draws on the histogram path would have made the default `plot_expected_purchases_ppc(model)` allocate on the order of a gigabyte on a large point-estimate fit for no benefit, and the hist path is now byte-identical to pre-PR behaviour.
- The `ax` check catching a length-two sequence of non-Axes (`plt.subplots(2, 2)[1]`) closes the actual failure mode, not just the length case.
- Integer eval grid `np.arange(min, max + 1)`, the `ValueError` on degenerate all-equal counts, the corrected `95% simultaneous band` label, and the proportion axis labels all read correctly now.
- The continuous-CDF caveat and the estimated-vs-known reference CDF caveat are both in the public docstring Notes, with the ratio warning backed by the coverage numbers in the `_MIN_PPC_RATIO` comment. That is the right place for it.
- Notebook diff down to +375/-45 with pre-existing cells restored from base is a large improvement in reviewability.

Nothing blocking from me. Approving.

juanitorduz:
I will take a look into the failing test

PR #2906, opened 2026-08-18
fix(mmm): keep the ambient layout engine in MMMPlotSuite

### LINKED ISSUE #2779: decomposition.waterfall calls fig.tight_layout() unconditionally, overriding the ambient layout engine

anevolbap:
Fixes #2779

The waterfall fix in #2793 covered `decomposition.waterfall`. The remaining scope, per [this comment](https://github.com/pymc-labs/pymc-marketing/issues/2779#issuecomment-5318375243): ten `tight_layout()` calls in `mmm/plot.py` had the same defect, replacing a layout engine installed by the caller's style.

Adds a `_tight_layout(fig, **kwargs)` helper that lays the figure out only when no engine is set, and routes all ten sites through it. Three were `plt.tight_layout()` targeting the current figure; they now take the figure the method built, since that's what the guard inspects.

----
📚 Documentation preview 📚: https://pymc-marketing--2906.org.readthedocs.build/en/2906/


juanitorduz (approved):
thanks @anevolbap !

anevolbap:
Merge queue kicked this out on an unrelated failure: `test_slow (numba, 2)` / `tests/clv/models/test_gamma_gamma.py::TestGammaGammaModel::test_model_convergence`, a seeded-MCMC tolerance check on the Gamma-Gamma CLV model (`p` posterior mean 8.12 vs expected 6.0 at rtol=0.3). Nothing here touches `clv/`. Looks like sampling variance, not a regression from this change.

PR #2904, opened 2026-08-18
Rename funnel MuEffect classes to reflect subgraph topology

drbenvincent:
## Summary

- Renames notebook-local `FunnelEffect` classes to topology-specific names aligned with PR #2885:
  - `mmm_funnel_mueffect.ipynb`: `MeasuredMediatorEffect` (one unobserved mediator, one measurement, mediated μ contribution)
  - `mmm_funnel_mueffect_advanced.ipynb`: `BudgetDualIndicatorEffect` (budget-linked spend node, two indicators, different DAG)
- Updates field names (`hub_adstock`, `hub_input_transform`, …), model variable prefixes (`mediator_*`, `budget_dual_*`), and factory helpers to mirror the YAML notebook implementation.
- Adds brief callouts that class names describe topology while “funnel” remains the use-case term.
- Re-executed both notebooks (mock-sampled, CI-style).

Closes #2892

## Notes

- Final names follow Daimon’s review on #2892 and PR #2885 (`MeasuredMediatorEffect` / `BudgetDualIndicatorEffect`), not the issue’s original `Latent*` / `Panel*` proposal.
- `mmm_funnel_from_yml.ipynb` and YAML specs are already renamed in #2885; this PR aligns the upstream hand-wired notebooks.
- `FUNNEL_EFFECT_SUBGRAPH_GUIDE.md` does not exist and was skipped.

## Test plan

- [x] Mock-executed `mmm_funnel_mueffect.ipynb`
- [x] Mock-executed `mmm_funnel_mueffect_advanced.ipynb`
- [x] Pre-commit hooks pass on both notebooks

Made with [Cursor](https://cursor.com)

----
📚 Documentation preview 📚: https://pymc-marketing--2904.org.readthedocs.build/en/2904/

Issue #2901, opened 2026-08-18
Rethinking MuEffects for funnel DAGs


drbenvincent:
## Why this

`MuEffect` is a powerful hook. It can add structure the base `MMM` does not know about: extra nodes, extra likelihoods, a mediated term in `μ`. The funnel notebooks prove it.

The problem is where that is heading. Each new subgraph shape is another custom class. The intro notebook has one `FunnelEffect`; the geo notebook has a different topology under the same name; the next client DAG is another `create_effect` and another YAML `class:` key. That does not scale.

(`FunnelEffect` over-promises a general funnel builder. [#2892](https://github.com/pymc-labs/pymc-marketing/issues/2892) proposes renaming while the classes are still notebook-local, so the name describes the topology rather than the use case. [#2885](https://github.com/pymc-labs/pymc-marketing/pull/2885) starts that with `MeasuredMediatorEffect` / `BudgetDualIndicatorEffect`. A rename is right. It does not fix the proliferation.)

The other route is **one class whose configuration is the DAG**. The next topology is more config, not a new Python class. A custom `create_effect` can already encode any graph you can write in PyMC. The claim is that **the next DAG should not require a next class**.

This stays on `MuEffect`. Direct media, intercept, seasonality, and `y` stay on `MMM`. Because the spec *is* nodes and edges, a NetworkX fragment for [#2886](https://github.com/pymc-labs/pymc-marketing/issues/2886) falls out of it rather than a hand-maintained `DAG_EDGES` list.

## Proposal

One `MuEffect` — working name `SubgraphEffect` — whose public language is **nodes and edges**. You name the extra nodes; for each, parents in, children out.

### Parents and children

Take the intro funnel. Unobserved demand has two parents (upper spend, a control) and two children (a noisy reading of demand, and sales).

```mermaid
%%{init: {"flowchart": {"curve": "basis", "defaultRenderer": "dagre"}}}%%
flowchart LR
  upper_spend -->|"parent"| demand
  lower_control -->|"parent"| demand
  demand -->|"child"| lower_spend
  demand -->|"child"| y
```

`parents` / `children` is the DAG vocabulary. We still distinguish **two kinds of child**, because they are different statistical objects:

| | **Measurement** | **Contribution** |
|---|---|---|
| Child node | An observed column (`lower_spend`) | Always `y` (a term in `μ`) |
| What it is | A likelihood: noisy reading of the latent | A deterministic added into the sales mean |
[...]

PR #2907, opened 2026-08-19
Fix funnel MuEffect footguns: shared data vars and channel-by-name API

### LINKED ISSUE #2871: FunnelEffect notebook: hardcoded channel index and DataVarMuEffect name collisions
## Summary

The notebook-local `FunnelEffect` in `mmm_funnel_mueffect.ipynb` demonstrates joint funnel estimation via `MuEffect`, but two implementation choices are easy to misuse if this pattern is copied into production code.

## Caveats

1. **Positional channel index.** Treatment is wired via `mmm.channel_data_scaled.isel(channel=0)`. Reordering `channel_columns` or adding channels silently targets the wrong upper-funnel series.

2. **`DataVarMuEffect` unprefixed data names.** `DataVarMuEffect.create_data` registers `pmd.Data` under unprefixed variable names (e.g. `lower_control`). Two effects sharing a control name collide at build time with `ValueError: Variable name ... already exists.`

## Context

Documented as a warning in `docs/source/notebooks/mmm/mmm_funnel_mueffect.ipynb` (see PR #2868). Broader architecture discussion in #2775.

## Possible follow-ups

- Resolve treatment by channel name instead of positional index (depends on a stable `MMM` API for scaled channel data).
- Namespace or prefix `DataVarMuEffect` data registration to avoid name collisions across multiple effects.

drbenvincent:
## Summary
- Add `MMM.scaled_channel(channel)` to select scaled spend by the model `channel` coordinate (the axis that actually indexes `channel_data_scaled`), not by `channel_columns` position. For DataFrame input those orders match; for `xr.Dataset` they can differ, and indexing by `channel_columns` silently returned the wrong channel.
- Register each `DataVarMuEffect` dataset column once as `pm.Data` under its dataset name so multiple effects can share a column. Skip only when the existing node is a data variable with matching dims and shape; raise on any other name collision. Issue #2871 asked for namespacing to avoid collisions, but prefixing `{prefix}_{var_name}` created duplicate graph nodes for the same underlying data. Sharing under the dataset name is the intended contract; genuine collisions stay loud.
- Update the funnel notebook to use the new APIs.

Closes #2871

## Test plan
- [x] `pytest tests/mmm/test_additive_effect.py::TestDataVarMuEffect`
- [x] `pytest tests/mmm/test_mmm.py::TestMultidimMMMEdgeCases::test_scaled_channel_selects_by_name`
- [x] `pytest tests/mmm/test_mmm.py::TestMultidimMMMEdgeCases::test_scaled_channel_follows_dataset_coord_order`
- [x] `pytest tests/mmm/test_mmm.py::TestMultidimMMMEdgeCases::test_mu_effects_share_one_pm_data_node`
- [x] `pytest tests/mmm/test_mmm.py::TestMultidimMMMEdgeCases::test_mu_effect_column_named_like_channel_data_raises`
- [x] `pytest tests/mmm/test_incrementality.py::TestPostFitMutationGuard`
- [x] `pytest tests/mmm/test_incrementality.py::TestMediatedMuEffects::test_evaluator_discovers_the_effect_s_own_data`
- [x] Pre-commit on changed files

----
📚 Documentation preview 📚: https://pymc-marketing--2907.org.readthedocs.build/en/2907/


drbenvincent:
## Review round 1 (Bugbot + manual)

### Bugbot
No bugs flagged on branch diff.

### Manual review notes

**Breaking change (intentional, closes #2871)**
- `DataVarMuEffect` now registers `pm.Data` as `{prefix}_{var_name}` instead of unprefixed dataset names.
- Custom subclasses must read inputs via `model[self.model_data_name(var)]` (built-ins `MediaMuEffect` / `ControlMuEffect` updated).
- Incrementality/counterfactual paths are compatible: graph discovery uses live `pm.Data` names (`funnel_lf_budget` in tests).

**`MMM.scaled_channel`**
- Selects by `channel_columns` index (same as previous `isel(channel=0)` pattern, but name-safe).
- Raises if model not built or channel missing.

**Coverage added in follow-up commit**
- `test_scaled_channel_selects_by_name` now also asserts error when model is unbuilt.
- `DataVarMuEffect` class docstring documents the namespacing contract.

**Out of scope / follow-ups**
- Open PRs #2885 / #2904 still carry notebook-local `isel(channel=0)` until rebased onto this API.
- Untracked `FUNNEL_EFFECT_SUBGRAPH_GUIDE.md` not updated here.

### Round 1 fixes pushed
- Clarified `DataVarMuEffect` docstring
- Added unbuilt-model guard test for `scaled_channel`

juanitorduz:
thanks @drbenvincent ! This is great! Can you please take a look into the failing notebook test :) ?

pymc_marketing/mmm/additive_effect.py
```diff
@@ -551,7 +573,7 @@ def create_data(self, mmm: Model) -> None:
         """
         for var_name in self.data_vars:
             da = mmm.xarray_dataset[var_name]
-            pmd.Data(var_name, da.values, dims=da.dims)
+            pmd.Data(self.model_data_name(var_name), da.values, dims=da.dims)
```
williambdean, line 576: If it is actually the same data, dont think we want differnet nodes in the graph

drbenvincent: Good catch - prefixing created duplicate graph nodes for the same underlying data. Switched to registering each dataset column once under its dataset name; subsequent effects reuse the existing `pm.Data` node if it's already there. `scaled_channel` is unchanged. Fix in https://github.com/pymc-labs/pymc-marketing/pull/2907/commits/de2c00ade48e7b9e847c9eebe692d432b79c1f94

drbenvincent:
So we touched core library code and the error was only caught by a failing doc build. So planning some more testing around this...

drbenvincent:
Following up on @williambdean's review: switched from prefixed `pm.Data` names to registering each dataset column once (shared across effects).

Added a few unit tests so we catch regressions before notebook CI:

- **Single node** — two `DataVarMuEffect`s sharing a column register one `pm.Data`, regardless of order
- **Graph wiring** — both effects' `create_effect` paths read the same registered node
- **`set_data`** — updating via one effect updates the shared node
- **`MMM.build_model` smoke** — thin integration test through the real build loop

These guard the contract that bit us in the funnel notebook: one graph node per dataset column, not one per effect prefix.

daimon-pymclabs:
## Review

Two substantive findings, plus some smaller notes.

### 1. `scaled_channel` reintroduces the positional footgun for `xr.Dataset` inputs

`pymc_marketing/mmm/mmm.py` maps name to position via `self.channel_columns.index(channel)`, but the `channel` axis of `channel_data_scaled` comes from `self.xarray_dataset._channel`, whose coord order is only forced to `channel_columns` on the DataFrame path (`_reindex_to_user_order` in `data_conversion.py`).

For `xr.Dataset` input, `_validate_mmm_structure` checks only that a `channel` coord exists, not its order, and `_reindex_dataset_to_user_order` is called on the prediction path, not in `_generate_and_preprocess_model_data`. So for a Dataset whose `channel` coord is ordered differently from `channel_columns`, `scaled_channel` silently returns the wrong channel: exactly the failure mode the helper is meant to remove.

Suggested fix: resolve the index against the coord that actually indexes the tensor, e.g. `list(self.model_coords["channel"]).index(channel)`, and validate membership against that coord too. Worth a test that builds with a Dataset whose channel coord order differs from `channel_columns`; the new `test_mu_effects_share_one_pm_data_node` passes a Dataset with matching order, so this case is currently untested.

### 2. The dedup guard converts a loud error into a silent wrong model

```python
for var_name in self.data_vars:
    if var_name in mmm.model.named_vars:
        continue
```

`named_vars` contains every named variable, not just `pm.Data` nodes registered from the same dataset column. `create_data` runs after `channel_scale`, `target_scale`, `channel_data` and `target_data` are registered, so a dataset column named e.g. `channel_data` skips registration and the subsequent `model[var_name]` inside `create_effect` silently binds to the base channel tensor. Before this PR that raised a duplicate-name error.

Suggested fix: skip only when the existing node is a data variable whose dims and shape match `mmm.xarray_dataset[var_name]`, and raise a clear error otherwise. That keeps the intended sharing while preserving a loud failure on a genuine collision.

### Smaller notes

- The assertions `sum(name == "shared" for name in mmm.model.named_vars) == 1` (in `tests/mmm/test_additive_effect.py` and `tests/mmm/test_mmm.py`) are vacuous: `named_vars` keys are unique by construction, so the sum can never exceed 1. Counting the model's data variables, or asserting node identity between the two effects' inputs, would test what is intended.
- `test_set_data_updates_shared_column` only calls `effect_b.set_data`. Calling both effects' `set_data` would exercise the idempotence of the shared-node update, which is the actual concern introduced here.
- The PR description says registration is prefixed as `{prefix}_{var_name}`, but the implementation dedupes under the plain dataset name instead. Those are opposite semantics (isolation vs sharing), and #2871 asked for namespacing, so the rationale for choosing sharing is worth a line in the description and in the changelog-facing text.
- The two `tests/mmm/conftest.py` hunks are pure formatting reflow with no functional change.
- The notebook block is still a `{warning}` directive, but its content is now a neutral "library helper" note. `{note}` reads better.

drbenvincent:
## Review round 2

Addressed @daimon-pymclabs's comments:

- **`scaled_channel`** now indexes `channel_data_scaled` by `model_coords["channel"]` (the tensor axis), not `channel_columns`. Dataset input whose channel coord order differs from `channel_columns` no longer silently returns the wrong channel.
- **`DataVarMuEffect.create_data`** still shares a column across effects, but only when the existing node is a `pm.Data` with matching dims and shape. A name collision with something else (e.g. a dataset column named `channel_data`) raises instead of silently binding to the wrong tensor.
- Tests cover the Dataset-order case, both collision kinds, shared-node identity, and idempotent `set_data` from both effects. Notebook `{warning}` → `{note}`; PR description now matches the sharing contract rather than the old prefixing plan.

drbenvincent:
As long as people are ok with this, this one can be merged independent of the broader discussion around #2886 and #2901

daimon-pymclabs:
## Round 2 review

Both round-1 findings are fixed, and correctly.

- `scaled_channel` now resolves the index against `model_coords["channel"]`, and `test_scaled_channel_follows_dataset_coord_order` pins it properly: it asserts the `channel_columns` index gives a *different* answer, so the test would actually fail if the old positional logic came back. That is the right shape for a regression test.
- `create_data` now distinguishes the three cases (absent, matching data var, collision) and raises on the last two kinds. `test_shared_data_wired_through_create_effect` asserting node identity via `ancestors` replaces the vacuous `sum(name == "shared" ...) == 1` assertion, and `set_data` is now called from both effects.

Not blocking, and fine as follow-ups:

**A same-shape collision with an MMM-internal data node is still silent.** The dims-and-shape check catches `channel_data` because that node is `(date, channel)` while the colliding column is `(date,)`. But `target_data` and `time_index` are both registered as `(date,)` with length `n_dates`, so a dataset column named either of those matches on dims *and* shape and gets silently shared with a semantically unrelated tensor. The check is structural where the real invariant is "this name belongs to my dataset column". Cheapest fix is to refuse the handful of reserved MMM-internal names (`channel_data`, `target_data`, `channel_scale`, `target_scale`, `time_index`, `control_data`) outright, since sharing with those is never what a `DataVarMuEffect` wants.

**`existing.get_value()` copies.** Default is `borrow=False`, so the shape check materializes a full copy of the array just to read `.shape`. `existing.get_value(borrow=True).shape` avoids it, or `existing.type.shape` where it is static.

Agreed on merging this independently of #2886 / #2901: it is a bug fix on the current API surface and does not presuppose which way the funnel `MuEffect` discussion lands.

juanitorduz:
## Round 3 review

Approve with one requested change. The round-1/round-2 findings are genuinely fixed, but the dedup guard still converts one loud error into a silent wrong model, and it reproduces.

### Verified

| Check | Result |
|---|---|
| `tests/mmm/test_additive_effect.py` + `tests/mmm/test_incrementality.py` at head | 207 passed, 6 skipped (74s) |
| `test_mmm.py::TestMultidimMMMEdgeCases` | 17 passed |
| `pre-commit run --files` on the 5 changed files | all hooks pass (ruff, mypy, nbformat, API-doc completeness) |
| Funnel notebook cells 1-16 (through `gen.build_model`, `do()`, prior predictive, `scaled_channel("upper_spend")`) | executes clean; `lower_spend` registered exactly once |
| `git merge-tree main <head>` | clean (branch is 8 commits behind main) |

`scaled_channel` is correct: `model_coords` is built from `xarray_dataset.coords` (`mmm.py:1809`), the same dataset that feeds `pmd.Data("channel_data", ..._channel)`, so `model_coords["channel"]` is the tensor axis. `XTensorVariable.__eq__` is `object.__eq__`, so `existing not in model.data_vars` is identity-based (no symbolic-truthiness trap), and `model.data_vars` is a `treelist` holding only real data nodes.

### Change requested: a `target_data` collision is silent, and it feeds the target into `mu`

`create_data` shares any existing data var whose dims *and* shape match. `target_data` is registered at `mmm.py:2303`, **before** the `mu_effect.create_data` loop at `mmm.py:2335`, with dims `("date", *self.dims)` -- exactly the shape of the most common `DataVarMuEffect` column.

One effect, one dataset column named `target_data`:

```
PR head: target_data: NO ERROR (shared)
           registered value == effect column?  False
           registered value == scaled target?  True
           effect contribution reads that node: True
main:    target_data: raised ValueError Variable name target_data already exists.
```

The effect's own series is never registered, its contribution reads the target tensor, and the target enters `mu` as a predictor. That contradicts the PR's own "genuine collisions stay loud" contract, and it is the only behavioral regression I found.

`time_index` is *not* affected: it is registered at `mmm.py:2339`, after the effect loop, so PyMC's duplicate-name check fires (`ValueError: Variable name time_index already exists.`). So the round-2 note was half right. `channel_data` is caught by the dims check, as the new test shows.

The check is structural where the invariant is ownership. Cheapest sound fix:

```python
# mmm.py, immediately before `for mu_effect in self.mu_effects:`
self._library_data_names = frozenset(self.model.named_vars)
```

```python
# DataVarMuEffect.create_data
if var_name in getattr(mmm, "_library_data_names", frozenset()):
    raise ValueError(
        f"{var_name!r} is registered by MMM itself; rename the dataset column."
    )
```

No reserved-name list to keep in sync. A cheaper stopgap is adding `np.array_equal(existing.get_value(borrow=True), da.values)` to the reuse condition, but that leaves the `set_data`-overwrites-a-library-node hazard when the arrays happen to be equal. Either way, the `target_data` case is a one-line regression test next to `test_mu_effect_column_named_like_channel_data_raises`.

### Nits

- `additive_effect.py:576`: `existing.get_value()` defaults to `borrow=False`, so it copies the whole array just to read `.shape`. Use `get_value(borrow=True)`; `existing.type.shape` is `(None,)` for `pmd.Data`, so that is not an option.
- `additive_effect.py:568`: the existence check reads `mmm.model` while the write uses the ambient model context. `pmd.Data` accepts `model=`; pass `model=model` so both target the same object.
- `docs/source/notebooks/mmm/mmm_funnel_from_yml.ipynb:774` still uses `mmm.channel_data_scaled.isel(channel=0)` inside a `DataVarMuEffect` -- the same footgun, in an executed doc. One-line switch to `scaled_channel`, squarely in this PR's scope (`tests/mmm/conftest.py:714` as well, lower value).
- `test_create_data_raises_on_non_data_name_collision` runs in a model with zero data vars, so `existing not in model.data_vars` is trivially true; the test would still pass if the membership check were broken. Registering a `pmd.Data` in that model first makes it exercise the real comparison.
- `test_mu_effect_column_named_like_channel_data_raises` pins the dims-mismatch branch, not anything `channel_data`-specific; the name oversells it.
- `scaled_channel` vs the existing `channel_data_scaled` attribute: adjacent names with reversed word order. `scaled_channel_data(channel)` reads as the obvious accessor for that tensor.

juanitorduz:
@drbenvincent this one is very close to be done :) I left the last review :)

drbenvincent (commented):
## Review round 1

**Review outcome: APPROVE.** GitHub does not allow this authenticated account to
formally approve its own PR, so this is submitted as a visible comment review.

The open correctness comments are addressed:

- `scaled_channel` resolves its index from the model `channel` coordinate, and the Dataset-coordinate-order regression test proves the old positional lookup would fail.
- `DataVarMuEffect` shares only its own previously registered `pm.Data` nodes; the build-time snapshot prevents both same-shape and different-shape MMM-internal collisions, including `target_data`.
- Registration now explicitly targets the inspected model and the shape probe avoids an unnecessary data copy.

Verified on remote head `c4de60dea48f1837e496ce363fa2ba7a0b904bd3`:

`uv run pytest tests/mmm/test_additive_effect.py::TestDataVarMuEffect tests/mmm/test_mmm.py::TestMultidimMMMEdgeCases` — 27 passed.

drbenvincent:
Addressed the remaining Round 3 review points in c4de60d: MMM-owned data-node names (including `target_data`) are now rejected before effects register data; shared effect-owned data remains supported. Added the `target_data` regression test, avoided the shape-check copy, bound `pmd.Data` to the same model explicitly, strengthened the non-data collision test, and replaced the funnel notebook’s positional channel selection with `scaled_channel("upper_spend")`. Focused MMM tests, lint, formatting, and notebook schema validation pass.

drbenvincent:
@juanitorduz Can you assess if this is ready. Agent addressed the last review and a critical self review (with a sub-agent) thinks it's ready to go. But happy if you want another round

daimon-pymclabs:
## Round 4 review

The Round 3 change request is fixed, and the fix is the sound one rather than the stopgap. Approving from my side.

### Verified against `c4de60d` (static review of the head diff, plus CI)

- `mmm.py:2416` snapshots `self._library_data_names = frozenset(self.model.named_vars)` immediately before the `mu_effects` loop, and `DataVarMuEffect.create_data` checks membership before any reuse. The ordering is right: `existing is None` registers, a library name raises, a non-data name raises, then dims/shape decide reuse. Names registered by an *earlier effect* are not in the snapshot, so effect-to-effect sharing still works, which is exactly the property the inline comment claims.
- `target_data` now raises `"registered by MMM itself"` instead of silently feeding the scaled target into `mu`, and `test_mu_effect_column_named_like_target_data_raises` pins it at the `build_model` level, where the ordering hazard actually lives. `test_mu_effect_column_named_like_mmm_data_raises` covers `channel_data` the same way.
- The three nits are done: `get_value(borrow=True)` (no full-array copy for a shape read), `pmd.Data(..., model=model)` so the existence check and the write target the same object, and `test_create_data_raises_on_non_data_name_collision` now registers a real `pmd.Data` first so the `existing not in model.data_vars` branch is genuinely exercised. `test_mu_effect_column_named_like_channel_data_raises` was replaced by `test_create_data_raises_on_mismatched_existing_data`, which is what it always tested.
- Both funnel notebooks use `mmm.scaled_channel("upper_spend")`; no `isel(channel=...)` positional selection remains in either, and the `mmm_funnel_mueffect.ipynb` admonition flipped from warning to note correctly. The `conftest.py` half of that nit is moot: the fixtures there use `channel_data_scaled.sum(dim="channel")` or the whole tensor, never a positional channel index.
- CI: tests, pre-commit, sdist and all non-MMM notebook jobs pass at head. The `notebooks (cvm, --notebooks mmm --end-idx 10)` job and the RTD build were still running when I looked; the funnel notebook change lands in that job, so worth a glance before merging.

### Residual, none blocking

- The snapshot covers *all* named vars at that point, not just data nodes, so a dataset column named `intercept` now reports `"registered by MMM itself"` rather than `"a non-data variable with that name already exists"`. Both are loud and the stricter one is the safer default; only the message is slightly less precise.
- Names MMM registers *after* the loop (`time_index`) still fall through to PyMC's own `Variable name ... already exists`. Loud, correct, but a different and less actionable message than the new one. Only worth unifying if someone hits it.
- `_library_data_names` is set in `build_model` only, and `create_data` reads it through `getattr(..., frozenset())`. That means the guard is exercised solely by the two `test_mmm.py` build-level tests, not by anything in `test_additive_effect.py`. That is the right place to test it, so this is a note rather than a gap.
- `scaled_channel` vs the existing `channel_data_scaled` attribute: the reversed word order stays a little confusable, and this is public API from the moment it merges. Not worth another round, but it is the last cheap moment to rename.

juanitorduz (approved):
Thanks @drbenvincent

PR #2908, opened 2026-08-21
feat(mmm): co-optimize monetary spend variables with the media budgets

PabloRoque:
## Description

**Lower-funnel spend becomes a decision rather than an input.** It is money, so it belongs in the same pot as the media budgets — but `LeverVariable` (#2884) cannot carry it: a lever is optimized in its own units and deliberately sits *outside* the budget-sum constraint. This adds a second variable on the media path over a different node.

```python
    start_date=..., end_date=...,
    response_variable="total_response_original_scale",
    spend_vars=["lf_budget"],
)
result = optimizer.allocate_budget(total_budget=200_000)

result.budgets                  # media, per channel
result.spend_var_allocations    # {"lf_budget": ...}, the other money
```

**`constraints.py` is untouched.** The default sum constraint already totals every variable that reports a `budget_contribution`, with a comment written in anticipation of exactly this. That seam shipped in #2884 with one implementor; this is the second, and the first evidence it was designed right.

## Commits

**1 — one owner for the window's three blocks.** `MMM.budget_optimizer` derived carry-in, decisions and carry-over inline from the date axis, three lines from the `BudgetOptimizer` guard that re-checks the same arithmetic. `_window_layout` names the split once; the guard stays, because it protects callers who set the three by hand.

Also drops dead defensiveness in `effective_carryover_lags`: `incrementality_spec` is concrete on the `MuEffect` ABC (`additive_effect.py:477`, returning `None`) and `mmm.py` already imports `MuEffect`, so neither the `getattr` nor the `NotImplementedError` branch could fire. They *could* fire for a duck-typed object that is not a `MuEffect` subclass — which is what removing them exposed in a test double, so the double now opts out the way the ABC's own default does. Production code tolerating a shape no registered effect can have is the wrong way round. **Flagging it as the one intentional narrowing here**, since `add_mu_effect` is annotated `MuEffect` but does not enforce it at runtime.

**2 — the spend variables.** `spend_vars` names additional monetary `pm.Data` nodes. Each becomes a `MediaVariable` over its own non-date dims, inheriting the window's three blocks, cost-per-unit conversion and mask scatter, and joins the budget constraint by reporting its spend. Three things it needed:

- **`scales_dims`**, so a scales array over `("geo",)` broadcasts. The hardcoded `("channel",)` only ever applied to *array* scales, which is why a scalar-scaled spend variable worked without it — smaller than expected.
- **`MediaVariable.from_model`**, reading dims, coords and dtype off the model as `LeverVariable.from_model` already does, so `budget_optimizer.py` gains no model introspection of its own.
- **A scatter that recognises a 0-dimensional mask.** A budget decided over date alone — a single national spend — has no cells, and pytensor cannot index with a scalar boolean. Found by running it, not by planning.

**`x0` now shares one budget** across the cells of every spending variable instead of handing each the whole of it, which would start the solver at twice the budget, outside an equality constraint. Unchanged with one spender.

**An unreachable spend variable is refused** like an unreachable lever, for a sharper reason: it still draws from the budget, so the solver funds it out of media and gets nothing back.

**3 — review fixes.** The spender set is settled once at construction rather than rebuilding throwaway symbolic tensors on every `x0` call; `_window_layout` documents that only the date axis comes from the model it is handed and the carry-over from `self`; `optimized_vars` stops calling itself "non-media".

**4 — `spend_var_allocations`.** `optimized_vars` carries levers and money alike, and a caller summing against a budget must not add a discount depth to it. `spend_var_names` records which is which and the property reads it back, so the budget check is self-contained in the result. Two accessors rather than one merged `monetary_allocations`: `budgets` is per-channel and a spend variable is per-spend-var, so merging needs either the media variable's name in the result or a heterogeneous mix. Uglier, honest about the structure.

## Measured

Media and lower-funnel spend reach a joint interior optimum on the funnel fixture, and the split is exact:

| total | media | `lf_budget` | sum |
|---|---|---|---|
| 10 | 10.000 | 0.000 | 10.000 |
| 50 | 36.514 | 13.486 | 50.000 |
| 200 | 123.204 | 76.796 | 200.000 |
| 1000 | 640.279 | 359.721 | 1000.000 |

At a small budget media wins outright; as it saturates the lower funnel becomes worth funding. The interior split is what shows the two are genuinely traded off — a corner solution would satisfy the sum constraint just as well while saying nothing about whether the second variable moves the objective.

## Verification

712 passed, 6 skipped across `test_utils.py`, `test_budget_optimizer.py`, `test_budget_optimizer_mmm.py`, `test_optimization_variables.py`, `test_mmm.py`, `test_additive_effect.py`, `test_cost_per_unit.py`, `test_incrementality.py`. Lint and mypy clean.

Mutation-verified, each failing the test that asserts its guarantee:

| mutation | caught by |
|---|---|
| spend variable stops reporting its spend | budget no longer sums to the total |
| `x0` hands each spender the whole budget | the guess sums to 2× |
| 0-d scatter special case removed | `NotImplementedError: Boolean scalar indexing` |
| duplicate-name guard removed | `DID NOT RAISE ValueError` |
| carry-in read from the channel node, not the variable's own | three tests |
| `spend_var_names` left unpopulated | two tests |
| `spend_var_allocations` returns every `optimized_vars` entry | the lever-exclusion unit test |

Two of these are worth calling out, because in both cases no end-to-end optimization would have caught the bug:

- **The `x0` fix.** SLSQP recovered from the infeasible start on this fixture, so every optimization test passed with it broken. It is asserted on the guess itself.
- **The lever-exclusion test.** As first written it declared `spend_vars=[]` and contained no lever, so it asserted the empty case and would have passed whatever the property returned. The distinction is only observable with a lever *and* a spend variable present at once, which the funnel fixture cannot produce (`lf_budget` is date-only, so it cannot be a lever), so it is now a unit test on a result built by hand.

## Not in this PR

- **Per-spend-variable configuration.** Scales, `cost_per_unit` and a cell mask all default; the plumbing accepts them but they are not exposed per entry. Bounds already work through the existing `bounds_overrides`, keyed by variable name.
- **`requires_full_axis`** — an effect declaring `evaluation_mode="full"` cannot be reproduced by any window, and `effective_carryover_lags` still sizes one anyway.
- **Carry-in dates inflating the reported objective.** Argmax is unaffected; the quoted value is not.

## Related

- Retires "it does not co-optimize the lower funnel" from #2890's caveats.
- Same machinery freq/reach needs: reach is bought with money, and `cost_per_unit_tensor` already exists to fold `assumed_frequency` into the rate (#1968).
- Builds on #2884 (the `budget_contribution` seam), #2895 (the objective that sees the mediated path) and #2898 (the window).

## Type of change

- [x] New feature (`spend_vars` defaults to empty; `x0` unchanged with a single spending variable; `spend_var_names` defaults to empty)
- [x] Minor breaking change: an effect that is not a `MuEffect` subclass and lacks `incrementality_spec` now raises instead of being skipped (see commit 1)

### DIFF
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -318,9 +318,13 @@ class BudgetOptimizationResult:
         The raw scipy optimization result (solver diagnostics, ``x``, ``fun``,
         convergence status).
     optimized_vars : dict[str, xarray.DataArray]
-        Optimal values of any non-media decision variables co-optimized
-        alongside the media budgets. Empty until such variables are declared
-        (see https://github.com/pymc-labs/pymc-marketing/pull/2621).
+        Optimal values of every decision variable other than the media budgets,
+        by name: ``optimizable_vars`` levers in their own units, and
+        ``spend_vars`` monetary variables in money. Empty when neither is
+        declared. Note that a monetary entry here and ``budgets`` are the same
+        kind of quantity, drawn from the same total -- ``budgets`` is singled
+        out because it is the allocation most callers want, not because it is
+        the only spend.
     callback_info : list[OptimizationIterationInfo] or None
         Per-iteration diagnostics (``x``, ``fun``, ``jac``, constraint values)
         when ``allocate_budget(callback=True)``; ``None`` otherwise.
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -325,6 +325,10 @@ class BudgetOptimizationResult:
         kind of quantity, drawn from the same total -- ``budgets`` is singled
         out because it is the allocation most callers want, not because it is
         the only spend.
+    spend_var_names : list[str]
+        Which ``optimized_vars`` entries are monetary, i.e. the declared
+        ``spend_vars``. Recorded so the result can say on its own which of its
+        allocations draw from the budget; see :attr:`spend_var_allocations`.
     callback_info : list[OptimizationIterationInfo] or None
         Per-iteration diagnostics (``x``, ``fun``, ``jac``, constraint values)
         when ``allocate_budget(callback=True)``; ``None`` otherwise.
@@ -333,7 +337,35 @@ class BudgetOptimizationResult:
     budgets: DataArray
     scipy_result: OptimizeResult
     optimized_vars: dict[str, DataArray] = field(default_factory=dict)
+    spend_var_names: list[str] = field(default_factory=list)
     callback_info: list[OptimizationIterationInfo] | None = None
+
+    @property
+    def spend_var_allocations(self) -> dict[str, DataArray]:
+        """The ``optimized_vars`` entries that are money rather than levers.
+
+        ``optimized_vars`` holds both kinds, and telling them apart matters:
+        these are drawn from the same total as :attr:`budgets`, so they are
+        what a caller sums against a budget, while a lever's units are its own
+        and adding them to money means nothing.
+
+        Returns
+        -------
+        dict[str, xarray.DataArray]
+            One entry per declared ``spend_vars`` name.
+
+        Examples
+        --------
+        Check that the plan spends the budget and no more:
+
+        .. code-block:: python
+
+            spent = float(result.budgets.sum()) + sum(
+                float(allocation.sum())
+                for allocation in result.spend_var_allocations.values()
+            )
+        """
+        return {name: self.optimized_vars[name] for name in self.spend_var_names}
 
     def __iter__(self):
         """Yield ``(budgets, scipy_result)`` for two-element unpacking."""
@@ -2243,6 +2275,7 @@ class BudgetOptimizer(BaseModel):
                 budgets=optimal_budgets,
                 scipy_result=result,
                 optimized_vars=unpacked,
+                spend_var_names=list(self.spend_vars),
                 callback_info=callback_info if callback else None,
             )
 
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -1313,6 +1313,17 @@ class BudgetOptimizer(BaseModel):
         ),
     )
 
+    spend_var_scales: dict[str, float | Sequence[float]] = Field(
+        default_factory=dict,
+        description=(
+            "Per-spend-variable scale factors converting monetary amounts into "
+            "the node's stored units, keyed by name. Omit a name to declare its "
+            "node stored in raw money. Mirrors channel_scales for the media "
+            "variable: a node stored scaled and left at the default 1.0 is "
+            "optimized in the wrong units, silently."
+        ),
+    )
+
     optimizable_vars: dict[str, Sequence[tuple[float | None, float | None]] | None] = (
         Field(
             default_factory=dict,
@@ -1681,13 +1692,24 @@ class BudgetOptimizer(BaseModel):
                 f"spend_vars may not name channel_data_var "
                 f"({self.channel_data_var!r}); it is already optimized."
             )
+        unknown = set(self.spend_var_scales) - set(self.spend_vars)
+        if unknown:
+            raise ValueError(
+                f"spend_var_scales names {sorted(unknown)}, which are not in "
+                f"spend_vars {list(self.spend_vars)}. A scale on a name that is "
+                "not optimized has no effect, so it is more likely a typo than "
+                "an intention."
+            )
         spends = [
             MediaVariable.from_model(
                 self.model,
                 name,
                 num_periods=self.num_periods,
                 adstock_periods=self.adstock_periods,
-                carry_in_values=carry_in_for(name),
+                carry_in_periods=self.carry_in_periods,
+                scales=np.asarray(self.spend_var_scales[name])
+                if name in self.spend_var_scales
+                else 1.0,
                 date_dim=self.date_dim,
             )
             for name in self.spend_vars
--- a/pymc_marketing/mmm/budget_optimizer.py
+++ b/pymc_marketing/mmm/budget_optimizer.py
@@ -1264,6 +1264,19 @@ class BudgetOptimizer(BaseModel):
         ),
     )
 
+    spend_vars: Sequence[str] = Field(
+        default_factory=tuple,
+        description=(
+            "Names of additional monetary pm.Data variables to co-optimize with "
+            "the media budgets. Each must vary over the date dimension and is "
+            "decided over its remaining dims, in money, so it competes with "
+            "media for the same total through the default budget-sum "
+            "constraint. Use optimizable_vars instead for a quantity that is "
+            "not money -- a discount depth, a frequency -- which is optimized "
+            "in its own units and stays outside that constraint."
+        ),
+    )
+
     optimizable_vars: dict[str, Sequence[tuple[float | None, float | None]] | None] = (
         Field(
             default_factory=dict,
@@ -1603,11 +1616,12 @@ class BudgetOptimizer(BaseModel):
         # model graph. One do() call over every variable keeps gradients joint.
         # Read from the model we were handed: it already carries the spend that
         # preceded the window, so nothing here needs to know where history lives.
-        carry_in_values = None
-        if self.carry_in_periods:
-            carry_in_values = np.asarray(self.model[self.channel_data_var].get_value())[
-                : self.carry_in_periods
-            ]
+        def carry_in_for(name: str) -> np.ndarray | None:
+            if not self.carry_in_periods:
+                return None
+            return np.asarray(self.model[name].get_value())[: self.carry_in_periods]
+
+        carry_in_values = carry_in_for(self.channel_data_var)
 
         media_variable = MediaVariable(
             name=self.channel_data_var,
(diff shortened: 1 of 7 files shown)


daimon-pymclabs:
## Review: co-optimize monetary spend variables with the media budgets

Reviewed at `a30df142`. The architectural claim in the description holds up: `constraints.py` really is untouched, and the extension lands entirely through `budget_contribution`, so the seam introduced in #2884 is doing the work it was designed for. The `_WindowLayout` / `_window_layout` consolidation in `mmm.py` is arithmetically identical to the code it replaces, and the new surface (`spend_vars`, `MediaVariable.from_model`, `scales_dims`, the 0-d branch in `scattered`, `spend_var_names` / `spend_var_allocations`) is coherent.

Two issues I would want addressed before merge, then four cleanups. Both blockers were reproduced against a checkout of this branch with a scratch test file: 4 of 9 tests failed on the unpatched code and all 9 passed after the fixes below, with `test_optimization_variables.py` (47) and `test_budget_optimizer.py` (49) staying green.

### Blocker 1: validation ordering means a bad `spend_vars` name never reaches its error message

At `budget_optimizer.py:1690`, `carry_in_values=carry_in_for(name)` is evaluated as an argument, so `self.model[name].get_value()` at line 1658 dereferences the name before `MediaVariable.from_model` validates it. Whenever `carry_in_periods > 0`, which is the normal `MMM.budget_optimizer` path since carry-in comes from `effective_carryover_lags`, two things fall out of that single line:

* A misspelled name raises a bare `KeyError` from `pymc/model/core.py` instead of the "is not a variable with named dims in the model" message written at `optimization_variables.py:397`. That message is currently unreachable on the default path.
* A dimensioned but non-shared node (e.g. `channel_contribution`) raises `AttributeError: 'XTensorVariable' object has no attribute 'get_value'` rather than anything actionable.

Compare `LeverVariable.from_model`, which validates first and calls `.get_value()` last.

Suggested fix: pass the period count rather than the value, so the read happens after the two existing `ValueError` checks. In `MediaVariable.from_model`, add a parameter next to `carry_in_values`:

```python
carry_in_values: np.ndarray | None = None,
carry_in_periods: int = 0,
```

and insert this after the `date_dim` check, before `cell_dims = ...`:

```python
if carry_in_periods and carry_in_values is None:
    node = model[name]
    if not hasattr(node, "get_value"):
        raise ValueError(
            f"spend variable '{name}' is not a shared variable, so the "
            "spend made before the window cannot be read from it. "
            "Spend variables must be pm.Data nodes."
        )
    carry_in_values = np.asarray(node.get_value())[:carry_in_periods]
```

At the caller, `carry_in_values=carry_in_for(name)` becomes `carry_in_periods=self.carry_in_periods`. Keeping `carry_in_values` as an accepted argument leaves the media variable's own construction at line 1667 untouched, so the change is additive; `carry_in_for` stays for `channel_data_var`, whose name is not user-supplied.

Tests worth adding, parametrized over `carry_in_periods` in `(0, 2)` since only the nonzero case discriminates:

```python
def test_a_misspelled_spend_var_says_so(fitted_mmm):
    with pytest.raises(ValueError, match="not a variable with named dims"):
        make_optimizer(fitted_mmm, spend_vars=["lf_bugdet"])


def test_a_deterministic_spend_var_says_so(fitted_mmm):
    with pytest.raises(ValueError, match="not a shared variable"):
        make_optimizer(fitted_mmm, spend_vars=["channel_contribution"])
```

### Blocker 2: spend variables are pinned to scale 1.0 with no way to change it

`MediaVariable.from_model` accepts `scales` (`optimization_variables.py:349`) and forwards it to `channel_scales`, but `model_post_init` never passes one (`budget_optimizer.py:1685`), so every spend variable gets `1.0`. Media itself needs `channel_scales` precisely because `channel_data` is stored scaled. If a user's spend node went through comparable scaling, the optimizer treats one stored unit as one dollar and the allocation is wrong with no error and no warning. A silently wrong default is a different thing from an absent one.

The downstream machinery already works: with the scale threaded through, the substituted node is divided correctly and the budget stays shared in money. This is pure plumbing.

Suggested fix, a new field next to `spend_vars`:

```python
spend_var_scales: dict[str, float | Sequence[float]] = Field(
    default_factory=dict,
    description=(
        "Per-spend-variable scale factors converting monetary amounts into "
        "the node's stored units, keyed by name. Omit a name to declare its "
        "node stored in raw money. Mirrors channel_scales for the media "
        "variable: a node stored scaled and left at the default 1.0 is "
        "optimized in the wrong units, silently."
    ),
)
```

and in `model_post_init`, ahead of the `spends` comprehension:

```python
unknown = set(self.spend_var_scales) - set(self.spend_vars)
if unknown:
    raise ValueError(
        f"spend_var_scales names {sorted(unknown)}, which are not in "
        f"spend_vars {list(self.spend_vars)}. A scale on a name that is "
        "not optimized has no effect, so it is more likely a typo than "
        "an intention."
    )
```

with `scales=np.asarray(self.spend_var_scales[name]) if name in self.spend_var_scales else 1.0` passed into `from_model`.

Flagging the unknown-key `ValueError` as a judgement call toward strictness: silently dropping the key produces exactly the wrong-units answer the field exists to prevent, so I think it should fail loudly, but it is a stricter API than the PR currently has and you may prefer a warning.

If you would rather not widen the API here, the minimum is a sentence in the `spend_vars` field description stating that the named node must be in raw monetary units.

### Cleanups, none of which change behaviour for a correct caller

1. **`test_a_lever_does_not_take_a_share_of_the_budget` contains no lever.** It builds one `_money("channel_data")` and asserts `spending_variables()` returns it. Nothing would fail if `spending_variables` returned every variable unconditionally. This is the same vacuity the description identifies and fixes for `test_spend_var_allocations_excludes_levers`; add a `LeverVariable` and assert it is absent.

2. **`budget_contributions()` does not reuse `_spending`.** `spending_variables` documents itself as the single answer to "does this spend?", but `budget_contributions()` at `optimization_variables.py:822` re-runs the `is not None` filter independently. Iterating `self._spending` there makes the claim structural rather than conventional and drops a second graph build.

3. **The commit-1 narrowing leaves one path inconsistent.** `spend_reach.py:231` still does `getattr(effect, "incrementality_spec", lambda: None)()`, the same defensiveness removed from `mmm.py`. After this PR one path tolerates a duck-typed effect without the method and the other raises `AttributeError`. Since the removal is already a documented minor break, making it uniform here is cheaper than leaving a second reading of the contract in the tree. Relatedly, an `isinstance` check in `add_mu_effect` would turn that `AttributeError` into a message that says what to do.

4. **`duplicated = [name for name in self.spend_vars if name == self.channel_data_var]`** (`budget_optimizer.py:1678`) is a list comprehension used as a boolean, and the message does not use the list. `if self.channel_data_var in self.spend_vars` reads better.

Minor: the nested `unreachable(names)` has an untyped parameter in a module annotated throughout, and `spend_var_allocations` will `KeyError` if `spend_var_names` and `optimized_vars` ever disagree. Only one construction site exists so that is fine today, but it is an invariant held by convention across a dataclass boundary rather than by construction.

I also checked whether `spend_vars: Sequence[str]` would accept a bare string and iterate it character by character. It does not; `spend_vars="lf_budget"` is rejected.

### Test coverage

Thorough and unusually explicit about what each test discriminates: sum constraint, interior split, empty case, unreachable refusal, duplicate refusal, per-variable carry-in, channel-name collision, 0-d scatter, geo-dimensioned `from_model`. The mutation table is real verification rather than decoration. Remaining gaps are the vacuous lever test above and the absence of an end-to-end case with two spend variables; only `x0` covers the multi-spender path, and symbolically.

### Risk

Low for existing users. `spend_vars` defaults empty, `x0` is unchanged with a single spender, `scales_dims` defaults to `("channel",)`, and the window refactor is behaviour-preserving. The one real break is the `incrementality_spec` narrowing, correctly flagged, and it only affects objects that are not `MuEffect` subclasses.

Approve with changes: the two blockers above, then the cleanups at your discretion.

juanitorduz:
Sorry @PabloRoque , daimon was supposed to just push a comment but also did a commit. Feel free to revert it if not relevant 🙏

daimon-pymclabs:
Apologies for the noise, @PabloRoque — @juanitorduz is right that I was only asked for a comment and pushed `55ad545` as well. That was not intentional, and reverting it costs you nothing: it is a leaf commit containing only the two fixes from my review (validate the `spend_vars` name before `model[name]` is dereferenced, and thread `spend_var_scales` through `MediaVariable.from_model`), plus their tests. `git revert 421205c 55ad545` or a force-push back to `a30df14` both leave the rest of the branch untouched.

On the red CI, so you are not chasing it:

* `pre-commit.ci` was mine. `ruff format` wanted a line break in the test I added at `test_budget_optimizer_mmm.py:2202`. I have pushed `421205c` with just that reformat, nothing else.
* `test_slow (numba, 2)` is unrelated to this branch. The failure is `tests/clv/models/test_gamma_gamma.py::TestGammaGammaModel::test_model_convergence`, a sampling convergence assertion at `rtol=0.3` that missed on one of three elements (8.12 actual vs 6.0 desired). No CLV code is touched here.
* `all_tests` is the aggregate gate reporting those two.

Happy to keep future reviews comment-only.

timbo112711:
Great catch here @PabloRoque! Looking solid to me

daimon-pymclabs:
The red `Test` run on 421205c is unrelated to this PR's diff.

The only failing job is `test_slow (numba, 2)`, and the only failing test in it is `tests/clv/models/test_gamma_gamma.py::TestGammaGammaModel::test_model_convergence`:

```
Not equal to tolerance rtol=0.3, atol=0
Mismatch at index [0]: 8.115318525372663 (ACTUAL), 6.0 (DESIRED)
ACTUAL:  array([ 8.115319,  3.7727  , 12.035771])
DESIRED: array([ 6.,  4., 15.])
```

That is a CLV sampling-convergence tolerance check with nothing in the call path this PR touches (MMM budget optimization), so it reads as a sampler flake rather than a regression. A re-run should clear it. If it reproduces on a re-run, it is worth a separate issue against the CLV test rather than blocking this PR.

juanitorduz (approved):

Issue #2740, opened 2026-07-16
using Index in order to solve the spends in natural case


williambdean:
This should be solvable

CC: @ianhi

PR #2910, opened 2026-08-21
feat(mmm): add OriginalScaleIndex and fix drop_scalar_coords for custom xarray indexes

williambdean:
## Summary

- Adds `OriginalScaleIndex`: a custom xarray `Index` that maps scaled `x` coordinates to original domain on `.sel()`, with support for multi-dim scale (e.g. geo × channel panel models)
- Fixes a latent bug in `plot.drop_scalar_coords` where `curve.indexes` triggered `to_pandas_indexes()`, failing for any custom xarray index — changed to `curve.xindexes`
- Exports `OriginalScaleIndex` from `pymc_marketing.mmm`

### DIFF
--- a/pymc_marketing/mmm/__init__.py
+++ b/pymc_marketing/mmm/__init__.py
@@ -77,6 +77,7 @@ from pymc_marketing.mmm.optimization_variables import (
     OptimizationVariable,
     OptimizationVariables,
 )
+from pymc_marketing.mmm.original_scale_index import OriginalScaleIndex
 from pymc_marketing.mmm.plotting import MMMPlotSuiteFacade
 from pymc_marketing.mmm.preprocessing import (
     preprocessing_method_X,
@@ -133,6 +134,7 @@ __all__ = [
     "NoSaturation",
     "OptimizationVariable",
     "OptimizationVariables",
+    "OriginalScaleIndex",
     "PeriodicCovFunc",
     "RootSaturation",
     "SaturationTransformation",
--- a/pymc_marketing/mmm/original_scale_index.py
+++ b/pymc_marketing/mmm/original_scale_index.py
@@ -1,0 +1,234 @@
+#   Copyright 2022 - 2026 The PyMC Labs Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""Custom xarray Index for saturation curve x coordinates in original scale."""
+
+from __future__ import annotations
+
+import numpy as np
+import pandas as pd
+import xarray as xr
+from xarray import Index
+from xarray.core.indexing import IndexSelResult
+from xarray.indexes import PandasIndex
+
+
+class OriginalScaleIndex(Index):
+    """Custom xarray Index that maps scaled x coordinates to original domain on selection.
+
+    Stores ``x_original = x_scaled * channel_scale`` as a pre-computed DataArray
+    via xarray broadcasting. Selecting on any scale dimension (e.g. ``"channel"``,
+    ``"geo"``) reduces the array; once all scale dimensions are resolved, the
+    ``"x"`` coordinate is replaced with original-domain values backed by a plain
+    :class:`~xarray.indexes.PandasIndex`. Partial selections return a new
+    ``OriginalScaleIndex`` over the remaining scale dimensions.
+
+    ``da.xindexes["x"]`` and ``da.xindexes["channel"]`` return the same
+    ``OriginalScaleIndex`` instance.
+
+    Parameters
+    ----------
+    x_original : xr.DataArray
+        Pre-computed original-domain x values. Must have ``"x"`` as one dimension;
+        all other dimensions are treated as scale dimensions (e.g. ``("x", "channel")``
+        or ``("x", "geo", "channel")``). The ``"x"`` coordinate holds the scaled
+        linspace values; the DataArray values are ``x_scaled * channel_scale``.
+
+    Examples
+    --------
+    Attach directly to a DataArray via :func:`~xarray.DataArray.set_xindex`:
+
+    .. code-block:: python
+
+        channel_scale = xr.DataArray(
+            [5000.0, 1200.0],
+            dims=["channel"],
+            coords={"channel": ["TV", "Radio"]},
+        )
+        # da has dims (chain, draw, channel, x) with x in [0, 1]
+        curve = da.drop_indexes(["x", "channel"]).set_xindex(
+            ["x", "channel"], OriginalScaleIndex, channel_scale=channel_scale
+        )
+        curve.sel(channel="TV").coords["x"]  # TV's original-domain x values
+
+    See :meth:`~pymc_marketing.mmm.MMM.sample_saturation_curve` for the primary
+    use-case where this index is attached automatically.
+    """
+
+    def __init__(self, x_original: xr.DataArray) -> None:
+        self.x_original = x_original
+
+    @property
+    def _scale_dims(self) -> tuple[str, ...]:
+        return tuple(d for d in self.x_original.dims if d != "x")
+
+    @classmethod
+    def _from_x_and_scale(
+        cls,
+        x_values: np.ndarray,
+        channel_scale: xr.DataArray,
+    ) -> OriginalScaleIndex:
+        x_da = xr.DataArray(x_values, dims=["x"], coords={"x": x_values})
+        return cls(x_original=x_da * channel_scale)
+
+    @classmethod
+    def from_variables(
+        cls,
+        variables: dict[str, xr.Variable],
+        *,
+        options: dict | None = None,
+    ) -> OriginalScaleIndex:
+        """Create an OriginalScaleIndex from coordinate variables.
+
+        Called by xarray when ``set_xindex([...], OriginalScaleIndex, channel_scale=da)``
+        is invoked.
+
+        Parameters
+        ----------
+        variables : dict[str, xr.Variable]
+            Must contain an ``"x"`` variable (scaled x linspace). Scale dimension
+            variables are derived from ``options["channel_scale"]``.
+        options : dict, optional
+            Must contain ``"channel_scale"``: an ``xr.DataArray`` with the per-channel
+            scale factors and whatever dims apply.
+
+        Returns
+        -------
+        OriginalScaleIndex
+        """
+        options = options or {}
+        x_var = variables.get("x")
+        channel_scale = options.get("channel_scale")
+        if x_var is None:
+            raise ValueError(
+                f"OriginalScaleIndex requires an 'x' variable. Got: {list(variables)}"
+            )
+        if channel_scale is None:
+            raise ValueError(
+                "OriginalScaleIndex requires 'channel_scale' in options. "
+                "Pass it via set_xindex(..., channel_scale=da)."
+            )
+        return cls._from_x_and_scale(x_var.values, channel_scale)
+
+    def create_variables(
+        self,
+        variables: dict | None = None,
+    ) -> dict[str, xr.Variable]:
+        """Return the coordinate variables managed by this index."""
+        result: dict[str, xr.Variable] = {
+            "x": xr.Variable("x", self.x_original.coords["x"].values)
+        }
+        for dim in self._scale_dims:
+            result[dim] = xr.Variable(dim, self.x_original.coords[dim].values)
+        return result
+
+    def sel(
+        self,
+        labels: dict,
+        method: str | None = None,
+        tolerance: float | None = None,
+    ) -> IndexSelResult:
+        """Handle label-based selection.
+
+        Selects on scale dimensions by indexing into the pre-computed ``x_original``
+        DataArray. When all scale dimensions are resolved the ``x`` coordinate is
+        replaced with original-domain values. Partial selections return a new
+        ``OriginalScaleIndex`` over the remaining dimensions.
+
+        Parameters
+        ----------
+        labels : dict
+            Selection labels. Keys may be any subset of the managed scale
+            dimension names or ``"x"``.
+        method : str, optional
+            Passed through to PandasIndex for ``"x"`` selection.
+        tolerance : float, optional
+            Passed through to PandasIndex for ``"x"`` selection.
+
+        Returns
+        -------
+        IndexSelResult
+        """
+        scale_labels = {k: v for k, v in labels.items() if k in self._scale_dims}
+        x_labels = {k: v for k, v in labels.items() if k == "x"}
+
+        if scale_labels:
+            dim_indexers: dict[str, int] = {}
+            for dim, val in scale_labels.items():
+                arr = self.x_original.coords[dim].values
+                matches = np.where(arr == val)[0]
+                if not len(matches):
+                    raise KeyError(f"{dim}={val!r} not found. Available: {list(arr)}")
+                dim_indexers[dim] = int(matches[0])
+
+            selected = self.x_original.sel(scale_labels)
+            remaining = [d for d in self._scale_dims if d not in scale_labels]
+
+            if not remaining:
+                x_natural = selected.values  # shape (n_x,) — already in original domain
+                return IndexSelResult(
+                    dim_indexers=dim_indexers,
+                    variables={"x": xr.Variable("x", x_natural)},
+                    indexes={"x": PandasIndex(pd.Index(x_natural), "x")},
+                )
+            else:
+                new_idx = OriginalScaleIndex(selected)
+                return IndexSelResult(
+                    dim_indexers=dim_indexers,
+                    indexes={"x": new_idx, **{d: new_idx for d in remaining}},
+                )
+
+        if x_labels:
+            x_scaled = self.x_original.coords["x"].values
+            return PandasIndex(pd.Index(x_scaled), "x").sel(
+                {"x": x_labels["x"]}, method=method, tolerance=tolerance
+            )
+
+        raise NotImplementedError(
+            f"OriginalScaleIndex.sel does not support labels: {list(labels)}"
+        )
+
+    def isel(self, indexers: dict) -> OriginalScaleIndex | PandasIndex:
+        """Handle integer-based selection.
+
+        Parameters
+        ----------
+        indexers : dict
+            Integer indexers for ``"x"`` and/or any scale dimensions.
+
+        Returns
+        -------
+        OriginalScaleIndex | PandasIndex
+            Returns a ``PandasIndex`` on original-domain x values once all scale
+            dimensions are reduced; otherwise returns a new ``OriginalScaleIndex``.
+        """
+        x_idx = indexers.get("x", slice(None))
+        scale_indexers = {d: indexers[d] for d in self._scale_dims if d in indexers}
+        selected = self.x_original.isel({"x": x_idx, **scale_indexers})
+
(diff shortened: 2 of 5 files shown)


williambdean:
Close #2740

CC: @ianhi

williambdean:
Please review this @daimon-pymclabs

daimon-pymclabs:
## Review: `OriginalScaleIndex` and the `drop_scalar_coords` fix

Reviewed at `252ec5f`. The `plot.py` change is right and I would take it on its own: `curve.indexes` calls `to_pandas_indexes()`, which raises for any index that is not pandas-backed, so `drop_scalar_coords` was unusable with a custom index. `curve.xindexes` is the correct accessor and the substitution is behaviour-preserving for every pandas-backed index. The export and the two negative-path tests in `test_plot.py` are fine.

The index itself needs work before merge. I probed `original_scale_index.py` in isolation against xarray 2026.7.0 (numpy/pandas/xarray only, no pymc import) with the same two-channel setup the tests use. Five distinct operations fail, and one of them fails silently.

### Blocker 1: a label on `x` combined with a label on a scale dim is silently discarded

```python
curve.sel(channel="TV", x=0.5)   # OriginalScaleIndex
# sizes {'x': 5}, values [0. 1. 2. 3. 4.]

da.sel(channel="TV", x=0.5)      # same data, default indexes
# 2.0
```

Because `x` and `channel` are managed by the same index, xarray hands both labels to one `sel` call. `sel` splits them at lines 172-173, and the `if scale_labels:` branch at line 175 returns before `x_labels` is ever looked at. Nothing raises. The caller asked for one point and got the whole axis back, with the `x` coordinate correctly rescaled so the result looks plausible.

This is the one I would not merge around. The semantics are a real design question, since after the channel is resolved the `x` coordinate is in the original domain while the label the user typed is presumably scaled, and I do not want to guess which you intend. The minimum safe behaviour is to refuse:

```python
if scale_labels and x_labels:
    raise NotImplementedError(
        "OriginalScaleIndex cannot resolve an 'x' label together with a scale "
        "label in one .sel(), because 'x' changes domain once the scale is "
        "applied. Chain them instead: .sel(channel=...).sel(x=...)."
    )
```

Chaining already works and gives the right answer, so the workaround is one character of punctuation.

### Blocker 2: `remaining` is computed from indexer names, not from which dims survived

`isel` (line 229) and `sel` (line 185) both decide "are all scale dims resolved?" by asking whether the dim was *mentioned*, not whether it was *reduced*. A list or slice indexer mentions the dim and keeps it, so the code takes the fully-resolved branch and builds a `PandasIndex` over a still-2-D array:

| call | result on this branch |
|---|---|
| `curve.isel(channel=[0, 1])` | `ValueError: Index data must be 1-dimensional` |
| `curve.isel(channel=slice(0, 1))` | `ValueError: Index data must be 1-dimensional` |
| `curve.sel(channel=["TV", "Radio"])` | `ValueError: dimensions ('x',) must have the same length as the number of data dimensions, ndim=2` |
| `curve.groupby("channel")` | `ValueError: Index data must be 1-dimensional` |

`groupby` is the one that worries me. It is the ordinary way to iterate a saturation curve per channel, and it goes through `isel` with a list, so it breaks on every curve this index is attached to.

The fix is to ask the selected array which dims it still has:

```python
if set(selected.dims) == {"x"}:
    return PandasIndex(pd.Index(selected.values), "x")
return OriginalScaleIndex(selected)
```

I applied exactly that to `isel` in my scratch copy: all four calls above stop raising, and `isel(x=slice(0, 3))`, `isel(channel=0)` and `sel(channel="TV")` stay correct. The same reasoning applies to the `remaining` computation in `sel`.

One caveat so the fix is not oversold: with it, `groupby("channel")` no longer raises but each group still carries **scaled** `x`, because groupby reduces via a one-element list rather than a scalar, so the index never reaches its fully-resolved branch. `sel(channel="TV")` gives original-domain `x` and `groupby` gives 0-1 `x` for the same channel. That divergence should be resolved, or documented loudly, before users hit it.

### Blocker 3: selecting on `x` alone never works

```python
curve.sel(x=0.5)
# ValueError: dimensions ('x',) must have the same length as the number of data dimensions, ndim=0
curve.sel(x=0.26, method="nearest")
# same
```

The `if x_labels:` branch at line 201 returns a scalar `dim_indexers`, the index survives the selection, and `create_variables` then tries `xr.Variable("x", <0-d array>)`. Slices are fine (`sel(x=slice(0, 0.5))` works), scalars are not. `method` and `tolerance` are documented as pass-throughs at lines 163-166 and are unreachable in practice for the case they exist to serve.

`create_variables` needs to handle an `x_original` that has lost its `x` dim, or `sel`/`isel` need to hand back a plain index when `x` is reduced to a scalar.

### Smaller points

1. **`sel` on a scale dim only accepts an exact scalar.** `np.where(arr == val)` at line 179 means `curve.sel(channel=slice("Radio", "TV"))` reports `KeyError: channel=slice(...) not found. Available: ['TV', 'Radio']`, which reads as if the slice were a missing label. `PandasIndex` handles slices and lists natively; delegating to one per scale dim would cover blocker 2 and this together.
2. **`int(matches[0])` at line 182 keeps only the first hit** and discards the rest without comment. With a scalar label there is only ever one, so it is not a bug today, but it is the line that lets a list indexer through into the wrong branch.
3. **`create_variables` ignores its `variables` argument** and rebuilds from `self.x_original.coords`, dropping any `attrs` and any dtype the caller had set on the coordinate.
4. **The class docstring points at a method that does not use it.** Lines 74-75 say `MMM.sample_saturation_curve` is "the primary use-case where this index is attached automatically". Nothing in `pymc_marketing/` references `OriginalScaleIndex` outside `mmm/__init__.py`, so that sentence is false on this branch and will render as such in the API docs. Either wire it up here or write it as a forward intention.
5. **`test_repr_contains_dims_and_n_x` barely discriminates.** `assert "5" in r` passes on any repr containing the digit, and `"channel" in r` would pass on the class name alone in a differently-named case. Asserting the full string is cheaper to read and actually pins the format.

### Test coverage

The panel/2-D scale cases, the `az.hdi` round-trip and the `to_series` case are good choices and cover the parts most likely to break on an xarray upgrade. The gap lines up exactly with the failures above: nothing calls `sel(x=...)`, nothing passes a list or slice to a scale dim, nothing calls `groupby`, and nothing combines an `x` label with a scale label. Four tests, one per row of the table plus the combined-label case, would have caught all of this.

### Risk

Contained. Nothing existing constructs an `OriginalScaleIndex`, so the only change reaching current users is the one-word `plot.py` fix, which is safe. The blockers matter for whoever wires this into `sample_saturation_curve` next, and blocker 1 matters most because it is the only one that returns an answer instead of raising.

Happy to push any of these if you would rather not, but I will wait to be asked.

williambdean:
Thanks @daimon-pymclabs — this is exactly the review this index needed. All points are addressed in 762e1a60d (rebased on current main):

**Blocker 1 (combined labels silently discarded):** scalar `x` labels combined with scale labels now raise `NotImplementedError` with chaining guidance. Slices/lists on `x` **are** supported combined (`sel(channel="TV", x=slice(0, 2500))` resolves the scale dims first and applies the spend-domain slice to the resolved axis) — we verified the scalar case cannot carry the original-domain coordinate through the sel machinery (`_overwrite_indexes` ignores `variables` once `indexes` is empty), so it refuses rather than guess.

**Blocker 2 (`remaining` from mentioned dims):** `sel`/`isel` now compute surviving dims from the selected array, and scale-dim labels resolve through a per-dim `PandasIndex`. `sel(channel=[...])`, `isel(channel=[...])`, slices and `groupby("channel")` all work. Your caveat stands: groupby groups keep scaled `x` (documented in *Known limitations* along with the `drop_indexes(["channel", "x"])` escape hatch).

**Blocker 3 (`sel(x=scalar)` crash):** x labels are now spend-domain by contract; scalar `x` requires the scale dims resolved (chained), and `isel` returns `None` when every managed dim reduces (PandasIndex semantics), so the crash path is gone and `method`/`tolerance` are reachable where they apply.

**Smaller points:** (1, 2) scale-dim label resolution delegated to `PandasIndex` — the `np.where`/`matches[0]` logic is gone; (3) `create_variables` honors passed variables; (4) docstring now forward-references #2740 instead of claiming `sample_saturation_curve` uses it; (5) repr test pins the exact string.

**Tests:** 18 new tests covering combined slice/list x (spend domain), list/slice scale indexers, groupby, all raise paths, exact repr, attrs preservation — which also closes the codecov gap.

Happy to take the groupby scaled-`x` divergence further in #2740 if you'd rather see it resolved than documented. Ready for re-review whenever you get a chance.

daimon-pymclabs:
## Re-review at `762e1a6`

Re-probed `original_scale_index.py` in isolation against xarray 2026.7.0 (numpy/pandas/xarray only, no pymc import), same two-channel setup plus a `("geo", "channel")` panel case.

Blockers 1-3 are genuinely closed. Every call from my table now behaves: `sel(channel=["TV","Radio"])`, `isel(channel=[0,1])`, `isel(channel=slice(0,1))` and `groupby("channel")` all return instead of raising; `sel(channel="TV").sel(x=2500)` gives `0.5` and `sel(channel="TV", x=slice(0, 2500))` gives spend-domain `[0, 1250, 2500]`; the two refusal paths raise with the guidance you describe; `method="nearest"` is reachable and correct. Smaller points 1-5 are all addressed as described, and `create_variables` really does hand back the caller's variable identity-preserved.

Two things I found on this commit. The first is the same failure mode as blocker 1, in the branch the new combined-label support opened.

### Blocker 1 has a surviving twin: an `x` slice or list is silently discarded when a scale dim stays unresolved

The new scalar guard fires before the selection runs, so it only sees whether `x` is scalar, not whether the scale dims will actually reduce. A list or slice indexer on a scale dim *mentions* the dim and keeps it, so the code takes the `if remaining:` early return at line 295 and the `x_labels` you validated at 263-282 are never applied:

```python
curve.sel(channel=["TV", "Radio"], x=slice(0, 2500))
# sizes {'channel': 2, 'x': 5}   -- x unfiltered
# x coords [0. 0.25 0.5 0.75 1.]  -- and still scaled

curve.sel(channel=["TV"], x=slice(0, 2500))          # same, size-1 list
curve.sel(channel=slice("TV", "Radio"), x=slice(0, 2500))  # same
```

It also hits the panel case with more than one scale dim, which is the one I would expect a user to trip over first, because resolving one of two scale dims looks like an ordinary partial selection:

```python
# scale dims ("geo", "channel")
c2.sel(channel="TV", x=slice(0, 2500))
# sizes {'geo': 2, 'x': 5}, x coords [0. 0.25 0.5 0.75 1.]
```

No exception, and `x` comes back in the 0-1 domain while the caller passed a spend-domain slice, so the wrong answer is a plausible-looking one. That is the property that made blocker 1 the one I would not merge around, and the docstring at lines 72-77 currently promises the opposite: "no `x` label can be used while any scale dimension is unresolved -- both raise `NotImplementedError`".

`test_sel_x_with_partially_resolved_scale_dims_raises` looks like it covers this, but it passes `x=2500.0`, a scalar, so it exercises the line-272 guard rather than this path. Nothing in the suite passes a slice or list `x` together with an unresolved scale dim.

The fix is to move the refusal after `remaining` is computed, where it can see what actually survived:

```python
selected = self.x_original.isel(dim_indexers)
remaining = [d for d in self._scale_dims if d in selected.dims]

if remaining:
    if x_labels:
        raise NotImplementedError(
            "OriginalScaleIndex cannot resolve an 'x' label while the scale "
            f"dimensions {remaining} are unresolved: 'x' labels are in the "
            "original (spend) domain, which differs per channel. A list or "
            "slice indexer keeps its scale dimension, so reduce it to a "
            "scalar first, e.g. .sel(channel='TV', x=slice(...))."
        )
    return IndexSelResult(...)
```

I applied exactly that in my scratch copy: all four calls above raise with the domain explanation, and every currently-passing behaviour I probed is unchanged, including `sel(channel="TV", x=slice(0, 2500))`, `sel(channel="TV", x=[0.0, 2500.0])`, `sel(channel="TV", x=slice(1000, None))` with `method="nearest"`, `groupby`, the list and slice scale selections, `to_series` and the `az.hdi` round-trip. Refusing is the conservative reading; if you would rather *support* it by resolving the scale dims first and applying the `x` label per remaining channel, that is a bigger semantic call and I would keep it out of this PR.

### `__repr__` raises `KeyError` after `isel` reduces `x`

`isel` now drops the `x` variable when `x` is a scalar indexer (line 363), and `create_variables` correctly handles that state at line 209, but `__repr__` still does `self.x_original.sizes["x"]`:

```python
repr(curve.isel(x=0).xindexes["channel"])
# KeyError: 'x'
```

Low severity, but it is on the display path, so a notebook user inspecting `.xindexes` after a positional `x` selection gets a traceback instead of a repr. `self.x_original.sizes.get("x")` clears it (renders `n_x=None`), or drop the field when `x` is gone. `test_repr_pins_exact_format` only covers the unreduced index.

### One observation, not a blocker

`create_variables` passing the caller's variable through means `attrs` now survive the rescale, so a coordinate carrying `units="scaled"` keeps that attr after `sel(channel="TV")` converts it to spend. That is the behaviour I asked for in point 3 and I would not change it here, but if `sample_saturation_curve` ever tags the scaled `x` with units, the label will be wrong in the original domain.

Everything else looks good to me, and the groupby scaled-`x` divergence is fine deferred to #2740 now that it is documented in *Known limitations*. Happy to push the two fixes above if you would rather not touch it again.

juanitorduz:
## Review at `762e1a6`

Probed the branch in a worktree against the repo's pinned stack (xarray 2026.4.0, pandas 3.0.5, numpy 2.4.6) rather than in isolation, so the plotting path is exercised end-to-end. The PR's own suite is green: **85 passed**.

The core idea works. With the convention this repo actually uses for saturation-style curves — `non_grid_names={x_coord_name}`, as in `fourier.py:672` and `hsgp.py:451` — `plot_curve` on a curve carrying `OriginalScaleIndex` produces genuinely per-channel spend axes:

```
subplot 'channel=TV':    x data range 0 .. 5000
subplot 'channel=Radio': x data range 0 .. 1200
```

That is the payoff of #2740 and it lands. @daimon-pymclabs's blockers 1-3 are closed as described. Two findings below are new; two are confirmations that the last round's open items are still open at this commit.

---

### Blocker A (new): `channel_scale` is never aligned to the data's own coord labels, so a reordered scale silently returns the wrong channel's data

`from_variables` receives the `channel` variable in `variables` but ignores it (line 188 uses only `x_var` and `options["channel_scale"]`). `_from_x_and_scale` then builds `x_original = x_da * channel_scale`, and xarray's multiplication aligns on labels — so `x_original`'s channel order is **`channel_scale`'s** order. The positional `dim_indexers` that `sel` returns are computed in that order, but xarray applies them to the *data*, which has its own order. Nothing checks the two agree.

Data rows encoded so mis-selection is visible (TV row = 1.0, Radio row = 2.0), data order `['TV','Radio']`, scale order `['Radio','TV']`:

```python
curve.sel(channel="TV")
# <xarray.DataArray (x: 3)>
# array([2., 2., 2.])            <- Radio's data
# Coordinates:
#   * x        (x) float64 0.0 2.5e+03 5e+03   <- TV's spend axis
#   * channel  <U5 'Radio'
```

Asked for TV, got Radio's data on TV's spend axis. No exception. The result is *internally* plausible — a `'Radio'`-labelled curve — so the tell is subtle: the axis silently belongs to a different channel. The same construction with default indexes returns `[1., 1., 1.]`.

This matters more than an exotic-input bug because label-order independence is xarray's baseline contract — `x_da * channel_scale` aligns correctly, which is precisely why the index's internal order can drift from the data's. `OriginalScaleIndex` is exported from `pymc_marketing.mmm`, so users will build one from a scaler whose channel order need not match the curve's.

Related, same root cause: a `channel_scale` carrying a label the data lacks raises `IndexError: index out of bounds` on `sel`, and one *missing* a label the data has builds fine and fails later.

Aligning in `from_variables` fixes all three:

```python
for dim, var in variables.items():
    if dim == "x" or dim not in channel_scale.dims:
        continue
    data_labels = pd.Index(np.asarray(var.values))
    scale_labels = pd.Index(np.asarray(channel_scale.coords[dim].values))
    if data_labels.equals(scale_labels):
        continue
    missing = data_labels.difference(scale_labels)
    if len(missing):
        raise ValueError(
            f"channel_scale is missing {dim} labels {list(missing)}; "
            f"it must cover every label on the data ({list(data_labels)})."
        )
    channel_scale = channel_scale.sel({dim: data_labels})
```

Applied: the reordered case returns `[1., 1., 1.]` with max x `5000` for TV and `[2., 2., 2.]` with max x `1200` for Radio; a surplus label is dropped and an absent one gives a clean `KeyError`; a missing one raises the `ValueError` at construction.

### Blocker B (new): the `drop_scalar_coords` fix is unreachable for combined-sample input, because `_unstack_sample` three lines below still calls `.indexes`

`plot.py:100` is `curve.indexes.get("sample")` — the same `to_pandas_indexes()` trap this PR fixes at line 82. It is called from `plot.py:543`, `:624` and `:802`, and at `:802` it runs *immediately before* `drop_scalar_coords(curve)` at `:803`. So for any curve with a `sample` dim, the old line raises before the fixed one is reached:

```python
combined = az.extract(curve.to_dataset(name="c"), var_names="c", combined=True)
plot_curve(combined, {"x"})
# TypeError: OriginalScaleIndex(scale_dims=['channel'], n_x=6) cannot be cast to a pandas.Index object
```

Same for `plot_hdi` and `plot_samples`. `arviz.extract(..., combined=True)` output is a supported input — `tests/test_plot.py::test_plot_functions_combined_sample` covers it — so this is half the plotting surface.

Worth flagging that the obvious one-word fix is wrong here: `xindexes` returns a `PandasMultiIndex`, not a `pd.MultiIndex`, so the `isinstance` on the next line goes `False` and the unstack **silently stops happening for every curve, including plain ones**:

```
naive .xindexes:  plain combined -> unstack triggered: False   <- silent regression
```

Unwrapping keeps both paths working:

```python
xindex = curve.xindexes.get("sample")
index = xindex.index if isinstance(xindex, PandasMultiIndex) else None
```

```
proposed:  plain combined -> True    custom-index combined -> True
```

With that in, all three plot functions succeed on combined input and the 85 tests still pass.

### Confirmations from the last round

Both of @daimon-pymclabs's re-review items still reproduce at `762e1a6` on xarray 2026.4.0 — flagging only that they are open, the diagnosis there is correct and I won't restate it.

**The `x` slice/list twin.** Still silent:

```python
curve.sel(channel=["TV", "Radio"], x=slice(0, 2500))
# sizes {'channel': 2, 'x': 5}; x coords [0., 0.25, 0.5, 0.75, 1.]  -- unfiltered and still scaled
```

Also `channel=["TV"]`, `channel=slice("TV","Radio")`, and `x=[0.0, 2500.0]`. The scalar guard at line 272 fires before `remaining` exists, and the `if remaining:` early return at 295 skips the `x_labels` entirely. The docstring at lines 72-77 promises the opposite. Moving the refusal below `remaining` (exactly as proposed) makes all four raise, and the suite stays green.

Worth noting `test_sel_x_with_partially_resolved_scale_dims_raises` passes `x=2500.0` — a scalar — so despite the name it exercises the line-272 guard, not this path.

**`__repr__` after `isel` reduces `x`.** `repr(curve.isel(x=0).xindexes["channel"])` → `KeyError: 'x'`. `sizes.get("x")` clears it (renders `n_x=None`).

### Test coverage

The telling result: **all four fixes above applied together leave all 85 tests passing.** Nothing in the suite pins any of the broken behaviour, so the gaps are exactly load-bearing. Worth adding, one each: a reordered `channel_scale` asserting `sel(channel="TV")` returns TV's *data* (not just TV's x coord — that part is already right and is what makes the bug invisible); `plot_curve` on `az.extract(..., combined=True)` output; a slice/list `x` with an unresolved scale dim; and `repr` after `isel(x=0)`.

### Risk

Unchanged from the last round for existing users — nothing in `pymc_marketing/` constructs an `OriginalScaleIndex` yet, so only the `plot.py` line reaches them today, and it is safe. Blockers A and B both land on whoever wires this into `sample_saturation_curve`, and A is the one I would hold the merge for, since it is the only finding that returns wrong *data* rather than a wrong coordinate or an exception.

Happy to push any of these if useful — the four fixes are small and I have them verified against the suite.

PR #2915, opened 2026-08-24
fix(docs): add persistent stale-version banner and fix pydata-sphinx-theme CSS collision

### LINKED ISSUE #942: Add "stale documentation" banner
Notify when version of the API docs being viewed are stale

williambdean:
## Summary

RTD's native notification toasts are hidden by pydata-sphinx-theme's `.toast:not(.show) { display: none }` CSS rule. This PR adds:

1. **CSS override** in `custom.css` to unhide native RTD toasts for new builds
2. **Custom Script addon JS** (`rtd-stale-banner.js`) that renders a persistent banner on non-stable versions, including all frozen builds that cannot be rebuilt

## Banner behavior

| Version | Style | Message |
|---------|-------|---------|
| Old tagged (e.g. 0.19.0) | Yellow warning | "You are viewing docs for version X, which may be outdated." |
| Dev (latest) | Blue info | "This is the development version of the documentation." |
| Stable | No banner | — |

- Links are **page-specific** (rewrites version segment to `/stable/`)
- Works across translations (`/en/`, `/es/`)
- Dismiss button persists for **30 days** per version in localStorage

## Root cause

The RTD notification addon sets `class="raised toast"` on its `<readthedocs-notification>` host element. Bootstrap's CSS (bundled in pydata-sphinx-theme) selects `.toast:not(.show)` and hides it. This is tracked in [readthedocs/addons#413](https://github.com/readthedocs/addons/issues/413) — open since October 2024 with no fix merged.

## RTD dashboard setup (required after merge)

Enable the Custom Script addon:
1. Go to Settings → Addons → Custom Script
2. Enter URL: `/en/latest/_static/rtd-stale-banner.js`

## Verification

- [x] `/en/0.19.0/...` → yellow banner, links to `/en/stable/...`
- [x] `/en/latest/` → blue banner, links to `/en/stable/`
- [x] `/en/stable/` → no banner
- [x] `/es/0.19.0/...` → yellow banner, links to `/es/stable/...`
- [x] Dismiss persists across page reloads (30-day localStorage)

Closes #942

----
📚 Documentation preview 📚: https://pymc-marketing--2915.org.readthedocs.build/en/2915/


juanitorduz (approved):

OriolAbril:
I am reviewing theme related things (just did a release for `labs-sphinx-theme` in fact which maybe I should have paused until going all related changes like this one but doing more is quite easy so).

The idea is great, in fact, this is something both readthedocs and the pydata-sphinx-theme (if using the theme versioning) do. The problem is we use rtd versioning and the theme kills most (all) of the newish add-ons (relevant upstream issue: https://github.com/pydata/pydata-sphinx-theme/issues/1933). However, it doesn't seem to be working, I see the built-in readthedocs pop up in `/latest/` but the styling doesn't match the description and I see nothing on older versions.

At least for `latest` the most probable reason is the javascript file is not doing anything as it looks like it wasn't added to the `html_js_files` list in `conf.py`. I am not sure this will fix anything with the older versions though as the js file won't be loaded or accessible for them.

Even the css alone I think is a critical fix that should go upstream instead of being restricted to pymc-marketing repo. I think ideally the pydata-sphinx-theme should not be adding this css into the theme (I can open an issue there to discuss that) and instead add that for projects that use the theme versioning (or document how to add it in the theme version switcher docs). Alternatively, the `labs-sphinx-theme` should be the place to store the fix for causalpy to also automatically benefit from it.

For the older versions, if we have multiple of these fixes to address: the version banner, canonical links and SEO stuff... we can also look at rebuilding them by publishing patch versions from the existing releases that only do doc updates (ideally only updating the labs-sphinx-theme version). It might be a bit time consuming but hopefully something easy to automate. RTD stores all the build logs, from https://app.readthedocs.org/projects/pymc-marketing/ one can click on each of the versions to see the logs of the build, which include installing dependencies so we can also try to reproduce a forzen build and instead add post processing scripts to inject the js script or clean-up the html metadata.

PR #2924, opened 2026-08-26
fix(deps): drop the matplotlib<3.11 upper bound

### LINKED ISSUE #2923: Drop the matplotlib<3.11 upper bound: the preliz chain that required it has cleared
`pyproject.toml:33` still carries `matplotlib>=3.5.1,<3.11`. The constraint that justified it is gone, so the bound now only holds the project back from matplotlib 3.11.x.

## Why it was added

#2626 added it (merged 2026-06-12, closing #2625). matplotlib 3.11.0 stopped auto-importing the deprecated `matplotlib.style.core` shim, and consumers in this dependency set touched it at import, so CI died with `AttributeError: module 'matplotlib.style' has no attribute 'core'`. The chain was `matplotlib<3.11` held by `preliz<0.27`, held by `pymc-extras<0.14`, held by `pymc<6.1`.

## Why it can go now

Every link has cleared on `main`:

- `pyproject.toml` pins `pymc>=6.2.0,<6.3.0` and `pymc-extras>=0.14.0,<0.15.0`.
- pymc-extras 0.14.0 itself requires `preliz>=0.27.0,<0.28.0`, so preliz cannot resolve below 0.27 regardless of the looser declared floor.
- preliz 0.27.0 (2026-06-18) carries the fix.
- `arviz-plots[matplotlib]>=1.2.0,<2.0` covers the other consumer that touched the shim.

matplotlib is at 3.11.1.

## One thing to check when removing it

The declared floor at `pyproject.toml:41` is `preliz>=0.20.0,<1.0`, looser than what pymc-extras forces. The oldest-deps CI shard resolves the lowest allowed versions, and that shard is where this pin broke things before. Worth proving the resolver cannot pick preliz 0.26 there, and raising the declared floor to `>=0.27` in the same change if it can.

An earlier measurement in a local venv with the bound removed: matplotlib 3.11.1 with preliz 0.26.0 gave 8 failures in `tests/mmm` (`test_hsgp`, `test_mmm`, `test_tvp`), all the `matplotlib.style.core` `AttributeError` raised through `create_constrained_inverse_gamma_prior` -> `pymc_extras.prior.constrain` -> `from preliz import maxent`, with 3020 passing; forcing preliz 0.27.1 made those 8 pass. That predates the current pins and should be re-run against `main`.

anevolbap:
Closes #2923

`matplotlib>=3.5.1,<3.11` came in with #2626 because matplotlib 3.11.0 stopped auto-importing the deprecated `matplotlib.style.core` submodule, so the attribute access that preliz below 0.27 relied on started raising `AttributeError`. The submodule itself is still importable explicitly and is scheduled for removal in 3.13. The chain was `matplotlib<3.11` held by `preliz<0.27`, held by `pymc-extras<0.14`, held by `pymc<6.1`.

All of it has cleared on `main`: `pymc-extras>=0.14.0` is pinned, and pymc-extras 0.14.0 itself requires `preliz>=0.27.0,<0.28.0`, so preliz cannot resolve below the fixed release. `arviz-plots[matplotlib]>=1.2.0` covers the other consumer.

Verified locally under matplotlib 3.11.1 and preliz 0.27.1: `from preliz import maxent`, the import that used to raise `AttributeError: module 'matplotlib.style' has no attribute 'core'`, now succeeds, and `matplotlib.style` no longer exposes `core` as an attribute. So the access pattern that broke is still broken, and nothing in the resolved set reaches for it any more. `arviz_plots` 1.2.0 and `arviz` 1.2.0 also import clean.

One thing worth a look in review: removing the bound alone does not move the lock. matplotlib stays at the already-locked 3.10.9, since that still satisfies `>=3.5.1`. The relock here used `--upgrade-package matplotlib`, which is what actually takes it to 3.11.1. The marker changes elsewhere in `uv.lock` come from that upgrade, not from lock churn: relocking pristine `main` produces an empty diff.

The declared floor `preliz>=0.20.0,<1.0` is looser than what pymc-extras forces. I left it alone since pymc-extras has a single release in the pinned `0.14.x` range and it hard-requires `preliz>=0.27.0,<0.28.0`, so 0.26 is unreachable, but tightening it to `>=0.27` would make the requirement self-evident rather than transitive.

----
📚 Documentation preview 📚: https://pymc-marketing--2924.org.readthedocs.build/en/2924/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     # the arviz meta-package to pull it in.
     "arviz-stats>=1.2.0,<2.0",
     "arviz>=1.2.0,<2.0",
-    "matplotlib>=3.5.1,<3.11",
+    "matplotlib>=3.5.1",
     "narwhals",
     # numba<=0.66 (a hard requirement of pytensor 3.x) does not support numpy 2.5,
     # and uv>=0.12 otherwise falls back to a numba pre-release. Lift together with
--- a/uv.lock
+++ b/uv.lock
@@ -1601,7 +1601,7 @@ name = "gunicorn"
 version = "26.0.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "packaging" },
+    { name = "packaging", marker = "sys_platform != 'win32'" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/6d/b7/a4a3f632f823e432ce6bc65f62961b7980c898c77f075a2f7118cb3846fe/gunicorn-26.0.0.tar.gz", hash = "sha256:ca9346f85e3a4aeeb64d491045c16b9a35647abd37ea15efe53080eb8b090baf", size = 727286, upload-time = "2026-05-05T06:38:25.529Z" }
 wheels = [
@@ -2360,7 +2360,7 @@ wheels = [
 
 [[package]]
 name = "matplotlib"
-version = "3.10.9"
+version = "3.11.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "contourpy" },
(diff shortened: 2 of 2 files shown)


juanitorduz (approved):
Thanks @anevolbap

PR #2929, opened 2026-08-28
fix(r2d2): register Fourier coord before the R2D2 decomposition builds

### LINKED ISSUE #2927: BUG: R2D2 with both `control` and `fourier` components raises `KeyError: 'fourier_mode'` inside `MMM`
## Summary

The `R2D2` prior added in #2833 documents a headline use case — one variance budget shared between the control block and the yearly seasonality — but that configuration cannot be built. It raises `KeyError: 'fourier_mode'`.

The example is advertised in two places in the `R2D2` docstring ([r2d2.py#L62-L75](https://github.com/pymc-labs/pymc-marketing/blob/66d575da/pymc_marketing/r2d2.py#L62-L75) and [r2d2.py#L316-L329](https://github.com/pymc-labs/pymc-marketing/blob/66d575da/pymc_marketing/r2d2.py#L316-L329)) and is not covered by any test. The one MMM integration test, `test_r2d2_with_mmm_build_model`, deliberately sidesteps it with the comment `# Use R2D2 only for control components (not fourier)`.

## Minimal reproducible example

Self-contained; the only additions to the docstring example are the data plumbing and `yearly_seasonality=1` (without the latter no Fourier term is created at all and `gamma_fourier` is never consumed).

```python
import numpy as np
import pandas as pd

from pymc_marketing.mmm.components.adstock import GeometricAdstock
from pymc_marketing.mmm.components.saturation import LogisticSaturation
from pymc_marketing.r2d2 import R2D2

r2d2 = R2D2(
    r2=Prior("Beta", mu=0.8, sigma=0.2),
    total_sigma=Prior("LogNormal", mu=0, sigma=1),
    dims={"control": "control", "fourier": "fourier_mode"},
)

mmm = MMM(
    date_column="date",
    channel_columns=["x1"],
    control_columns=["c1"],
    target_column="y",
    yearly_seasonality=1,
    model_config={
        "likelihood": Prior("Normal", sigma=r2d2.error_sigma, dims="date"),
        "gamma_control": r2d2.split("control"),
        "gamma_fourier": r2d2.split("fourier"),
    },
)

n = 60
rng = np.random.default_rng(0)
X = pd.DataFrame(
    {
        "date": pd.date_range("2021-01-04", freq="W-MON", periods=n),
        "x1": rng.random(n),
        "c1": rng.normal(size=n),
    }
)
y = pd.Series(rng.random(n))

mmm.build_model(X, y)
```

Note that at least one control column is required to trigger this. With `dims={"fourier": "fourier_mode"}` alone the model builds fine, because `FourierBase.apply` registers the coord immediately before creating the variable.

### Actual

```
  File "mre.py", line 42, in <module>
    mmm.build_model(X, y)
    ~~~~~~~~~~~~~~~^^^^^^
  File "pymc_marketing/mmm/mmm.py", line 2410, in build_model
    gamma_control = self.model_config["gamma_control"].create_variable(
        name="gamma_control", xdist=True
    )
  File "pymc_marketing/r2d2.py", line 175, in create_variable
    self.decomposition.create_variable("r2d2")
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "pymc_marketing/r2d2.py", line 417, in create_variable
    dim_sizes[comp] = int(model.dim_lengths[dim].eval())
                          ~~~~~~~~~~~~~~~~~^^^^^
KeyError: 'fourier_mode'
```

### Expected

The model builds, with a single Dirichlet spanning every coefficient across both components — `('c1', 'sin_1', 'cos_1')` in this example.

## Root cause

An ordering conflict between when the decomposition needs its dim lengths and when `MMM` registers the Fourier coord.

`R2D2.create_variable` needs *every* component's dim length up front, because the Dirichlet spans all components at once ([r2d2.py#L414-L427](https://github.com/pymc-labs/pymc-marketing/blob/66d575da/pymc_marketing/r2d2.py#L414-L427)). It fires lazily on the first `split(...).create_variable(...)` call.

In `MMM.build_model`, that first call is `gamma_control` at [mmm.py#L2410](https://github.com/pymc-labs/pymc-marketing/blob/66d575da/pymc_marketing/mmm/mmm.py#L2410). The `fourier_mode` coord is only registered afterwards, by `FourierBase.apply` calling `model.add_coord(self.prefix, self.nodes)`, reached at [mmm.py#L2434](https://github.com/pymc-labs/pymc-marketing/blob/66d575da/pymc_marketing/mmm/mmm.py#L2434). So the coord does not exist yet when the decomposition asks for its length.

Two workarounds that do **not** help, for the record:

- Listing `fourier` first in `dims` — the dict order only controls how the Dirichlet is sliced, not when dim lengths are looked up.
- Pre-building the decomposition in a separate model context that has both coords — `R2D2Split.create_variable` guards on `self.decomposition._built_model_id != id(model)`, so it rebuilds from scratch inside the MMM's own model and hits the same missing coord. Even if it did not, the variables would belong to the other model.

## Suggested fix

Register the coord alongside the others when `model_coords` is assembled, near [mmm.py#L1809](https://github.com/pymc-labs/pymc-marketing/blob/66d575da/pymc_marketing/mmm/mmm.py#L1809):

```python
if self.yearly_seasonality is not None:
    self.model_coords[self.yearly_fourier.prefix] = self.yearly_fourier.nodes
```

`pm.Model.add_coord` returns early when the coord already exists with equal values, so the later `apply` call becomes a no-op. `self.model_coords` is read in only four places in `mmm.py` (the `pm.Model(coords=...)` call, the `dayofyear` lookup, date-overlap validation, and the time index), none of which iterate it or are sensitive to an extra key. `fourier_mode` already ends up in `model.coords` today, just later.

## Also worth fixing

1. **The docstring examples omit `yearly_seasonality`**, so even after the coord fix they fail as written — no Fourier term is created and `gamma_fourier` goes unused. Both copies need the argument.
2. **The error message is opaque.** A bare `KeyError` on a dim name gives no hint about which component is at fault. Wrapping the lookup in `R2D2.create_variable` to report something like *component `'fourier'` maps to dim `'fourier_mode'`, which is not present in the model; available dims are ...* would make this diagnosable.

## Versions

| | |
| --- | --- |
| pymc-marketing | 1.1.0 (`main` at 66d575da) |
| pymc | 6.2.0 |
| pytensor | 3.2.4 |
| Python | 3.13.13 |

TeemuSailynoja:
Closes #2927

## Summary

`R2D2` advertises sharing one variance budget between an MMM's control block and its yearly Fourier seasonality, via `dims={"control": "control", "fourier": "fourier_mode"}`. That configuration could not be built — it raised `KeyError: 'fourier_mode'`.

`R2D2.create_variable` needs every component's dim length up front, because the Dirichlet spans all components at once. It fires lazily on the first `split(...).create_variable(...)` call, which in `MMM.build_model` is `gamma_control`. But the `fourier_mode` coord was only registered later, by `FourierBase.apply`. So the coord did not exist yet when the decomposition asked for its length.

The fix registers the coord alongside the others in `_generate_and_preprocess_model_data`:

```python
if self.yearly_seasonality is not None:
    self.model_coords[self.yearly_fourier.prefix] = self.yearly_fourier.nodes
```

This is safe because `pm.Model.add_coord` returns early when the coord already exists with equal values, so the later `apply` call becomes a no-op. `fourier_mode` already ended up in `model.coords` before this change — it is now simply available earlier, before `gamma_control` is created. `self.model_coords` is read in only four places in `mmm.py` (the `pm.Model(coords=...)` call, the `dayofyear` lookup, date-overlap validation, and the time index), none of which iterate it or are sensitive to an extra key.

Also fixed: both copies of the MMM integration example in the `R2D2` docstrings omitted `yearly_seasonality`, so even after the coord fix they would create no Fourier term and leave `gamma_fourier` unused.

## Test plan

- [x] New regression test `test_r2d2_with_mmm_control_and_fourier_build_model` asserts the shared Dirichlet spans every coefficient across both components, in order: `["event_1", "event_2", "t", "sin_1", "cos_1"]`. It also checks both `gamma_control` and `gamma_fourier` are free random variables and that `fourier_mode` has length 2.
- [x] Confirmed the test fails without the fix, with `KeyError: 'fourier_mode'` from the `dim_sizes` lookup in `r2d2.py`, and passes with it.
- [x] `uv run pytest tests/test_r2d2.py` — 29 passed
- [x] `uv run pytest tests/mmm/test_mmm.py` — 247 passed. This is the important one for the default path: it covers `yearly_seasonality` of `None` and `4` crossed with the TVP settings over single-dim and multi-dim fixtures, plus `test_fourier_contribution`, `test_yearly_seasonality_contribution` and `test_with_controls_and_seasonality`. A user who just passes a number and no custom prior sees no behavioural change.
- [x] `uv run pytest tests` — 4469 passed, 52 skipped. The 9 failures in `tests/test_mlflow.py` are pre-existing and environmental: they were reproduced identically on clean `main`, and are caused by the Graphviz `dot` binary being absent from the local machine, so `model_graph.pdf` is never logged. CI installs `graphviz` explicitly.
- [x] `uv run pre-commit run --files ...` — all hooks pass, including mypy.

----
📚 Documentation preview 📚: https://pymc-marketing--2929.org.readthedocs.build/en/2929/

### DIFF
--- a/pymc_marketing/mmm/mmm.py
+++ b/pymc_marketing/mmm/mmm.py
@@ -1811,6 +1811,9 @@ class MMM(RegressionModelBuilder):
             for dim in self.xarray_dataset.coords.dims
         }
 
+        if self.yearly_seasonality is not None:
+            self.model_coords[self.yearly_fourier.prefix] = self.yearly_fourier.nodes
+
         if bool(self.time_varying_intercept) or bool(self.time_varying_media):
             dates = pd.DatetimeIndex(self.xarray_dataset.coords["date"].values)
             self._time_index = xr.DataArray(np.arange(len(dates)), dims=("date",))
--- a/pymc_marketing/r2d2.py
+++ b/pymc_marketing/r2d2.py
@@ -67,6 +67,7 @@ MMM integration:
 >>> mmm = MMM(
 ...     adstock=GeometricAdstock(l_max=8),
 ...     saturation=LogisticSaturation(),
+...     yearly_seasonality=1,
 ...     model_config={
 ...         "likelihood": Prior("Normal", sigma=r2d2.error_sigma),
 ...         "gamma_control": r2d2.split("control"),
@@ -321,6 +322,7 @@ class R2D2:
     >>> mmm = MMM(
     ...     adstock=GeometricAdstock(l_max=8),
     ...     saturation=LogisticSaturation(),
+    ...     yearly_seasonality=1,
     ...     model_config={
     ...         "likelihood": Prior("Normal", sigma=r2d2.error_sigma),
     ...         "gamma_control": r2d2.split("control"),
--- a/pymc_marketing/r2d2.py
+++ b/pymc_marketing/r2d2.py
@@ -67,7 +67,7 @@ MMM integration:
 >>> mmm = MMM(
 ...     adstock=GeometricAdstock(l_max=8),
 ...     saturation=LogisticSaturation(),
-...     yearly_seasonality=1,
+...     yearly_seasonality=4,
 ...     model_config={
 ...         "likelihood": Prior("Normal", sigma=r2d2.error_sigma),
 ...         "gamma_control": r2d2.split("control"),
@@ -322,7 +322,7 @@ class R2D2:
     >>> mmm = MMM(
     ...     adstock=GeometricAdstock(l_max=8),
     ...     saturation=LogisticSaturation(),
-    ...     yearly_seasonality=1,
+    ...     yearly_seasonality=4,
     ...     model_config={
     ...         "likelihood": Prior("Normal", sigma=r2d2.error_sigma),
     ...         "gamma_control": r2d2.split("control"),
--- a/tests/test_r2d2.py
+++ b/tests/test_r2d2.py
@@ -647,6 +647,44 @@ class TestR2D2WithMMM:
         # Verify the decomposition was built
         assert r2d2.built
 
+    def test_r2d2_with_mmm_control_and_fourier_build_model(self, mmm_data):
+        """R2D2 should build when sharing budget across control and Fourier."""
+        from pymc_marketing.mmm import MMM
+        from pymc_marketing.mmm.components.adstock import GeometricAdstock
+        from pymc_marketing.mmm.components.saturation import LogisticSaturation
+
+        r2d2 = R2D2(
+            r2=Prior("Beta", mu=0.8, sigma=0.2),
+            total_sigma=Prior("LogNormal", mu=0, sigma=1),
+            dims={"control": "control", "fourier": "fourier_mode"},
+        )
+        mmm = MMM(
+            date_column="date_week",
+            channel_columns=["x1", "x2"],
+            control_columns=["event_1", "event_2", "t"],
+            target_column="y",
+            adstock=GeometricAdstock(l_max=8),
+            saturation=LogisticSaturation(),
+            yearly_seasonality=1,
+            model_config={
+                "likelihood": Prior("Normal", sigma=r2d2.error_sigma, dims="date"),
+                "gamma_control": r2d2.split("control"),
+                "gamma_fourier": r2d2.split("fourier"),
+            },
+        )
+
+        X = mmm_data.drop("y", axis=1)
+        y = mmm_data["y"]
+
+        mmm.build_model(X, y)
+
+        expected_split = ["event_1", "event_2", "t", "sin_1", "cos_1"]
+        assert list(mmm.model.coords["r2d2_split"]) == expected_split
+        free_rv_names = {v.name for v in mmm.model.free_RVs}
+        assert "gamma_control" in free_rv_names
+        assert "gamma_fourier" in free_rv_names
+        assert mmm.model.dim_lengths["fourier_mode"].eval() == 2
+
     def test_r2d2_with_mmm_contribution_names(self, mmm_data):
         """R2D2 should produce variables that work with MMM contribution plotting."""
         from pymc_marketing.mmm import MMM


pymc_marketing/r2d2.py
```diff
@@ -67,6 +67,7 @@
 >>> mmm = MMM(
 ...     adstock=GeometricAdstock(l_max=8),
 ...     saturation=LogisticSaturation(),
+...     yearly_seasonality=1,
```
williambdean: This is in docstring, right? Not sure how 1 is as a recommendation

pymc_marketing/mmm/mmm.py
```diff
@@ -1811,6 +1811,9 @@ def _generate_and_preprocess_model_data(
             for dim in self.xarray_dataset.coords.dims
         }
 
+        if self.yearly_seasonality is not None:
+            self.model_coords[self.yearly_fourier.prefix] = self.yearly_fourier.nodes
```
williambdean, line 1815: Anyway to add into the r2d2 logic / delay there or by using fourier mu effect. Seems a bit too unrelated to mmm implemenation

williambdean (commented):
What are your thoughts?

pymc_marketing/r2d2.py
TeemuSailynoja: Definitely a poorly chosen value from me. Bumping up to 4.

pymc_marketing/mmm/mmm.py
TeemuSailynoja: I thought this fits quite nicely here as we are registering other coordinates here too, and `fourier_mode` is listed as a reserved coordinate in `__init__`: https://github.com/pymc-labs/pymc-marketing/blob/fabba92c96aa6a4ec6d42fb3241a8ed725995d0a/pymc_marketing/mmm/mmm.py#L515
I hadn't realized that `fourier_mode` was never added to `mmm.model_coords`. It does go to `mmm.model.coords` though, so perhaps we could add it when the first entry to model context happens here:
https://github.com/pymc-labs/pymc-marketing/blob/fabba92c96aa6a4ec6d42fb3241a8ed725995d0a/pymc_marketing/mmm/mmm.py#L2293?
 
To have it contained in the R2D2 class, the model scope would need to know the coords for all components when the first of the component splits needs to create a variable. Now I'm hitting the issue with gamma_control here: https://github.com/pymc-labs/pymc-marketing/blob/fabba92c96aa6a4ec6d42fb3241a8ed725995d0a/pymc_marketing/mmm/mmm.py#L2410

It could also happen earlier though, when a time varying intercept is created, as I think we could also use R2D2 to regularize a HSGP.

juanitorduz:
@williambdean @TeemuSailynoja is this one ready ?

williambdean (approved):

PR #2936, opened 2026-08-31
MMM Case Study II: Funnel-Aware Modeling, ROAS Calibration, and Budget Planning

juanitorduz:
## Description

Second iteration of the MMM end-to-end case study (`mmm_case_study`), delivering the follow-ups its next steps promised:

- **Funnel model**: SEM (26.9% of the modeled media spend, deliberately excluded in iteration 1) enters as a lower-funnel lever through a custom `MuEffect`. TV, Radio, and Newspaper build a latent demand pool, SEM harvests it, and SEM impressions anchor the mediator through a second likelihood.
- **ROAS calibration**: experiment ROAS ranges for Online Display and Direct Mail via `add_cost_per_target_calibration(..., target_per_cost=True)`; both posteriors land inside their ranges and the decision-critical uncertainty collapses.
- **Time-slice cross-validation** replaces the train/test split: four quarterly folds with a custom fold builder (per-fold scaling and calibration), stable parameters, no overfitting signature; the final model is fit on all 209 weeks.
- **Funnel-aware budget optimization** of a future 13-week window with `response_variable="total_response_original_scale"` and `spend_vars=["sem_spend"]`, contrasted against the default direct-only objective (they disagree exactly where the funnel is). Share-space re-weighting of the training data shows a 23% media-driven uplift with 99.8% probability of being positive.

Also registers the notebook in the gallery and adds a forward link from the first case study's next steps.

The notebook was executed end-to-end with real MCMC (zero divergences, max R-hat 1.01) and passes the mocked notebook runner (~1 min). Note for reviewers: the notebook CI workflow's path filters exclude `docs/source/notebooks/mmm/**`, so this PR does not trigger it; the local mocked run is the gate.

## Related Issue
- [ ] Closes #
- [ ] Related to #

## Checklist
- [x] Checked that the [pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html)
- [x] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html)

----
📚 Documentation preview 📚: https://pymc-marketing--2936.org.readthedocs.build/en/2936/


daimon-pymclabs (commented):
## Detailed review

I read the notebook end to end (all 103 cells plus the executed outputs), checked the rendered RTD preview, and verified the arithmetic against `data/case_study_data.csv` and the current `budget_optimizer` / `optimization_variables` source. This is a strong notebook: the funnel wiring is correct, the leakage discipline in the CV builder is right, the gradient-decomposition check in the optimization section is a genuinely good idea, and the "Critical Assessment" callouts are honest. Nothing here is blocking. Comments below are ordered by how much they change what a reader takes away.

Things I checked and found fine, so nobody re-checks them: all `{ref}` targets resolve on the preview build (`mmm_funnel_mueffect`, `..._advanced`, `mmm_roas_calibration`, `mmm_time_slice_cross_validation`, `mmm_lift_test`, `mmm_funnel_from_yml`, `mmm_example`, `mmm_case_study`); the `seealso` appended to case study I renders as a directive, not literal text; `gallery.md` and `gallery.yaml` are in sync and the thumbnail path is fine because `docs/source/gallery/images/` is gitignored and generated by `scripts/generate_gallery.py`; the CV fold arithmetic (157 + 4x13 = 209) is right; the experiment dates do sit before fold 0's training window ends (~2017-08), so the leakage note holds; `total_budget` and `budget_bounds` really are per-period levels (`MediaVariable.to_model` repeats `z` across `num_periods`), so the "weekly" labels on the allocation table are correct.

---

### 1. Every media cell in the funnel-aware plan sits on a guardrail (cell 88/89, narrative in cell 90)

In the funnel-aware optimum, all seven channels are at a bound: Direct Mail at its floor (283,234 = 0.5x historical) and the other six at their ceilings (1.5x historical). Conditional on Direct Mail at the floor, the six ceilings sum to 1,197,004, which is exactly `1,480,238 - 283,234`. So the media allocation is a vertex of the +/-50% box, and the only genuinely interior number in the whole plan is the SEM level, whose optimality condition is "marginal SEM return equals marginal Direct Mail return", the two levers not pinned by the box.

That reframes the headline. "Trims the SEM budget by 16% to an interior optimum of about 607K per week" reads as a property of the response surface; it is really the residual left over once six channels hit their ceilings, and it is a function of `percentage_change = 0.5`, which the notebook picks without justification.

Suggestion: add a short sensitivity block re-running `allocate_budget` at `percentage_change` in {0.2, 0.5, 1.0} and tabulating the SEM level and the Direct Mail level. If the SEM cut survives a +/-20% box it is a real finding; if the number swings with the guardrail, say so. One extra paragraph, and it converts the most quotable number in the notebook from an artifact into a result. At minimum, add a sentence to cell 90 naming the corner solution explicitly.

### 2. The funnel-vs-direct comparison confounds the objective with the decision space (cells 84-93)

Cell 93 says "the difference is purely which response paths the objective is allowed to count". It is not. The two runs differ in two ways at once: the objective (`total_response_original_scale` vs `total_media_contribution_original_scale`) *and* the feasible set (the funnel run may move `sem_spend`, the direct run cannot, by library design). The funnel plan ends up with 116K/week more media money because it took it out of SEM, which is why it reaches ceilings the direct plan cannot afford (Online Display 422,301 vs 361,759; Social Media 263,996 vs 241,394). Those two rows are a budget-size effect, not a funnel effect, but cell 90 lists them alongside Newspaper as evidence for the funnel.

Suggestion: add a third run, funnel objective with SEM pinned at `sem_hist` (same media pot as the direct run). Then you get a clean two-way decomposition: run 2 vs run 3 isolates the objective, run 3 vs run 1 isolates the SEM lever. The Newspaper flip (floor vs ceiling) is the one comparison that already isolates the objective correctly, and it would survive this cleanly.

Separately, worth naming that the +10M expected gain is a within-model quantity: the funnel plan is by construction the argmax of the metric it is scored on, so a positive mean is guaranteed. The 75.4% (not 100%) only reflects posterior variation around the mean-utility optimum. Cell 93 is honest about the interval spanning zero, but a reader can easily take "beats the direct-only plan with 75% probability" as validation rather than internal consistency.

Also note the Newspaper flip, which the narrative leads with, is worth about 33K/week on a 2.09M pot, roughly 1.5% of the budget. Fine as a mechanism illustration, but the framing currently implies a material money difference.

### 3. `funnel_lambda` is a spend-to-impressions coefficient, not evidence that SEM drives sales (cells 57, 100)

The conclusion states: "SEM budget moves the mediator: `funnel_lambda` concentrates around 0.68, clearly away from zero." True, but `lambda` is identified almost entirely by the impressions likelihood, where `sem_impressions ~ TruncatedNormal(impressions_scale * (D + lambda * sem_spend / sem_scale), sigma_s)`. "Search spend buys search impressions" is close to tautological and is not the decision-relevant claim. The parameter that carries the mediated path into revenue is `sat_lf_beta` (plus `sat_lf_lam`, `adstock_lf_alpha`), and the notebook plots it in the trace but never reports its posterior or discusses its identification.

Suggestion: report `sat_lf_beta` alongside `funnel_lambda` in the diagnostics section, and say in one sentence what a weak `sat_lf_beta` would have meant. Right now the reader's confidence in the funnel rests on the one parameter that was never in doubt. The mediated-share result in cell 72 (6% / 11% / 25%, intervals reaching zero) is the honest version of the same story and could be foreshadowed here.

Related, and worth a sentence somewhere: because the measurement coefficient is fixed at 1 and `sigma_s` is fitted, `M*` tracks observed `sem_impressions` fairly tightly, so `D` is close to `(impressions - lambda * spend) / scale`. A skeptical reader will ask how this differs from putting `sem_impressions` in as a channel, which is the failure mode cell 1 opens by warning against. The answer is good (the counterfactual perturbs upper spend, which moves `D` and hence `M*`, without re-conditioning on the observed impressions) but it is exactly the point that deserves to be stated, not left implicit.

### 4. `funnel_baseline` and the sales intercept are structurally confounded (cell 20)

`funnel_baseline` is a constant added inside `demand`, and `demand` flows into the sales mean through `sat_lf(adstock_lf(M*))`. A constant shift in `D` produces a near-constant shift in the mediated sales contribution, which competes directly with `intercept`. In this fit the tight `Normal(0.2, 0.05)` intercept prior is doing the work of resolving that ridge. It sampled cleanly here (0 divergences, max R-hat 1.010), so this is not a bug, but it is a fragility a reader copying the pattern will hit with a wider intercept prior.

Suggestion: one sentence in cell 19's "three design points" list acknowledging it, next to the existing (good) note about why the measurement coefficient is fixed. It would also be informative to report how much of the mediator level is baseline vs upper-channel-driven; if baseline dominates, that is the quantitative reason the mediated shares in cell 72 come out small with intervals touching zero.

### 5. No return number is reported for `sem_spend` itself

SEM is 26.9% of media money and the plan cuts it 16%, but the notebook never reports SEM's ROAS or marginal return. `contribution_over_spend` covers the seven channels only, and cell 75's table has no SEM row. Readers will ask why the cut is justified.

Suggestion: report the marginal response per SEM dollar at the historical level (you already have `optimizer._objective_and_grad` at `x0` in cell 87; the last element of `grad_joint` is exactly this), and contrast it with the same quantity for Direct Mail. That single comparison *is* the mechanism behind the 16% cut and would make the recommendation defensible rather than asserted.

### 6. Calibration results are partly true by construction (cell 79)

"Both calibrated posteriors sit inside their experiment ranges" is presented as a finding, but it is the expected outcome when the calibration likelihood is doing its job and the model does not contradict the experiments. The genuine information is the *shape*: whether the posterior sits at the center of the range (model and experiment agree) or is pulled to an edge (they were fighting). Online Display at 27.8 vs a midpoint of 28.0 is a real agreement signal, and Direct Mail at 4.82 vs 5.0 likewise. Worth saying so, and worth stating what a failure would have looked like.

Cell 36's caveat about calibrated channels being stable "by construction" already gets this right, so it is a one-line consistency fix.

### 7. The scale plausibility check is imprecise (cell 25)

"Total sales are about 22.6B against 355M of direct channel spend, so if media drive roughly a third of sales, the blended direct ROAS is in the low tens." I recomputed from the CSV: sales 22.58B, direct spend 355.4M, and 22.58B / 3 / 355.4M = 21.2. The fitted result agrees: applying the cell 75 posterior means to actual per-channel spend gives an implied media contribution of 6.98B, a 30.9% media share and a spend-weighted ROAS of 19.6. So the check works nicely, but the number is "around 20", not "low tens". Since this sentence is what licenses the 20-36 and 2-8 experiment ranges, it is worth getting right.

### 8. Hard `raise` on stochastic quantities will break future doc builds

Cells 54 (`n_divergences != 0`), 68 (`indirect.mean() <= 0`), and 87 (`grad_diff[upper_index] < 0`) raise `RuntimeError`. A future PyMC/nutpie/pytensor bump that produces a single divergence takes down the docs build for a reason unrelated to this notebook. Cell 30 (`n_splits != 4`) and the gradient-decomposition check in cell 87 are deterministic and fine to keep as hard assertions; the divergence and posterior-sign checks are not.

Suggestion: `warnings.warn(...)` plus a printed value for the stochastic ones, keep the `raise` for the structural ones. The `np.testing.assert_allclose` budget-sum checks are deterministic and fine as is.

### 9. `FunnelMMM` silently degrades on the exact failure mode cell 22 warns about

```python
present = [var for var in funnel_vars if var in X.columns]
```

If a user calls a prediction path with a frame that is missing `sem_spend` / `sem_impressions`, `present` is empty, the funnel data stays frozen at training values, and nothing complains. That is precisely the silent failure cell 22 spends a paragraph explaining. Since the class exists to prevent it, it should enforce it:

```python
missing = [var for var in funnel_vars if var not in X.columns]
if missing:
    raise ValueError(f"funnel columns missing from prediction frame: {missing}")
```

### 10. Private API surface in a flagship notebook

The notebook depends on `optimizer._objective_and_grad`, `optimizer.optimization_variables.pack`, overriding `MMM._posterior_predictive_data_transformation`, `mmm.mu_effects[0]`, and `mmm.channel_data_scaled`. Readers copy flagship notebooks verbatim, so this pins us to those internals or breaks users on the next refactor.

Suggestions: (a) `mmm.mu_effects[0]` in cell 61 should look the effect up by prefix rather than by index, which is a two-line fix and removes a real ordering hazard; (b) for the optimizer internals, either add a public accessor (happy to open a separate library issue for `objective_and_grad` / `pack`, since the gradient-decomposition check in cell 87 is a pattern worth making first-class), or add a short admonition saying these are internals used for a didactic check and are not part of the stable API. Same for the `save`/load limitation you already document in cell 22, which is worth a tracking issue rather than only a notebook caveat.

### 11. Smaller items

- Cell 18 computes `spend_shares` on the full data and stores it in `model_config`, then `make_mmm` overwrites `saturation_beta` with a per-fold version. The dead assignment makes the leakage story harder to verify than it needs to be. Drop it from cell 18 and let `make_mmm` own it.
- Cell 70 uses `direct_total` (full posterior contribution) as the denominator of `mediated_share`, while cell 76 uses `direct_deltas` (the zeroing counterfactual) for the same channels. They agree numerically because `saturation(0) = 0`, but using one definition throughout would be cleaner.
- Cell 98's `sales_uplift` is a ratio of two *independent* posterior-predictive draws of `y_original_scale`, so likelihood noise enters numerator and denominator uncorrelated and the plotted distribution is wider than the counterfactual uncertainty. The mean (9.5%) is unaffected. `media_uplift` in cell 96 uses deterministics and is clean. Consider using the expected value rather than the predictive draw, or noting the inflation in the caption.
- `sample_plan` consumes the shared `rng` sequentially, so re-running a single cell out of order changes results. Worth a fixed integer seed per call for a docs notebook.
- Cell 5 has `%load_ext autoreload` / `%autoreload 2`, which is development scaffolding that should not ship in a published notebook. Cell 56 mutates `az.rcParams["plot.max_subplots"]` globally without restoring it.
- Three different titles for the same document: PR title "Add MMM case study II: funnel-aware modeling, ROAS calibration, and budget planning", notebook H1 "MMM Case Study, Second Iteration: Funnel Effects, ROAS Calibration, and Budget Planning", gallery card "MMM Case Study II: Funnel-Aware Modeling, ROAS Calibration and Budget Planning". Pick one.
- `l_max_lf = 6` is justified only by the comment "demand takes time to convert". Since it sets the extra carryover the whole `incrementality_spec` / optimizer window arithmetic depends on, one sentence on why 6 (or a note that the results are insensitive to it) would help.
- The CI note in the PR description (path filters exclude `docs/source/notebooks/mmm/**`, so the notebook workflow does not run) is useful and should probably become a follow-up issue rather than living only in a PR body.

juanitorduz:
All 11 items addressed in 21ceaf6e. Point by point:

1. **Corner solution / guardrail sensitivity.** The corner solution is now named explicitly, and a new sensitivity block re-solves the plan at +/-20%, +/-50%, and +/-100% guardrails (reusing the compiled optimizer via `set_constraints`, so the cost is two extra SLSQP solves). The SEM cut survives every box and deepens with it: -6% at +/-20%, -16% at +/-50%, -32% at +/-100%; Direct Mail sits at its floor in every box and all media cells are on a bound in every solve. The prose now reports the direction as a property of the model and the magnitudes as properties of the box.

2. **Objective confounded with the decision space.** Added the third run you proposed: funnel objective with SEM pinned at its historical level via an equality constraint, which forces exactly the direct run's media pot. The experiment overturned the expectation in the review (and in my own draft of the fix): the pinned plan is numerically identical to the direct-only plan (differences of a few dollars, solver tolerance), so at a fixed pot the two objectives choose the same corner, and the mediated marginals raise the upper channels' returns without reordering them. The Newspaper flip is therefore a budget-size effect, not an objective effect. The gain decomposition in the notebook now attributes essentially zero to the objective and the full ~10.0M to the SEM lever, and the prose states that a positive expected gain is by construction, with the 75% reflecting posterior variation rather than validation.

3. **`funnel_lambda` vs `sat_lf_beta`.** Diagnostics now print both: `funnel_lambda` 0.67 (89% interval 0.60 to 0.74) labeled as the near-tautological spend-to-impressions link, `sat_lf_beta` 0.27 (0.03 to 0.62) as the parameter that carries the mediated path into sales, positive-leaning but wide, with the connection to the imprecise mediated shares made explicit. The measurement-check section gains the impressions-as-channel distinction (the counterfactual moves latent demand without re-conditioning on observed impressions). The conclusion is reworded accordingly.

4. **`funnel_baseline` / intercept confound.** Added as a fourth design point, plus a printed mediator-level decomposition: 40% baseline, 10% upper funnel media, 49% SEM term, linked to why the mediated shares are small.

5. **SEM return number.** The gradient at the historical plan is now a labeled marginal-return table for all eight lines: SEM ~159 per weekly dollar against Direct Mail 61, Insert 760, TV 631. The prose presents this comparison as the mechanism behind the trim.

6. **Calibration by construction.** Rewritten around position rather than containment: both posteriors sit at their range centers (Online Display 27.9 vs midpoint 28.0; Direct Mail 4.8 vs 5.0), with edge-pull named as what failure would have looked like, and Insert (4.0 to 157.7) and Radio (3.3 to 110.9) as the uncalibrated contrast.

7. **Scale check.** "Low tens" replaced by "around 20" with the arithmetic shown (22.58B / 3 / 355.4M is about 21).

8. **Stochastic raises.** The final-fit divergence check and the mediated-reachability guard are now warnings with printed values; the CV divergence cell gained a warning it did not have. The deterministic guards (fold count, gradient decomposition, budget-sum allclose) stay hard, as suggested.

9. **`FunnelMMM` silent degradation.** The override now raises `ValueError` naming the missing funnel columns. Verified no internal path in the notebook's flow passes a column-less DataFrame (optimization models arrive as Datasets, `contribution_over_spend` intervenes on the graph, the plot suite reads idata). The Dataset-side gap (`set_data` silently skipping absent variables) is noted in #2938.

10. **Private API surface.** `mmm.mu_effects[0]` replaced by a prefix lookup. Added an admonition naming the internals used by the didactic checks, linked to the new tracking issues: #2937 (public plan-evaluation API) and #2938 (funnel save/load round-trip). The CI path-filter note became #2939.

11. **Smaller items.** Dead `spend_shares`/`model_config` assignment removed (`make_model_config(X)` owns it, per-fold locality easier to verify); mediated-share denominator unified on the zeroing counterfactual; `sample_plan` and `contribution_over_spend` now take fixed seeds for out-of-order reproducibility, which also means the sales-uplift ratio is scored on paired noise draws, so the caption now says the spread mostly reflects parameter uncertainty (the width-inflation reading applied to the old seeding); autoreload magics removed; `plot.max_subplots` restored after the trace; the title is unified across the notebook H1, the gallery card, and the PR; `l_max_lf = 6` carries its justification.

Not changed: the save/load limitation stays a notebook caveat with #2938 tracking it, and the notebook keeps `_objective_and_grad`/`pack` for the didactic checks until #2937 lands.

daimon-pymclabs (commented):
## Review round 2

Re-read the notebook against commit `21ceaf6`, focusing on the 13 new cells and the re-executed outputs, and re-verified the numbers that changed with the re-run. Every item from the first round is addressed, and three of the fixes changed the notebook's own conclusions in the right direction. The pinned-SEM run is the best thing in this commit: it turned a claim I doubted into a clean negative result, and the notebook now says out loud that the funnel's planning value is pricing SEM rather than reordering media. Follow-up issues #2937, #2938 and #2939 are filed and linked.

Three findings below, then nits. Nothing blocking.

---

### 1. The marginal-response table is in per-weekly-dollar units, and the prose reads it as per-dollar (cells 96, 97)

`-grad_joint` is the derivative of the windowed objective with respect to a **weekly spend level**, and `MediaVariable.to_model` repeats that level across all 13 periods of the window (`optimization_variables.py:484-498`). So each entry is the total quarter response bought by one extra dollar per week sustained for 13 weeks, not by one extra dollar.

The check is in the notebook already. Divide the column by 13 and it reproduces the cell 80 ROAS table:

| channel | marginal (cell 96) | /13 | ROAS mean (cell 80) |
|---|---|---|---|
| Direct Mail | 61.3 | 4.7 | 4.75 |
| Online Display | 306.3 | 23.6 | 27.94 |
| TV | 631.0 | 48.5 | 47.66 |
| Insert | 760.2 | 58.5 | 54.35 |
| Newspaper | 208.0 | 16.0 | 14.37 |

All seven land within 15% of the average ROAS, which is what you would expect for marginal-vs-average near the historical operating point. Without the divide they sit 13x above a table three cells earlier in the same notebook, with no explanation, and cell 97's "a marginal SEM dollar returns about 159 in response" is the sentence a reader will quote.

The ratio arguments in cell 97 and cell 100 are all unaffected, so this is a units sentence, not a rework. Suggestion: one clause after the table saying these are per sustained weekly dollar across the 13-week window, divide by 13 to compare with the ROAS table. That also gives you the number the section is really after: **SEM's marginal return is about 12 per dollar, between Direct Mail's 4.7 and Online Display's 24.** That single line answers "why trim SEM" better than the current paragraph does, and it is directly comparable to the calibrated ROAS numbers the stakeholder already has.

### 2. SEM is the arithmetic residual in every box, so the sensitivity table is one observation, not three (cells 105, 106)

In all three solves every media cell is at a bound, so the media total is fully determined by the box and SEM is whatever is left of the 2,086,890 pot:

| box | media at bounds | implied SEM | reported SEM |
|---|---|---|---|
| ±20% | 453,175 floor + 957,603 ceilings = 1,410,778 | 676,112 | 676,112 |
| ±50% | 283,234 + 1,197,004 = 1,480,238 | 606,652 | 606,652 |
| ±100% | 0 + 1,596,004 = 1,596,004 | 490,886 | 490,885 |

Matches to the dollar. So the "cut deepens as the box widens" pattern is mechanical: a wider box means more media ceiling to fund, and SEM pays for all of it. It is not three independent looks at SEM's value.

There is still a real robustness result here, and it is the one worth stating: in every box the optimizer picks the same corner and leaves SEM as the residual, meaning SEM's marginal return sits below every media channel's over the whole range tested. That is a statement about the ranking, and it survives. The -6% / -16% / -32% magnitudes are the box, and cell 106 half-says this already ("magnitudes are properties of the box"); the fix is to make the surviving claim the ranking rather than the cut.

Related and worth one line: **the SEM band never binds.** SEM lands interior in all three solves (676,112 inside [577,935, 866,903]; 606,652 inside [361,209, 1,083,628]; 490,885 inside [0, 1,444,838]). The custom `sem_lower`/`sem_upper` constraint machinery in cell 87 is inactive at every reported optimum. That is fine, and the machinery is still the right thing to demonstrate, but a reader who works through those two `Constraint` objects deserves to know they are not what produced the answer.

A sensitivity that would move independently: hold the media box at ±50% and vary the SEM band alone, or vary the total pot. Both leave the media corner in place and probe the one number that is actually interior.

### 3. `gain_from_objective` prints as `-0` (cell 102)

The decomposition is the most interesting output in the section and it renders as `gain from the objective alone (pinned vs direct): -0`, which reads like a formatting bug rather than the finding. Print it as a share of the total gain, or with enough precision to show it is solver noise rather than a rounding artifact:

```python
print(
    f"gain from the objective alone (pinned vs direct): {gain_from_objective.mean():,.0f} "
    f"({gain_from_objective.mean() / response_gain.mean():.2%} of the total)"
)
```

### Nits

- Cell 59 says `funnel_lambda` "concentrates around 0.68" and then, two sentences later, "is tight (0.67, 89% interval 0.60 to 0.74)". The table says 0.67 and the conclusion says 0.67. Drop the 0.68.
- Cell 100 says the pinned and direct columns differ by "a few dollars". They differ by 21.18 on Online Display and -21.18 on Social Media, offsetting exactly, which is a nicer tell than "a few dollars": the objective is flat between two channels it values identically. The `gain_from_objective` print is the real proof and it is already there.
- Cell 85's admonition lists `set_constraints` among library internals. It is public (`budget_optimizer.py:1810`, documented, no underscore) and is the intended way to swap constraint sets, which is exactly what cells 92 and 105 use it for. The other four in that list are correctly flagged.
- Cell 100's "its marginal of 208 per dollar is the lowest after Direct Mail" is true among the media lines but not among all lines in the table, since SEM sits at 158.8. Adding "media line" removes the ambiguity.
- Cell 76 reports the mediator level as 40% baseline / 10% upper funnel / 49% SEM term. Worth naming that the 49% is the direct `lambda * B` term at historical SEM spend, so half the pool SEM is said to harvest is SEM's own spend. That is exactly what the DAG in cell 2 draws, and stating it closes the loop with the "how is this different from putting impressions in as a channel" answer you added to cell 64.
- Title now matches the gallery card except for the Oxford comma ("ROAS Calibration, and Budget Planning" vs "ROAS Calibration and Budget Planning").

### Re-verified after the re-run

Everything the re-execution touched still agrees with the prose: ROAS table (Online Display 27.94 [20.47, 35.79] against a 28.0 midpoint, Direct Mail 4.75 [2.09, 7.46] against 5.0, Insert 4.03 to 157.70, Radio 3.31 to 110.85); `az.summary` returns `eti89`, so cell 59's "89% interval" labels are right and 0.034 to 0.62 for `sat_lf_beta` is the actual lower edge; mediated shares unchanged at 6% / 25% / 11% after the denominator switch to `direct_delta_da`, and the dims align because `direct_deltas` is already summed over date; the mediator decomposition sums to 1 by construction, so 40/10/49 is a genuine partition; 0 divergences in both the CV loop and the final fit, max R-hat 1.010; uplift 23.0% / 99.8% / 9.5%; pinned media total 1,364,471 equals the direct pot exactly; `warnings` is imported in cell 5 and the `filterwarnings` there only suppresses `FutureWarning`, so the three downgraded checks will actually surface; `set_constraints` does not recompile the objective (`_compile_objective_and_grad` runs before it at line 1784), so cell 94's gradient is unaffected by the pinned constraint set installed in cell 92; and the loop in cell 105 leaves the optimizer on the ±100% constraint set, which is harmless because nothing after it uses the optimizer.

juanitorduz:
All nine items addressed in 8083ac5f. Point by point:

**F1 (units).** Correct. The marginal table now carries two columns, per sustained weekly dollar and per dollar (divided by the 13 planning weeks), and the prose states the units, including the carry-over tail in the response sum, and the comparability with the ROAS table. One precision on your check: Online Display lands 15.7% below its average ROAS, so it is "within about 16%", not 15%. And one correction to the ranking as phrased in the review: SEM's ~12 per dollar sits between Direct Mail's 4.7 and Online Display's 24, so SEM is not below every media line; it is below every media line still free to give (Direct Mail is already at its floor). The Newspaper marginal sentence is requalified the same way (208 per sustained weekly dollar, about 16 per dollar, lowest media line after Direct Mail).

**F2 (residual arithmetic).** The sensitivity conclusion now states that the three rows are one observation (same corner in every box, SEM equal to the pot minus the media corner to the cent), makes the ranking the surviving claim, and names the band machinery as inactive at every reported optimum. Your independent probe is adopted in the form that binds: the SEM lower band binds for any band tighter than about +/-16% ((1-p) x 722,419 > 606,652), so the notebook adds a +/-10% band solve at the +/-50% media box. The result is informative: SEM is forced to the band floor (650,177) and the unwind is ordered, with Newspaper returning its entire ceiling premium (32,624, landing exactly on its floor) before Online Display concedes the remaining 10,901 interiorly. The pot sweep is not added: with the media corner intact, SEM moves one-for-one with the pot, which is the same residual arithmetic.

**F3 (-0 print).** Now prints the level with two decimals and the share of the total: -0.24 (-0.00%) for the objective, 9,954,859.19 (100.00%) for the SEM lever.

**Nits.** (a) The 0.68 is gone; 0.67 appears once, with its interval. (b) "A few dollars" replaced by the 21.18 tell, Online Display against Social Media offsetting exactly, with your reading that the objective is flat between two channels it values identically. (c) `set_constraints` removed from the internals list and noted as public. (d) "Media line" added. (e) Added: the 49% SEM term is the direct lambda-times-spend contribution at historical levels, so about half of the pool SEM harvests is demand its own spend created, which closes the loop with the DAG and the impressions-as-channel answer. (f) Not changed: at the reviewed commit 21ceaf6e the Oxford comma is already present in all four locations (gallery.yaml line 117, gallery.md line 285, the notebook H1, and the PR title), and `git grep "Calibration and Budget" 21ceaf6e` returns nothing, so the mismatch appears to be a stale read.

Also reworded the CRPS paragraph into jitter-proof ranges: the per-fold CRPS values move at the third digit between runs because the CV posterior predictive is unseeded upstream (`_time_slice_step` passes no `random_seed`); every other reported number reproduces exactly across the re-runs.

daimon-pymclabs (commented):
## Review round 3

Round-2 items are all addressed and the fixes are the right ones. The marginal table now carries both units and the per-dollar column reproduces the ROAS table; `set_constraints` is correctly reclassified as public; the divergence checks are `warnings.warn` rather than `raise`, so a future doc build will not hard-fail on a stray divergence; `0.68` is gone from the `funnel_lambda` paragraph; the cell 25 plausibility check now says "around 20" and the arithmetic backs it; and the tightened-band probe in cells 106 to 108 is exactly the independent observation the sensitivity table could not supply. I re-verified the new arithmetic: `-grad_joint / 13` matches the ROAS table to within 16% on all seven channels with Online Display widest at 15.5%, `32,624 + 10,901 = 43,525` matches the SEM band floor move, Newspaper lands on `16,311.79` which is its floor to the dollar, and `606,652 / 722,419 - 1 = -16.03%` so the band really does start binding at about plus or minus 16%.

Three findings this round. None blocking, and one of them is a genuine numerical problem with a headline number.

### 1. The share-space counterfactual applies a corner solution far outside the box that produced it (cell 111, cell 115)

`optimal_shares` is derived from `result`, which is the argmax at **one** budget level: a 2.087M per-week pot, inside a plus or minus 50% box drawn around the last-52-week channel means. Cell 111 then applies those fixed shares to all 209 weeks, whose pots run from 0.29x to 3.42x that level (min 615K, max 7.14M, full-history mean 2.33M).

The consequence, computed from `case_study_data.csv` and the shares printed in cell 111:

| channel | box ceiling | reweighted mean | weeks above the ceiling | weeks above that channel's own historical max |
|---|---|---|---|---|
| Insert | 74,068 | 81,430 | 85 / 209 | 0 |
| Newspaper | 48,935 | 53,511 | 82 / 209 | 0 |
| Radio | 182,953 | 204,739 | 90 / 209 | 11 |
| TV | 204,750 | 228,005 | 89 / 209 | 0 |
| Social Media | 263,996 | 295,475 | 90 / 209 | 15 |
| Online Display | 422,301 | 469,969 | 89 / 209 | 29 |
| Direct Mail | 849,703 | 316,415 | 2 / 209 | 0 |

Six of the seven media cells sit above the optimizer's own ceiling in roughly 40% of weeks, and for six of them the **average** reweighted week is already above the ceiling, because the full-history mean pot (2.33M) exceeds the 52-week pot the shares were solved at. Online Display exceeds its largest observed weekly spend in 29 weeks.

So the 23% is not "the optimal mix run over history". It is a fixed-share rule that is the argmax at approximately no week in the sample, evaluated in a spend region where the saturation curves are extrapolating for three channels. The warning box gets half of this ("the optimal shares push some channels beyond spend levels the model has seen") but not the part that actually undermines the framing: the shares themselves are budget-level-specific, so calling them "the optimal mix" and holding them fixed across a 12x range of weekly pots is the assumption doing the work.

Two ways out, either is fine:
- Cheap: put the numbers in the warning. "The optimal shares are the argmax at a 2.09M weekly pot; historical pots span 615K to 7.14M, so the reweighted plan sits above the optimizer's own ceiling in about 40% of weeks and above Online Display's largest observed spend in 29 of them."
- Honest: re-solve per week with each week's own pot and its own box, then reweight. That is 209 solves of an already-compiled optimizer, so it may be too slow for a doc build, in which case bucket the weeks into three or four pot sizes and solve once per bucket.

### 2. The marginal-versus-ROAS agreement is real but it is not evidence for the unit conversion (cell 95)

Cell 95 says the per-dollar column is "directly comparable with the ROAS table" and that every channel lands within about 16%. Both statements are true, but the sentence invites the reader to read the agreement as a check on the `/13`, and it is not one.

A marginal return cannot exceed an average return at the same spend level under a concave saturation curve. Three channels do: Newspaper +11.3%, Insert +7.6%, TV +1.8%. Nothing is broken; the two quantities are evaluated at different spend levels (marginal at the last-52-week mean, ROAS over all 209 weeks). The deviations track that difference almost exactly:

| channel | 52w mean / full-history mean | marginal vs ROAS |
|---|---|---|
| Newspaper | 0.13 | +11.3% |
| Insert | 0.62 | +7.6% |
| Direct Mail | 0.75 | -1.1% |
| TV | 0.81 | +1.8% |
| Radio | 0.99 | -10.2% |
| Online Display | 1.30 | -15.5% |
| Social Media | 1.73 | -12.7% |

Spearman -0.93 across the seven (p = 0.003). Online Display is the widest gap because its recent spend is 1.3x its historical mean, not because of anything about the conversion.

The `/13` is right, and I checked it independently: the decision variable is a weekly level held for 13 weeks, so total spend is `13 x`, and `d(response)/d(total spend) = (1/13) d(response)/dx`. Suggest saying so directly and replacing "directly comparable" with something like "comparable once you allow for the spend levels the two are evaluated at: the gaps are ordered by how far each channel's recent spend sits from its four-year mean, which is why Newspaper's marginal exceeds its average and Online Display's falls short."

### 3. The unwind order in cell 108 is explained at a spend level nothing in the notebook evaluates

Cell 106 makes exactly the right point: "the marginal table is evaluated at the historical plan, not at the optimum, so the unwind order cannot be read off it." Cell 108 then reads it off it anyway ("the order matches the per-dollar ranking") and explains the second step with "at these levels its saturation curve offers the least response per dollar reclaimed", which is an assertion, not a check.

It matters for the second step specifically. Newspaper going first is safe: 16.0 per dollar against 23.6 for the next line, and it is the cell furthest from any competitor. But Online Display supplying the remaining 10,901 rather than Social Media rests on 23.6 versus 24.2, a 2.5% gap at the historical plan, inside the rounding of the table as printed. And at the 50%-box optimum both are at their ceilings, which is a different point on both saturation curves and is where the solver is actually working.

One cell closes this, reusing machinery already in the notebook:

```python
_, grad_at_opt = optimizer._objective_and_grad(x_funnel_aware)
pd.Series(-grad_at_opt / n_plan_weeks, index=[*channel_columns, "sem_spend"]).round(1)
```

That gives the at-the-optimum marginals cell 106 correctly says are missing, turns "offers the least response per dollar reclaimed" into a verified statement, and shows the reader how much the ranking moves between the historical plan and the corner. If Online Display and Social Media have separated at the optimum, that is a better sentence than the one currently there. If they have not, the honest version is that the solver picked between two near-tied cells and the split is not information.

### Nits

- Cell 102 output: `gain from the objective alone (pinned vs direct): -0.24 (-0.00% of the total)`. The `.2f` fixed the `-0`, but `-0.00%` reads as the same class of artifact. Either drop the percentage on that line or print "under 0.01% of the total".
- Cell 108: "Newspaper gives back its entire ceiling premium (32,624)". Its ceiling premium over the historical level is 16,312. 32,624 is the full ceiling-to-floor swing, which happens to equal the historical level itself. "gives back its entire plus or minus 50% range" is the exact statement.
- Cell 115 warning: "The SEM impressions series stays at its observed values because it is an outcome of the plan, not an input." In this counterfactual the impressions series is inert rather than held fixed as a choice: cell 112 samples `y` only, `uplift_var_names` excludes the impressions likelihood, and the mediator is a deterministic function of spend and parameters. The observed impressions mattered at fit time and do nothing here. Worth rewording so a reader does not go looking for where the series enters.

timbo112711:
@juanitorduz just review the notebook! Over all the flow is great and adds onto the first case study nicely 👌🏻. Left some comments!

daimon-pymclabs:
@timbo112711 on your question about the overlapping titles in the `cv.plot.param_stability()` plots: it is the former, and `figsize` will not fix it.

`MMMCVPlotFactory.param_stability` hardcodes `"layout": "none"` into its default figure kwargs (`pymc_marketing/mmm/plotting/cv.py:259`), so matplotlib runs no constrained or tight layout pass and the labelled title collides with the y-axis channel labels. Growing the figure moves the collision around rather than resolving it, because nothing is reflowing the axes.

Worth noting it is the only place in the package that sets `layout` at all, so it reads as a leftover rather than a house default. Passing `figure_kwargs={"layout": "constrained"}` at the call site should confirm the diagnosis, and if it does, the fix is to make that the default and keep `"none"` reachable through `figure_kwargs` for anyone placing axes by hand. Happy for you to take it.

juanitorduz:
@timbo112711 all seven threads addressed in the latest commit, plus a set of additions from a second read. Point by point on your threads:

1. **Escaped underscores in the equations.** Replaced `\_` with `_` inside `\text{}` in both SEM equations. For the record: the RTD build was passing the escape straight to MathJax (which renders `\_` as an underscore), so the published page was fine; the literal backslash was a ReviewNB rendering artifact. The plain underscore renders on all three.
2. **"Data registration" sentence.** Adopted the intent with one precision: `DataVarMuEffect` registers the two columns as `pm.Data` containers (not coordinates), and only `sem_impressions` is observed by a likelihood. The sentence now reads "which registers `sem_spend` and `sem_impressions` as `pm.Data` containers in the model, so prediction calls can swap in new values without rebuilding the model".
3. **Baking the data plumbing into the API.** Opened #2945 with the full proposal: the model derives `extra_vars` from its effects' `data_vars` and threads them through build, prediction, CV, and `fit_data`, `set_data` stops skipping silently, plus the tests that do not exist today. It cross-links #2938 (the save/load symptom) and lists the file:line evidence for each gap.
4. **Overlapping titles in `param_stability`.** It is the library, not `figsize`: the method hardcodes `"layout": "none"` (`pymc_marketing/mmm/plotting/cv.py:259`), as Daimon diagnosed. Verified on synthetic CV data that `figure_kwargs={"layout": "constrained"}` removes the collision; both call sites in the notebook now pass it, and #2944 is open for the default change, with you named as the volunteer.
5. **Numbered caveats.** Done.
6. **"media does not control".** Done, both occurrences, plus the one remaining plural elsewhere ("if media drives roughly a third of sales").
7. **Next step 5.** It became an appendix: the notebook now builds the same model from a YAML registry (`build_mmm_from_yaml`, with the notebook's classes registered on `REGISTRY`) and proves equivalence, 42 named variables in both graphs and a log-probability match to the last digit at the same point. Next step 5 keeps MLflow and points to the appendix.

Additions in the same commit, from a second read of the notebook as a tutorial:

- **94% HDI everywhere.** `az.rcParams["stats.ci_kind"] = "hdi"` and `["stats.ci_prob"] = 0.94` in the setup cell (ArviZ 1.x defaults to an 89% equal-tailed interval), the ROAS table and the gain interval moved from quantiles to `az.hdi`, and every "interval" in the prose now says which one it is. The experiment ranges remain "95%" because they are inputs, and the prose says so once. `cv.summary.crps` has no interval, so nothing changed there.
- **Prose-to-output consistency.** Every number the prose quotes now has a printed source (a "Quoted Quantities" cell dumps them as JSON at the end), and a local checker asserts about 110 quoted values against that dump. The re-run moved several numbers, all re-read: the expected gain of the funnel plan is now 11.3M (77% win probability), the tightened-band unwind is shared by Newspaper, Online Display, and Social Media, and the Insert and Radio ROAS endpoints changed with the switch from quantiles to HDI.
- **New section "Comparison with the Baseline MMM from Part I"** refits Part I's specification on the same 209 weeks and compares results, not only methods: calibration cut the Online Display and Direct Mail interval widths to 13% and 28% of the uncalibrated widths, the upper funnel means moved in step with their mediated shares (Radio by 6.2 ROAS points), the direct-only channels did not move, and SEM gets a return figure only in Part II.
- **Model diagnostics** start with `az.diagnose`; nutpie reports tree-depth saturation under a name the summary does not look for, so that check is printed separately, and the R-hat line now uses the unrounded maximum (1.009).
- **SEM impressions posterior predictive with noise** added next to the implied-mean plot, with printed coverage: 204 of 209 weeks inside the 94% HDI, 43% inside the 50% HDI, and the prose explains the slow level drift behind the second number.
- **`M*` explanation**, business readings of the direct-vs-mediated plots, the SEM plot's second axis without its grid, `azp.plot_dist` for the calibration figure, and the CV warning removed by passing the prior width as a labelled `xr.DataArray`.

Not changed: `cv.summary.param_stability` cannot tabulate a scalar parameter (`funnel_lambda`) and raises on a 0-dimensional dataset, so the per-fold table is computed by hand in the notebook; worth a small library fix later.

daimon-pymclabs (commented):
## Final review

Re-read the notebook at `dc2b788` (the content is unchanged since `0a77fea9`; the two later commits are merges from main), checked every executed output against the prose, and re-verified the arithmetic against `data/case_study_data.csv`.

State of play first, so nobody re-checks it. CI is green, and it now means something: [#2939](https://github.com/pymc-labs/pymc-marketing/issues/2939) is closed, the notebook workflow's path filters cover `docs/source/notebooks/mmm/**`, and both `mmm` shards pass on cvm and numba, so this notebook is actually executed in CI. The RTD preview builds. The title matches in all four places (PR, notebook H1, `gallery.md`, `gallery.yaml`), the gallery card and yaml entry are in sync, and the `seealso` in case study I renders as a directive.

The round-3 fixes that landed are right: the marginal table carries both unit columns plus the ratio-to-average-ROAS column (largest gap 16.3%, Social Media), the sensitivity conclusion is now the ranking rather than the cut, the tightened-band probe is in, and `gain_from_objective` prints a real number (4.21, 0.00% of the total) instead of `-0`. The two additions from that commit are the best material in the PR: the Part I refit gives the comparison actual numbers (Online Display 58.0 [0.1, 119.1] to 27.7 [20.6, 36.3], width ratio 0.13) and the YAML appendix proves the handover path with 42 matching variables and an identical log-probability.

Four items below. The first is a labeling problem on a headline number. The other three are round-3 findings that did not make it into the last commit, restated with the numbers so they are cheap to close. Nothing blocking.

---

### 1. The funnel contribution is credited to media, but at least 41% of it is ambient demand (cells 71, 117, 125, 126)

`funnel_effect_contribution` is `sat_lf(adstock_lf(M*))` and `M*` contains `funnel_baseline`, so the term does not vanish when every spend line goes to zero. Cell 80 puts the mediator level at 41% baseline, 10% upper funnel media, 49% SEM term. Because `sat_lf` is concave with `sat_lf(0) = 0`, the floor's share of the *contribution* is strictly larger than its share of the *level*: `sat(b) / sat(b + x) > b / (b + x)` week by week. So more than 41% of the 14.0% "mediated SEM (funnel)" row, over 5.7 points of sales, is ambient demand converting rather than anything a budget can move.

Three places read that row as media:

- cell 71's table plus cell 72's sentence that the waterfall "understates the true media-driven share";
- cell 125's derived row, "media-driven (direct plus mediated)", at 43.2%;
- cell 126's conclusion that "the flat model both undercounts media and misattributes the part that travels through search", and the same sentence in the conclusion.

The tell is inside the comparison table. Baseline goes from 62.6% (Part I) to 56.6% (Part II) while the funnel term appears at 14.0%, so about 6 of those 14 points came out of the intercept, which is what the concavity argument predicts. Part I books ambient search demand as baseline; Part II books it inside a line labeled media. The 43.2% against 36.7% gap is therefore not a like-for-like undercount, and it is the one number in the comparison section that does not survive its own framing.

Worth saying that the mediated-share section (cells 74 to 82) does this correctly, by zeroing each channel, and it is exactly why those shares come out at 5% to 24% rather than anything resembling 14% of sales. The decomposition table is the only place the non-incremental version is presented as media mass.

One cell closes it, reusing machinery already in the notebook: set every column of `pot_columns` to zero, run `sample_plan`, and read the residual `funnel_effect_contribution_original_scale`. That gives the ambient floor explicitly, splits the 14.0% into floor plus incremental, and makes the Part I versus Part II baseline rows comparable. The same floor sits in both the numerator and the denominator of cell 117's `media_driven_response`, where it dilutes the ratio, so the 23% uplift is conservative rather than wrong; only the label needs care there.

### 2. Round-3 finding 1 is still open: the share-space counterfactual applies a corner solution far outside the box that produced it (cells 116, 120)

`optimal_shares` comes from `result`, the argmax at one budget level (a 2.087M weekly pot inside a plus or minus 50% box around the last-52-week means). Cell 116 applies those fixed shares to all 209 weeks, whose pots run from 615K to 7.14M, that is 0.29x to 3.42x the level they were solved at, with a full-history mean of 2.33M. Recomputed from the CSV and the shares printed in cell 116:

| channel | box ceiling | reweighted mean | weeks above the ceiling | weeks above that channel's own historical max |
|---|---|---|---|---|
| Insert | 74,068 | 81,430 | 85 / 209 | 0 |
| Newspaper | 48,935 | 53,511 | 82 / 209 | 0 |
| Radio | 182,953 | 204,739 | 90 / 209 | 11 |
| TV | 204,750 | 228,005 | 89 / 209 | 0 |
| Social Media | 263,996 | 295,475 | 90 / 209 | 15 |
| Online Display | 422,301 | 469,969 | 89 / 209 | 29 |
| Direct Mail | 849,703 | 316,415 | 2 / 209 | 0 |

Six of the seven media cells sit above the optimizer's own ceiling in roughly 40% of weeks, and for six of them the average reweighted week is already above the ceiling, because the full-history mean pot exceeds the pot the shares were solved at. Online Display exceeds its largest observed weekly spend in 29 weeks.

The warning in cell 120 still says only that "the optimal shares push some channels beyond spend levels the model has seen". The part that undermines the framing is not in it: the shares are budget-level-specific, so calling them "the optimal mix" and holding them fixed across a 12x range of weekly pots is the assumption doing the work behind the 23%. The cheap fix stands, put the numbers in the warning; the honest fix is to bucket the weeks into three or four pot sizes and re-solve once per bucket.

### 3. Round-3 finding 3 is still open: cell 113 explains the unwind at a spend level nothing evaluates

Cell 111 makes the right point ("the marginal table is evaluated at the historical plan, not at the optimum, so the unwind order cannot be read off it"), and cell 113 then reads it off it anyway ("the order matches the per-dollar ranking") and explains the second step with "at these levels its saturation curve offers the least response per dollar reclaimed", which is an assertion rather than a check. It matters for the second step specifically: Online Display supplying 6,836 and Social Media 4,065 rests on 23.94 against 24.11 per dollar at the historical plan, a 0.7% gap, while at the 50%-box optimum both sit at their ceilings, a different point on both curves. One cell:

```python
_, grad_at_opt = optimizer._objective_and_grad(x_funnel_aware)
pd.Series(-grad_at_opt / n_plan_weeks, index=[*channel_columns, "sem_spend"]).round(1)
```

If the two have separated at the optimum, that is a better sentence than the current one. If they have not, the honest version is that the solver split a near-tie and the split is not information.

### 4. Stale note: #2944 is fixed on main and already merged into this branch (cell 36)

Cell 36 tells the reader that the plots "use a constrained figure layout because the default layout of this plot method lets the variable label collide with the channel labels ([#2944](https://github.com/pymc-labs/pymc-marketing/issues/2944))". [#3005](https://github.com/pymc-labs/pymc-marketing/pull/3005) made `layout: "constrained"` the default (`pymc_marketing/mmm/plotting/cv.py:259-263`), #2944 was closed on 2026-09-12, and the Sep 14 merge brought it into this branch. So the note describes a defect that no longer exists and the explicit `figure_kwargs` in cells 37 and 38 is now a no-op. The executed figures are unaffected either way. Either drop the explanation and keep the kwarg silently, or drop both.

### Nits

- Cell 113: "Newspaper gives back its entire ceiling premium (32,624)". Its premium over the historical level is 16,312; 32,624 is the full ceiling-to-floor swing, which equals the historical level. "gives back its entire plus or minus 50% range" is the exact statement. This one carried over from round 3.
- Cell 120: "The SEM impressions series stays at its observed values because it is an outcome of the plan, not an input." In this counterfactual the series is inert rather than held fixed as a choice: `sample_plan` draws `y` only, the impressions likelihood is not in `uplift_var_names`, and the mediator is a deterministic function of spend and parameters. The observed impressions mattered at fit time and do nothing here.
- Cell 126 calls Insert and Social Media "the control group" that "should not move, and they do not", and the conclusion says the direct-only channels "did not move". Social Media moves 0.8 ROAS points, but Insert moves 5.41 (53.35 to 58.76, about 10% of its mean). It is well inside a 130-wide interval, so the reading is fine, but "within a ROAS point or five" and "did not move" are doing different jobs in the same paragraph.
- Cell 53 fits with `random_seed=rng` while everything else in the notebook is seeded with the integer `seed`. Since `rng` is consumed sequentially, re-running that one cell changes the final fit and its posterior predictive. A fixed integer there would make the whole notebook order-independent.
- The PR description is now stale in two ways worth a refresh before merge: the CI note ("path filters exclude `docs/source/notebooks/mmm/**`, so this PR does not trigger it") is no longer true, and the uplift figures quoted there (23%, 99.8%) predate the re-run that produced 23.2% and 99.7%.

### Re-verified after the merges

Every number the prose quotes reproduces against the executed outputs: mediator split 41/10/49; mediated shares 5%, 11%, 24% with HDIs 0 to 16%, 0 to 38%, 0 to 76%; calibrated ROAS 27.70 [20.60, 36.35] against a 28.0 midpoint and 4.63 [1.85, 7.31] against 5.0; `funnel_lambda` 0.67 [0.59, 0.75] with fold means 0.683 to 0.696; `sat_lf_beta` 0.27 [0.013, 0.74]; max unrounded R-hat 1.0088 and zero divergences in the CV loop, the final fit, and the Part I refit; impressions coverage 204 of 209 at the 94% HDI and 43.1% at the 50%; CRPS 17.2 to 18.0M train against 15.0, 29.2, 15.0, 17.7M test; per-dollar marginals 4.70 (Direct Mail), 11.63 (SEM), 16.43 (Newspaper), 23.94, 24.11, 33.81, 47.79, 59.82; the 44.84 offsetting difference between the pinned and direct plans; expected gain 11,337,083 [-15.98M, 40.34M] at 77.2%; uplift 23.2% [6.2%, 41.4%] at 99.7% and 9.4% [2.3%, 16.6%]; width ratios 0.13, 0.28, 0.94, 0.95, 0.95, 1.01, 1.02. I also re-derived the residual identity behind the sensitivity table at all three boxes (1,410,778 + 676,112, 1,480,238 + 606,652, 1,596,004 + 490,886 against a 2,086,890 pot) and the plus or minus 16% band-binding threshold. The notebook is 11.0MB with outputs, in line with the other MMM case studies (20.5MB and 12.9MB), so nothing to do there.

daimon-pymclabs (approved):
Approving per request.

Nothing here is blocking: CI is green on both mmm shards, the RTD preview builds, gallery wiring and titles are consistent, and the numbers the prose quotes reproduce against the executed outputs.

The one item I would still fix, ideally before merge but fine as a follow-up: `funnel_effect_contribution` includes the `funnel_baseline` floor, so the 14.0% "mediated SEM" row and the 43.2% "media-driven" total attribute some ambient demand to media, which makes cell 126's "the flat model undercounts media" not like-for-like. A zero-all-spend counterfactual cell resolves it. The remaining three (share-space shares applied outside the box ceiling in ~40% of weeks, cell 113's unwind order without marginals at the optimum, and cell 36 describing #2944 as open after #3005 fixed it) are all editable post-merge.

Details in the previous review: https://github.com/pymc-labs/pymc-marketing/pull/2936#pullrequestreview-5207780256

juanitorduz:
I will merge this one as it has been open for a while and we can iterate in the future.

daimon-pymclabs:
@lancechua on the discrete x-axis point: it is a package bug rather than a notebook one, and it is worse than a cosmetic nitpick.

`MMMCVPlotFactory.crps` builds its x coordinate as `np.arange(len(cv_labels))` and hands it to `azp.visuals.line_xy` (`pymc_marketing/mmm/plotting/cv.py:388-392`). Nothing maps the ticks back to `cv_labels`, and no integer locator is set, so the axis shows float-formatted positional indices and the actual fold labels, which the function already has in hand, are dropped. Anyone reading the plot cannot tell which fold is which; they have to count. Fixing the labelling is the substantive part, and switching to markers, or a scatter over a line, follows from it.

Merging train and test into one panel is a separate change: `PlotCollection.grid` is called with `cols=["split"]` (`:377-384`), so the two panels currently get independent y-scales, and merging them makes the comparison honest but also changes what the existing layout means. Worth doing, worth doing on its own.

This PR is merged, so the four threads here need a follow-up. I can open one issue covering the `crps` labelling and the single-panel option, and leave the notation rename ($s_{sem}$ to $s_{spd}$, and $s_{imp} = \max(\text{sem\_impressions}_t)$), the upper-funnel-calibration forward reference, and the guardrail forest plot as notebook items for whoever picks up the next case study. Say the word and I'll file it.

PR #2943, opened 2026-09-03
Fix/r2d2 fourier coord ordering

TeemuSailynoja:
Tidies up #2929 which already was merged, but we agreed with @williambdean that this is a cleaner way to achieve this.

----
📚 Documentation preview 📚: https://pymc-marketing--2943.org.readthedocs.build/en/2943/


williambdean:
oh, I was thinking about having it not in the mmm.py module at all but somewhere closer to the fourier or MuEffect for fourier

williambdean:
Might not be possible at the moment due to the MMM implementation. But just a thought for longer term vision

juanitorduz (approved):

juanitorduz:
> Might not be possible at the moment due to the MMM implementation. But just a thought for longer term vision

ahhg! I merged before seeing this, @williambdean ! ~~Shall we revert?~~ 

Do we want to merge it ?

williambdean:
Either way is fine. Just dont think this is that much of improvement

PR #2946, opened 2026-09-04
Add media inflation MMM notebook

drbenvincent:
## Summary
Adds a new example notebook showing how media inflation can make nominal spend a misleading input for media mix models (MMMs), and how to account for it during data preparation.

## What the notebook covers
- Explains why rising CPM or cost-per-GRP can cause spend and media delivery to diverge.
- Uses a simulated example with a known reference truth to compare nominal and inflation-adjusted media inputs.
- Shows why impressions or GRPs are preferable when available, and how to proceed with spend plus CPM or spend-only data.
- Demonstrates the effect of media-unit choice on inferred media and trend contributions.
- Includes reproducible random seeding and clear, accessible explanations of marketing and modelling terminology.

## Validation
- Real notebook execution
- Mock notebook runner
- Pre-commit
- Linting


daimon-pymclabs:
## Full-spectrum review

Read the notebook end to end plus the gallery entries. Verdict up front: the premise is sound and worth a docs page, the prose is unusually clear, but as it stands the demonstration is weaker than the claims it makes. Three things need fixing before merge (missing sensitivity section, oracle deflator, simulator/model mismatch), and there is a framing sharpening that matters given where `pymc-marketing` is heading.

---

### a) Premise

**The premise holds.** Non-stationary CPM breaking the spend-as-exposure proxy is a real, common, under-documented failure mode, and the choice to demonstrate it via decomposition drift (media up, trend down, total flat) rather than a headline ROAS number is the right call. I would keep it.

**But the causal framing is slightly off-target.** The notebook argues from mediation: spend → impressions → sales, impressions is the mediator, therefore prefer impressions. That is true but it is not the actual problem here. Given that DAG, spend is a genuine cause of sales and regressing on spend estimates a well-defined total effect. What breaks is narrower and more interesting: **the spend → delivery map is non-stationary, and adstock+saturation assume it is fixed.** If CPM were constant at any level, nominal spend would be a perfectly good input and nothing in this notebook would fire.

That distinction matters for two reasons. First, "spend is not the mediator" would imply spend is always the wrong input, which contradicts the notebook's own table row 4. Second, for a library moving causal-first, "identify the estimand, then check whether your exposure mapping is stationary over the window" is the reusable lesson; "use the mediator" is not. I would rewrite the second paragraph of *What media inflation changes* around stationarity and keep the DAG as support rather than as the argument.

### b) Logical flow

**The sensitivity analysis is promised twice and never delivered.** This is the biggest single gap.

- Cell 4 table, row 3: "then test sensitivity to the index"
- *Fit 2* intro: "In practice, the index is an estimate, so we will test how different assumptions affect the attribution"
- Summary bullet 4: "assess whether your attribution conclusions change across plausible assumptions"

There is no such section. A reader following the notebook is told three times to do something the notebook never shows. This is also the section that would carry the most practical weight, because in the field the index is always wrong by some margin and nobody knows how much error is tolerable. Refitting at, say, 4%, 8%, and 12% annual and overlaying the three trend posteriors would answer that and would cost one extra cell plus two fits.

**The deflator is the oracle, and the "validation" is algebra.** `index` is built from `ANNUAL_CPM_INFLATION`, the same constant that generated `cpm`. So `X_today = spend / index` is *exactly* proportional to `impressions` by construction, and the "Deflated spend tracks impressions" figure cannot show anything else. It is presented as "a simple validation" that "checks that relationship in the simulated data," which reads as empirical confirmation of a step that is definitionally true.

This compounds with the missing sensitivity section: the notebook demonstrates the spend-only case using information (the exact CPM path) that a spend-only analyst does not have, and then never shows what happens when that information is imperfect. Two fixes, both cheap: say plainly that the index is exact here so the figure is a sanity check on the arithmetic rather than evidence, and make at least the headline Fit 2 use a deliberately imperfect index so the recovered decomposition is honest.

**Section ordering.** The paragraph beginning "We now fit a baseline MMM using nominal spend" sits in its own markdown cell *above* the `## Fit 1: using nominal spend` heading, so the section intro is orphaned outside its section. Merge it into the Fit 1 cell.

**Caption placement is inconsistent.** Captions precede their figure at cells 8/9, 10/11, 25/26, and 32/33, but follow it at 18/19 and 27–29. Pick one. Following the figure is the more common reading order and matches the rest of the MMM gallery.

### c) Clarity

- `plot_media_vs_trend` returns `(fig, axes)` and cells 18 and 28 do not capture it, so each plot is followed by a printed `(<Figure size ...>, array([<Axes...>], dtype=object))`. Add a trailing semicolon or drop the return.
- `years_elapsed` is computed in the data-generating cell and then recomputed identically in the Fit 2 cell. The second one is either redundant or a latent bug if the first ever changes. Same for `y`, which is a numpy array from the simulator and then rebound to `df["y"]`.
- `%load_ext autoreload` / `%autoreload 2` is development cruft in a docs notebook.
- The ylim values are computed from the nominal fit and reused for Fit 2 so the panels are comparable. That is the right choice, but it is only mentioned in a caption. One sentence in the cell would help.
- Admonition density is high: eight callouts (2 note, 3 warning, 2 important, 1 tip) in a notebook of this length, and the "use a smooth index" point appears three separate times (opening warning, Fit 2 tip, summary table). The repetition makes the genuinely important ones (the `set_cost_per_unit` warning, the `time_varying_intercept` warning) blend in. I would cut to about four.
- The two decision tables (cell 4 and the summary) share the same first three rows with reworded justifications. Keep the summary table, cut the early one down to a single sentence pointing forward.
- The closing "**Reflection.** ... Which input would best represent exposure for paid-social auctions, television, or out-of-home media?" reads as course material rather than library docs. I would cut it.

### d) Missing and begging to be included

**A verifiable simulator/model mismatch.** `geometric_adstock_np` uses unnormalized weights (`alpha ** arange(l_max)`, summing to ~1.67 at `alpha=0.4`), while `GeometricAdstock` in the fitted model defaults to `normalize=True` (`pymc_marketing/mmm/components/adstock.py`, `normalize: bool = Field(True, ...)`). The reference series and the model therefore disagree about the adstock scale before saturation ever runs. Saturation absorbs part of this, which is presumably why nothing looks broken, but it means the notebook's own line "the level is only approximately recovered" is partly measuring a self-inflicted mismatch rather than an inherent limit of the method. In a notebook whose whole argument rests on comparing fitted output against a planted truth, the simulator should match the model exactly. Either pass `normalize=False` to `GeometricAdstock` or normalize the numpy weights.

**Where does the cost index actually come from?** "An external media-cost index" appears five times and is never made concrete. This is the first question a reader will have and the notebook is the natural place to answer it: platform-reported CPM aggregated to quarterly, agency or industry benchmarks, auction-level cost pulled from the ad platform API, and so on. Two or three named options with their tradeoffs would do it.

**Media inflation is not general inflation.** Nothing distinguishes media-cost inflation from CPI, and readers will reach for CPI because it is easy to get. Worth one explicit line saying CPI is the wrong deflator for this.

**What about the target?** The example simulates flat sales, so the question never arises, but in production `y` is usually revenue in nominal dollars, which is also inflating, often at a different rate from media. Deflating `X` while leaving `y` nominal is a mismatch the notebook currently leaves the reader to discover. Even a scoped-out sentence would help.

**No evidence that the two fits are indistinguishable in-sample.** The notebook asserts "a good in-sample fit does not establish that the model recovered the correct mechanism," which is the load-bearing claim for why anyone should care, and then never demonstrates it. A LOO comparison or an overlaid posterior predictive check showing the two models fit `y` about equally well would turn that assertion into the notebook's strongest result: *you cannot detect this problem from model fit, only from getting the units right.*

**How does this plug into the budget optimizer?** Deflated units are handled as a one-line "outside the scope of this attribution example" plus a closing note. But budget optimization is the thing most users do immediately after fitting, and a model fitted on today's-dollars or impressions needs a cost mapping before `optimize_budget` means anything. Given `mmm_cost_per_unit` already exists and is cross-referenced, a short worked handoff (or at minimum a concrete pointer to the exact method to call) would close the loop rather than leaving the user at the cliff edge.

**Multi-channel differential inflation** gets no mention. Channels inflate at very different rates, which is exactly when this bites hardest. One line in the scope note is enough.

### Small confirmations

Cross-references (`mmm_quickstart`, `mmm_example`, `mmm_cost_per_unit`) all resolve. The `arviz` v1 `prob=` / `ci_bound` usage is correct for the current pin. The numpy `logistic_saturation_np` matches `LogisticSaturation` exactly.

### What I would do

Blocking: add the sensitivity section, fix the adstock normalization mismatch, and either use a deliberately imperfect index for Fit 2 or state plainly that the deflator is exact.

High value, cheap: the LOO or PPC comparison, and naming real sources for the cost index.

Cuts: one of the two tables, about half the admonitions, the reflection prompt, the autoreload cell, the duplicate `years_elapsed`.

drbenvincent:
<p>Any more specifics on this. If there is only the direct path of spend -&gt; impressions -&gt; KPI, then you'd presumably only want to condition on impressions. I am vaguely aware that this is murky territory, and I'm happy  to have way less prescriptive wording. But do you have any more specifics or a pointer to more details on this?</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/2946/discussion/#comment-5540326534'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
Thanks @drbenvincent ! LGTM... before we merge ... what is `docs/source/_inv/scipy.objects.inv` ?

daimon-pymclabs:
@juanitorduz it is a vendored copy of scipy's intersphinx inventory, added in this PR alongside the `conf.py` change:

```python
_intersphinx_inv = Path(__file__).parent / "_inv"
...
    # Vendored: docs.scipy.org/objects.inv intermittently times out on CI/RTD.
    "scipy": (
        "https://docs.scipy.org/doc/scipy/",
        str(_intersphinx_inv / "scipy.objects.inv"),
    ),
```

So it is a build-reliability workaround for `docs.scipy.org` timing out, not notebook content. Two things worth deciding before merge:

1. **As written it is a pin, not a fallback.** When the second tuple element is a path string, Sphinx loads *only* that inventory and never contacts `docs.scipy.org`. Every scipy cross-reference is then frozen at whatever scipy version this file was downloaded from, and it will silently rot as scipy moves. If the intent is "try upstream, fall back to the local copy", the second element should be a tuple with `None` first:

   ```python
   "scipy": (
       "https://docs.scipy.org/doc/scipy/",
       (None, str(_intersphinx_inv / "scipy.objects.inv")),
   ),
   ```

   `None` resolves to `<uri>/objects.inv`, and Sphinx falls through to the vendored file only when the fetch fails. That keeps links current and still survives an outage.

2. **It adds a 125 KB binary to the repo** with no recorded provenance. Worth a one-line comment in `conf.py` or a `docs/source/_inv/README.md` saying which scipy version and date it came from, otherwise nobody will know when to refresh it.

Neither blocks the notebook itself, which is the actual subject of the PR. If you would rather not carry the vendored inventory at all, the other common fix is bumping `intersphinx_timeout` and letting the reference go unresolved on a flaky build, but the vendored copy is the more deterministic option.

juanitorduz:
@daimon-pymclabs @drbenvincent why do we need the change in conffters and this inv_folder ?

daimon-pymclabs:
@juanitorduz short version: it is unrelated to the notebook. @drbenvincent added it in [28a66ed](https://github.com/pymc-labs/pymc-marketing/commit/28a66eda) because `docs.scipy.org/objects.inv` was intermittently timing out on CI and RTD, so the docs build for this branch kept failing on the intersphinx fetch. The `conf.py` change points `intersphinx_mapping["scipy"]` at a vendored copy of the inventory in `docs/source/_inv/` instead of fetching it at build time.

So it is a build-reliability fix that happened to land on a notebook branch. Two options:

1. **Split it out** into its own PR and rebase this one on top. Cleanest, and it lets the fix be reviewed on its own merits.
2. **Keep it here** but fix the semantics first. As written it is a hard pin, not a fallback: when the second tuple element is a plain path string, Sphinx loads only the local inventory and never contacts upstream, so every scipy cross-reference freezes at whatever version that file came from. Making it `(None, str(...))` gives you fetch-with-fallback instead, plus a line recording which scipy version the file is from.

I would go with 1, since it is genuinely a separate concern from the media inflation notebook and it is the part you keep tripping over in review.

juanitorduz (approved):
Thanks @drbenvincent !

PR #2949, opened 2026-09-06
fix(clv): broadcast x in BetaGeoBetaBinom.logp so short values don't truncate the scan

### LINKED ISSUE #2919: BUG: BetaGeoBetaBinom.logp is silently wrong for a length-1 value (collapses BetaGeoBetaBinomModel posterior draws)

Golopmoui3:
Fixes #2919

## Problem

`x` was missing from the `pt.broadcast_arrays` call in `BetaGeoBetaBinom.logp`:

```python
t_x, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
    t_x, alpha, beta, gamma, delta, T
)
...
unnorm_logp = scan(
    fn=logp_customer_died,
    outputs_info=[None],
    sequences=[t_x, x, alpha, beta, gamma, delta, T],
    return_updates=False,
)
```

Since `x` (like `t_x`) comes from `value`, a length-1 value against a longer parameter batch kept its length while the other scan sequences were broadcast to the batch length. `scan` truncates `n_steps` to the shortest sequence, so the data-dependent term was computed once — effectively for draw 0 — and reused for every other draw, with only the `-betaln(alpha, beta) - betaln(gamma, delta)` normalizer varying. The output shape stayed correct, so nothing warned.

`BetaGeoBetaBinomModel._logp` passes exactly such a length-1 dummy value, so everything flowing through `_logp` was affected: `expected_purchases`, `expected_probability_alive`, `expected_purchase_probability` — posterior means looked fine while posterior *variation* was collapsed, making the credible intervals meaningless.

## Fix

Add `x` to the `broadcast_arrays` call (one line + comment).

## Testing

New regression test `test_logp_length_one_value_broadcasts_to_parameter_batch` (6 parameter draws, one customer `t_x=3, x=2`):

- length-1 value now broadcasts to the full batch: shape `(6,)`, identical to the tiled-value logp (rtol 1e-10),
- agrees with the scalar-parameter logp of the same customer (~9e-10 rel. diff, asserted at 1e-8).

Full `tests/clv/test_distributions.py` suite: **51 passed** in docker (python:3.12), including `test_logp_matches_excel` against the Bruce Hardie reference values. (The MCMC-heavy `tests/clv/models/test_beta_geo_beta_binom.py` suite wasn't run locally to completion — left for CI.)

----
📚 Documentation preview 📚: https://pymc-marketing--2949.org.readthedocs.build/en/2949/

### DIFF
--- a/pymc_marketing/clv/distributions.py
+++ b/pymc_marketing/clv/distributions.py
@@ -310,8 +310,13 @@ class BetaGeoBetaBinom(Discrete):
         # Broadcast all the parameters so they are sequences.
         # Potentially inefficient, but otherwise ugly logic needed to unpack arguments in the scan function,
         # since sequences always precede non-sequences.
-        t_x, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
-            t_x, alpha, beta, gamma, delta, T
+        # `x` must be broadcast too: it comes from `value`, whose length may
+        # differ from the parameter batch. If it keeps a shorter length (e.g. a
+        # single customer against a batch of parameter draws), `scan` truncates
+        # to the shortest sequence and the logp silently computes over a
+        # prefix of the data (see #2919).
+        t_x, x, alpha, beta, gamma, delta, T = pt.broadcast_arrays(
+            t_x, x, alpha, beta, gamma, delta, T
         )
 
         def logp_customer_died(t_x_i, x_i, alpha_i, beta_i, gamma_i, delta_i, T_i):
--- a/tests/clv/test_distributions.py
+++ b/tests/clv/test_distributions.py
@@ -197,6 +197,39 @@ class TestBetaGeoBetaBinom:
             rtol=1e-3,
         )
 
+    def test_logp_length_one_value_broadcasts_to_parameter_batch(self):
+        # Regression test for #2919: with a length-1 value against a batched
+        # parameter set, `x` was not broadcast to the batch length, so `scan`
+        # truncated to a single step and the data-dependent term was computed
+        # once (effectively for draw 0) and reused for every draw, collapsing
+        # the posterior variation of everything flowing through `_logp`.
+        alpha = np.full(6, 1.204)
+        beta = np.full(6, 0.750)
+        gamma = np.full(6, 0.657)
+        delta = np.full(6, 2.783)
+        T = np.full(6, 6)
+
+        # Single customer: t_x=3, x=2
+        value_one = np.array([[3.0, 2.0]])
+        value_tiled = np.repeat(value_one, 6, axis=0)
+
+        dist = BetaGeoBetaBinom.dist(
+            alpha=alpha, beta=beta, gamma=gamma, delta=delta, T=T
+        )
+        logp_one = pm.logp(dist, value_one).eval()
+        logp_tiled = pm.logp(dist, value_tiled).eval()
+
+        # A length-1 value must broadcast to the full parameter batch.
+        assert logp_one.shape == (6,)
+        np.testing.assert_allclose(logp_one, logp_tiled, rtol=1e-10)
+
+        # And it must agree with the scalar-parameter logp of that customer.
+        dist_scalar = BetaGeoBetaBinom.dist(
+            alpha=1.204, beta=0.750, gamma=0.657, delta=2.783, T=6
+        )
+        expected_first = pm.logp(dist_scalar, value_one[0]).eval()
+        np.testing.assert_allclose(logp_tiled[0], expected_first, rtol=1e-8)
+
     def test_invalid_value_logp(self):
         beta_geo_beta_binom = BetaGeoBetaBinom.dist(
             alpha=1.20, beta=0.75, gamma=0.66, delta=2.78, T=6


daimon-pymclabs:
Thanks for the fix. I verified the source change empirically and it is correct: with `x` broadcast, a length-1 `value` against a batch of 6 varying parameter draws matches a scipy transcription of Fader/Hardie/Shang eq. (5) to ~2e-15, and the model-level `BetaGeoBetaBinomModel._logp` path (2 chains x 4 draws x 4 customers) matches to ~6e-15 with per-draw variation restored.

One issue with the regression test: as written it passes on unfixed `main`, so it does not lock the bug out.

All six parameter draws are built with `np.full(...)`, i.e. they are identical. The pre-fix behavior reuses the data-dependent term of draw 0 for every draw, which with identical draws produces exactly the correct numbers. I ran `test_logp_length_one_value_broadcasts_to_parameter_batch` verbatim against `main` without the one-line fix and it passed.

Suggested replacement, which makes the draws differ and adds an independent draw-by-draw check against the scalar-parameter logp:

```python
    def test_logp_length_one_value_broadcasts_to_parameter_batch(self):
        # Regression test for #2919: with a length-1 value against a batch of
        # parameter draws, `x` was not broadcast to the batch length, so `scan`
        # truncated to a single step and the data-dependent term of draw 0 was
        # reused for every draw, collapsing the posterior variation of
        # everything flowing through `_logp`.
        # The draws below must differ from each other: with identical draws the
        # buggy and the correct computation agree numerically.
        rng = np.random.default_rng(42)
        n_draws = 6
        alpha = 1.204 * np.exp(rng.normal(scale=0.3, size=n_draws))
        beta = 0.750 * np.exp(rng.normal(scale=0.3, size=n_draws))
        gamma = 0.657 * np.exp(rng.normal(scale=0.3, size=n_draws))
        delta = 2.783 * np.exp(rng.normal(scale=0.3, size=n_draws))
        T = np.full(n_draws, 6)

        # Single customer: t_x=3, x=2
        value_one = np.array([[3.0, 2.0]])
        value_tiled = np.repeat(value_one, n_draws, axis=0)

        dist = BetaGeoBetaBinom.dist(
            alpha=alpha, beta=beta, gamma=gamma, delta=delta, T=T
        )
        logp_one = pm.logp(dist, value_one).eval()
        logp_tiled = pm.logp(dist, value_tiled).eval()

        # A length-1 value must broadcast to the full parameter batch...
        assert logp_one.shape == (n_draws,)
        # ...and agree with tiling the customer by hand.
        np.testing.assert_allclose(logp_one, logp_tiled, rtol=1e-10)

        # Both must vary across draws and match the scalar-parameter logp
        # draw by draw, so the two batched paths cannot pass by sharing a bug.
        expected = np.array(
            [
                pm.logp(
                    BetaGeoBetaBinom.dist(
                        alpha=alpha[i],
                        beta=beta[i],
                        gamma=gamma[i],
                        delta=delta[i],
                        T=6,
                    ),
                    value_one[0],
                ).eval()
                for i in range(n_draws)
            ]
        )
        assert np.ptp(expected) > 0.1
        np.testing.assert_allclose(logp_one, expected, rtol=1e-8)
```

Results of running exactly this test:

- with the fix applied: passes
- with the one-line fix reverted on the same checkout: fails, 5 of 6 elements mismatched, max absolute difference 0.896 in log space

```
E       Mismatched elements: 5 / 6 (83.3%)
E        ACTUAL: array([-4.478415, -4.610404, -4.454373, -5.42293 , -4.638398, -5.400374])
E        DESIRED: array([-4.478415, -4.576792, -4.542264, -4.739898, -4.507346, -4.504161])
E       Max absolute difference among violations: 0.89621339
```

The `np.ptp(expected) > 0.1` guard is there so a later edit cannot silently return the test to identical draws.

Two smaller notes:

- The test is slow, dominated by `scan` graph compilation (about 2 minutes here for the failing path, longer when all seven logp graphs compile). Worth knowing before it lands in the default test run.
- PR #2902 fixes the same issue (#2919) by rewriting the logp with `OpFromGraph` plus `pt.vectorize` and caching `_logp_fn`. This PR is the minimal fix and I would land it first, but whoever merges should decide what happens to #2902 rather than leaving both open.

shivamlalakiya:
Heads up for whoever reviews this: #2902 is an older open PR against the same bug (opened 2026-08-18, before #2919 was filed; it started from #1033 and picked up #2919 once that was reported). Flagging the overlap so the two are not merged independently, not staking a claim.

The diagnosis here is right and matches what #2902 landed on: `x` comes from `value`, was missing from the `pt.broadcast_arrays` call, and `scan` truncated `n_steps` to the shortest sequence, so the data-dependent term was computed once and reused across draws.

The difference is the fix. In review of #2902, @ricardoV94 pointed out that the `scan` was never doing what it appeared to do and that adding `x` to the broadcast keeps a per-customer `scan` that does not need to exist. #2902 now removes the `scan` entirely, wrapping the single-customer logp in an `OpFromGraph` and lifting it with `pt.vectorize`, which fixes the same truncation as a side effect of there no longer being a sequence to truncate. @daimon-pymclabs reproduced that version rather than reading the diff and confirmed it holds.

So the one-line broadcast here is correct but is a subset of #2902. If maintainers prefer the minimal change for backport reasons that is a reasonable call, it is just worth making it deliberately rather than merging both.

ricardoV94 (approved):
I think it's fine to go with the easy patch and follow up with removing the Scan next

ricardoV94:
Whomever merges this may want to make sure relevant co-authors are specified

ricardoV94:
and it needs a rebase

daimon-pymclabs:
Re co-authors and the rebase, to save whoever merges the digging:

The one-line change here was authored by @Golopmoui3, the same diagnosis was reached independently in #2902 by @shivamlalakiya, and #2919 is where the bug was reported. Ready-to-paste trailers for the squash commit:

```
Co-authored-by: Shivam Lalakiya <50960482+shivamlalakiya@users.noreply.github.com>
```

(@Golopmoui3 is already the commit author, so no trailer needed there.)

On the rebase: this branch is a single commit touching one line in `BetaGeoBetaBinom.logp` plus the regression test, so a rebase onto current `main` should be clean. The open question is still the one from earlier in this thread and in #2902: this PR is the minimal fix, #2902 removes the `scan` entirely via `OpFromGraph` + `pt.vectorize`. Both fix #2919, and they should not land independently.

One thing worth not losing in the rebase: as written, the regression test also passes on unfixed `main`, because all six parameter draws are built with `np.full(...)` and are therefore identical, which is exactly the case where the buggy and correct computations agree. The replacement test in [my earlier comment](https://github.com/pymc-labs/pymc-marketing/pull/2949#issuecomment-5567541656) fails on unfixed `main` (5 of 6 elements mismatched, max absolute difference 0.896 in log space) and passes with the fix.

ricardoV94:
yeah test should be a regression test

PR #2950, opened 2026-09-07
docs: interactive priors with modist in the prior predictive notebook

juanitorduz:
## Description

This PR adds [modist](https://github.com/williambdean/modist) to the docs dependencies and uses its draggable distribution widgets in the prior predictive notebook (`docs/source/notebooks/general/prior_predictive.ipynb`). The goal is to turn prior elicitation into an interactive loop: drag the density, re-run the cells below, and look at the implications on the prior predictive samples and on the ROAS before fitting anything.

modist works in plain Jupyter (only `anywidget` is required, no marimo). Each widget exposes `.params`, a frozen `.scipy` distribution, and a `.prior` adapter that returns a `pymc_extras.prior.Prior`, which plugs directly into the PyMC-Marketing model configuration.

### Changes

- **Dependencies**: `modist>=0.6.0` in the `docs` extra (`uv.lock` updated; it pulls `anywidget` and `psygnal`).
- **New intro section** "Interactive Priors with modist": the four families, the drag affordances, the three adapters, and the drag-then-re-run workflow. A note explains that the rendered docs page is draggable but does not re-execute code.
- **Simple example**: a "Shape the Prior Interactively" subsection with a `Normal` widget compared against the observed data; the priors on `mu` and `sigma` are now widgets (`Normal(0, 2)` and `Gamma(1, 1/3)`, which is the same distribution as `Exponential(1/3)`), sampled through `.scipy.rvs` and splatted into the PyMC model with `**widget.params`.
- **MMM section**: widgets for `adstock_alpha` (Beta), `saturation_lam` (Gamma), `intercept` (Normal) and `gamma_control` (Normal). The preliz plots, the adstock and saturation curves, `LogisticSaturation`, and `my_model_config` all read from the widgets through `.prior` (or `Prior("Gamma", **widget.params, dims="channel")` when dims are needed). An `ipywidgets.Tab` control panel gathers the four MMM widgets in one place and a table records the priors at execution time.
- **Consistency fix**: the adstock section showed `Beta(1, 3)` while the model configuration used `Beta(2, 3)`. Both now use `Beta(2, 3)`, so the fitted model is unchanged.
- **Key takeaways**: a bullet on the drag and re-run loop and a tip pointing marimo users to `md.pymc.create_priors`.

### How the widgets render in the docs

The notebook was re-executed with `jupyter execute`, which stores the widget state (including the anywidget ESM bundle) in `metadata.widgets`. myst-nb already injects the ipywidgets embed loader for notebooks with widget state, and jsdelivr serves the `anywidget@~0.11.*` AMD bundle the loader requests, so the widgets render draggable on the docs page. Every widget is paired with a static preliz plot fed from the same widget, so the page stays informative if the CDN load fails. The notebook grows by about 1.3 MB (7 widget instances).

Maintainer note (also left as an HTML comment in the notebook): regenerate this notebook with `jupyter execute --inplace`. Saving from an editor that does not store widget state strips the widget outputs.

### Verification

- `uv lock` resolves with the bare pin and `uv lock --check` passes.
- Mocked CI runner (`scripts/run_notebooks/runner.py --notebooks general/prior_predictive.ipynb`) passes.
- Full execution end to end without errors; zero divergences in the MMM fit; all 7 anywidget views plus the Tab are present in the saved widget state.
- `pre-commit` passes on the changed files (ruff, nbformat, watermark, gallery, large files).
- Local `sphinx-build -W` (Python 3.14): the built page contains the widget-state script, the ipywidgets embed loader and all 15 widget views, and no widget text repr leaks into the HTML. The local build reports 106 autodoc warnings (`error while formatting signature` for the `CovFunc` and `WeibullType` string enums), none about this notebook or modist; they come from modules this PR does not touch and look like a Python 3.14 autodoc artifact, so the CI docs build on 3.12 is the real gate.

## Related Issue

- [ ] Closes #
- [x] Related to https://github.com/williambdean/modist

## Checklist


juanitorduz:
I will close this one as we need better integration with priors and saturation.

PR #2953, opened 2026-09-07
Support pandas 3 and lift the pandas<3 pin

### LINKED ISSUE #2240: Compatibility with pandas 3.0

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/2240

## Description

pandas 3 makes full-column `.loc` setitem strictly in-place and infers the `str` dtype for string columns. The date-column coercion `X.loc[:, date_column] = pd.to_datetime(X[date_column])` in `mmm/data_conversion.py` and `mmm/builders/yaml.py` therefore raised `TypeError` for string and int64 date columns. This is what broke CI in #2951: 30 of the 31 pandas 3 test failures come from that one line, and the last one is a `dtype == object` assertion in `tests/pie/test_model.py`.

Changes:

- `to_mmm_dataset` and `build_mmm_from_yaml` coerce the date column with `X.assign(...)`. This works on pandas 2 and 3 and no longer mutates the caller's DataFrame. Regression tests assert the input frame is unchanged for string and int64 date columns, in both functions.
- pandas 3 removed the `H` period alias and the `M` and `H` offset aliases. The CLV utilities now normalise the public `time_unit` (`H` to `h`) once at every entry point, which also fixes the `np.timedelta64` divisions that never accepted `H`, and translate `M` to `ME` only for the `pd.date_range` call. Regression tests cover monthly and hourly `time_unit` values on both pandas majors. The public `time_unit` vocabulary is unchanged.
- `GammaGammaModel.expected_customer_lifetime_value` builds `future_spend` on a copy instead of adding the column to the caller's frame.
- `docs/source/notebooks/mmm/mmm_counterfactuals.ipynb` and `mmm_example.ipynb`: `Series.dt.to_pydatetime()` returns a Series of Python datetimes on pandas 3, which matplotlib cannot mask in `fill_between`. The datetime Series is now passed directly. Outputs were not regenerated; both notebooks run end to end under the mocked CI runner.
- `tests/pie/test_model.py` checks string columns with `is_numeric_dtype` instead of `dtype == object`.
- `pyproject.toml`: `pandas<3` becomes `pandas>=2.2` with no upper bound. The lock moves to pandas 3.0.5 and mlflow 3.16.0, the first mlflow release that allows pandas 3; pytz drops out because pandas 3 no longer needs it.
- CI runs pandas 3 on every leg. `pandas>=2.2` stays declared but is not exercised by CI (maintainer decision).

**Behaviour changes for the release notes:** `to_mmm_dataset`, `build_mmm_from_yaml` and `GammaGammaModel.expected_customer_lifetime_value` no longer mutate the DataFrame passed in. Callers that relied on the date column being coerced in place, or on `future_spend` being added to their frame, must use the return values instead.

Verified locally: the affected test modules pass on pandas 2.3.3 and 3.0.5, the full suite passes on the pandas 3 venv, pre-commit is clean, `uv lock --check` passes, and a sweep of the pandas-heavy test modules with `-W error::pandas.errors.PandasChangeWarning` reports no pandas deprecations.

Not in scope (pre-existing on pandas 2, follow-up issue): `_convert_frequency_to_timedelta` in `mmm/utils.py` does not recognise the `h`, `min` and `s` aliases that `pd.infer_freq` returns.

## Related Issue
- [x] Closes #2240
- [ ] Related to #

## Checklist


.github/workflows/test.yml
```diff
@@ -53,12 +53,13 @@ jobs:
         run: |
           sudo apt-get install graphviz graphviz-dev
           uv pip install -e .[test]
-      - name: Pin oldest versions of PyMC and extras
+      - name: Pin oldest versions of PyMC, extras and pandas
```
juanitorduz: remove the checks fdor pandas 2.x

juanitorduz (commented):
comment

daimon-pymclabs:
## Review: pandas 3 support

Overall this is the right shape of fix: the `.loc[:, col] =` to `assign` change is the correct pandas 3 idiom, it drops an unwanted caller-mutation side effect, and it works on both major versions. Two things I think need to be resolved before merge (one is a hard CI failure), plus some smaller notes.

### 1. Blocking: the MMM notebooks fail on pandas 3

`Test Notebooks / notebooks (cvm, --notebooks mmm --start-idx 10)` and the numba twin both fail on this branch, while the same jobs pass on the `pandas<3` branch. The venv there resolves fresh from `pyproject.toml`, so lifting the pin is what put pandas 3.0.5 in it.

Failure is in `docs/source/notebooks/mmm/mmm_counterfactuals.ipynb`, cell 14:

```
X_out_of_sample["date_week"].dt.to_pydatetime()
...
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs
could not be safely coerced to any supported types according to the casting rule 'safe'
```

pandas 2.2 deprecated `DatetimeProperties.to_pydatetime` returning an ndarray; pandas 3 returns an object-dtype `Series`, and `Axes.fill_between` cannot unit-convert that. The fix is to drop the call and pass the datetime series (or `.to_numpy()`) straight to `fill_between`.

`docs/source/notebooks/mmm/mmm_example.ipynb` has the same `.dt.to_pydatetime()` call, so I would expect the `--end-idx 10` split to fail the same way once it finishes. Those two files are the only `to_pydatetime` hits in the repo; library code is clean.

### 2. `time_unit="H"` is still broken on pandas 3, and the new map hides it

`_OFFSET_ALIASES` translates the `date_range` freq, but the surrounding period conversions still receive the raw `time_unit`:

- `clv/utils.py:884` `start_date.to_period(time_unit)`
- `clv/utils.py:907` `date_range.to_period(time_unit)`
- `clv/utils.py:247` inside `_find_first_transactions`

pandas 3 removed `H` as a period alias too (2.2 deprecated `H` in favour of `h` for `Period` and `date_range` alike), so `time_unit="H"` raises at the very next line even with the offset mapping in place. `M` is fine, because `M` is still the canonical monthly *period* alias, which is exactly why a single map cannot serve both call sites.

Suggestion: keep `_OFFSET_ALIASES = {"M": "ME"}` for `date_range` and add a separate `_PERIOD_ALIASES = {"H": "h"}` applied at every `to_period`, or normalise `time_unit` once at the public entry points. If hourly is not worth fixing in this PR (the body already lists two other hourly breakages as out of scope), then I would drop the `"H"` entry and raise a clear error for `time_unit="H"`, so the mapping does not read as hourly support that then fails one line later. The docstrings in `customer_lifetime_value` and `GammaGammaModel` still advertise `"H"`.

### 3. The `pandas>=2.2` floor is never exercised

The pandas 2 leg does `uv pip install ... "pandas<3"`, which resolves the newest 2.x (2.3.3), so nothing tests 2.2. Either pin `"pandas==2.2.*"` on that leg to test the declared floor, or set the floor to a version CI actually runs. Related: pandas 2 coverage now rides entirely on the `oldest-pymc` leg, so retiring that leg silently drops it. A short comment in the matrix, or a dedicated `pandas: "2"` axis entry, would make that dependency explicit.

### 4. No upper bound on pandas

`numpy` is capped at `<2.5` in the same block, and this PR exists because a pandas major bump broke the library. `"pandas>=2.2,<4"` would keep the same protection without re-pinning to 2.x. Your call if the project prefers uncapped runtime deps.

### 5. Two full-column `.loc` setitems remain, one of them mutating a user frame

- `clv/models/gamma_gamma.py:216` `data.loc[:, "future_spend"] = ...` adds a column to the caller's DataFrame
- `clv/utils.py:270` `period_transactions.loc[:, "first"] = False` (on an explicit `.copy()`, so it is only the idiom)

Both are new-column enlargement, which pandas 3 still allows, so this is not a correctness bug. But after this PR two call sites no longer mutate the caller and one still does, which is an inconsistency users will hit. `data = data.assign(future_spend=...)` is a one-line follow-up.

### 6. Test coverage of the fix

`test_input_dataframe_is_not_mutated` covers `to_mmm_dataset` only. The identical fix in `build_mmm_from_yaml` has no non-mutation assertion; the yaml builder tests already construct frames, so an `assert_frame_equal(X, X_before)` there is cheap and pins the second half of the fix.

### 7. Smaller notes

- On pandas 2 (copy-on-write off by default) `assign` copies the frame, and the yaml path now copies twice: once in `build_mmm_from_yaml`, again inside `to_mmm_dataset`. Negligible on pandas 3, worth knowing for wide panels on pandas 2.
- The `date_col_in_X` dance in `yaml.py` re-checks the condition the enclosing `if` already tested. Since you are touching the block, `if date_column in X.columns: ... else: raise ValueError(...)` reads better.
- `to_mmm_dataset` and `build_mmm_from_yaml` no longer coercing the caller's frame in place is a user-visible behaviour change. Worth one line in the release notes, since callers that relied on the side effect will now pass string dates onward.
- `_expected_cumulative_transactions`'s `time_unit` docstring still points at numpy datetime units, which is not the vocabulary the function accepts. Drive-by cleanup if you feel like it.

### What I checked and liked

- `assign` preserves column position for an existing column, so no reordering of `X`.
- The monthly regression test asserts the `PeriodIndex` and its `freqstr`, which is the part that would silently drift if the alias mapping were applied to `to_period` as well.
- The lockfile diff is consistent with the claim: pandas 3.0.5, mlflow 3.16.0, `pytz` dropped, `tzdata` narrowed to win32/emscripten.
- The `dtype == object` to `is_numeric_dtype` swap in `tests/pie/test_model.py` is the right assertion for pandas 3's string dtype.

Happy to take another look once the notebook fix is in.

juanitorduz:
Thanks for the thorough review. Addressed in 7c38984e, item by item:

1. **Notebooks.** Both `mmm_counterfactuals.ipynb` and `mmm_example.ipynb` now pass the datetime Series straight to `fill_between`. The stale `to_pydatetime` FutureWarning outputs were removed from `mmm_counterfactuals.ipynb`. Both notebooks run end to end under the mocked CI runner locally on pandas 3.0.5. Outputs were deliberately not regenerated (the change is a one-token plotting fix and a full re-run would rewrite every sampled output), so the watermark cells still show pandas 2.3.3.
2. **Hourly.** `time_unit` is now normalised once at every public CLV entry point (`_PERIOD_ALIASES = {"H": "h"}`), which fixes `to_period`, `date_range` and the `np.timedelta64` division at the same time; `_OFFSET_ALIASES` is reduced to `{"M": "ME"}` and only touches the `date_range` call. Hourly regression tests were added for `rfm_summary` and `_expected_cumulative_transactions` (both `"H"` and `"h"`), plus an `"h"` case in `test_time_unit_scaling`. The three docstrings that pointed at numpy datetime units now list the accepted values.
3. **pandas floor.** Per the maintainer decision on the inline thread, the pandas 2 pin on the oldest-pymc leg was removed. CI runs pandas 3 only; `pandas>=2.2` stays declared, with a comment in `pyproject.toml` saying it is not exercised by CI.
4. **Upper bound.** Left uncapped, matching the other runtime dependencies.
5. **Remaining `.loc` sites.** `expected_customer_lifetime_value` now uses `assign` and no longer adds `future_spend` to the caller's frame (verified that the old DataArray `.loc` assignment was positional, so `.to_numpy()` is behaviour-preserving), with a non-mutation test. `period_transactions["first"] = False` for the other site.
6. **Yaml coverage.** Added `test_build_mmm_from_yaml_does_not_mutate_X`, which feeds a string date column so the coercion actually runs.
7. **Smaller notes.** The `date_col_in_X` block is now a plain `if`/`else`. The PR description gained a behaviour-changes line for the release notes. The double copy on pandas 2 is accepted.

pymc_marketing/clv/utils.py
```diff
@@ -897,7 +901,9 @@ def _expected_cumulative_transactions(
     repeated_transactions = repeated_and_first_transactions[~first_trans_mask]
     first_transactions = repeated_and_first_transactions[first_trans_mask]
 
-    date_range = pandas.date_range(start_date, periods=t + 1, freq=time_unit)
+    date_range = pandas.date_range(
+        start_date, periods=t + 1, freq=_OFFSET_ALIASES.get(time_unit, time_unit)
+    )
```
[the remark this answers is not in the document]
juanitorduz: Fixed in 7c38984e: `time_unit` is normalised (`H` to `h`) at the start of every public CLV function via `_PERIOD_ALIASES`, so all `to_period` calls and the `np.timedelta64` divisions receive a valid alias. `_OFFSET_ALIASES` now only maps `M` to `ME` for the `date_range` call. Hourly regression tests cover `rfm_summary` and `_expected_cumulative_transactions` on both pandas majors.

.github/workflows/test.yml
juanitorduz: Done in 7c38984e: the `pandas<3` pin was removed from the oldest-pymc step, so every leg now resolves pandas 3.

daimon-pymclabs:
## Second pass on 7c38984e

All seven items from the previous review are addressed, and I re-checked each one against the branch rather than the description. One new substantive finding, and CI is not fully green yet.

### What I verified as fixed

- **Notebooks.** `to_pydatetime` is gone from the whole repo (`grep` over `*.py` and `*.ipynb` is clean). `notebooks (cvm, --notebooks mmm --end-idx 10)` now passes in 8m19s, which is the `mmm_example.ipynb` split that was expected to fail next. The `--start-idx 10` legs (the `mmm_counterfactuals.ipynb` split, the original hard failure) are still running as I write this.
- **Hourly `time_unit`.** Confirmed by loading `pymc_marketing/clv/utils.py` from this branch on pandas 3.0.5 / numpy 2.5.3: `rfm_summary`, `rfm_train_test_split` and `_expected_cumulative_transactions` all accept `"H"` and `"h"` and agree on the result. Splitting `_PERIOD_ALIASES` from `_OFFSET_ALIASES` is the right shape, and normalising at the entry points rather than at each call site means the `np.timedelta64` divisions get fixed for free, as the description claims.
- **`assign` in `expected_customer_lifetime_value`.** `to_xarray` builds its `customer_id` coord from `data["customer_id"]` in row order and does not sort, so `.to_numpy()` is positional in the same way the old DataArray `.loc` setitem was. The change is behaviour-preserving, and no docs notebook reads `future_spend` back off the input frame after the call, so the mutation drop is safe for the examples too.
- **Yaml coverage and the `if`/`else` cleanup.** `test_build_mmm_from_yaml_does_not_mutate_X` feeds a string date column, so it exercises the coercion rather than skipping it.
- **Docstrings.** The four numpy-datetime-units pointers are replaced with the actual accepted vocabulary.

### New finding: `time_unit="M"` raises in `rfm_summary` and `rfm_train_test_split`

This is the `"H"` problem from the last round with the roles reversed, and this PR's docstring rewrite newly advertises it. `clv/utils.py:410` and `:416` divide by `np.timedelta64(1, time_unit)`, and numpy has no unambiguous month duration:

```python
from pymc_marketing.clv.utils import rfm_summary
tx = pd.DataFrame({"id": [1, 1, 2, 2, 3],
                   "date": ["2024-01-05", "2024-03-10", "2024-01-20", "2024-05-02", "2024-02-11"]})
rfm_summary(tx, customer_id_col="id", datetime_col="date", time_unit="M")
```

```
ValueError: Unit M is not supported. Only unambiguous timedelta values durations are
supported. Allowed units are 'W', 'D', 'h', 'm', 's', 'ms', 'us', 'ns'
```

Same on pandas 2.3.3 / numpy 2.2.6, so it is pre-existing rather than a pandas 3 regression, and `rfm_train_test_split` raises identically because it goes through `rfm_summary`. `rfm_segments` inherits it as well. `customer_lifetime_value(time_unit="M")` is fine, because that path scales through `_DAYS_PER_MONTH` instead of a numpy divisor, and `_expected_cumulative_transactions(time_unit="M")` is fine because it goes through `to_period`, which is exactly why the monthly regression test added in the first commit passes and hides this.

The reason it matters for this PR specifically: the three docstrings rewritten here now read `'M' (monthly)` as an accepted value for functions that raise on it. Before, they pointed at the numpy datetime-unit page, which at least described numpy's vocabulary honestly.

Two ways out, either is fine by me:

1. Compute the recency and T deltas from period counts rather than a timedelta division, which is what `rfm_train_test_split` already does one screen down at `:616` (`test_period_end.to_period(time_unit) - train_period_end.to_period(time_unit)).n`). That keeps calendar-month semantics consistent with the `frequency` column, which counts `to_period("M")` buckets.
2. If monthly RFM is not worth fixing here, drop `'M'` from the `rfm_summary` / `rfm_train_test_split` / `rfm_segments` docstrings and raise a clear error, so the vocabulary matches reality. Same argument I made for `'H'` last round, and I would apply it consistently.

Either way, adding `"M"` to the two new `@pytest.mark.parametrize("time_unit", ["H", "h"])` lists is the test that would have caught this, and it belongs next to the hourly cases.

### CI is not green yet

- `notebooks (cvm | numba, --notebooks mmm --start-idx 10)` and `notebooks (numba, --notebooks mmm --end-idx 10)` are still pending. The blocker cannot be called cleared until at least the `--start-idx 10` pair reports.
- `codecov/patch` is red at 66.67% with 5 lines missing, but that report has 1 of 28 uploads and predates most of the test legs finishing, so I read it as incomplete rather than a real gap. Worth a second look once the matrix drains.

### pandas 2 is now completely untested

Removing the `pandas<3` pin from the oldest-pymc leg means no leg installs pandas 2 at all: `pymc==6.2.0` does not cap pandas, so that leg resolves 3.x like every other. Combined with `pandas-stubs>=3.0.3.260530` in the lint extra, neither the test suite nor mypy exercises the declared `pandas>=2.2` floor. I understand this is a maintainer decision and the `pyproject.toml` comment is honest about it, so I am not blocking on it. But the consequence is that "works on pandas 2 and 3" is now a local, one-off verification that cannot be repeated, and the two-version idioms in this diff (`assign` over `.loc` setitem) are exactly the kind of thing that quietly rots. If pandas 2 is not going to be tested, bumping the floor to `>=3` and dropping the compatibility framing would at least be self-consistent; if it is meant to stay supported, one `pandas==2.2.*` leg buys it back cheaply.

### Smaller notes

- `test_expected_cumulative_transactions_hourly_time_unit` asserts `(df_cum["predicted"].diff().dropna() > 0).all()`. I checked the `actual` expectations by hand against the fixture and they are right (5 repeat transactions land in hours 3, 22, 29, 33 and 36, and the 08:00 day-3 repeat correctly falls outside the 48-hour window). The strict inequality on `predicted` is the only assertion here that depends on float increments staying above zero at hourly granularity; `>= 0` with a separate check that the series ends above where it starts would be less brittle.
- `clv/utils.py:34` still lists `"D"`, `"W"` and `"H"` in the `_DAYS_PER_MONTH` comment while the dict below it is now keyed `"h"`. One-word drive-by.
- Not regenerating the notebook outputs is the right call for a one-token plotting fix, but it does leave the rendered docs showing plots produced under pandas 2.3.3. Fine by me, flagging it only so it is a decision on the record rather than an oversight.
- The out-of-scope note about `_convert_frequency_to_timedelta` is accurate and I would keep it out of this PR. For the follow-up issue: both call sites feed it `pd.infer_freq(...)`, which returns `h`, `min` and `s` on pandas 2.2 and later, so sub-daily data does not hit the `"H"` / `"T"` / `"S"` branches at all and silently lands in the "Defaulting to weeks" fallback with a `UserWarning`.

Happy to approve once the mmm notebook legs are green and the monthly `time_unit` question is settled either way.

juanitorduz:
@bugbot review

juanitorduz:
Thanks for the second pass. Addressed in 38d7ee63.

**Monthly `time_unit`.** Went with option 1. `rfm_summary` now counts recency and T as period-ordinal differences (`to_period(time_unit)` on the aligned timestamps) instead of dividing by `np.timedelta64(1, time_unit)`, which is what `rfm_train_test_split` already did for `test_T` and what the `frequency` column already counts. Because every timestamp in that path is aligned to a period start, the daily, weekly and hourly values are unchanged (checked on cdnow for `D`, `W`, `D` with `time_scaler=7`, a string `observation_period_end`, and a `pd.Period` observation end: identical to the old output), and `M` now yields calendar months. `rfm_train_test_split` and `rfm_segments` inherit the fix.

**Regression tests.** `test_rfm_summary_time_unit` parametrizes your five-row frame over `D`, `W`, `M` and `H` with hand-checked recency and T per unit, so the `M` row is the regression test and the other rows pin the no-change claim. `test_rfm_train_test_split_monthly_time_unit` covers the split with `time_unit="M"` (`test_T == 2`). The whole `tests/clv` package passes on pandas 3.0.5 and `tests/clv/test_utils.py` on pandas 2.3.3 with FutureWarning promoted to error.

**Smaller notes.** The hourly `predicted` assertion is now `>= 0` plus "ends above where it starts". The `_DAYS_PER_MONTH` comment says `"h"`. Not regenerating the notebook outputs is deliberate.

**pandas 2.** Understood, and I agree it is a real trade-off. The maintainer decision on this PR is to run CI on pandas 3 only and keep `pandas>=2.2` declared with the comment saying it is not exercised; I am leaving that as is rather than re-litigating it here.

daimon-pymclabs (approved):
## Final pass on 38d7ee63 — approving

The monthly `time_unit` finding from the previous round is fixed the way I would have fixed it (option 1: count recency and T in calendar periods instead of dividing by a numpy duration), and the original pandas 3 blocker is confirmed cleared in CI. Nothing left blocking from my side.

### The blocker is genuinely cleared

Every notebook leg on `7c38984e` reported success, including `notebooks (cvm | numba, --notebooks mmm --start-idx 10)`, which is the `mmm_counterfactuals.ipynb` split that failed on the first commit. All 52 check runs on that commit were green (the two skipped ones are the PyPI upload jobs). So the `to_pydatetime` removal is verified end to end on pandas 3, not just inferred from the `--end-idx 10` leg.

### The monthly fix is behaviour-preserving where it matters

I checked this rather than reading it, by loading `clv/utils.py` from both `7c38984e` and `38d7ee63` side by side on pandas 3.0.5 and diffing `rfm_summary` outputs over 4000 synthetic transactions across 400 customers, for `time_unit` in `{"D", "W", "h", "H"}` x `time_scaler` in `{1, 7}` x `include_first_transaction` in `{False, True}` x `observation_period_end` in `{None, "2025-02-14 13:37"}`. All 32 combinations are frame-for-frame identical (`DataFrame.equals`, so dtypes too). That is the property the change needs: `_find_first_transactions` already returns period-start-aligned timestamps, so the ordinal difference and the timedelta division agree exactly for the fixed-length units.

The new `test_rfm_summary_time_unit` expectations are also right. I recomputed all four rows by hand off the fixture and they match, including the hourly `2832` / `2472` / `1944` and the monthly `[2, 4, 0]` / `[4, 4, 3]`. `rfm_train_test_split(time_unit="M")` returns the documented `test_T` of 2.0 for all three customers and `test_frequency` of 1 for customer 2, and `rfm_segments(time_unit="M")` now runs instead of raising. Parametrising over `"D"`, `"W"`, `"M"` and `"H"` rather than adding a monthly-only case is the better version of what I suggested, since it pins the equivalence rather than just the absence of an exception.

Two side benefits worth having on the record, both improvements rather than regressions:

- Timezone-aware input with DST in the window no longer produces a fractional `T`. A day-unit span across the March DST transition used to divide 23 or 25 real hours by a flat 24; it now counts calendar days. (In practice both versions agree here because `to_period` drops the timezone first, which I confirmed, so this is latent rather than active.)
- `recency` and `T` are now integer-valued multiples of the period for every code path, which matches how `frequency` has always been counted.

### One behaviour change I found, which I read as a fix

When `observation_period_end` is passed as a `pandas.Period` whose freq is finer than `time_unit`, `T` changes:

```python
rfm_summary(tx, "id", "date", time_unit="W", observation_period_end=pd.Period("2024-05-17", freq="D"))
# 7c38984e: T = [19.571428, 17.571428, 14.571428]
# 38d7ee63: T = [19.0, 17.0, 14.0]
```

This is the only difference I could find between the two commits, and it only affects the `isinstance(observation_period_end, pandas.Period)` branch, because that branch calls `.to_timestamp()` without re-flooring to `time_unit`. Passing the same instant as a string or a `datetime` already floored on both commits, so the Period branch was the inconsistent one. Flooring it agrees with the string branch and with the whole-periods semantics above, so I would keep the new behaviour. Flagging it only because it is a silent numeric change to a model input for anyone hitting that branch, and it is the kind of thing worth one line in the release notes.

### Also confirmed fixed from the last round

- `_DAYS_PER_MONTH` comment now says `"h"` instead of `"H"`.
- `test_expected_cumulative_transactions_hourly_time_unit` relaxed to `>= 0` plus an explicit end-above-start assertion, which is the shape I asked for.

### Approval condition

CI on `38d7ee63` is entirely pending, since the commit landed a few minutes ago. This commit touches only `clv/utils.py` and `tests/clv/test_utils.py`, and my local equivalence check covers the behavioural risk in it, so I am approving now rather than gating on the matrix. Please just confirm the run is green before merging, in particular the `clv` test shards and the `notebooks --notebooks clv` legs, since the CDNOW walkthrough tests compare against reference values from the R BTYD package and are the only thing my synthetic comparison does not reproduce.

I am not re-raising the untested pandas 2 floor. It is a maintainer call, it is documented honestly in `pyproject.toml`, and it should not hold up this PR.

daimon-pymclabs (approved):
## Third pass on `4a9821d3`: approving

Both open items are closed and I checked the monthly fix independently rather than reading the description.

**Monthly `time_unit`.** Option 1 is the right choice and the implementation is correct. I reimplemented the period-ordinal path standalone on pandas 3.0.5 against the five-row frame from my last comment and compared it to the old `np.timedelta64` division:

| `time_unit` | old | new |
|---|---|---|
| `D` | recency 65 / 103 / 0, T 118 / 103 / 81 | identical |
| `W` | recency 9 / 15 / 0, T 17 / 15 / 12 | identical |
| `h` | recency 1560 / 2472 / 0, T 2832 / 2472 / 1944 | identical |
| `M` | `ValueError: Unit M is not supported` | recency 2 / 4 / 0, T 4 / 4 / 3 |

So the no-change claim for the fixed-duration units holds exactly, and the monthly numbers are the calendar counts they should be (customer 1 Jan to Mar is 2 months with an observation end in May giving T = 4). The reason it is exact rather than approximately equal is worth having in the comment, which it now is: `_find_first_transactions` periodizes and calls `.to_timestamp()` before the `min`/`max` aggregation, so every timestamp reaching the ordinal difference is already on a period boundary and there is no sub-period remainder for the division to have counted. `rfm_train_test_split` and `rfm_segments` inherit it, and `test_rfm_summary_time_unit` parametrizing over `D`/`W`/`M`/`H` pins both halves of that: the `M` row as the regression and the others as the no-change guard.

**CI.** Green on the head commit: 55 success, 2 skipped, no failures, including both `notebooks (cvm | numba, --notebooks mmm --start-idx 10)` legs that were the original blocker, and `codecov/patch` now passes. The two failures on `f68da88b` were the supersede cascade from the next push rather than anything real. Cursor Bugbot passes.

**Smaller notes** all addressed: `>=` plus the ends-above-start check on the hourly `predicted` series, and the `_DAYS_PER_MONTH` comment matching the `"h"` key.

**pandas 2.** Your call, and I am not relitigating it. Recording the consequence once so it is in the history rather than in this thread: with no leg resolving pandas 2, the declared `pandas>=2.2` floor and the two-version idioms in this diff are unverified from here on, so a future change that breaks pandas 2 will merge green. If that floor is meant to hold, the cheapest insurance later is one `pandas==2.2.*` job; if it is not, bumping to `>=3` would make the metadata honest. Neither belongs in this PR.

The `_convert_frequency_to_timedelta` follow-up is still worth an issue. I can open it with the `pd.infer_freq` detail from my last comment if you have not already.

PR #2954, opened 2026-09-07
Case study/lift test roas information text

juanitorduz:
Add text to the original notebook by @TeemuSailynoja https://github.com/pymc-labs/pymc-marketing/commit/ca637b39e796ff740c2f6f1b22fd5dba5506531e

----
📚 Documentation preview 📚: https://pymc-marketing--2954.org.readthedocs.build/en/2954/


daimon-pymclabs (approved):
Read through the notebook and the gallery wiring. The methodology write-up is clear and, as far as I can check, accurate: the pre-posterior framing is correct, the four steps map cleanly onto the code, and the "when does the shortcut break?" admonition is the right caveat to lead with. Approving.

Verification I ran, so it's on the record:

- `{ref}` targets resolve: `(mmm_case_study)=` and `(mmm_lift_test)=` both exist as labels in their notebooks.
- `{meth}` targets resolve: `Incrementality.contribution_over_spend` (`pymc_marketing/mmm/incrementality.py:1501`) and `MMM.add_lift_test_measurements` (`pymc_marketing/mmm/mmm.py:3436`).
- `gallery.md` and `gallery.yaml` are in sync, so `generate_gallery.py --check` should pass. The missing `images/mmm_roas_experimentation.png` is fine, thumbnails are extracted at build time.
- Watermark cell is last and carries all of `-n -u -v -iv -w`.
- `dollarmath`, `amsmath` and `colon_fence` are enabled in `conf.py`, so the `$$` blocks and the grid card render.
- Prose figures match the committed outputs: mean 52.34, sd before 36.63, sd after 15.19, 58.53 percent, central interval 18.98 to 27.36.

Minor items, none blocking:

1. **The runtime claim contradicts the committed output.** The helper description says "the loop over 1000 experiments below takes a few minutes", but the `%%time` output on that cell reads `Wall time: 16.4 s`. Worth changing to "about 15 seconds" or dropping the estimate, otherwise the first thing a reader checks disagrees with the text.

2. **`k_hat` is never shown.** The admonition tells the reader to "check the `k_hat` value stored for each simulated experiment (values above 0.7 signal unreliable weights)", but no cell surfaces it. A one-liner after the simulation cell, e.g. `max(e["k_hat"] for e in experiments)`, would let the reader confirm the diagnostic passes for this configuration instead of taking it on trust. It also demonstrates the check the admonition is asking them to perform on their own data.

3. **The sign on `psislw` reads like a bug.** `psislw(-Normal(mu=lift_obs, sigma=sigma).logpdf(...))` is correct here (the results concentrate the way proper likelihood weights would, and degenerate 1/likelihood weights could not produce a clean 59 percent reduction with a tight interval), but the leading minus follows the PSIS-LOO convention rather than the "weights proportional to the likelihood" story the markdown just told. One inline comment saying why the negation is there would save the next reader the detour I just took.

4. **The tutorial teaches private attributes.** `mmm.scalers._target` and `mmm.scalers._channel` appear in the helper and, more importantly, in the admonition as something the reader should compute themselves. Underscore-prefixed attributes in a public tutorial tend to break quietly on refactor. If there is no public accessor, a short note that this is internal and may move would help; if one is planned, this notebook is a good argument for it.

5. **Cosmetic:** `import inspect` inside `sample_lift_experiments` shadows the top-level import in the Prepare Notebook cell.

6. **Optional:** the six hardcoded figures in the prose (52, 37, 15, 59 percent, 19 and 27) all match today's outputs, but they will drift silently if the notebook is ever re-executed against a different sampler or seed. Not worth restructuring for, just noting that a future re-run needs a prose pass too.

PR #2959, opened 2026-09-08
Case study: How Much Would One Lift Test Reduce ROAS Uncertainty?

TeemuSailynoja:
## Description

Adds a new case study notebook: "How Much Would One Lift Test Reduce ROAS Uncertainty?"

This notebook demonstrates a pre-posterior analysis for MMM experimentation planning – answering the question "before we run a lift test, how much information gain should we expect for our ROAS estimates?"

### What it does

Using the fitted case-study MMM, the notebook:

1. Defines ROAS as incremental contribution over spend (via Incrementality.contribution_over_spend), with adstock carry-over included.
2. Simulates a hypothetical lift test on one channel (e.g., Online Display at 50k → 100k weekly spend) and computes the implied lift for every posterior draw.
3. Updates the posterior via importance sampling (PSIS-weighted resampling) instead of refitting — the test likelihood acts as a single extra term in Bayes' rule.
4. Averages over unknown test outcomes by drawing plausible lifts from the prior predictive, then reports the expected reduction in ROAS standard deviation per channel.

In the case study, one lift test on Online Display is expected to cut its ROAS standard deviation by ~56% while leaving other channels unaffected.

## Checklist
 - [x] Each commit corresponds to a relevant logical change

----
📚 Documentation preview 📚: https://pymc-marketing--2959.org.readthedocs.build/en/2959/


daimon-pymclabs (commented):
Nice notebook, the pre-posterior framing is clear and the four-step structure reads well. A few things I'd fix before merge.

**1. The spend levels in the prose don't match the code.** "Choose the Experiment" says the test moves Online Display "from 50,000 to 100,000 per week" (and the PR description says the same), but the code is

```python
test_budget = 50_000
start_spend = 100_000
```

and the call is `sample_lift_experiments(1_000, start_spend, start_spend + test_budget, ...)`, i.e. 100,000 → 150,000. Either swap the two values or fix the sentence. The follow-on claim that this is "at the low end of the spend levels seen in the training data" should be re-checked against whichever pair you keep.

**2. Broken sentence in the Conclusion.** "The standard deviation is one criterion. Expected information gain, are alternatives that the same simulation supports." Looks like a lost clause. Also, the text above the first plot says the Online Display SD is "about 37" while the results section says "about 36" for the same number (the table gives 36.25) — worth making those agree.

**3. You can drop the resampling noise entirely.** Both the results text and the summary discussion caveat that the near-zero reductions for untested channels are mostly resampling noise. The post-test SD is a weighted moment, so you can compute it directly from the PSIS weights instead of drawing 1000 indices:

```python
w = weights / weights.sum()
m = (roas * w).sum(("chain", "draw"))
sd = np.sqrt((roas**2 * w).sum(("chain", "draw")) - m**2)
```

That makes the untested channels sit exactly where the weighting puts them rather than at zero-plus-noise, which strengthens the "the transfer is negligible" claim, and it removes the need to justify `num_samples=1000` against 4800 draws. Keep `resample_posterior` for the forest plot, where you do want actual draws.

**4. `resample_posterior` copies every posterior variable inside a 1000-iteration loop.** Only `ROAS` is consumed by `stds`, and the notebook has `channel_contribution`, the original-scale contributions and the seasonality terms in there too. A `var_names` argument defaulting to all variables would cut the 36 seconds a long way, and the sentence explaining why the loop is slow could then go.

**5. The `k_hat` check cell is bare.** `[e["k_hat"] for e in experiments if e["k_hat"]>0.7]` renders as `[]` with nothing telling the reader that empty is the pass condition. Something like printing `max(k_hat)` alongside the count over 0.7 makes the diagnostic self-explanatory in the rendered docs. That cell is also not ruff-formatted (missing spaces around `>`), which suggests the linting box on the checklist may not be reflecting the current state of the notebook.

**6. `from scipy.stats import Normal`** is the new distribution-object API, SciPy 1.15 and up. Worth confirming the docs environment floors SciPy that high, otherwise `scipy.stats.norm(loc=..., scale=...).logpdf(...)` is the safer call here.

**7. Minor: the PSIS sign comment.** `psislw(-Normal(...).logpdf(...))` with a note that "psislw expects negative log importance weights" reads inverted against the `az.stats.psislw` convention, where LOO passes `-log_likelihood` precisely because its target weights go as one over the likelihood. The results behave the way correct weighting should (only the tested channel sharpens, and its mean moves toward the simulated result), so I assume the `arviz.base` entry point takes the negated convention. A link to that in the comment would save the next reader the same double-take.

**8. Minor: reproducibility.** `sample_lift_experiments` and `resample_posterior` both close over the module-level `rng`, so re-running just the "Run the Simulations" section gives different numbers than the rendered output. A `random_seed` or `rng` argument would make the section rerunnable on its own.

Nothing here is structural, the method and the write-up are sound. Items 1 and 2 are the ones that need fixing.

juanitorduz (approved):

Issue #2966, opened 2026-09-08
docs: BTYD logp is not a density over recency (missing t_x^(x-1)/Gamma(x) factor)


williambdean:
Found while validating a vectorized `rng_fn` for `BetaGeoNBD` (perf PR to follow). Short version: the likelihood is correct for inference, but `exp(logp)` is not a density over recency, so anything that integrates or interprets it as one is off.

## Expected behavior

Integrating `exp(logp(t, x))` over recency recovers the probability of `x` repeat purchases:

```
P(X = x) = integral of exp(logp(t, x)) dt   over t in [0, T]
```

## Actual behavior

For the BTYD models the likelihood (Fader, Hardie, Lee 2005 convention, also used by `lifetimes`) equals the joint density of `(recency, frequency)` divided by `t_x^(x-1) / Gamma(x)`:

* [BetaGeoNBD.logp](https://github.com/pymc-labs/pymc-marketing/blob/f86e1229dce26554be4e3b57a7c49ee94bd75e79/pymc_marketing/clv/distributions.py#L465-L512)
* [ModifiedBetaGeoNBD.logp](https://github.com/pymc-labs/pymc-marketing/blob/f86e1229dce26554be4e3b57a7c49ee94bd75e79/pymc_marketing/clv/distributions.py#L616-L662)
* [ParetoNBD.logp](https://github.com/pymc-labs/pymc-marketing/blob/f86e1229dce26554be4e3b57a7c49ee94bd75e79/pymc_marketing/clv/distributions.py#L143-L207)

The factor depends only on the observed data, so it is constant with respect to the parameters: fits, posteriors, and predictions are all unaffected. But `exp(logp)` cannot be treated as a density. The correct relation is:

```
P(X = x) = integral of exp(logp(t, x)) * t^(x-1) / Gamma(x) dt   over t in [0, T]
```

Verified numerically for `BetaGeoNBD`, `ModifiedBetaGeoNBD`, and `ParetoNBD` (same factor for all three).

## Steps to reproduce

Save as `mre.py`:

```python
import numpy as np
import pytensor
import pytensor.tensor as pt
import pymc as pm
from scipy.special import gamma as gamma_fn

from pymc_marketing.clv.distributions import BetaGeoNBD

a, b, r, alpha, T = 0.793, 2.426, 0.243, 4.414, 78.0
N = 200_000

# 1. simulate (recency, frequency) pairs from the model itself
rec, freq = pm.draw(
[...]

PR #2967, opened 2026-09-08
perf: vectorize BetaGeoNBD rng_fn

williambdean:
Prior and posterior predictive sampling for the BG/NBD model spent all its time in a per-customer Python simulation loop. This replaces it with an exact vectorized decomposition, so predictive sampling is ~7x faster at CDNOW scale (0.72s -> 0.097s for 100 draws x 2357 customers) and, more importantly, no longer scales with the expected number of purchases: the old cost grows with E[X] (0.72s -> 4.92s on a heavy-purchase parameter set) while the new one stays flat around 0.1-0.2s. Interactive prior predictive workflows (e.g. marimo/modist notebooks that re-sample on every widget drag) go from minutes to sub-second.

The identity is exact, not an approximation: the purchase count is `X = min(K, N_T)` with `K ~ Geometric(p)` and `N_T ~ Poisson(lam * T)`, and the recency is an inverse-CDF sampled truncated Erlang (churn-limited) or a `T * Beta(N_T, 1)` order statistic (time-limited). There is no truncation parameter to get wrong.

Same seed now produces different (but identically distributed) draws since the RNG consumption order changed. No test pinned exact `rng_fn` values. Distributional equivalence was validated against the previous implementation, which is kept as a test oracle in `tests/clv/test_distributions.py`, plus exact analytic checks for `P(X = 0)` and `E[X]`, invariants, broadcasting shapes, and a parameter recovery parity exercise between the two samplers.

While cross-validating the sampler against its own logp, a convention question about the BTYD likelihood surfaced and is reported separately in #2966.

Refs #2966

----
📚 Documentation preview 📚: https://pymc-marketing--2967.org.readthedocs.build/en/2967/

### DIFF
--- a/pymc_marketing/clv/distributions.py
+++ b/pymc_marketing/clv/distributions.py
@@ -22,6 +22,7 @@ from pymc.distributions.shape_utils import rv_size_is_none
 from pytensor import scan
 from pytensor.graph import vectorize_graph
 from pytensor.tensor.random.op import RandomVariable
+from scipy.special import gammainc, gammaincinv
 
 __all__ = [
     "BetaGeoBetaBinom",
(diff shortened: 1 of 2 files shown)


juanitorduz:
## Review

Checked out `perf/vectorize-bgnbd-rng` at `48c8135` and verified this independently. **The identity is exact and the perf claims hold.** A few nits below, none blocking.

### Correctness — verified, including the joint law

I re-derived the decomposition against the old loop rather than trusting the marginals:

- The old loop draws the churn Bernoulli *after* accepting a purchase, so the churn index is `K ~ Geometric(p)` on `{1,2,...}` and `X = min(K, N_T)`. ✔
- Churn-limited branch: conditioning on `K = k` the acceptance event is exactly `{N_T >= k} = {t_k <= T}`, and `P(N_T >= k) = gammainc(k, lam*T) = cdf_T`. The truncated-Erlang density then cancels `cdf_T` against the acceptance probability, so `P(K=k) * cdf_T * truncErlang = P(K=k) * Erlang`. The draw-then-condition ordering is legitimate, not an approximation. ✔
- Time-limited branch: given `N_T = n` the arrivals are uniform order statistics independent of `K`, so `t_n = T * Beta(n, 1)`. ✔

The added tests check the `X` marginal (chi²) and the `recency | X > 0` marginal (KS) *separately*, which would not catch a swap between the two branches. So I checked the **joint**: 200k draws per parameter set, old vs new, two-sample KS on `recency | X = k` for `k ∈ {1,2,3,5,10}` across all three parametrizations in the test matrix — 13 comparisons, min p-value 0.11, conditional means agreeing to ~0.5%. Also confirmed `rec <= T` with no `rec == T` boundary hits in 500k draws (and `logp` uses `pt.gt(t_x, T)`, so `T` would be in support anyway).

**Suggestion:** promote one of those conditional checks into `test_rng_fn_matches_reference` — e.g. KS on `rec[x == 1]` — so the branch assignment is actually pinned. Cheap: the whole `-k rng_fn` set runs in 2.16s.

### Perf — reproduced

M4 Pro, min of 3 reps, `rng_fn` called directly:

| case | old | new | speedup |
|---|---|---|---|
| `(0.793, 2.426, 0.243, 4.414, T=78)`, 100×2357 | 0.318s | 0.043s | 7.4× |
| `(0.5, 10, 2, 2, T=50)` (E[X]≈27), 100×2357 | 2.233s | 0.067s | 33× |

Flat in `E[X]` as claimed.

### Nits

1. **The time-limited branch can reuse `u`.** `Beta(n, 1) == U^(1/n)`, and `u` is already drawn and unused for those entries:
   ```python
   recency[time_limited] = T[time_limited] * u[time_limited] ** (1.0 / N_T[time_limited])
   ```
   Drops an RNG call and an allocation, and makes both branches inverse-CDF transforms of the same uniform, which reads more uniformly. I validated the substitution distributionally.

2. **Both `np.maximum(..., 1)` guards are dead code.** `rng.geometric` returns `>= 1` by construction, and `time_limited` already implies `X > 0` hence `N_T >= 1`. They read as if a real edge case exists.

3. **`p_safe` deserves one more line of comment.** Clipping to `np.finfo(np.float64).tiny` looks like it should blow up — `ceil(-log1p(-u) / log1p(-2.2e-308))` is ~1e307, far past `INT64_MAX`. It is safe only because NumPy explicitly saturates: `random_geometric_inversion` in `distributions.c` returns `INT64_MAX` when `z >= 9.223372036854776e+18` (present since well before the `numpy>=2.0` pin). That saturated `K` then makes `X = N_T`, which *is* the correct `p -> 0` limit. Confirmed empirically with `a=1e-8, b=1e8` (100% of Beta draws underflow to exactly 0.0): no crash, no negative `X`, `E[X] = E[N_T]`. Worth saying "NumPy saturates geometric at INT64_MAX, which yields X = N_T, the p→0 limit" so a future reader doesn't 'fix' it.

4. **New failure mode, informational.** `rng.poisson` raises `ValueError: lam value too large` once `lam * T > 9.22e18`; the old loop would instead spin ~`lam*T` Python iterations, i.e. hang. Needs something like `alpha ~ 1e-18` to hit, and failing loudly beats hanging, so I would not change it — just noting the change.

5. **`ModifiedBetaGeoNBDRV.rng_fn` (`distributions.py:540`) still has the loop.** The same decomposition applies with `K = Geometric(p) - 1` (MBG checks churn before purchase 1). I implemented it and cross-checked against the current MBG loop: `P(X=0)`, `E[X]`, and `recency | X = k` for k∈{1,2,3} all match across the same three parametrizations. Obvious follow-up if you want it in a separate PR.

6. **Description/diff mismatch.** The body lists "broadcasting shapes, and a parameter recovery parity exercise" as part of the added validation; neither is in the diff. Shapes are covered by the pre-existing `test_bg_nbd_sample_prior`, so this is just the description overselling slightly.

### Checks run

- `pytest tests/clv/test_distributions.py -k BetaGeoNBD` → 30 passed.
- `pytest tests/clv/models/test_beta_geo.py` → 40 passed, 3 skipped, 2 failed with `ModuleNotFoundError: No module named 'beartype'` — local env gap, unrelated to this PR.
- `pre-commit run --files pymc_marketing/clv/distributions.py tests/clv/test_distributions.py` → all hooks pass (ruff, ruff-format, mypy).

LGTM modulo the nits; #1/#2/#3 are quick and #5 is the natural next step.

juanitorduz (approved):
LGTM (minor nits from the agent)

PR #2969, opened 2026-09-08
fix: render the model graph in a temporary directory

### LINKED ISSUE #2968: log_model_graph leaves the DOT source in the working directory when graphviz is not installed
`autolog` writes its artifacts into the current working directory and deletes each one after logging. `log_model_graph` is the one case where the delete can be skipped.

`graph.render(path)` (`pymc_marketing/mlflow.py:553`) writes the DOT source before it invokes `dot`. Without the graphviz binary the render raises, the `except` logs and returns, and the `os.remove(path)` at line 560 is never reached. The call site passes a relative path, `log_model_graph(model, "model_graph")` at line 627, so the leftover lands wherever the process happens to be.

Repro on `main`, in an empty directory:

```python
import logging

import numpy as np
import pymc as pm

from pymc_marketing.mlflow import log_model_graph

logging.basicConfig(level=logging.INFO)

with pm.Model() as model:
    mu = pm.Normal("mu", mu=0, sigma=1)
    pm.Normal("obs", mu=mu, sigma=1, observed=np.ones(3))

log_model_graph(model, "model_graph")
```

```
INFO:pymc_marketing.mlflow:Unable to render the model graph. failed to execute PosixPath('dot'), make sure the Graphviz executables are on your systems' PATH
$ ls
model_graph
```

The file is untracked and no `.gitignore` entry covers it, so it is easy to sweep into a commit (been there, done that). `.gitignore` line 9 already carries `*.nc` under `# InferenceData` for the same class of leftover.

Rendering into a `tempfile.TemporaryDirectory` and logging from there would keep every artifact out of the user's directory, not only this one. `summary.html` (line 1386) and `idata.nc` (line 1426) take the same route and survive only because nothing between the write and the delete raises. `tempfile` is already imported at line 193 and used that way at line 1167.

anevolbap:
Fixes #2968

## Description

`graph.render` saves the DOT source before it invokes `dot`, and the `os.remove` cleanup sat after the call. Without the graphviz binary the render raises and the source file stays in the working directory, where no `.gitignore` pattern covers it.

The render now runs inside a `tempfile.TemporaryDirectory`, so neither the DOT source nor the PDF reaches the working directory and no explicit cleanup is needed. This also covers a failure in `mlflow.log_artifact` itself, which used to leave both files. The logged artifact keeps its name, `model_graph.pdf`.

`path` now names the artifact instead of a filesystem location, and the docstring says so. Nothing observable changes: the file was deleted right after logging, and `mlflow.log_artifact` uses the basename.

`log_arviz_summary` and `log_inference_data` write `summary.html` and `idata.nc` the same way and survive only because nothing between the write and the delete raises. Should they move into a temporary directory too, in a follow-up?

The `dot_missing` case of `test_log_model_graph_no_graphviz` patches `graphviz.backend.rendering.render`, so `save` still writes the DOT source before the failure. The test runs in `tmp_path` and checks that no `model_graph*` file is left, for every case. The `dot_missing` case fails on `main` and passes here.

## Related Issue

- [x] Closes #2968

## Checklist


anevolbap:
@juanitorduz this is ready for review when you have time. There is one open question in the description: should `summary.html` and `idata.nc` also move into a temporary directory, in a follow-up PR?

PR #1989, opened 2025-10-09
fix notebook runner action

williambdean:
## Description

Aim to fix the failing actions for notebooks

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://www.pymc-marketing.io/en/latest/contributing/index.html). Feel free to comment [`pre-commit.ci autofix` to auto-fix](https://pre-commit.ci/#configuration-autofix_prs).
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

----
📚 Documentation preview 📚: https://pymc-marketing--1989.org.readthedocs.build/en/1989/

PR #2914, opened 2026-08-24
feat(ci): add LLM-based PR release notes labeler

williambdean:
## What this does

Adds a Claude Code-based labeling step that classifies every new PR into one of the release-note categories defined in  (major, deprecation, enhancement, bug, docs, maintenance) so that the automatically-generated release notes stop dumping unlabeled PRs into **Maintenance 🔧** via the catch-all .

## Design

| Decision | Detail |
|---|---|
| **Trigger** |  only — no re-labeling on pushes |
| **Eligibility** | Drafts skipped; labeled when readied. Bot PRs skipped (already handled by Bot PRs workflow). Fork PRs from unaffiliated authors skipped (GitHub withholds secrets/write tokens for those anyway). |
| **Trust gate** | Same-repo PRs always proceed. Fork PRs only if author association ∈ OWNER/MEMBER/COLLABORATOR/CONTRIBUTOR (tested on 40 recent PRs: PabloRoque + anevolbap use forks despite being MEMBER/CONTRIBUTOR). |
| **Agent prompt** | Prefix-as-hypothesis, content-as-verdict. Semantic title prefixes (`feat:`, `fix:`, ...) are checked against the actual diff and file patterns. Explicit override rules for common mislabels (fix: on docs-only, chore: with user-facing behavior, deprecation vs removal-of-deprecated-API). |
| **Fallback** | `maintenance` (appears in notes) vs `no-releasenotes` (excluded). |
| **Security** | Labels filtered against allowlist before `gh pr edit`. Labels reach bash only via `env:` indirection — never inline interpolation. |
| **Comment** | Sticky comment via hidden `<!-- pr-labeler -->` marker; edit-in-place on re-runs. Collapsed `<details>` block, capped at ≤3 sentences. |

## Validation

- [x] Pre-commit passes on both files
- [x] Extraction pipeline tested against hostile inputs (injection payloads filtered, valid labels survive)
- [x] `-F body=@file` syntax verified via live markdown-echo probe
- [ ] Replay test against 8 historical mislabeled PRs — blocked locally (no `ANTHROPIC_API_KEY`); validate via first real PR in CI
- [ ] Live smoke test on next PR after merge

## Known follow-up

- `triage.yml` has the same `${{ }}` injection bug (shell-interpolating LLM output). Separate PR recommended.

----
📚 Documentation preview 📚: https://pymc-marketing--2914.org.readthedocs.build/en/2914/

PR #2971, opened 2026-09-08
ci: run notebook tests on mmm notebook changes

### LINKED ISSUE #2939: Notebook CI path filters exclude docs/source/notebooks/mmm/**
All three path filter blocks in `.github/workflows/test_notebook.yml` exclude `docs/source/notebooks/mmm/**.ipynb`, so a PR that only changes mmm notebooks (for example #2936) never triggers the mocked notebook workflow; the local runner is the only gate, and a broken mmm notebook is first caught on some later unrelated PR. Flagged in the review of #2936 (https://github.com/pymc-labs/pymc-marketing/pull/2936#pullrequestreview-5068287770, item 11).

Either include mmm notebooks in the trigger paths (accepting the CI cost per mmm-notebook PR) or document the exclusion and the expectation that authors run `scripts/run_notebooks/runner.py` locally.

shivamlalakiya:
Closes #2939.

## The exclusion is orphaned, not a policy

The issue frames this as a choice between including mmm notebooks in the trigger paths and documenting the exclusion. The history says it is neither: the exclusion lost its reason eleven months ago and nobody removed it.

- #1986 (`41c53e29`) renamed this workflow to "Test Other Notebooks", added a sibling `.github/workflows/test_notebook_mmm.yml`, and added `!docs/source/notebooks/mmm/**.ipynb` here so the two workflows would not run the same notebooks twice.
- #1989 (`9af7f676`), one day later, deleted `test_notebook_mmm.yml`, renamed this workflow back to "Test Notebooks", and folded the mmm runs into its own matrix:

```yaml
- "--notebooks mmm --end-idx 10"
- "--notebooks mmm --start-idx 10"
```

The path exclusion was left behind. Since then this workflow has run every mmm notebook without ever being triggered by a change to one.

## What this does not cost

The mmm splits already run on every PR that touches `pymc_marketing/**`, `tests/**.py`, `pyproject.toml`, `scripts/run_notebooks/**.py`, or this workflow file. Removing the exclusion does not add a job to those runs. It only adds coverage for the case the issue describes: a PR that changes an mmm notebook and nothing else, which today is gated solely by whoever remembers to run `scripts/run_notebooks/runner.py` locally.

`!docs/source/notebooks/*/dev/**.ipynb` is untouched. Dev notebooks stay excluded.

## Verification

GitHub path filters cannot be dry-run locally, so this is not verified end to end here. `.github/workflows/test_notebook.yml` is itself in the trigger paths, so this PR runs the full matrix and shows the workflow still works after the edit. The behaviour this fixes is only observable on the next PR that changes an mmm notebook and nothing else.

No release note: the issue carries `no releasenotes`.

----
📚 Documentation preview 📚: https://pymc-marketing--2971.org.readthedocs.build/en/2971/

### DIFF
--- a/.github/workflows/test_notebook.yml
+++ b/.github/workflows/test_notebook.yml
@@ -10,7 +10,6 @@ on:
      - ".github/workflows/test_notebook.yml"
      - "scripts/run_notebooks/**.py"
      - "docs/source/notebooks/**.ipynb"
-     - "!docs/source/notebooks/mmm/**.ipynb"
      - "!docs/source/notebooks/*/dev/**.ipynb"
   push:
     branches: [main, v1.0.0]
@@ -21,7 +20,6 @@ on:
      - ".github/workflows/test_notebook.yml"
      - "scripts/run_notebooks/**.py"
      - "docs/source/notebooks/**.ipynb"
-     - "!docs/source/notebooks/mmm/**.ipynb"
      - "!docs/source/notebooks/*/dev/**.ipynb"
 
 concurrency:


shivamlalakiya:
The red `label` check here is not from this diff. It is the eligibility gate in `pr-labeler.yml` (added in #2914) running the labeler step on fork PRs whose author is a `CONTRIBUTOR`, where `ANTHROPIC_API_KEY` is unavailable:

```
env:
  ANTHROPIC_API_KEY:
  GH_TOKEN: ***
Not logged in · Please run /login
```

Same failure on #2969 today. #2972 fixes the gate and its own `label` check now reports `Skipping LLM labeling: fork PR (CONTRIBUTOR); secrets are unavailable to pull_request runs from forks`. Nothing to change on this branch.

Once #2972 merges, this PR needs a draft/ready-for-review toggle to clear the check: the labeler triggers on `[opened, ready_for_review]`, so a plain re-run of the failed job will not pick up the fixed gate.

Correcting something I wrote in the first version of this comment: the notebook matrix running here is **not** evidence that the trigger fix works. This PR changes no notebook. The matrix ran because `.github/workflows/test_notebook.yml` is itself in the workflow's `paths` list, so it shows the edited workflow still executes end to end and nothing more. As the PR body says, the behaviour this fixes is only observable on the next PR that changes an mmm notebook and nothing else.

juanitorduz (approved):

Issue #2974, opened 2026-09-08
LogLinkSpec.validate_target rejects an all-negative target that the model can fit once scaled


shivamlalakiya:
Split out of #2835 at @daimon-pymclabs's request. It came up as the argument for *not* collapsing `LogLinkSpec.validate_target` into the new likelihood support check, and it deserves its own decision.

## The rejection

`LogLinkSpec.validate_target` rejects any non-positive target (`mmm/link.py`):

```python
def validate_target(self, y: np.ndarray) -> None:
    if np.any(y <= 0):
        raise ValueError(
            "All target values must be strictly positive when using "
            "link='log' (LogNormal likelihood). ..."
        )
```

It reads the raw target, before scaling.

## The case it rejects that the model can fit

An all-negative target under the default scaling:

```
raw target : [-5. -3. -7. -2. -9. -4. -6. -8.]
scale      : -2.0
observed   : [2.5  1.5  3.5  1.  4.5  2.  3.  4. ]
```

The `LogNormal` observes `[2.5 ... 4.0]`, which is strictly inside its support. `original_scale_transform` is `exp(mu) * target_scale`, and with a negative `target_scale` it maps back onto the negative originals. The graph is coherent end to end; the rejection happens before any of it runs.

Confirmed by building the same data under `link='identity'` with a `Gamma` likelihood, which has the same positive support and no link-level rule: it builds and gives a finite logp of `-17.66`.

## Why this is a real question and not a bug report

Two defensible positions, and I do not think this should be settled quietly inside a validation PR:

- **Keep the rejection.** A negative target under a log link is nonsense as a model regardless of whether the arithmetic closes. The error is early and clear, and the negative scale that makes it fittable is arguably itself a bug (#2973).
- **Narrow it.** Check the scaled target instead, which is what the likelihood sees. Then the rule stops firing on data the model handles, and it stays correct for the ordinary mixed-sign target, which is the case users actually hit.

The second is a behaviour change that makes previously rejected models buildable, which is why #2835 left it alone.

Note the interaction: if #2973 lands with a non-positive scale rejected at build time, this case disappears on its own, because the negative scale becomes unreachable. Sequencing #2973 first may make this a no-op.

Related: #2835, #2973.

daimon-pymclabs:
Taking the call: **keep the rejection, and do not narrow it.**

[...]

Issue #2973, opened 2026-09-08
DataDerivedScaling reduces with a signed max/mean while its docstring says max-absolute, and a zero scale silently zeroes the target


shivamlalakiya:
Split out of #2835 at @daimon-pymclabs's request, so it gets argued on its own terms rather than riding along in a validation PR.

## The docstring and the code disagree

`DataDerivedScaling`'s docstring calls `method="max"` "Max-absolute scaling (default behaviour)":

```python
    Examples
    --------
    Max-absolute scaling (default behaviour):

    .. code-block:: python

        DataDerivedScaling(method="max", dims=())
```

`_compute_scale_for_variable` reduces with the plain reducer, no absolute value (`mmm/mmm.py:1926-1929`):

```python
method_fn = getattr(data, scaling.method)
scale = method_fn(dim=reduce_dims)
```

So `max` is `data.max(...)`, not `abs(data).max(...)`.

## What it does

A target whose maximum is negative gets a negative scale, and the observed values flip sign. Measured on a single-dim MMM with the default scaling:

```
raw target : [-5. -3. -7. -2. -9. -4. -6. -8.]
scale      : -2.0
observed   : [2.5  1.5  3.5  1.  4.5  2.  3.  4. ]
logp       : -17.661266674163816
```

The model fits. `original_scale_transform` multiplies by `target_scale` again, so `*_original_scale` comes back negative and the round trip is consistent. Nothing is numerically wrong; the point is that the sign flip is invisible and the docstring says it does not happen.

A maximum of exactly zero is the sharper case. Then the scale is zero, every ratio is NaN or infinite, and the switch at `mmm/mmm.py:2323-2330` rewrites them to `0.0`. The target is silently replaced by zeros rather than scaled. Reachable per-slice under `DataDerivedScaling(dims=())` with a heterogeneous panel:

```
scale    : [ 0. 50.]
raw      : [-2. 50. -2. 50. -2. 50.  0. 50.]
observed : [ 0.  1.  0.  1.  0.  1.  0.  1.]
```

Country `A`'s data is gone. Under a likelihood with positive support that now raises at build time (#2835), but under `Normal` it fits on zeros and reports nothing.

## Options, in increasing blast radius

1. **Docs only.** Change the docstring to say `max` and `mean` are signed reductions. Zero behaviour change, and the zero-scale case stays silent.
[...]

Issue #2956, opened 2026-09-08
Stale observed target: fit() after sample_prior_predictive() trains on the placeholder y


daimon-pymclabs:
### Summary

If `sample_prior_predictive(X)` is called before `fit(X, y)` on the same model instance, the subsequent fit trains against the placeholder target created during the prior-predictive call, not against `y`.

### Reproduction

```python
mmm = MMM(...)
mmm.sample_prior_predictive(X)        # no y -> placeholder target is baked in
mmm.fit(X, y)

mmm.model.rvs_to_values[mmm.model["y"]].eval()   # -> [1. 1. 1. 1. 1.] with LogNormalPrior
mmm.xarray_dataset["_target"]                    # -> ones, not y
```

With the default `Normal` likelihood the same script gives zeros instead of ones. Either way the posterior is conditioned on the placeholder. Verified by @anevolbap at `ef0ae478` while reviewing #2875; the behaviour is not specific to that branch.

### Cause

`RegressionModelBuilder.fit` only builds the model when one does not already exist, and nothing downstream refreshes the observed target or its scaler:

```python
# pymc_marketing/model_builder.py:1612-1613
if not hasattr(self, "model"):
    self.build_model(X, y)
```

Any call that builds the model first (`sample_prior_predictive`, `graphviz`, manual `build_model`) therefore fixes the observed data for the rest of the instance's life. It has always been silent; `LogNormalPrior` only surfaced it because `validate_observed` rejected the zeros placeholder and raised.

### Suggested fix

A rebuild is not required. Both quantities are already mutable data containers in `build_model`:

```python
_target_scale = pmd.Data("target_scale", self.scalers._target)        # mmm.py:2274
_target       = pmd.Data("target_data", self.xarray_dataset._target)  # mmm.py:2278
```

So when a model already exists, `fit` can re-scale against the real `y` and `pm.set_data({"target_data": ..., "target_scale": ...})`, which keeps the graph and any existing prior-predictive draws intact. `target_scale` must be recomputed from the real `y` rather than carried over from the placeholder (it is `1.0` in the repro), otherwise the contribution decomposition stays on the placeholder's scale.

A regression test asserting that `rvs_to_values[model["y"]]` matches `y` after a prior-predictive-then-fit sequence would lock this down for every likelihood.

Happy to take this one.

shivamlalakiya:
Context for whoever picks this up, plus a correction to my own first version of this comment.

[...]

PR #2975, opened 2026-09-08
feat(mmm): reject target values outside the likelihood support at build time

### LINKED ISSUE #2835: Target values outside the likelihood support build fine and die as -inf logp

shivamlalakiya:
Closes #2835.

A target outside the likelihood's support builds without complaint and dies later as an `-inf` logp with nothing naming the cause. `LinkSpec.validate_likelihood_support` rejects it at build time instead, naming the distribution, the violated bound, and how many values violate it.

The design discussion is on the issue. Two points from it drive the shape of this diff.

## The check runs on what the likelihood observes, not on the target

`build_model` observes `target / target_scale` (`mmm/mmm.py:2322-2331`, `2479`), so the check has to live after the scaling rather than next to the existing `validate_target` call. The two arrays genuinely differ. `DataDerivedScaling` reduces with `data.max(...)` or `data.mean(...)`, not their absolute values, so the scale can be negative and the ratio can flip sign:

```
raw target : [-5. -3. -7. -2. -9. -4. -6. -8.]
scale      : -2.0
observed   : [2.5  1.5  3.5  1.  4.5  2.  3.  4. ]
logp       : -17.661266674163816
```

Every raw value is outside `Gamma`'s support and the model fits, because the likelihood never sees those values. A rule written against the target rejects a model that works. That case is `test_negative_target_with_negative_scale_builds`.

## `LogLinkSpec.validate_target` is unchanged

The issue asks whether it should delegate to the new check. I would say no, and this PR does not change it.

Under `link='log'` the likelihood is always `LogNormal`, so at a glance the support check subsumes it. It does not: the data above is rejected by `validate_target` (it reads the raw target) and would be accepted by the support check (the scaled values are positive, and `exp(mu) * target_scale` inverts back to the negative originals correctly). Delegating would silently start allowing all-negative targets under a log link. Whether that should be allowed is a modelling decision for the maintainers, not a side effect of adding a support check, so I left the behaviour alone and narrowed the docstring to say which array each rule reads. `test_log_link_still_rejects_the_raw_target` pins that.

Happy to flip it if you would rather the two collapse into one rule.

## Coverage

| distribution | constraint on the observed values |
|---|---|
| `LogNormal`, `Gamma`, `InverseGamma` | `> 0` |
| `Beta` | `0 < y < 1` |
| `Poisson`, `NegativeBinomial` | non-negative integers |
| `TruncatedNormal` | within whichever of `lower`/`upper` is a number |

Unrecognised distributions are not checked and do not warn: `validate_likelihood_compatibility` already warns for those.

## Three things deliberately skipped

**The no-`y` placeholder.** `fit` and `sample_prior_predictive` substitute `np.zeros` when no target is given (`model_builder.py:1546`, `:1772`). The scale is then zero, every ratio is NaN, and the existing switch clamps them to `0.0`, which is outside every positive-only support. Checking that breaks `mmm.sample_prior_predictive(X)` under `Gamma` or `LogNormal`:

```console
ValueError: Likelihood 'Gamma' requires the observed target to be strictly positive,
but 8 of 8 values are not. ...
```

At `build_model` nothing distinguishes the placeholder from a user's genuinely all-zero target, so an all-zero target is skipped. This is the same zeros constraint @daimon-pymclabs wrote up as point 3 on #2956, reached from the other side. `test_prior_predictive_without_y_builds` pins it, because it is the thing a naive version of this check breaks.

The clamp is still caught whenever it is *not* the placeholder, which is the case that actually destroys data. A per-slice zero scale under `DataDerivedScaling(dims=())`:

```
scale    : [ 0. 50.]
raw      : [-2. 50. -2. 50. -2. 50.  0. 50.]
observed : [ 0.  1.  0.  1.  0.  1.  0.  1.]
```

Country `A`'s target is gone. That raises, and when every violating value is exactly `0.0` the error names the scale instead of reporting a bare support violation. `test_degenerate_scale_names_the_scale`.

**Wrappers and special priors.** `_distribution_name` unwraps `Censored(Prior("LogNormal", ...))` to `LogNormal`, but censoring at zero is exactly what makes a zero observation valid, so applying the inner support would reject the data the wrapper exists for. The guard is `isinstance(getattr(likelihood, "distribution", None), str)`, which is also False for the `SpecialPrior` subclasses; those validate their own observations.

**A `TruncatedNormal` bound given as a `Prior` or an array.** The support then varies per draw or per element and there is no single interval to report, so it is left alone.

## Cost

The observed values are evaluated only after a distribution with a checkable support has been found, so the default `Normal` pays nothing. `build_model` on a 520-week single-dim model, median of three:

```
without the check : 0.014s
with the check    : 0.016s
```

An earlier version evaluated unconditionally and cost 0.022s with a ~190ms first-call compile. Flagging the numbers because #2874 was parked over build-time cost.

The eval is also guarded: if it raises, the check returns rather than failing `build_model`, per review on the issue. A check that can break model construction for a graph it did not anticipate is worse than no check. `test_unevaluable_observed_is_skipped`.

## Tests

18 new tests in `TestLikelihoodSupport`, covering each support rule directly, the build-time path for single and multi dimensional targets, the negative-scale case that must keep building, the placeholder path through `sample_prior_predictive`, the degenerate per-slice scale, the guarded eval, the `Censored` wrapper, and the error message naming the distribution and the count.

```console
$ uv run pytest tests/mmm/test_link.py -q
104 passed, 2 skipped

$ uv run pytest tests/mmm/test_mmm.py -q
247 passed

$ uv run pytest tests/test_model_builder.py -q
85 passed
```

The whole of `tests/mmm` plus `tests/test_model_builder.py` also ran green at 3267 passed, 9 skipped, 1 xfailed, 1 xpassed, on the commit before the last two review changes (the guarded eval and the zero-scale message). Those two only add a skip path and extra text on an already-raising branch, and the three files above cover them, but I would rather state which run covered what than round it up. CI has the full picture.

## Out of scope, filed separately

Both split out at @daimon-pymclabs's request rather than folded in, since each is a behaviour change with its own blast radius:

- #2973: `DataDerivedScaling` reduces with a signed `max`/`mean` while its docstring says max-absolute, and a zero scale silently zeroes a target slice.
- #2974: the `link='log'` rejection of an all-negative target the model can fit once scaled.

----
📚 Documentation preview 📚: https://pymc-marketing--2975.org.readthedocs.build/en/2975/

### DIFF
--- a/pymc_marketing/mmm/link.py
+++ b/pymc_marketing/mmm/link.py
@@ -50,8 +50,11 @@ class LinkFunction(StrEnum):
 #: likelihood is observed on the target divided by ``target_scale``, which is
 #: not integer-valued, so they cannot be used under this model at all.
 #: ``Beta`` needs the target inside ``(0, 1)``, which the scaling does not
-#: guarantee, and nothing checks the target against the likelihood support yet.
-#: See issue #2835.
+#: guarantee.  :meth:`LinkSpec.validate_likelihood_support` now rejects a
+#: target outside any of those supports at build time (issue #2835) rather
+#: than letting it through to an ``-inf`` logp, but that check says nothing
+#: about whether ``mu`` is on the response scale, which is what this set is
+#: for.
 RESPONSE_SCALE_LIKELIHOODS = frozenset(
     {
         "Normal",
@@ -85,6 +88,60 @@ def _distribution_name(likelihood: Prior) -> str:
     while dist is not None and not isinstance(dist, str):
         dist = getattr(dist, "distribution", None)
     return dist if dist is not None else type(likelihood).__name__
+
+
+def _positive(likelihood: Prior, observed: np.ndarray):
+    return observed <= 0, "strictly positive"
+
+
+def _unit_interval(likelihood: Prior, observed: np.ndarray):
+    return (observed <= 0) | (observed >= 1), "strictly inside (0, 1)"
+
+
+def _non_negative_integer(likelihood: Prior, observed: np.ndarray):
+    return (observed < 0) | (observed != np.floor(observed)), "a non-negative integer"
+
+
+def _within_truncation(likelihood: Prior, observed: np.ndarray):
+    """Check *observed* against whichever of ``lower``/``upper`` is a number.
+
+    A bound given as a ``Prior`` or an array is left alone: the support then
+    varies per draw or per element, so there is no single interval to report.
+    """
+    parameters = getattr(likelihood, "parameters", {})
+    bounds = {
+        name: parameters[name]
+        for name in ("lower", "upper")
+        if isinstance(parameters.get(name), int | float)
+    }
+    if not bounds:
+        return None
+
+    mask = np.zeros_like(observed, dtype=bool)
+    if "lower" in bounds:
+        mask |= observed < bounds["lower"]
+    if "upper" in bounds:
+        mask |= observed > bounds["upper"]
+
+    described = " and ".join(f"{name} {value}" for name, value in bounds.items())
+    return mask, f"within its truncation ({described})"
+
+
+#: Support checks by distribution name.  Each returns the mask of violating
+#: values and a description of what was required, or ``None`` when the
+#: distribution carries no checkable bound.  Distributions absent from this
+#: mapping are not checked: ``Normal``, ``StudentT`` and ``Laplace`` are
+#: unbounded, and anything unrecognised already draws a warning from
+#: :meth:`LinkSpec.validate_likelihood_compatibility`.
+_SUPPORT_CHECKS = {
+    "LogNormal": _positive,
+    "Gamma": _positive,
+    "InverseGamma": _positive,
+    "Beta": _unit_interval,
+    "Poisson": _non_negative_integer,
+    "NegativeBinomial": _non_negative_integer,
+    "TruncatedNormal": _within_truncation,
+}
 
 
 class LinkSpec(ABC):
@@ -337,6 +394,100 @@ class LinkSpec(ABC):
                 f"decomposition and optimisation results."
             )
 
+    @staticmethod
+    def validate_likelihood_support(likelihood: Prior, observed) -> None:
+        """Raise if *observed* falls outside the support of *likelihood*.
+
+        *observed* is the value the likelihood is given, which is the target
+        divided by ``target_scale`` rather than the target itself.  The two
+        differ in sign whenever the scale is negative, which
+        ``DataDerivedScaling`` allows because it reduces with ``max``/``mean``
+        rather than their absolute values, so this check cannot be run against
+        the raw target.  ``build_model`` clamps a NaN or infinite ratio to
+        ``0.0``, which is finite but outside every positive-only support, and
+        this check reports that too.
+
+        Without it a target outside the support builds without complaint and
+        fails later as an ``-inf`` logp with nothing naming the cause.
+
+        Unrecognised distributions are not checked and do not warn:
+        :meth:`validate_likelihood_compatibility` already warns for those, and
+        a second warning on the same line would say nothing new.
+
+        Parameters
+        ----------
+        likelihood : Prior
+            The likelihood distribution prior.
+        observed : np.ndarray or XTensorVariable
+            The values handed to the likelihood.  A symbolic variable is
+            evaluated only once a distribution with a checkable support has
+            been found, so the default ``Normal`` costs no compile.
+
+        Raises
+        ------
+        ValueError
+            If any observed value lies outside the support.
+        """
+        if not isinstance(getattr(likelihood, "distribution", None), str):
+            # Anything whose own ``distribution`` is not a name is either a
+            # wrapper such as ``Censored`` or a ``SpecialPrior`` subclass.
+            # ``_distribution_name`` reports the inner name for those, but the
+            # inner support is not theirs: censoring at zero is precisely what
+            # makes a zero observation valid under ``LogNormal``, so applying
+            # it would reject the data the wrapper exists for.  The special
+            # priors validate their own observations instead.
+            return
+
+        dist_name = _distribution_name(likelihood)
+        check = _SUPPORT_CHECKS.get(dist_name)
+        if check is None:
+            return
+
+        if hasattr(observed, "eval"):
+            try:
+                observed = observed.eval()
+            except Exception:
+                # A check that can break model construction for a graph it did
+                # not anticipate is worse than no check.
+                return
+
+        violation = check(likelihood, np.asarray(observed, dtype=float))
+        if violation is None:
+            return
+
+        mask, requirement = violation
+        count = int(np.count_nonzero(mask))
+        if not count:
+            return
+
+        message = (
+            f"Likelihood '{dist_name}' requires the observed target to be "
+            f"{requirement}, but {count} of {mask.size} values are not. "
+            "The likelihood observes the target divided by 'target_scale', "
+            "not the raw target, so a value that looks valid in the original "
+            "units can still fall outside the support. Without this check the "
+            "model would build and then sample an '-inf' logp."
+        )
+
+        values = np.asarray(observed, dtype=float)
+        if np.all(values[mask] == 0.0):
+            # `build_model` rewrites a NaN or infinite ratio to 0.0, so a
+            # violation made entirely of zeros usually means the scale is the
+            # problem rather than the target's own values.
+            message += (
+                " Every violating value is exactly 0.0, which is what "
+                "build_model writes when 'target / target_scale' is NaN or "
+                "infinite. Check 'target_scale' for a zero entry, which a "
+                "target slice whose maximum is zero produces, rather than "
+                "looking for bad values in the target itself."
+            )
+        else:
+            message += (
+                " Fix the target, or choose a likelihood whose support covers it."
+            )
+
+        raise ValueError(message)
+
 
 class IdentityLinkSpec(LinkSpec):
     """Identity link: ``E[y] = mu * target_scale``."""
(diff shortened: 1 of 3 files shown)


daimon-pymclabs:
Reviewed. Answering the deviation question here rather than back on #2835 since the code is what settles it.

**Your call on the all-zero case is right, keep it.** The placeholder and a genuine all-zero target are the same array at `build_model`, so any check that rejects one rejects the other, and breaking `sample_prior_predictive` to catch a target that carries no information is the wrong trade. Threading `y`-was-supplied down from the callers is #2956's change and should not be pulled in here. `test_prior_predictive_without_y_builds` pinning the thing a naive version breaks is the right shape for that test. Not flipping it.

The structure is what I asked for: `_SUPPORT_CHECKS` reads cleanly, the guarded `.eval()` is what I wanted, the `Censored`/`SpecialPrior` early return is a case I had not thought about and the comment explains it better than I would have, and `test_negative_target_with_negative_scale_builds` asserting a finite logp is the right way to pin case 2. One real problem, then small stuff.

## The zero-attribution branch misfires on the common case

```python
if np.all(values[mask] == 0.0):
    message += " Every violating value is exactly 0.0, which is what build_model writes when ..."
```

The inference is "all violations are zero, therefore the clamp wrote them, therefore the scale is the problem". But the clamp is not the only way to get an exact `0.0` in `target_data_scaled` — a raw target containing genuine zeros produces them under a perfectly healthy scale. `y = [0, 5, 10, 0, 8]` under `Gamma` with `method="max"` gives `target_scale = 10`, no NaN and no inf anywhere, `observed = [0, 0.5, 1.0, 0, 0.8]`, and the mask is both exact zeros. The user then gets:

> Check 'target_scale' for a zero entry, which a target slice whose maximum is zero produces, rather than looking for bad values in the target itself.

Their scale is `10`. There is no zero entry to find, and the advice actively steers them away from the actual cause, which is that their target has zero weeks and `Gamma` cannot observe those. A target with occasional zeros under a positive-support likelihood strikes me as a lot more common than a degenerate per-slice scale, so this is the branch's likely-hit path rather than an edge case.

It is also cheaply distinguishable, because you have the thing that decides it at the call site: the clamp fired iff `_target / _target_scale` was NaN or inf, which for a finite target means a zero entry in `_target_scale`. Passing the scale into `validate_likelihood_support` (or just testing `np.any(scale == 0)` in `mmm.py` and passing a flag) turns the heuristic into a fact. If you would rather keep the signature as it is, the fallback is to soften the wording to name both causes rather than asserting one.

Note the same branch is reachable for `Beta` and a `lower`-bounded `TruncatedNormal`, where `0.0` is an unremarkable target value and the clamp story is even less likely to be the explanation.

## Smaller notes

- `_positive` and `_unit_interval` let NaN through: `np.nan <= 0` is `False`, so a NaN observation passes both silently, while `_non_negative_integer` catches it by accident via `observed != np.floor(observed)`. Inside `build_model` the clamp means NaN cannot reach the check, so this is not a live bug there, but `validate_likelihood_support` is a public staticmethod and the tests call it directly with bare arrays. An explicit `~np.isfinite(observed)` term in the mask would make the guarantee hold for any caller.
- `np.asarray(observed, dtype=float)` is built twice, once for `check(...)` and again as `values` for the zero test. Hoist it above the `check` call and reuse it.
- `_within_truncation` accepts `int | float`, which lets `bool` through as a bound (`lower=False` becomes `0`) and rejects `np.int64`, so a bound that came out of numpy indexing is silently skipped rather than checked. `numbers.Real` plus an explicit `bool` exclusion covers both.
- The `link='log'` docstring on `validate_target` is exactly the narrowing I asked for, and `test_log_link_still_rejects_the_raw_target` pins it. Nothing to change; flagging it so you know it landed as intended.

The 0.016s versus 0.014s number on the default `Normal` answers the #2874 cost concern as far as I am concerned. Once the zero-attribution point is resolved, I am happy for this to go in.

shivamlalakiya:
All four fixed in `ba2e1e10`. The zero-attribution point was a real bug and your counterexample reproduces exactly as you wrote it.

## Zero attribution

Confirmed first:

```
y      = [0, 5, 10, 0, 8, 3, 7, 2]   under Gamma
scale  = 10.0
message: "... Check 'target_scale' for a zero entry ..."
```

No zero entry to find. You are also right that this is the likely-hit path rather than the edge case; a target with zero weeks under a positive-support likelihood is ordinary.

Took the version you preferred over the wording fallback: `validate_likelihood_support` now takes `target_scale`, and `build_model` passes `self.scalers["_target"].values`, so the attribution is a fact rather than an inference. Three outcomes:

- scale has a zero entry: names the scale as the cause, as before.
- scale has no zero entry: says these are zeros in the target, and to remove, impute, or pick a likelihood whose support includes zero.
- no scale passed (the staticmethod called directly): names both and asserts neither.

`test_target_zeros_are_not_blamed_on_the_scale` is your counterexample, and `test_zero_attribution_names_both_without_the_scale` pins the third case.

## The three smaller notes

**NaN.** Fixed by writing each mask as the negation of the support condition rather than as the violation, so `~(observed > 0)` catches NaN instead of `observed <= 0` letting it through. Same for the unit interval and the truncation bounds. `test_nan_is_never_silently_accepted` covers `Gamma`, `Beta` and `TruncatedNormal`.

**Double `np.asarray`.** Hoisted, built once.

**Truncation bound types.** Now `numbers.Real` with `bool` excluded. One correction to that note, though: only the `bool` half was reachable. `Prior` validates its own parameter types and rejects `np.int64` and `np.float32` outright:

```
int64      REJECTED: Parameters must be one of the following types: (int, float, np.array, Prior, pt.TensorVariable)
float64    OK
float32    REJECTED
bool       OK
```

So `np.float64` was the only numpy bound that could reach the check, and it subclasses `float`, so `int | float` already accepted it. `lower=False` being read as the bound `0` was the live bug. The `numbers.Real` change is still worth having since this is a public staticmethod that can be called with a hand-built object, but I did not want to claim it fixed something that was not reachable. `test_numpy_truncation_bounds_are_honoured` and `test_boolean_truncation_bound_is_not_treated_as_a_number`.

## Verification

```console
$ uv run pytest tests/mmm/test_link.py -q
111 passed, 2 skipped

$ uv run pytest tests/mmm/test_mmm.py -q
247 passed
```

25 tests in `TestLikelihoodSupport` now. Also fixed a flake of my own that CI caught before you reviewed: `test_prior_predictive_without_y_builds` drew a negative scale from the `Gamma` prior under `link='identity'` on two matrix jobs and passed locally on the seed. Only the build is under test there, so the sampling failure is tolerated and the support error is not. I checked the assertion still fails when the all-zero guard is removed from `build_model`, so it is not passing vacuously.

Thanks for the review, particularly for catching that the zero branch was reasoning from a correlation.

daimon-pymclabs (approved):
Checked `ba2e1e1` against all four. The attribution rewrite is what I wanted: passing `self.scalers["_target"].values` in and branching on `np.any(scale == 0.0)` makes it a fact, and the three-way split with the no-scale case naming both is better than what I proposed, since the staticmethod really cannot know. `test_target_zeros_are_not_blamed_on_the_scale` pins my counterexample and `test_zero_attribution_names_both_without_the_scale` pins the branch I would have forgotten to test.

Writing each mask as the negation of the support condition instead of as the violation is also the nicer version of the NaN fix. Thanks for the correction on the truncation bound types, and for saying plainly that only the `bool` half was reachable rather than letting the note stand as written.

This is approved. Two notes below, neither blocking, and the first is a follow-up rather than a change to this PR.

## The third branch can still misattribute, one layer up

The zero-scale/no-zero-scale split is exact for a finite target, which is what the `_attribute_violation` docstring says. But nothing rejects a NaN in the raw target, and `xarray`'s reductions skip NaN by default, so `data.max(dim=...)` returns a healthy positive scale from a target that contains one:

```
y      = [4.0, nan, 10.0, 8.0]   under Gamma, link='identity'
scale  = 10.0                     (max skips the NaN)
ratio  = [0.4, nan, 1.0, 0.8]  -> clamp -> [0.4, 0.0, 1.0, 0.8]
```

The violation is a single exact `0.0`, the scale has no zero entry, so the message says these are zeros in the target and to remove or impute them. The remedy is right by luck, but the diagnosis names a value that is not in their target, so a user who greps for zeros finds none, which is the same wrong-turn the old branch caused. `Beta` and a `lower`-bounded `TruncatedNormal` reach it the same way.

I have not run this one, so treat the trace as read off the code rather than reproduced: `_compute_scale_for_variable` calls `getattr(data, scaling.method)(dim=reduce_dims)` with no `skipna=False`, and I found no NaN check on `y` on the path to `build_model` (the `np.isnan` at `mmm.py:2008` guards a fixed scale, not the target). Worth confirming before acting on it.

Two ways to close it, and I do not think either belongs in this PR. Either name both in the third branch, which costs a clause and gives up the precision you just bought, or reject a non-finite raw target outright, which is a better fix and is really a separate change with its own error message. My preference is the second, as a follow-up issue. Say if you would rather I open it.

## Infinite observations still pass two of the checks

The negation rewrite catches NaN everywhere, but not `+inf`, because `+inf` satisfies the support condition:

- `_positive`: `~(inf > 0)` is `False`, so `inf` passes for `Gamma`, `LogNormal`, and the rest.
- `_non_negative_integer`: `inf >= 0` and `inf == np.floor(inf)` are both `True`, so `inf` passes for `Poisson` and `NegativeBinomial`.
- `_unit_interval` catches it, via `inf < 1`, and `_within_truncation` catches it explicitly through the `~np.isfinite` term you added.

So the guarantee holds for three checks and not for two, which is the kind of asymmetry someone hits later and reads as a bug in the mask rather than as a gap in the rule. Not live inside `build_model`, same as the NaN case, and `test_nan_is_never_silently_accepted` is the natural place to extend: adding `np.inf` to the parametrisation would fail on `Gamma` today. An `~np.isfinite(observed)` term in `_positive` and `_non_negative_integer`, matching `_within_truncation`, closes it, and the test name then covers what it claims.

Issue #2962, opened 2026-09-08
Port CLV models to Terms


juanitorduz:
Port CLV models to Temrs

PR #2976, opened 2026-09-08
feat: compose BG/NBD parameters as terms outside the model

### LINKED ISSUE #2964: BG/NBD use Terms
Use `terms` to add covariates.

williambdean:
Closes #2964 and starts the `Terms` milestone (#17): the first dogfooded consumer of `pymc_marketing.terms` (#2965).

`BetaGeoModel` now composes its parameters as terms **outside** the model context. The model block itself is reduced to register-data, build, and likelihood.

Parameter recipes pass through `model_config` and serialize with the model:

```python
from pymc_marketing.clv.models.beta_geo import (
    BetaGeoModel,
    create_purchase_covariates,
    create_dropout_covariates,
)

model = BetaGeoModel(
    model_config={
        "alpha": create_purchase_covariates(["income", "age"]),
        "a": a_recipe,
        "b": b_recipe,
    }
)
```

Details:

- `model_config["alpha" | "a" | "b"]` accepts terms (`Parameter`, `Named`, compositions). Missing recipes fall back to defaults derived from the existing priors, so current behavior is unchanged.
- `build_model` also accepts an `xr.Dataset` (with `T`/`recency`/`frequency` variables, `customer_id` coordinate, and `purchase_data`/`dropout_data`); a DataFrame is reconstructed for predictions.
- Deprecated and now warn: `purchase_covariate_cols`, `dropout_covariate_cols`, `purchase_coefficient`, `dropout_coefficient` config keys. The columns and coefficient priors become part of the recipe and its serialization instead.
- New terms vocabulary in `pymc_marketing.terms`: `Named` (expression into a named deterministic), `Ref` (reference a built variable, e.g. hierarchical `a_scale` -> `a`), `xdist=False` escape hatch (pymc.dims lacks some distributions, e.g. Pareto), and `to_dict`/`from_dict` registration so recipes survive `fit()` (attrs serialize at sampling time) and `save()`/`load()`.
- Recipes can reference the same data group: multiple `Dot` terms with one `var_name` share a single registered variable, persisted once in `fit_data`. The built-in helpers keep separate `purchase_data`/`dropout_data` variables (and dims) for posterior-name back-compat; cross-branch sharing is available through custom recipes.
- `fit_data` now stores the **merged modeling dataset**: the customer-level frame (with `customer_id` as a coordinate) merged with every recipe-bound variable, so custom recipes binding new data variables survive `fit()`/`save()`/`load()`. Old saved models (previous `fit_data` shape) keep loading through the DataFrame path.
- Stage 1 boundary: predictions keep the closed-form xarray path for default/helper recipes; arbitrary custom recipes raise `NotImplementedError`. Next milestone steps: recipe-driven predictions, then removing the deprecated keys; MBG/NBD keeps its legacy config surface until #2962 ports it.

Dogfooding findings so far: the xtensor -> tensor boundary for legacy distributions (`allow_xtensor_conversion`), register-before-build ordering, and the `pymc.dims` distribution coverage gap are all documented as Gotchas in the terms module docstring.

----
📚 Documentation preview 📚: https://pymc-marketing--2976.org.readthedocs.build/en/2976/

## Flexibility

The config contract is exercised as a combination matrix in `TestAllCombinations`: every `alpha` (default / scalar `Parameter` / `create_purchase_covariates` / custom `Dot` recipe) x `dropout` (hierarchical / nested / helper / custom `Named`) x data (`DataFrame` / `Dataset`) combination builds, with per-cell expected free-RV structure and the one principled boundary (custom `Dot` recipes need a `Dataset` or helper-embedded columns on the `DataFrame` path). Migration mixes (recipe + deprecated keys) are covered too.

Model structure is data: inspect, serialize, tweak, rebuild:

```python
from pymc_marketing.serialization import serialization

model = BetaGeoModel(model_config={"alpha": create_purchase_covariates(["income"])})

# inspect the tree, no PyMC context needed
collect_terms(model.model_config["alpha"].expr)

# serialize -> tweak a prior -> deserialize -> rebuild
spec = json.loads(json.dumps(serialization.serialize(model.model_config["alpha"])))
spec["expr"]["right"]["inner"]["right"]["prior"]["kwargs"]["sigma"] = 10.0
swapped = BetaGeoModel(
    model_config={"alpha": serialization.deserialize(spec)}
)
```

The same class, different graphs: `BetaGeoModel()`, covariate helpers, custom priors, or hand-rolled `Named`/`Ref` compositions — `build_model` never changes. `TestAllCombinations` is the executable contract for this space; `TestStructureAsData` covers inspecting and rewriting the serialized tree.

### DIFF
--- a/pymc_marketing/clv/models/basic.py
+++ b/pymc_marketing/clv/models/basic.py
@@ -122,7 +122,7 @@ class CLVModel(ModelBuilder):
         else:
             return f"{self._model_type}\n{self.model.str_repr()}"
 
-    def _prepare_fit(self, data: pd.DataFrame | None = None) -> None:
+    def _prepare_fit(self, data: pd.DataFrame | xr.Dataset | None = None) -> None:
         """Build the model on first fit, and reject a second fit on different data.
 
         CLV models bake the training data into the model graph, so refitting a built
@@ -135,11 +135,17 @@ class CLVModel(ModelBuilder):
                     "or call `build_model(data)` first."
                 )
             self.build_model(data)  # type: ignore[call-arg]
-        elif data is not None and not self.data.equals(data):  # type: ignore[attr-defined]
-            raise ValueError(
-                "The model was built with different data. "
-                "Create a new model instance to fit new data."
-            )
+        elif data is not None:
+            if isinstance(data, xr.Dataset):
+                modeling_data = getattr(self, "_modeling_data", None)
+                same = modeling_data is not None and bool(data.equals(modeling_data))
+            else:
+                same = bool(self.data.equals(data))  # type: ignore[attr-defined]
+            if not same:
+                raise ValueError(
+                    "The model was built with different data. "
+                    "Create a new model instance to fit new data."
+                )
 
     @classmethod
     def build_from_idata(cls, idata: xr.DataTree) -> None:
@@ -148,9 +154,18 @@ class CLVModel(ModelBuilder):
         model = cls(**kwargs)
 
         model.idata = idata
-        model.data = idata.fit_data.dataset.to_dataframe()
+        fit_data = idata.fit_data.dataset
+        if "customer_id" in fit_data.coords:
+            # New-shape fit_data: the modeling dataset (recipes may bind
+            # arbitrary data variables). The Dataset path validates and
+            # reconstructs the customer-level DataFrame.
+            model.build_model(fit_data)  # type: ignore[call-arg]
+        else:
+            # Old-shape fit_data: a plain DataFrame dump (customer_id as a
+            # variable on the "index" dimension).
+            model.data = fit_data.to_dataframe()  # type: ignore[attr-defined]
+            model.build_model(model.data)  # type: ignore[call-arg,attr-defined]
 
-        model.build_model(model.data)  # type: ignore
         if model.id != idata.attrs["id"]:
             msg = (
                 "The model id in the DataTree does not match the model id. "
@@ -159,7 +174,7 @@ class CLVModel(ModelBuilder):
                 "Investigate if the model structure or configuration has changed."
             )
             raise DifferentModelError(msg)
-        return model
+        return model  # type: ignore[return-value]
 
     def thin_fit_result(self, keep_every: int):
         """Return a copy of the model with a thinned fit result.
@@ -200,6 +215,20 @@ class CLVModel(ModelBuilder):
     @property
     def _serializable_model_config(self) -> dict:
         return self.model_config
+
+    def create_fit_data_group(self) -> xr.Dataset | None:
+        """Build the ``fit_data`` group stored alongside the posterior.
+
+        When the model persisted a modeling dataset (``_modeling_data`` set
+        during ``build_model``), it is merged with the customer-level
+        DataFrame so every recipe-bound data variable survives
+        serialization. Otherwise the base implementation is used.
+        """
+        modeling_data = getattr(self, "_modeling_data", None)
+        if modeling_data is None or getattr(self, "data", None) is None:
+            return super().create_fit_data_group()
+        customer_part = self.data.set_index("customer_id").to_xarray()  # type: ignore[attr-defined]
+        return xr.merge([customer_part, modeling_data], compat="no_conflicts")
 
     def fit_summary(self, **kwargs):
         """Compute the summary of the fit result."""
--- a/pymc_marketing/clv/models/beta_geo.py
+++ b/pymc_marketing/clv/models/beta_geo.py
@@ -435,6 +435,15 @@ def _dataset_to_dataframe(
         *[(col, "dropout_data", "dropout_covariate") for col in dropout_cols],
     ]:
         data[col] = np.asarray(ds[var].sel({dim: col}))
+    # Carry over remaining customer-level variables (e.g. extra user
+    # columns such as dates) so the round-trip preserves them.
+    for name in map(str, ds.data_vars):
+        if name in data:
+            continue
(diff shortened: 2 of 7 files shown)


williambdean:
Please review @daimon-pymclabs

daimon-pymclabs:
Reviewed. The decomposition is the right shape and `TestAllCombinations` is a genuinely good executable contract for the config surface. Two things I think need fixing before merge, both on the MBG/NBD side, plus some smaller notes.

## 1. `ModifiedBetaGeoModel()` now emits two spurious `DeprecationWarning`s on every instantiation

`_warn_deprecated_covariate_config` fires on mere key presence:

```python
if self.model_config.get(key) or key in self.model_config:
```

and `ModelBuilder.__init__` sets `self.model_config = default_model_config | model_config`. `BetaGeoModel.default_model_config` dropped the four deprecated keys, so BG/NBD is clean. But this PR adds them back for MBG:

```python
return {
    **super().default_model_config,
    "purchase_coefficient": Prior("Normal", mu=0, sigma=1),
    "dropout_coefficient": Prior("Normal", mu=0, sigma=1),
}
```

So a bare `ModifiedBetaGeoModel()` has both keys in `model_config` from its own defaults and warns about both, telling the user to pass `coefficient_prior=` to `create_purchase_covariates` — advice MBG cannot act on, since its legacy `build_model` reads those keys and the terms port is deferred to #2962. There is no way for an MBG user to silence it.

CI does not catch this: `filterwarnings` in `pyproject.toml` only ignores two unrelated entries, and none of the ~17 `ModifiedBetaGeoModel(...)` call sites in `tests/clv/models/test_modified_beta_geo.py` assert on warnings, so the tests emit them and pass.

The check should key off what the *user* passed, not the merged config. Something like capturing the user keys in `__init__` before the merge, or `set(model_config) & self._DEPRECATED_CONFIG_KEYS`, and warning only on those.

## 2. MBG silently accepts term recipes and then dies with an opaque `TypeError`

`_resolve_parameter_recipes()` runs in `BetaGeoModel.__init__`, which MBG inherits, so this constructs without complaint:

```python
ModifiedBetaGeoModel(model_config={"alpha": create_purchase_covariates(["income"])})
```

`self._alpha_recipe` becomes the `_Covariates`, `purchase_covariate_cols` returns `["income"]` off the recipe, so MBG's unchanged `build_model` takes the covariate branch and reaches:

```python
alpha_scale = self.model_config["alpha"].create_variable("alpha_scale")
```

`_Covariates.create_variable()` takes no arguments, so this is a `TypeError` at build time with nothing naming the cause. Same for `a`/`b` via `create_dropout_covariates`. Given the PR explicitly scopes MBG out until #2962, I would rather it say so: raise `NotImplementedError` in `ModifiedBetaGeoModel.__init__` when any of `alpha`/`a`/`b` is a `ModelTerm`, pointing at #2962.

## 3. The covariate helpers mutate the caller's prior, and share one instance across `a` and `b`

```python
coefficient = coefficient_prior or Prior("Normal", mu=0, sigma=1)
if hasattr(coefficient, "dims"):
    coefficient.dims = "dropout_covariate"
```

When a caller passes `coefficient_prior=`, its `dims` is overwritten in place, so a `Prior` reused across two helper calls (or held elsewhere) is silently rewritten. `create_dropout_covariates` then puts that same object into both the `dropout_coefficient_a` and `dropout_coefficient_b` `Dot`s, so later mutation of one recipe's coefficient changes the other. Build-time behavior matches the legacy path, but recipes are now user-facing objects that get inspected, serialized, and tweaked — the PR description sells exactly that — so shared mutable state is a sharper edge than it used to be. A copy before setting `dims` would close both.

The same in-place mutation exists in `_default_purchase_recipe` / `_default_*_dropout_recipes` against `config[...]`, which is pre-existing behavior, so I would leave those unless you want to fix it in the same pass.

## Smaller notes

- The `Named` docstring example does not run: `Named("sigma", Intercept(name="sigma_scale"), func=None)` — `Named` has fields `name`, `expr`, `dims`, no `func`. Looks like it was copied off `Transform`.
- `Named.dims` does not round-trip when it is a tuple: `to_dict` writes the tuple, JSON turns it into a list, and `from_dict` returns it unchanged. `pm.Deterministic` is fine with a list, but dataclass equality is not, so a `save()`/`load()` assertion like `test_save_load_recipe_roundtrip`'s `loaded.model_config["alpha"] == model.model_config["alpha"]` would fail. Only str dims are covered today; normalizing to a tuple in `from_dict` would make the existing test meaningful for the tuple case too.
- In `_warn_deprecated_covariate_config`, `self.model_config.get(key) or key in self.model_config` — the second clause subsumes the first, so the `.get` is dead. Worth dropping whichever one is not the intended semantics (see point 1).
- `purchase_covariate_cols` returns `list(self._alpha_recipe.cols)` whenever the recipe is a `_Covariates`, while `dropout_covariate_cols` additionally requires `recipe.cols` to be truthy before returning. Both land on `[]` today so nothing breaks, but the asymmetry reads as accidental.
- `create_purchase_covariates([])` builds a recipe whose expr binds `purchase_data` with no columns, which surfaces later as the `_check_dataframe_recipe_support` error about a recipe binding `purchase_data`. An empty-`cols` guard in the helper would point at the real mistake.

Nothing above touches the terms core, the serialization registration, or the BG/NBD build path, which all look right to me. The `xdist=False` escape hatch and the `_as_plain_tensor` boundary are documented in the Gotchas where I would want to find them.

williambdean:
COnverting to draft and trying to break it up: 

- xarray support for CLV
- pymc.dims support for CLV

PR #2978, opened 2026-09-09
docs(mmm): say that DataDerivedScaling reductions are signed

shivamlalakiya:
Step 1 of the sequencing agreed on #2973. Docs only, no behaviour change.

`DataDerivedScaling`'s docstring called `method="max"` "Max-absolute scaling", but `_compute_scale_for_variable` reduces with the plain reducer (`mmm/mmm.py:1926-1929`):

```python
method_fn = getattr(data, scaling.method)
scale = method_fn(dim=reduce_dims)
```

No absolute value. `FixedScaling` documents and enforces "All values must be positive"; nothing constrains a computed scale, so two things are reachable and neither is reported:

- A negative reduction gives a scale that flips the sign of the scaled data. The round trip closes, since `original_scale_transform` multiplies by the same scale again, but the flip is invisible.
- A zero reduction gives a scale of zero, so every ratio is NaN or infinite and the clamp in `build_model` rewrites them to `0.0`. That slice's data is replaced by zeros rather than scaled. Under a positive-support likelihood this now raises (#2975); under `Normal` it fits on the zeros silently.

The docstring now says all of that, including that an all-zero target reaches the second case, which is what `sample_prior_predictive` and `fit` produce when no `y` is given.

## What this deliberately does not do

Rejecting a non-positive computed scale is step 2 on #2973 and has to wait on #2956, because the no-`y` placeholder produces a scale of exactly zero by construction, so rejecting it today would make `mmm.sample_prior_predictive(X)` impossible under the default scaling.

Switching the reductions to `abs` stays off the table per that discussion: `target_scale` is baked into saved idata, so it would change what a reloaded model's `*_original_scale` means with nothing in the graph flagging it.

```console
$ uv run pytest tests/mmm/test_scaling.py -q
33 passed
```

Related: #2973, #2974, #2956.

----
📚 Documentation preview 📚: https://pymc-marketing--2978.org.readthedocs.build/en/2978/

### DIFF
--- a/pymc_marketing/mmm/scaling.py
+++ b/pymc_marketing/mmm/scaling.py
@@ -135,23 +135,42 @@ class VariableScaling(SerializableBaseModel, ABC):
 class DataDerivedScaling(VariableScaling):
     """Scale by a statistic of the data, computed at fit time.
 
+    Both reductions are **signed**: ``method="max"`` is ``data.max(...)`` and
+    ``method="mean"`` is ``data.mean(...)``, not the maximum or mean of the
+    absolute values.  Unlike :class:`FixedScaling`, which requires positive
+    values, nothing constrains the sign or magnitude of a computed scale, so
+    two cases are reachable and neither is reported:
+
+    * A slice whose reduction is **negative** gives a scale that flips the
+      sign of the scaled data.  The round trip still closes, because
+      ``original_scale_transform`` multiplies by the same scale again, but
+      the flip is invisible.
+    * A slice whose reduction is **exactly zero** gives a scale of zero, so
+      every scaled value is NaN or infinite.  ``MMM.build_model`` rewrites
+      those to ``0.0``, which replaces that slice's data with zeros rather
+      than scaling it.  Under a likelihood with positive support this raises
+      at build time; under ``Normal`` it fits on the zeros silently.
+
+    An all-zero target reaches the second case, which is what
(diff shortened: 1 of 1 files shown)


daimon-pymclabs (approved):
Approving. This is the docstring I asked for, and the wording is more careful than mine would have been: saying the round trip still closes under a negative scale, and separating "raises at build time under a positive-support likelihood" from "fits on the zeros silently under `Normal`", is exactly the distinction that makes the second bullet worth reading. Dropping "absolute" from the two example captions is the right call too, since those captions were the source of the wrong claim as much as the parameter description was.

One thing to fold in before merge, same class of error and it will outlive this PR if it stays. `MMM` repeats the claim in its own docstring, which is the copy most users read:

- `pymc_marketing/mmm/mmm.py:335` — "(by default using max-absolute scaling)"
- `pymc_marketing/mmm/mmm.py:376` — "Defaults to max-absolute scaling for both."

Both describe `DataDerivedScaling(method="max")`, so both are wrong in the same way, and 376 is the `scaling` parameter's own description. "max scaling" in each is enough; no need to restate the sign discussion there when the class docstring now carries it. That leaves no "max-absolute" left in the package.

Nothing else. Agreed on the sequencing you laid out on the issue, and I have replied there on the interim warning.

williambdean:
Thanks!

PR #2979, opened 2026-09-09
docs(mmm): separate the markdown blocks in the time-varying notebook

### LINKED ISSUE #2977: Time-varying media notebook math is not formatted correctly
The Model Specification section of [this page](https://www.pymc-marketing.io/en/stable/notebooks/mmm/mmm_time_varying_media_example.html) (source [here](https://github.com/pymc-labs/pymc-marketing/blob/main/docs/source/notebooks/mmm/mmm_time_varying_media_example.ipynb)) renders incorrectly:

Image

It appears ok in the notebook, so I guess it's in the rendering?

dchaudhari7177:
Closes #2977.

You guessed right that it is the rendering, not the notebook — but the notebook is what has to change. The first markdown cell has **no blank line between any two blocks**, and CommonMark needs one. The diff is 23 inserted blank lines in cell 0; no wording, no equation, no output.

## Why the maths breaks

```
In `pymc-marketing` we provide an API … linear model of the form
$$
y_{t} = \alpha + \sum …
$$
where $\alpha$ is the intercept, …
```

With no blank line in front of it, `$$` is not the start of a display-math block — it is two `$` on a line that is still part of the paragraph above. So the dollars pair up across the *prose*: the equation is typeset as text, and the sentences between the two equations are typeset as maths. That is exactly the screenshot.

You can see it in the published HTML — `\)` and `\(` land in the middle of English sentences:

```html
$\(
y_{t} = \alpha + \sum_{m=1}^{M}\beta_{m}f(x_{m, t}) + …
\)\(
where \)\alpha\( is the intercept, \)f\( is a media transformation function …
```

`mmm_example.ipynb` has the same equation and renders correctly, and the only difference is the blank lines:

```python
12 '… we consider a linear model of the form\n'
13 '\n'                                            # <- this
14 '$$\n'
15 'y_{t} = \alpha + …\n'
16 '$$\n'
17 '\n'                                            # <- and this
18 'where $\alpha$ is the intercept, …'
```

I counted `$$` lines across both notebooks:

| notebook | `$$` lines | preceded by a blank line |
| --- | --- | --- |
| `mmm_example.ipynb` (renders fine) | 2 | 1 (the other opens the cell) |
| `mmm_time_varying_media_example.ipynb` | 4 | **0** |

## A second bug in the same cell, worth knowing about

The same missing blank lines have made a paragraph into a heading:

```
The notebook assumes the reader has knowledge of the essential functionalities of PyMC-Marketing. …
---
## Part I: Data Generation Process
```

`---` directly under a paragraph is a **setext H2**, not a horizontal rule. So on the live page that entire 250-character sentence is an `<h2>`, with a `headerlink` anchor and an entry in the table of contents:

```html
<h2>The notebook assumes the reader has knowledge of the essential functionalities of
PyMC-Marketing. If one is unfamiliar, the <a …>“MMM Example Notebook”</a> serves as an
excellent starting point…<a class="headerlink" href="#the-notebook-assumes-the-reader-has-…">#</a></h2>
```

The blank line before `---` turns it back into the rule it was meant to be, and the sentence back into a paragraph.

## What changed, exactly

Blank lines inserted before every block that starts one: paragraphs, headings, the two `$$` blocks, the `---`, the blockquote, and the two numbered lists. Nothing inserted *between* list items, so the lists stay tight and their spacing does not change.

Verified mechanically rather than by eye:

- Re-serialising the untouched notebook with `json.dumps(…, indent=1, ensure_ascii=False)` reproduces the file on disk **byte for byte**, so the only difference in this commit is the inserted lines — outputs, images and execution counts are identical.
- Comparing cell by cell, exactly one cell differs: cell 0.
- Every `$$` now has a blank line before the opening one and after the closing one.
- No doubled blank lines.

I did not build the docs locally (it needs the full doc toolchain), so the HTML above is quoted from the currently published page as evidence of the bug rather than of the fix. The fix is the shape `mmm_example.ipynb` already uses successfully.

----
📚 Documentation preview 📚: https://pymc-marketing--2979.org.readthedocs.build/en/2979/


juanitorduz (approved):
Thanks @dchaudhari7177

PR #2986, opened 2026-09-09
feat(r2d2): add non-centered parameterization

TeemuSailynoja:
## Summary
- add an optional non-centered R2D2 parameterization
- retain centered behavior as the default
- add naming and serialization coverage

## Validation
- `uv run pytest tests/test_r2d2.py tests/test_serialization.py -q`
- `uv run ruff check pymc_marketing/r2d2.py tests/test_r2d2.py`

Needed by the control-dimensionality MMM case-study notebook, which will remain separate and merge after this API change.

----
📚 Documentation preview 📚: https://pymc-marketing--2986.org.readthedocs.build/en/2986/

### DIFF
--- a/pymc_marketing/r2d2.py
+++ b/pymc_marketing/r2d2.py
@@ -117,8 +117,8 @@ class R2D2Split:
     wherever a Prior is accepted (e.g., in model_config priors).
 
     When create_variable() is called, auto-builds the decomposition
-    if not already built, then creates a pmd.Normal coefficient with
-    mu=0, sigma=split, dims=component_dim.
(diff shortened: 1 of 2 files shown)


TeemuSailynoja:
@williambdean I had a load of divergences when trying to fit an MMM with an R2D2 prior and realized that we need a non-centered option :)

williambdean (approved):

williambdean:
Nice. Does it help there then?

TeemuSailynoja:
> Nice. Does it help there then?

very much 😉

PR #2987, opened 2026-09-09
feat: serialize terms vocabulary

### LINKED ISSUE #2980: serialization support for terms
use pymc-marketing serialization in order to support Terms

williambdean:
Closes #2980. First PR in the terms stack (serialization → Bass components → Bass port).

The terms vocabulary serializes through `pymc_marketing.serialization`: `to_dict`/`from_dict` are registered for `Parameter`, `Intercept`, `Dot`, `Transform`, `Product`, and `Sum`, so recipes stored in `model_config` survive `fit()` (attrs serialize at sampling time) and `save()`/`load()`.

`Transform` functions resolve by name dynamically against `pytensor.xtensor.math` / `pytensor.xtensor.linalg` and transforms registered with `pymc_extras.prior.register_tensor_transform` — no hard-coded whitelist. Numeric literals inside compositions serialize as literals, and nested priors round-trip through `pymc_extras.deserialize`.

----
📚 Documentation preview 📚: https://pymc-marketing--2987.org.readthedocs.build/en/2987/

### DIFF
--- a/pymc_marketing/serialization.py
+++ b/pymc_marketing/serialization.py
@@ -26,7 +26,7 @@ from __future__ import annotations
 
 import importlib
 import inspect
-from dataclasses import dataclass
+from dataclasses import dataclass, fields, is_dataclass
 from typing import Any, Protocol, Self, runtime_checkable
 
 from pydantic import BaseModel, Field
@@ -192,6 +192,11 @@ class TypeRegistry:
                 actual_cls.to_dict = _wrapped_to_dict  # type: ignore[attr-defined]
 
         return actual_cls
+
+    def is_registered(self, obj: Any) -> bool:
+        """Check whether an object's type is registered in the registry."""
+        type_key = f"{obj.__class__.__module__}.{obj.__class__.__qualname__}"
+        return type_key in self._registry
 
     def serialize(self, obj: Serializable) -> dict[str, Any]:
         """Serialize an object to a JSON-safe dict with ``__type__`` key."""
@@ -387,6 +392,16 @@ def _merge_shared_decompositions(config: dict[str, Any]) -> dict[str, Any]:
             return obj
         if isinstance(obj, R2D2Sigma):
             obj.decomposition = walk(obj.decomposition)
+            return obj
+        if is_dataclass(obj) and not isinstance(obj, type):
+            # Term compositions (Sum, Product, Parameter, Dot, Transform)
+            # carry R2D2Split priors in their fields; descend so shared
+            # decompositions merge instead of duplicating per field.
+            dataclass_params = getattr(obj, "__dataclass_params__", None)
+            if dataclass_params is not None and dataclass_params.frozen:
+                return obj
+            for dc_field in fields(obj):
+                setattr(obj, dc_field.name, walk(getattr(obj, dc_field.name)))
             return obj
         if isinstance(obj, dict):
             return {k: walk(v) for k, v in obj.items()}
--- a/pymc_marketing/serialization.py
+++ b/pymc_marketing/serialization.py
@@ -26,7 +26,7 @@ from __future__ import annotations
 
 import importlib
 import inspect
-from dataclasses import dataclass, fields, is_dataclass
+from dataclasses import dataclass, fields, is_dataclass, replace
 from typing import Any, Protocol, Self, runtime_checkable
 
 from pydantic import BaseModel, Field
@@ -397,11 +397,14 @@ def _merge_shared_decompositions(config: dict[str, Any]) -> dict[str, Any]:
             # Term compositions (Sum, Product, Parameter, Dot, Transform)
             # carry R2D2Split priors in their fields; descend so shared
             # decompositions merge instead of duplicating per field.
-            dataclass_params = getattr(obj, "__dataclass_params__", None)
-            if dataclass_params is not None and dataclass_params.frozen:
-                return obj
-            for dc_field in fields(obj):
-                setattr(obj, dc_field.name, walk(getattr(obj, dc_field.name)))
+            walked = {
+                dc_field.name: walk(getattr(obj, dc_field.name))
+                for dc_field in fields(obj)
+            }
+            if obj.__dataclass_params__.frozen:  # type: ignore[attr-defined]
+                return replace(obj, **walked)
+            for key, value in walked.items():
+                setattr(obj, key, value)
             return obj
         if isinstance(obj, dict):
             return {k: walk(v) for k, v in obj.items()}
--- a/pymc_marketing/terms.py
+++ b/pymc_marketing/terms.py
@@ -231,6 +231,13 @@ Gotchas
           model = pm.modelcontext(None)
           unique = list(dict.fromkeys(ds[self.data_source].values))
           model.add_coords({self.data_source: unique})
+- The built-in terms serialize via ``pymc_marketing.serialization``
+  (``to_dict`` / ``from_dict`` are registered), so recipes stored in
+  ``model_config`` survive ``fit()`` (attrs are JSON-serialized at
+  sampling time) and ``save()`` / ``load()``. ``Transform`` functions are
+  serialized by name, resolved against ``pytensor.xtensor.math`` /
+  ``pytensor.xtensor.linalg`` and transforms registered with
+  ``pymc_extras.prior.register_tensor_transform``.
 """
 
 from __future__ import annotations
@@ -242,9 +249,14 @@ from typing import Any, cast
 import pymc as pm
 import pymc.dims as pmd
 import pytensor.tensor as pt
+import pytensor.xtensor as ptx
 import xarray as xr
-from pymc_extras.prior import Prior, VariableFactory
+from pymc_extras.deserialize import DeserializableError
+from pymc_extras.deserialize import deserialize as pymc_extras_deserialize
+from pymc_extras.prior import CUSTOM_TRANSFORMS, Prior, VariableFactory
 from pytensor.xtensor.type import as_xtensor
+
+from pymc_marketing.serialization import SerializationError, serialization
 
 __all__ = [
     "Dot",
@@ -261,6 +273,81 @@ __all__ = [
     "register_data",
     "set_data",
 ]
+
+# Modules searched when serializing ``Transform`` functions. Registered
+# custom transforms (``pymc_extras.prior.register_tensor_transform``) take
+# precedence; otherwise the function is looked up by name in these modules.
+_TRANSFORM_MODULES = ("math", "linalg")
+
+
+def _func_name(func: Callable) -> str:
+    """Resolve a transform function to its serializable name."""
+    for name, registered in CUSTOM_TRANSFORMS.items():
+        if registered is func:
+            return name
+    for module_name in _TRANSFORM_MODULES:
+        module = getattr(ptx, module_name)
+        for attr in dir(module):
+            if getattr(module, attr, None) is func:
+                return attr
+    name = getattr(func, "__name__", None)
+    if name is not None and getattr(ptx, name, None) is func:
+        return name
+    raise SerializationError(
+        f"Function {func!r} is not serializable. Use a "
+        "pytensor.xtensor.math function, or register it with "
+        "pymc_extras.prior.register_tensor_transform."
+    )
+
+
+def _lookup_func(name: str) -> Callable:
+    """Resolve a serialized transform function name."""
+    if name in CUSTOM_TRANSFORMS:
+        return CUSTOM_TRANSFORMS[name]
+    for module_name in _TRANSFORM_MODULES:
+        try:
+            return getattr(getattr(ptx, module_name), name)
+        except AttributeError:
+            continue
+    try:
+        return getattr(ptx, name)
+    except AttributeError as err:
+        raise SerializationError(
+            f"Unknown serialized function {name!r}. Use a "
+            "pytensor.xtensor.math function, or register it with "
+            "pymc_extras.prior.register_tensor_transform."
+        ) from err
+
+
+def _serialize_child(value: Any) -> Any:
+    """Serialize a child of a composed term to a JSON-safe value.
+
+    Numeric literals become ``{"__type__": "literal", ...}`` wrappers;
+    variable factories that are not registered in the type registry
+    (``Prior``, ``Censored``, ...) serialize via their own ``to_dict``.
+    """
+    if isinstance(value, (int, float)) and not isinstance(value, bool):
+        return {"__type__": "literal", "value": value}
+    try:
+        return serialization.serialize(value)
+    except KeyError:
+        if hasattr(value, "to_dict"):
+            return value.to_dict()
+        raise
+
+
+def _deserialize_child(value: Any) -> Any:
+    """Deserialize a serialized term child back to a live object."""
+    if isinstance(value, dict):
+        if value.get("__type__") == "literal":
+            return value["value"]
+        if "__type__" in value:
+            return serialization.deserialize(value)
+        try:
+            return pymc_extras_deserialize(value)
+        except DeserializableError as err:
+            raise SerializationError(f"Cannot deserialize term child: {value}") from err
+    return value
 
 
 @dataclass
@@ -352,6 +439,7 @@ class ModelTerm:
         return Product(-1, self)
 
 
+@serialization.register
 @dataclass
 class Sum:
     """Container for additive composition via ``+``.
@@ -406,7 +494,17 @@ class Sum:
         for term in self.terms:
             set_data(term, ds=ds, model=model)
 
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize the sum and its children."""
+        return {"terms": [_serialize_child(term) for term in self.terms]}
 
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> Sum:
+        """Reconstruct a sum from its serialized form."""
+        return cls(terms=[_deserialize_child(term) for term in data["terms"]])
+
+
+@serialization.register
 @dataclass
 class Product:
     """Container for multiplicative composition via ``*``.
@@ -448,7 +546,23 @@ class Product:
         set_data(self.left, ds=ds, model=model)
         set_data(self.right, ds=ds, model=model)
 
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize the product and both operands."""
+        return {
+            "left": _serialize_child(self.left),
+            "right": _serialize_child(self.right),
+        }
 
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> Product:
+        """Reconstruct a product from its serialized form."""
+        return cls(
+            left=_deserialize_child(data["left"]),
+            right=_deserialize_child(data["right"]),
+        )
+
+
+@serialization.register
 @dataclass
 class Parameter(ModelTerm):
     """Named free parameter (scalar or dimensional via ``prior.dims``).
@@ -513,7 +627,17 @@ class Parameter(ModelTerm):
         """Build a free parameter variable."""
         return self.prior.create_variable(self.name, xdist=True)
 
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize the parameter name and prior."""
+        return {"name": self.name, "prior": _serialize_child(self.prior)}
 
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> Parameter:
+        """Reconstruct a parameter from its serialized form."""
+        return cls(name=data["name"], prior=_deserialize_child(data["prior"]))
+
+
+@serialization.register
 @dataclass
 class Intercept(Parameter):
     """GLM-style intercept term; same as :class:`Parameter` with a default name.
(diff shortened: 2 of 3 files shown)


juanitorduz:
@williambdean , this is my first experiment review with oh-my-pi :)

## Review

Reviewed at `c2cf224`. Checked out the PR head, built a venv (pymc 6.3.2 / pytensor 3.3.1 / pymc-extras 0.12.1) and ran the suite plus targeted probes — `tests/test_terms.py`: **74 passed**. The mechanism works end-to-end (evidence below), but there are three concrete defects and one clear duplication.

### Verified working

```
json round trip of Intercept("a") + 5 - Intercept("b")  -> equal after json.dumps/loads  OK
serialize_model_config / deserialize_model_config with Sum, Dot, Transform -> equal      OK
Censored(Prior(...)) child, nested Prior params, xr.DataArray child -> round trip        OK
_func_name/_lookup_func identity round trip: math.exp, linalg.solve, ptx.dot, ptx.where  OK
Intercept subclass gets correct __type__ via Parameter's injected to_dict wrapper        OK
terms is imported in pymc_marketing/__init__.py -> no deferred-registration hazard       OK
```

The `create_idata_attrs._json_default` -> `serialization.serialize` fallback means terms in `model_config` also JSON-serialize in models that don't route through `serialize_model_config`, and `_model_config_formatting` restores them. So the docstring claim about `fit()`/`save()`/`load()` is mechanically true — it is just not covered by a test (see gaps).

### Must fix

**1. `_lookup_func` duplicates `pymc_extras.prior._get_transform` verbatim.** Same CUSTOM_TRANSFORMS-first, then `(ptx.math, ptx.linalg, ptx)` order. That private import is already an accepted convention here (`pymc_marketing/mmm/hsgp.py:32`).

```python
>>> _get_transform("exp", xdist=True) is ptx.math.exp       # True
>>> _get_transform("solve", xdist=True) is ptx.linalg.solve # True
```

Delete `_lookup_func` + `_TRANSFORM_MODULES` and call `_get_transform(data["func"], xdist=True)`, converting `UnknownTransformError` -> `SerializationError`. Removes ~18 lines and keeps a single resolution rule as pytensor's namespace evolves.

**2. R2D2 decomposition sharing is lost when the split lives inside a term — the loaded model raises.** `_merge_shared_decompositions` (serialization.py:379) walks `dict`, `list`, `R2D2*`, and `.parameters`; it never descends into `Sum.terms` / `Product.left|right` / `Parameter.prior` / `Transform.inner`. Each `R2D2Split` then deserializes its own `R2D2`, and the `_built_model_id` guard in `R2D2Split.create_variable` (r2d2.py:158) builds `r2d2_*` twice:

```
live cfg:      ok
restored cfg:  ValueError: Variable name r2d2_r2 already exists.
```

Repro: `{"a": Parameter("a", prior=r2d2.split("control")), "b": Parameter("b", prior=r2d2.split("fourier"))}` through `serialize_model_config` -> `json.dumps`/`loads` -> `deserialize_model_config` -> `build_param` both. The identical config *without* the `Parameter` wrapper merges correctly, so this is specific to terms.

R2D2-in-terms is exactly the MMM use case this stack is heading for. Fix: give `walk` a dataclass-field branch (all five term classes are dataclasses), placed after the `R2D2*` branches, plus a regression test asserting `a.prior.decomposition is b.prior.decomposition`.

**3. Serialize succeeds, load fails — asymmetric for several accepted child types.**

| child | `serialize` | `deserialize` |
|---|---|---|
| custom `ModelTerm` with `to_dict`, unregistered | writes `{"k": 3}` (no `__type__`) | `SerializationError: Cannot deserialize term child` |
| `DeferredFactory` prior | writes `{"__deferred__": true, ...}` | `SerializationError` (the registry handles `__deferred__`, `_deserialize_child` does not) |
| custom term without `to_dict` | raw `KeyError` from the registry | — |
| `bool` (accepted by `build_param`) | raw ``KeyError: "Type 'builtins.bool' …"`` | — |
| `np.int64` | raw ``KeyError: "Type 'numpy.int64' …"`` | — |

All verified. Minimal fixes: in `_deserialize_child`, check `value.get("__deferred__")` first and delegate to `serialization.deserialize`; in `_serialize_child`, treat `numbers.Real`/`np.number` (and `bool`) as literals via `int()`/`float()`, and raise `SerializationError` naming `@serialization.register` instead of leaking `KeyError` — a bare `KeyError` surfacing from inside `fit()`'s attrs dump is a bad diagnostic. The `hasattr(value, "to_dict")` fallback should be narrowed to `VariableFactory` instances (which pymc-extras can actually read back); for a `ModelTerm` it guarantees an unloadable artifact.

### Should fix

**4. `_lookup_func` never checks the result is callable.** Names come from a file on disk:

```python
>>> _lookup_func("pytensor")     # <module 'pytensor' …>
>>> _lookup_func("basic")        # <module 'pytensor.xtensor.basic' …>
>>> _lookup_func("__loader__")   # <SourceFileLoader object …>
```

Not RCE (no arbitrary import), but `Transform.func` ends up as a module and fails much later at build time. Reject names starting with `_` and require `callable(...)` in whatever wrapper survives item 1.

**5. `{"__type__": "literal"}` squats the registry key namespace.** `serialization.deserialize({"__type__": "literal", "value": 5})` reports `Unknown type 'literal'. The class may not have been registered…`, which is misleading, and it means only `_deserialize_child` can read term payloads. A disjoint key (`{"__literal__": 5}`) keeps the invariant "`__type__` is always a registry key".

**6. Docstrings.** `Transform`'s docstring (terms.py:777) still advertises ``pt.math.sigmoid``, which cannot work at all — verified `TypeError: PyTensor forbids automatic conversion of XTensorVariable to TensorVariable` — and now also cannot serialize; change it to `ptx.math.sigmoid`. The **Extending** section (`Indexed` example) should state that custom terms need `@serialization.register` + `to_dict`/`from_dict` to survive `save()`/`load()`; that is now part of the term contract.

**7. Test gaps — they line up exactly with the failing `codecov/patch` (86.6%, 11 lines).** The uncovered lines are the negative paths: `_lookup_func`'s unknown-name `SerializationError`, its `AttributeError`/`continue` fallthrough, `_serialize_child`'s bare re-`raise`, and `_deserialize_child`'s `DeserializableError` branch. Missing behavioral coverage for the PR's own justification:

- a `json.dumps` round trip — every current test round-trips in-memory dicts only, so a non-JSON-safe leaf (e.g. `np.int64`) would pass CI and break `fit()`;
- `serialize_model_config`/`deserialize_model_config` with a term, and one `save()`/`load()` smoke test through `ModelBuilder` attrs;
- `_lookup_func("nope")` raising `SerializationError`.

**8. Nits.**

- `Dot.from_dict` does `data["name"]` although `name` is optional in the constructor; dropping/renaming that key later makes old artifacts unloadable. Prefer `data.get("name")` (same for `Parameter.prior`).
- `_serialize_child` catching `KeyError` couples to a registry implementation detail — `TypeRegistry.serialize` raises `KeyError` while `deserialize` raises `SerializationError`. Worth adding `serialization.is_registered(obj)` (or making `serialize` raise `SerializationError`) instead of exception-driven probing.
- `test_serialize_registered_custom_transform_roundtrip` mutates the global `CUSTOM_TRANSFORMS` and pops in `finally`; a `monkeypatch.setitem` fixture prevents state leaking into other tests if the body fails.
- The `ptx.exp` -> `ptx.math.exp` churn is diff noise but correct: `pytensor.xtensor` exposes no top-level `exp` (only `dot`, `where`, `concat`, `full_like`, …), so the old `import pytensor.xtensor.math as ptx` alias was the thing being fixed.

### Design suggestion

`Prior(transform=...)` and `HSGP.transform` store the transform **as a name string** and resolve it with `_get_transform` at build time. `Transform.func` stores a live callable, which is the only reason `_func_name`'s `dir()` identity scan exists. Accepting `func: str | Callable` and resolving in `create_variable` would delete `_func_name` entirely, make serialization trivially symmetric, align terms with the established convention, and reject unserializable lambdas at construction rather than at `fit()` time. Bigger than this PR, but it is the direction the rest of the codebase already went.

### Verdict

Approve after item 2 (it produces an unloadable model) plus items 1 and 3; the rest are cheap follow-ups in the same file. The design is otherwise sound: registering `Intercept` separately so the injected `to_dict` picks up the subclass key is right, and delegating priors to `pymc_extras.deserialize` avoids re-implementing prior serialization.

williambdean:
taking a look

juanitorduz:
## Re-review at `b38b30f`

Checked out the PR head in a worktree (pymc 6.2.0 / pytensor 3.2.4 / pymc-extras 0.14.0). `tests/test_terms.py`: **90 passed**; `tests/test_serialization.py` + `tests/test_r2d2.py`: **60 passed**. All eight items from the previous round are addressed. Three small residues below; none of them is in a path the built-in terms hit today.

### Verified fixed

| prev. item | check | result |
|---|---|---|
| 1 `_lookup_func` duplication | `_resolve_func` -> `_get_transform(name, xdist=True)` | gone; property test over **140** callables in `ptx.math`/`ptx.linalg`/`ptx`: `_resolve_func(_func_name(f)) is f` for all, 0 mismatches |
| 2 R2D2 sharing through terms | dataclass branch in `walk` | fixed, and holds deeper than the new test: `Sum([Parameter(split), Transform(Parameter(split))])` -> `shared=True`, both `create_variable` calls build |
| 3 codec asymmetry | `__literal__`, `__deferred__`, `is_registered`, numpy/bool literals | fixed; `bool`/`np.int64`/`np.float64`/`np.bool_` now JSON-safe and load back as Python scalars; unregistered custom terms raise `SerializationError` at serialize time |
| 4 callable check | `name.startswith("_") or not callable(func)` | `_resolve_func("basic")`, `("np")`, `("__loader__")`, `("nope")` all raise `SerializationError` |
| 5 `__type__` squatting | `{"__literal__": v}` | fixed |
| 6 docstrings | `ptx.math.sigmoid` (`hasattr(ptx.math, "sigmoid") is True`), custom-term contract documented | fixed |
| 7 test gaps | json round trips, `serialize/deserialize_model_config`, `ModelBuilder.create_idata_attrs` smoke test, unknown/non-callable func | fixed; only new uncovered lines in `terms.py` are 381-383 (see item 3 below) |
| 8 nits | `data.get("name")`, `is_registered`, `monkeypatch.setitem` | fixed |

Blast radius of the new `is_dataclass` branch is small and I confirmed it is not a regression for real config objects: `GeometricAdstock`, `LogisticSaturation`, `YearlyFourier`, `LinearTrend`, `EventEffect`, `HSGPKwargs`, `Scaled`, `MaskedPrior` are none of them dataclasses (pydantic `BaseModel` or plain classes), so they still fall through to the `parameters` branch and round-trip unchanged. Only `R2D2*` (matched earlier) and the terms reach it. Branch ordering is right: the dataclass case sits **after** the three `R2D2*` cases, so decomposition dedup still keys on `R2D2` identity.

### Should fix

**1. The frozen-dataclass early return reintroduces the exact bug it guards.** `serialization.py:401-402` returns the object without descending, so a frozen custom term silently loses decomposition sharing and the loaded model dies at build time — the same failure the PR fixes for non-frozen terms. This is also the only new line in `serialization.py` that no test covers.

```python
@serialization.register
@dataclass(frozen=True)
class FrozenTerm:            # to_dict/from_dict via _serialize_child/_deserialize_child
    prior: object

cfg = {"a": FrozenTerm(r2d2.split("control")), "b": Parameter("b", r2d2.split("fourier"))}
loaded = deserialize_model_config(json.loads(json.dumps(serialize_model_config(cfg))))
loaded["a"].prior.decomposition is loaded["b"].prior.decomposition   # False
# create_variable x2 -> ValueError: Variable name r2d2_r2 already exists.
```

The non-frozen equivalent shares correctly, so the guard is the whole difference. `dataclasses.replace` handles it without the special case:

```python
if is_dataclass(obj) and not isinstance(obj, type):
    walked = {f.name: walk(getattr(obj, f.name)) for f in fields(obj)}
    if obj.__dataclass_params__.frozen:
        return replace(obj, **walked)
    for key, value in walked.items():
        setattr(obj, key, value)
    return obj
```

(That also drops the `getattr(obj, "__dataclass_params__", None)` dance — every dataclass has the attribute, so the `is not None` test is dead.)

**2. Underscore-prefixed `CUSTOM_TRANSFORMS` entries serialize but cannot load.** The `name.startswith("_")` check is applied after `_get_transform`, so it also rejects names the user explicitly registered:

```python
CUSTOM_TRANSFORMS["_secret"] = _secret
serialization.serialize(Transform(Parameter("x"), func=_secret))["func"]   # '_secret'
serialization.deserialize(...)
# SerializationError: Serialized function name '_secret' must resolve to a callable
#                     pytensor function, got <function _secret ...>
```

The message contradicts itself — the object is callable. The underscore rule only exists to stop dunder/module attributes coming out of the module scan, so gate it on that path: reject `_`-prefixed names only when `name not in CUSTOM_TRANSFORMS`, and give the non-callable case its own message.

**3. Custom `VariableFactory` children are still write-only, and the error does not say how to fix it.** Narrowing the fallback to `VariableFactory` was the right move, but a protocol-satisfying factory that pymc-extras cannot deserialize still produces an unloadable artifact:

```python
class MyFactory:                      # dims + create_variable + to_dict -> isinstance(VariableFactory) is True
    ...
serialization.serialize(Parameter("a", prior=MyFactory(2)))
# {'__type__': 'pymc_marketing.terms.Parameter', 'name': 'a', 'prior': {'k': 2}}
serialization.deserialize(...)
# SerializationError: Cannot deserialize term child: {'k': 2}
```

Point the message at `pymc_extras.deserialize.register_deserialization` (the counterpart to the `@serialization.register` advice in `_serialize_child`'s message). Adding that test also closes `terms.py:381-382`, the `DeserializableError` branch — the only new lines without coverage. While there: `terms.py:383` (`return value` for a non-dict child) is unreachable from anything `_serialize_child` emits now that literals are wrapped; drop it or cover it.

### Nits

- The `serialization.py` changes (`TypeRegistry.is_registered`, the dataclass branch in `_merge_shared_decompositions`) are exercised only from `tests/test_terms.py`. A frozen/non-frozen dataclass-walk test in `tests/test_serialization.py` keeps them honest if terms ever move.
- `_func_name` still scans `dir()` on three modules per serialize; correct (property-tested above) but the `func: str | Callable` idea from the last round would delete it. Fine as a follow-up.
- Unrelated to this PR, noticed while round-tripping compositions: `-Intercept("a") + 3 * Parameter("b")` raises `TypeError: unsupported operand type(s) for +: 'Product' and 'Product'` — `__neg__` returns a `Product` and `Product` has no `__add__`. Pre-existing on `main`, worth a separate issue since the vocabulary now advertises these compositions as serializable recipes.

### Verdict

Approve. Item 1 is three lines and prevents a silent unloadable-model trap in the contract you just documented; items 2 and 3 are message/gating fixes in the same file. Nothing else blocks — the R2D2 fix is verified beyond the case the new test pins, and the transform name resolution is now provably symmetric.

williambdean:
All three residues applied in `a5ff984c5`:

- `_merge_shared_decompositions` handles frozen dataclasses via `dataclasses.replace` (per field walk unconditional, `getattr`/`__dataclass_params__` dance dropped)
- `_resolve_func`: the underscore gate now applies only to names outside `CUSTOM_TRANSFORMS` (explicit registrations opt out), and non-callables get their own message
- Unreadable factory children point at `pymc_extras.deserialize.register_deserialization`; the `DeserializableError` branch and the non-dict passthrough now have tests

Plus a frozen-walk regression test (`FrozenHolder` registered with `to_dict`/`from_dict`) in `tests/test_terms.py`. `tests/test_terms.py` 94 passed.

juanitorduz:
## Re-review at `a5ff984`

Worktree on the PR head, merged with current `main` (clean, `r2d2.py` picked up the new `centered` field). pymc 6.2.0 / pytensor 3.2.4 / pymc-extras 0.14.0. `tests/test_terms.py` + `tests/test_serialization.py` + `tests/test_r2d2.py`: **157 passed** post-merge. All three residues from the last round are fixed. One new residue, introduced by the `replace()` change itself.

### Verified fixed

| prev. item | check | result |
|---|---|---|
| 1 frozen-dataclass walk | `dataclasses.replace` per field | `FrozenHolder(r2d2.split(...))` shares; also holds nested (`FrozenOuter(FrozenHolder(split))` -> `shared=True`), which the new test does not pin |
| 2 underscore gate | `name not in CUSTOM_TRANSFORMS` guard | `CUSTOM_TRANSFORMS["_secret"]` round-trips, `restored.func is _secret` |
| 3 unreadable factory child | message names `register_deserialization` | `SerializationError: ... register it with pymc_extras.deserialize.register_deserialization ...` |

Also re-confirmed after the merge: `_resolve_func(_func_name(f)) is f` for **145/145** callables across `ptx.math`/`ptx.linalg`/`ptx`, 0 mismatches; literals (`bool`, `np.bool_`, `np.int64`, `np.float32`, `2**63`, `inf`, `nan`) survive `json.dumps`; and a real netCDF attr round trip of `Sum([Transform(Intercept("a"), ptx.math.exp), Parameter("b", Prior("Normal")), 3])` loads back equal with `func is ptx.math.exp` — the `save()`/`load()` claim in the PR body now holds through the file format, not just in memory.

### Should fix

**1. `replace()` reintroduces the write-only artifact for `init=False` fields — no R2D2 required.** `serialization.py:400-405` passes *every* field to `replace`, and `replace` rejects `init=False` fields. Serialize succeeds, load raises a raw `dataclasses` `TypeError`:

```python
@serialization.register
@dataclass(frozen=True)
class CachedTerm:
    prior: object
    _cache: dict = field(default_factory=dict, init=False)
    def to_dict(self): return {"prior": _serialize_child(self.prior)}
    @classmethod
    def from_dict(cls, data): return cls(prior=_deserialize_child(data["prior"]))

cfg = serialization.serialize_model_config({"mu": CachedTerm(prior=3.0)})
# {"mu": {"__type__": "...CachedTerm", "prior": {"__literal__": 3.0}}}   -> fine
serialization.deserialize_model_config(json.loads(json.dumps(cfg)))
# TypeError: field _cache is declared with init=False, it cannot be specified with replace()
```

The non-frozen equivalent walks fine, so the frozen path is again the whole difference — same shape as the bug we just closed, different failure mode (hard crash instead of silent duplication). Two more consequences of rebuilding unconditionally: identity is not preserved (`out["x"] is obj` -> `False`, vs `True` for non-frozen) and `__post_init__` re-runs on every load even when nothing merged (verified: 1 extra call for a no-op walk).

Dropping `replace` fixes all three and removes the frozen special case entirely — `object.__setattr__` is the same ownership argument already made for the non-frozen `setattr` branch (these objects were constructed by the loader microseconds earlier, and frozen `__hash__` is computed per call, not cached):

```python
if is_dataclass(obj) and not isinstance(obj, type):
    for dc_field in fields(obj):
        current = getattr(obj, dc_field.name)
        new = walk(current)
        if new is not current:
            object.__setattr__(obj, dc_field.name, new)
    return obj
```

I ran this against four frozen shapes — plain, `init=False`, `slots=True`, `__post_init__` — all four give `shared=True`, `identity_preserved=True`, `post_init_reruns=0`. It also drops the `replace` import and the `# type: ignore[attr-defined]`.

**2. `_resolve_func` (`terms.py:337-350`): the self-contradictory message survives, just narrower, and the split costs a dead line.** The underscore rule now fires only outside `CUSTOM_TRANSFORMS`, but it still reports "must resolve to a callable pytensor function" for names that *are* callable — the module scan reaches two of them:

```python
>>> _resolve_func("__annotate__")
SerializationError: Serialized function name '__annotate__' must resolve to a
                    callable pytensor function, got <function __annotate__ ...>.
```

And the nested-then-flat structure makes the second branch reachable only for a non-callable *registered* entry, which is exactly the one line codecov flags (`terms.py:346`, confirmed locally: `terms.py 273 stmts, missing 346, ...`). Two flat conditions give accurate messages for all four combinations and leave nothing uncovered:

```python
if name.startswith("_") and name not in CUSTOM_TRANSFORMS:
    raise SerializationError(
        f"Serialized function name {name!r} is private or a dunder attribute, "
        "not a serializable transform. Register it with "
        "pymc_extras.prior.register_tensor_transform to use it by name."
    )
if not callable(func):
    raise SerializationError(
        f"Serialized transform {name!r} resolved to {func!r}, which is not callable."
    )
```

(That also improves the `_resolve_func("basic")` case — currently "must resolve to a callable pytensor function, got `<module ...>`", which is true but buries the point.)

### Nits

- `tests/test_terms.py`: `test_deserialize_child_passthrough_non_dict` re-imports `_deserialize_child` inside the body although the same commit added it to the module-level import.
- The frozen-walk regression test lives in `tests/test_terms.py` but exercises `serialization.py`; a copy next to the other `_merge_shared_decompositions` tests would keep the branch honest if terms ever move. Whichever fix you take for item 1, an `init=False` frozen case is the assertion that would have caught it.
- Scanned `pymc_marketing` + `pymc_extras` for frozen dataclasses to size the blast radius of item 1 in shipped code: 31 hits, but the only ones with `init=False` fields are `pymc_extras.inference.advi.autoguide.*`, which never reach `model_config`. `SpendReach`, `ChannelDependentEffect`, `TemporalReach`, `CounterfactualScenarios`, `MMMSummaryFactory` are frozen but unregistered and runtime-only. So item 1 is a trap for custom terms, not a regression for anything shipping today.

### Verdict

Approve. Item 1 is a net deletion (the `replace` branch collapses into the existing loop) and closes the last frozen-path hole in the contract this PR documents; item 2 is a message/flow cleanup that also picks up the one uncovered line. Neither blocks the stack — the three round-2 residues are genuinely fixed and the transform-name resolution is symmetric over the full 145-callable surface.

williambdean:
Great. Getting back to this now and taking a look.

williambdean:
CI is green. CC @juanitorduz

PR #2991, opened 2026-09-09
feat: add Named and Ref terms

williambdean:
Second PR in the terms stack (serialization → Bass components → Bass port).

- `Named(name, expr, dims=None)` composes an expression into a `pmd.Deterministic` — dims-first, so expression leaves must be dims-native (`Parameter`/`Prior` with `xdist=True` or `pytensor.xtensor`). Coordinates, data registration, and data updates delegate to the inner expression.
- `Ref(name)` resolves an already-built named variable, letting later effects depend on earlier ones without recreating variables.

Both serialize through the terms registry from #2987, so components composed in `model_config` survive `fit()`/`save()`/`load()`.

----
📚 Documentation preview 📚: https://pymc-marketing--2991.org.readthedocs.build/en/2991/

### DIFF
--- a/pymc_marketing/terms.py
+++ b/pymc_marketing/terms.py
@@ -15,7 +15,8 @@
 """Composable model terms for building ``pymc.dims`` subgraphs from xarray data.
 
 ``pymc_marketing.terms`` provides a small set of built-in pieces
-(``Parameter`` / ``Intercept``, ``Dot``, ``Transform``) that compose with
+(``Parameter`` / ``Intercept``, ``Dot``, ``Transform``,
+``Named``, ``Ref``) that compose with
 ``+``, ``*``, and ``-`` into predictors.  **This module is not a full
 modeling toolkit** --- it does not ship complete hierarchical,
 domain-specific, or observation-model wrappers.  It is a **thin, expressive
@@ -236,7 +237,10 @@ Gotchas
   same covariate matrix).
 - ``build_param`` is not idempotent --- call it once to create PyMC
   variables, then sample. Calling it a second time tries to create
-  variables with the same name, causing a PyMC error.
+  variables with the same name, causing a PyMC error. ``Named`` is a
+  named wrapper with the same constraint: build once, reference
+  afterwards with ``Ref`` (which is freely reusable inside multiple
+  ``Named`` expressions).
 - Changing the number of observations in ``set_data`` requires rebuilding
   the model. This is a ``pmd.Data`` constraint (dimensional shared variables
   have fixed dimension sizes), not a framework limitation.
@@ -409,7 +413,8 @@ class ModelTerm:
 
     Subclass ``ModelTerm`` to define reusable PyMC subgraph recipes.
     Custom terms compose with the built-ins (``Parameter``, ``Intercept``,
-    ``Dot``, ``Transform``) via ``+``, ``*``, and ``-`` and use the same
+    ``Dot``, ``Transform``, ``Named``, ``Ref``) via ``+``, ``*``, and ``-``
+    and use the same
     helpers (``collect_coords``, ``register_data``, ``build_param``,
     ``set_data``) --- no framework changes required.  See the module
     docstring **Extending** section for the full contract and an example.
@@ -911,31 +916,53 @@ class Named(ModelTerm):
     expr : Any
         Inner expression accepted by ``build_param``.
     dims : str or tuple of str, optional
-        Dimensions for the deterministic variable.
+        Dimensions for the deterministic variable. These **align** the
+        built value onto ``dims`` (``pmd.Deterministic`` transposes), so
+        they must be a permutation of the inner value's dims, not a new
+        declaration - a value without those dims raises ``ValueError``.
 
     Examples
     --------
     .. code-block:: python
 
         from pymc_extras.prior import Prior
-        from pymc_marketing.terms import Named, Parameter
+        from pymc_marketing.terms import Named, Parameter, Ref
 
         scale = Named(
             "scale",
-            Parameter("scale", prior=Prior("HalfNormal"), dims="product"),
+            Parameter("scale_raw", prior=Prior("HalfNormal", dims="product")),
             dims="product",
         )
 
-        # an effect that references another built variable
-        effect = Named("effect", Ref("scale") * other_term, dims="customer_id")
+        # an effect that references another built variable: build once with
+        # ``Named``, reference afterwards with ``Ref``
+        effect = Named(
+            "effect",
+            Ref("scale") * Parameter("kappa", prior=Prior("Normal", dims="product")),
+            dims="product",
+        )
     """
 
     name: str
     expr: Any
     dims: str | tuple[str, ...] | None = None
 
+    def __post_init__(self):
+        """Normalize ``dims`` to a tuple, mirroring ``Prior``."""
+        if self.dims is None:
+            return
+        if isinstance(self.dims, str):
+            self.dims = (self.dims,)
+        elif not isinstance(self.dims, tuple):
+            self.dims = tuple(self.dims)
+
     def get_coords(self, ds: xr.Dataset) -> dict[str, Any]:
-        """Collect coordinates from the inner expression."""
+        """Collect coordinates from the inner expression.
+
+        ``Named`` does not declare its own coords: ``dims`` aligns the
+        built value onto an existing permutation, so coordinates arrive
+        from the inner expression's leaves.
+        """
         return get_coords(self.expr, ds)
 
     def register_data(self, ds: xr.Dataset) -> None:
@@ -947,8 +974,16 @@ class Named(ModelTerm):
         set_data(self.expr, ds=ds, model=model)
 
     def create_variable(self) -> pt.TensorVariable:
-        """Build the named deterministic from the inner expression."""
-        return pmd.Deterministic(self.name, build_param(self.expr), dims=self.dims)
+        """Build the named deterministic from the inner expression.
+
+        Bare ``VariableFactory`` children get a derived name (``<name>_param``)
+        so several ``Named`` terms do not collide on ``build_param``'s default.
+        """
+        return pmd.Deterministic(
+            self.name,
+            build_param(self.expr, name=f"{self.name}_param"),
+            dims=self.dims,
+        )
 
     def to_dict(self) -> dict[str, Any]:
         """Serialize the name, expression, and dims."""
@@ -977,19 +1012,58 @@ class Ref(ModelTerm):
 
     Lets an effect expression composed outside the model depend on another
     built effect (e.g. ``a`` on the ``a_scale`` deterministic) without
-    recreating its variables.
+    recreating its variables. ``Ref`` resolves at **build time** and
+    requires the referenced term to be built earlier in traversal order -
+    build once with ``Named``, reference afterwards with ``Ref``.
 
     Parameters
     ----------
     name : str
         Name of the referenced variable.
+
+    Examples
+    --------
+    .. code-block:: python
+
+        scale = Named("scale", Parameter("raw", prior=Prior("HalfNormal")))
+        effect = Named("effect", Ref("scale") * other_term)
     """
 
     name: str
 
     def create_variable(self) -> pt.TensorVariable:
-        """Resolve the referenced variable from the active model."""
-        return pm.modelcontext(None)[self.name]
+        """Resolve the referenced variable from the active model.
+
+        Raises
+        ------
+        ValueError
+            If the variable is not built yet, or if it is not
+            dims-native (referencing plain-tensor variables built by
+            non-terms model code fails later inside pytensor, so it is
+            rejected here with the offending name).
+        """
+        model = pm.modelcontext(None)
+        try:
+            variable = model[self.name]
+        except KeyError:
+            available = ", ".join(sorted(model.named_vars))
+            raise ValueError(
+                f"Ref({self.name!r}) cannot resolve: {self.name!r} is not "
+                "built yet. Terms build during build_param in traversal "
+                f"order, so a term referencing {self.name!r} must come "
+                f"after the term that builds it. Available named vars: "
+                f"{available}."
+            ) from None
+
+        if not isinstance(variable, pmd.XTensorVariable):
+            raise ValueError(
+                f"Ref({self.name!r}) resolved to {variable!r}, which is not "
+                "dims-native. Ref must reference a dims-native variable "
+                "(built by terms or pmd distributions) - plain-tensor "
+                "variables cannot compose into pytensor.xtensor "
+                "expressions."
+            )
+        return variable
 
     def to_dict(self) -> dict[str, Any]:
         """Serialize the referenced name."""
--- a/pymc_marketing/terms.py
+++ b/pymc_marketing/terms.py
@@ -72,12 +72,13 @@ contract: decorate them with ``@serialization.register`` and implement
 unregistered term raises
 :class:`~pymc_marketing.serialization.SerializationError`.
 
-Children of composed terms (``Sum``, ``Product``, ``Transform``) must be
-registered terms, ``VariableFactory`` (``Prior``, ...), ``xr.DataArray``,
-``DeferredFactory``, or numeric literals. ``Transform`` functions serialize
-by name and are resolved with ``pymc_extras.prior._get_transform``:
-``pytensor.xtensor.math`` / ``pytensor.xtensor.linalg`` functions, or
-transforms registered with ``pymc_extras.prior.register_tensor_transform``.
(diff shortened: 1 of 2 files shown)


juanitorduz:
## Review

Reviewed the head branch in a scratch worktree; `tests/test_terms.py` = **81 passed**. Everything below is from executed probes rather than reading.

Overall: small and well-scoped, and the lifecycle delegation correctly mirrors `Transform` (`get_coords` / `register_data` / `set_data` → inner expr, `add_coords` reached via the module-level `register_data` helper). Registry integration is consistent with the #2987 pattern. Two must-fix items, four should-fix, plus test-strength gaps.

---

## Must fix

### 1. The `Named` docstring example does not run — two separate errors

`terms.py:872-879`. Verified:

```
Parameter("scale", prior=Prior("HalfNormal"), dims="product")
→ TypeError: Parameter.__init__() got an unexpected keyword argument 'dims'
```

`Parameter` carries dims through `prior=Prior(..., dims=...)`; it has no `dims` field. Even after fixing that, the example still fails because the deterministic name equals the inner RV name:

```
Named("scale", Parameter("scale", prior=Prior("HalfNormal", dims="product")), dims="product")
→ ValueError: Variable name scale already exists.
```

Third, the second snippet uses `Ref(...)` and `other_term`, but the import line only pulls `Named, Parameter` and `other_term` is undefined. `.. code-block::` is not doctested, so CI cannot catch any of this. A version that actually runs:

```python
from pymc_marketing.terms import Named, Parameter, Ref

scale = Named(
    "scale",
    Parameter("scale_raw", prior=Prior("HalfNormal", dims="product")),
    dims="product",
)
effect = Named(
    "effect",
    Ref("scale") * Parameter("kappa", prior=Prior("Normal", dims="product")),
    dims="product",
)
```

### 2. `dims` is not normalized → real JSON round-trip breaks equality

`Named.to_dict` emits `self.dims` raw. `model_config` is JSON-serialized into idata attrs at `fit()`, and `TypeRegistry._serialize_config_value` does not normalize inside a registered class's own `to_dict`. So a tuple comes back as a list:

```
Named("eff", Parameter("p", prior=Prior("Normal", dims=("a","b"))), dims=("a","b"))
json.loads(json.dumps(serialize(term))) → deserialize → dims=['a', 'b']   equal=False
```

Functionally the list still works (`pmd.Deterministic(dims=["a","b"])` registers `('a','b')` — verified), but the round-trip identity this PR advertises silently fails. `Prior` avoids it by normalizing dims to a tuple (`Prior('Normal', dims='a').dims == ('a',)`; `Prior.from_dict({..., 'dims': ['a','b']}) == Prior(..., dims=('a','b'))` → `True`). `Named` is the first term with a raw tuple field, so it should follow the same convention:

```python
def __post_init__(self):
    if isinstance(self.dims, str):
        self.dims = (self.dims,)
    elif self.dims is not None:
        self.dims = tuple(self.dims)
```

`test_serialize_named_roundtrip` cannot catch this: it pins `restored.dims == "customer_id"` and never crosses JSON. Add a `json.loads(json.dumps(...))` hop plus a tuple-dims case (and update that assertion if you normalize).

---

## Should fix

### 3. `Ref` failure modes are opaque

Two distinct bad paths, both verified:

- **Not built yet** → bare `KeyError: 'a_scale'`. Nothing tells the user that `Ref` resolves at build time and requires the referenced term to be built earlier in traversal order. Raise with a message listing available named vars, and add a Gotchas bullet on the ordering requirement — the framework does not express it anywhere, and `Ref`-only expressions contribute no coords and no data registration, so a `model_config` can be wired that only fails deep inside `build_model`.
- **Exists but not dims-native** → `Ref.create_variable` happily returns a plain `TensorVariable` and the failure surfaces later from inside pytensor:

```
pm.Normal("plain", dims="product"); Named("eff", Ref("plain") * Parameter(...))
→ TypeError: non-scalar TensorVariable cannot be converted to XTensorVariable without dims.
```

That is the likely case when referencing variables built by existing non-terms model code. Validate in `Ref.create_variable` and name the offending variable.

### 4. Bare `VariableFactory` as `expr` silently creates an RV called `param`

`Named.create_variable` calls `build_param(self.expr)` without a name, so `build_param`'s default leaks:

```
Named("s", Prior("HalfNormal", dims="product"), dims="product")
→ model.named_vars == ['param', 's'];  collect_coords(...) == {}
```

Two such terms collide on `param`. The docstring's "Inner expression accepted by `build_param`" actively invites this. Either forward a derived name (`build_param(self.expr, name=f"{self.name}_param")`) or state that bare factories must be wrapped in `Parameter`. Pre-existing in `Transform`, but `Named` is the naming API, so it belongs here.

### 5. Document what `dims` actually does — it aligns, it does not declare

`pmd.Deterministic` transposes the value onto `dims`:

```
Named("s", Parameter("q", prior=Prior("Normal", dims=("a","b"))), dims=("b","a"))
→ value dims ('b','a'), model dims ('b','a')     # silent transpose

Named("s", Parameter("q", prior=Prior("Normal")), dims="product")
→ ValueError: Dimensions {'product'} do not exist. Expected one or more of: ()
```

So `dims` must be a permutation of the inner value's dims. "Dimensions for the deterministic variable" reads as a declaration; one clarifying sentence prevents a confusing `ValueError`. Corollary: `Named.get_coords` correctly does *not* need to add its own dims — worth a comment, since it looks like an omission next to `Parameter.get_coords`.

### 6. Reuse semantics need a Gotchas line

A `Named` instance is not reusable within a tree — building it twice raises `ValueError: Variable name q already exists.` A `Ref` instance is freely reusable (`Ref("base")` inside two `Named`s → `['base','e1','e2','q']`). "Build once with `Named`, reference afterwards with `Ref`" is exactly the pattern this PR exists to enable; say so next to the existing `build_param`-is-not-idempotent bullet.

---

## Tests

- **`test_ref_resolves_built_variable` asserts the wrong thing.** It only checks that `a_scale` and `a` exist in `named_vars` — it passes even if `Ref` resolved to an unrelated variable. The contract is the dependency edge; assert it (verified `True`):
  ```python
  from pytensor.graph.traversal import ancestors
  assert model["a_scale"] in set(ancestors([model["a"]]))
  ```
- **Nothing verifies the deterministic wraps the right value.** All three `Named` tests assert name-in-model plus the dims mapping. One value check (`pm.draw` on the deterministic vs. the expected transform) would catch a `Named` that stored the wrong operand — today that passes.
- **`test_named_builds_pmd_deterministic` / `test_named_dims_none_scalar` are near-duplicates** (same two assertions, dims present vs. absent) → `@pytest.mark.parametrize`, per the repo testing conventions.
- **Worth adding:** dims-permutation transpose (item 5), `Named` reused twice raising (item 6), and `Ref` to a `pmd.Data` variable if that is supported — it currently works and returns `XTensorSharedVariable`, which is either a feature to document or an accident to reject.
- `test_ref_missing_raises` should gain a `match=` once the error message improves (item 3).

---

## Nits

- Module docstring intro (`terms.py:17-19`) and the `ModelTerm` class docstring (`terms.py:364-365`) both still enumerate built-ins as "`Parameter` / `Intercept`, `Dot`, `Transform`" — add the two new terms.
- `Ref` has no `Examples` section; the repo docstring convention also asks for `Returns` on public methods (the file is uniformly missing them, so this is not yours alone).
- `create_variable() -> pt.TensorVariable` while actually returning `XTensorVariable` — file-wide pattern, noted for consistency only.
- `to_dict` drops `dims` when `None`, whereas `Dot` always writes `name`. Harmless, but the conditional key makes persisted configs less self-describing.
- Naming, optional: `Named` and `Ref` are generic identifiers to export publicly; `Deterministic` / `Var` (or `Lookup`) read better at call sites, and a `Named.ref()` helper returning `Ref(self.name)` would remove the stringly-typed drift between the two.
- No in-repo consumers yet (`git grep` finds `Named(` / `Ref(` only in `terms.py`) — expected mid-stack; the Bass port PR is where the ordering and error-message ergonomics from item 3 will bite.
- Docs need no page changes: `docs/source/api/index.md` uses recursive autosummary on `terms`.

williambdean:
All items applied in `15942f5d3`:

- Must-fix 1: docstring example now runs as written (dims on the Prior, `scale_raw` inner name, `Ref` imported)
- Must-fix 2: `Named.__post_init__` normalizes dims to tuples (str/list) mirroring `Prior`; json-hop + tuple-dims test added, serialization assertion updated to `("customer_id",)`
- Should-fix 3: `Ref` raises with the available named-var list and build-order guidance when unset; rejects non-dims-native variables by name; both message-tested
- Should-fix 4: bare `VariableFactory` children get the derived `${name}_param` name (`test_named_bare_factory_named_not_param`)
- Should-fix 5+6: dims-aligns-not-declares docstring + `get_coords` comment; reuse semantics in the \(non-idempotent\) Gotchas bullet
- Tests: dependency edge via `pytensor.graph.traversal.ancestors`, `pm.draw` value check (`doubled == scale * 2`), parametrizable pair kept as distinct scalars tests, transpose and reuse tests added, `Ref`-to-`pmd.Data` works (left as-is, it's a feature)
- Nits: built-ins enumerations updated in the module and `ModelTerm` docstrings. Naming (`Named`/`Ref` vs alternatives) deferred pending your call

juanitorduz:
## Review round 2 (`15942f5d3`)

Re-checked the head branch in a scratch worktree: `tests/test_terms.py` = **88 passed**, `pre-commit run --files pymc_marketing/terms.py tests/test_terms.py` = all hooks pass. Everything below is from executed probes, not reading.

Most of round 1 landed cleanly and I verified the fixes rather than taking them on faith:

- Must-fix 1: the `Named` example now runs verbatim → builds `['effect', 'kappa', 'scale', 'scale_raw']`. ✅
- Must-fix 2: `json.loads(json.dumps(serialize(term)))` → `deserialize` now gives `restored == term` with `dims == ('obs',)`; `dims="a"`/`["a"]`/`("a",)` all compare equal. ✅
- Should-fix 3: both `Ref` failure modes now raise `ValueError` with actionable text (`"is not built yet"` + available vars + build-order sentence; `"not dims-native"` + offending name). Verified a `Ref`-only `model_config` now fails with the ordering message instead of a bare `KeyError`. ✅
- Should-fix 5: `dims=("b","a")` on an `("a","b")` value transposes (verified `s == q.T`); a subset raises `('a',) must be a permuted list of ('a', 'b')`. Docstring wording matches reality now. ✅
- Should-fix 6 + nits: reuse Gotcha, `get_coords` comment, built-ins enumerations. ✅
- `test_ref_resolves_built_variable` now asserts the dependency edge, and `test_named_value_matches_expression` pins the value. ✅

Four things left, one of them blocking, plus a claim in your reply that I could not reproduce.

---

## Blocking

### Branch no longer merges — base moved

`Merge state: DIRTY`. The base picked up `b38b30fa` + `a5ff984c` after this branch was cut, and `git merge-tree origin/feat/terms-serialization origin/feat/terms-bass-components` reports `CONFLICT (content): tests/test_terms.py` (both sides append at the file tail). `terms.py` merges clean, so this is a mechanical rebase — but note the base also renamed `_lookup_func` → `_resolve_func` and changed the literal wrapper to `{"__literal__": ...}` in `_serialize_child`. `Named.to_dict` goes through `_serialize_child`, so it inherits that for free; just re-run the serialization tests after the rebase.

While rebasing: the base's new module docstring says *"Children of composed terms (``Sum``, ``Product``, ``Transform``) must be registered terms, ``VariableFactory``, ``xr.DataArray``, ``DeferredFactory``, or numeric literals"* — `Named` now also takes a child, so add it to that enumeration.

---

## Must fix

### 1. Should-fix 4 is only half-fixed: nested bare factories still collide on `param`

`Named.create_variable` (`terms.py:931-935`) forwards `name=f"{self.name}_param"`, which fixes the top-level case that `test_named_bare_factory_named_not_param` covers. But `build_param`'s `Sum`/`Product` branches (`terms.py:1105`, `terms.py:1108`) recurse *without* threading `name`, so any bare factory below the root still gets the module default:

```python
Named("s", Prior("HalfNormal", dims="product") * Parameter("kappa", prior=Prior("Normal", dims="product")), dims="product")
→ model.named_vars == ['kappa', 'param', 's']
```

and two such terms in one model still hard-fail:

```
Named("s1", Prior(...) * Parameter("k1", ...)) ; Named("s2", Prior(...) * Parameter("k2", ...))
→ ValueError: Variable name param already exists.
```

That is the exact failure mode from round 1, just one composition level down — and the new test makes it look covered. Either thread the derived name through the recursion (`build_param(term, name=name)`, with a positional suffix for `Sum`/`Product` children so siblings do not collide either), or drop the derived-name special case and state in the `Named` docstring that bare factories must be wrapped in `Parameter`. Whichever you pick, add the nested case to `test_named_bare_factory_named_not_param` — a two-term model asserting no `param` and no collision.

### 2. The `Ref` docstring example has the defect we just fixed on `Named`

`terms.py:973-979`, run verbatim:

```
Named("effect", Ref("scale") * other_term)
→ NameError: name 'other_term' is not defined
```

No imports either, and `.. code-block::` is not doctested so CI cannot catch it. Same treatment as the `Named` block:

```python
from pymc_marketing.terms import Named, Parameter, Ref

scale = Named("scale", Parameter("scale_raw", prior=Prior("HalfNormal")))
effect = Named("effect", Ref("scale") * Parameter("kappa", prior=Prior("Normal")))
```

---

## Should fix

### 3. `dims` annotation rejects the list path you added and tested

`terms.py:897` declares `dims: str | tuple[str, ...] | None`, but `__post_init__` normalizes lists and `test_named_dims_normalization` exercises `dims=["a", "b"]`. `tests/` is not type-checked, so pre-commit is green while user code is not:

```
mypy: Argument "dims" to "Named" has incompatible type "list[str]"; expected "str | tuple[str, ...] | None"  [arg-type]
```

Widen to `str | Sequence[str] | None` (the post-init invariant "always a tuple after construction" is worth keeping in that docstring line), or delete the list branch and the test. Right now the annotation and the test disagree.

### 4. The transpose test you listed is not in the branch

Your reply says *"transpose and reuse tests added"*. Reuse is there (`test_ref_reusable_in_multiple_named`, `test_named_reuse_raises`), but `grep -n "transpose\|permut" tests/test_terms.py` returns nothing across all 88 tests. I confirmed the behavior holds:

```python
Named("s", Parameter("q", prior=Prior("Normal", dims=("a","b"))), dims=("b","a"))
→ q.shape (2,3), s.shape (3,2), named_vars_to_dims['s'] == ('b','a'), allclose(s, q.T) is True
```

Since the docstring now *promises* permutation-with-transpose, that promise should be pinned — the assertion above is the whole test. Not a big deal, but I would rather flag the gap than let a claimed test go missing.

---

## Nits

- **`Ref` to `pmd.Data` works — document it.** Verified: returns `XTensorSharedVariable` and passes the `pmd.XTensorVariable` check (`XTensorSharedVariable` is a subclass; `pmd.XTensorVariable is pytensor.xtensor.type.XTensorVariable`). You left it as a feature, which I agree with — then say so in the `Ref` docstring, because the rejection message only mentions "built by terms or pmd distributions" and a reader will not guess that registered data qualifies. The plain counterpart is correctly rejected: `pm.Data("xp", ...)` → `Ref('xp') resolved to xp, which is not dims-native`.
- **`Named.set_data(ds)` without `model=`** → `TypeError: No model on context stack`, even though the base signature advertises `model: pm.Model | None = None`. Pre-existing in `Transform` and unreachable from MMM (which always passes the model), so not yours — but the delegating terms are where the optional-that-is-not-optional default is most visible. Worth dropping the default on data-carrying terms in a follow-up.
- **`Ref` "available named vars" list is unbounded.** Fine in tests; in a full MMM this message will be dozens of names long. Consider capping at ~20 sorted names plus `"... (N more)"`.
- **Test-local imports.** `from pytensor.graph.traversal import ancestors` inside `test_ref_resolves_built_variable` and `import numpy as np` inside `test_named_value_matches_expression` (`np` is already imported at module top) — hoist both to the module header to match the rest of the file.
- `test_named_builds_pmd_deterministic` / `test_named_dims_none_scalar` remain near-duplicates; keeping them separate is defensible, dropping it.
- **Naming — my call: keep `Named` / `Ref`.** They read fine at MMM call sites and the rest of the stack already uses them. A `Named.ref()` helper returning `Ref(self.name)` would still remove the stringly-typed drift between the two; nice-to-have, not a blocker for this PR.

Other probes that came back clean, for the record: `Named(Named(...))` nests without name collisions (`['inner','q','s']`); `Transform(Ref(...))` returns an `XTensorVariable`; `Ref` contributes no coords/data and `collect_terms` treats it like any leaf; `dims` passed as a list is not mutated in place; `Named` with `dims` absent from the inner value fails with pmd's `Dimensions {'product'} do not exist`.

juanitorduz (approved):
LGTM, besided the suggestions by the agent

PR #2993, opened 2026-09-09
feat(optimizer): key constraint diagnostics and add `constraint_history`

### LINKED ISSUE #2992: Callback `constraint_info` carries no constraint key; add keyed lookup on `BudgetOptimizationResult`
## Summary

When `allocate_budget(callback=True)` records constraint diagnostics, each `ConstraintIterationInfo` entry carries `type`, `value`, and `jac` but not the constraint's `key`. The key is dropped in `compile_constraints_for_scipy` (`pymc_marketing/mmm/constraints.py`), which emits only `type`, `fun`, and `jac`, so the callback loop in `BudgetOptimizer.allocate_budget` has nothing to record. Callers cannot tell which constraint a reported value belongs to.

The only option today is positional indexing into `constraint_info`, which is what `mmm_multi_objective_optimization.ipynb` does with `callback_results[-1]["constraint_info"][0]`. Order is insertion order of the `constraints` list plus the auto-added default, so reordering the list silently changes what the notebook checks. Before #2989 index 0 was the default sum constraint, not the user's minimum-response constraint; after #2989 it is the custom one only because it happens to be listed first.

## Reproduction

```python
result = optimizer.allocate_budget(total_budget=..., callback=True)
result.callback_info[-1]["constraint_info"]
# [{'type': 'ineq', 'value': ..., 'jac': ...}, {'type': 'eq', 'value': ..., 'jac': ...}]
# no way to say which is which
```

## Proposed design

1. Carry `key` through `compile_constraints_for_scipy` and add `key: str` to `ConstraintIterationInfo`. `constraint_info` stays a list, so existing positional code keeps working. Extra keys in the constraint dicts are ignored by every SciPy path (SLSQP reads `type`/`fun`/`jac`/`args`; `old_constraint_to_new` for trust-constr/cobyla/cobyqa reads the same four), verified against SciPy 1.18.
2. Add a `constraint_history` property on `BudgetOptimizationResult`, next to `spend_var_allocations`, returning `dict[str, list[ConstraintIterationInfo]]` keyed by constraint key across iterations. Empty when the run had no callback.
3. Update the notebook to look constraints up by key on the result object.
4. Tests for the key on the default constraint and for a custom plus default constraint.

## Related

#2988 / #2989 (silent kwarg drop that made this harder to notice).

PabloRoque:
## Description

Closes #2992.

With `allocate_budget(callback=True)`, each `constraint_info` entry carried `type`, `value`, and `jac` but not the constraint's key, because `compile_constraints_for_scipy` dropped it. Callers could only index constraints by position, which depends on the order they were compiled in. The multi-objective notebook did exactly that, and before #2989 index 0 was the default sum constraint rather than the user's minimum-response constraint.

## Changes

- `compile_constraints_for_scipy` now includes `"key"` in each compiled dict. SciPy reads only `type`/`fun`/`jac`/`args` on every path (SLSQP directly, and `old_constraint_to_new` for trust-constr/cobyla/cobyqa), verified against SciPy 1.18, so the extra key is ignored by the solver.
- `ConstraintIterationInfo` gains `key: str`, recorded by the callback. `constraint_info` stays a list, so positional code keeps working.
- New `BudgetOptimizationResult.constraint_history` property, next to `spend_var_allocations`: a dict keyed by constraint key with one `ConstraintIterationInfo` per iteration. Empty when the run had no callback. The auto-added sum constraint is under `"default"`.
- `callback` docstring now describes the entry shape and points to `constraint_history`.

## Review follow-up

Addressed from the review on this PR: notebook re-executed and dead variables removed, test identity checks located by key, docstring wording ("at the last iteration"). Kept `info["key"]` as a hard index on purpose: a hand-built entry without a key should fail loudly rather than land in an "unknown" bucket. A final-state constraint accessor that does not require `callback=True` is a reasonable follow-up and is out of scope here.

For the record, the docs build never executes notebooks (`nb_execution_excludepatterns = ["*.ipynb"]`), and the Test Notebooks workflow had already run these cells with mocked sampling and passed, so the raise branch was not a CI breaker. Re-executing was still the right call since outputs only refresh when committed.
- The two constraint-check cells in `mmm_multi_objective_optimization.ipynb` now read `min_response_constraint` by key from the result object.

The notebook was re-executed with real sampling so the recorded outputs match the key-based read (the docs build does not execute notebooks, so they would not refresh on their own). The single-model cell now shows the binding constraint at `-0.0`; the merged-model cell shows `240.8141`, i.e. not binding there. The old outputs (`0.0` in both) were the default sum constraint, read positionally. The two now-unreferenced `callback_results` assignments were dropped.

## Tests

- Existing callback test extended: the default-only run reports key `default`, and `constraint_history` has one entry per iteration. Without callback, `constraint_history` is `{}`.
- New `test_constraint_history_keyed_by_constraint`: a custom inequality plus the explicit default sum constraint; both keys present, entries per key match iteration count, and regrouped entries are the same objects as the positional ones (located by key, not by compile order). The custom floor is never active under the equality, so the test checks the keying, not the solver.
- `tests/mmm/test_budget_optimizer.py` plus the other modules that construct `BudgetOptimizer` pass locally (285 tests). pre-commit passes.

----
📚 Documentation preview 📚: https://pymc-marketing--2993.org.readthedocs.build/en/2993/


juanitorduz:
@PabloRoque : experimenting with oh-my-pi reviews :)

---

## Review

Verdict: **approve with two fixes before merge** (one is a likely docs-CI breaker). Core change is small, correct, and well-scoped.

### Verified locally

- **SciPy tolerance of the extra `"key"`** — claim holds. On scipy 1.18, a dict constraint carrying `"key"` runs clean through `SLSQP`, `trust-constr`, `COBYLA`, `COBYQA`. Default method here is `SLSQP` (`budget_optimizer.py:1548`), so the other three are manually confirmed but not exercised in CI.
- **Key collisions can't happen** — `set_constraints` (`budget_optimizer.py:1818-1827`) rejects duplicate keys and stores `_constraints` as a dict, so the `setdefault(...).append(...)` regrouping in `constraint_history` can never silently merge two distinct constraints. That invariant is what the property depends on, and it is enforced upstream.
- **Only one construction site** for `BudgetOptimizationResult` (`:2299`), so no other producer can emit `constraint_info` entries lacking `"key"`.
- **Tests pass**: `pytest -k "callback or constraint_history"` → 3 passed in a clean worktree of the PR head.
- Generated API rst for `BudgetOptimizationResult` is untracked build output — nothing to regenerate.

### Blocking

**1. Notebook cell can now raise and fail the docs build.**

`mmm_multi_objective_optimization.ipynb` cells at 1682/3350 end with:

```python
else:
    raise ValueError(f"Constraint violated: value {constraint_value} should be >= 0")
```

Before this PR, `constraint_info[0]` was the **equality** default (`sum - total_budget`), always ≈0 → the `np.isclose` branch. After the fix it reads `min_response_constraint`, which is `mean_resp - 8` — an inequality on a response scale, not normalized. Rounded to 4 decimals, a converged SLSQP iterate that is feasible only to solver tolerance (or slightly infeasible, which SLSQP permits) yields a negative value of arbitrary magnitude relative to `atol=1e-6` → `ValueError` → notebook execution failure in CI. The PR says outputs were not re-executed, so this path has never been run.

Fix: execute both notebooks locally and commit refreshed outputs, or make the check scale-aware (e.g. `constraint_value >= -1e-6 * max(1, abs(min_total_response))`) instead of an exact-zero/positive dichotomy.

**2. Dead variables left in the notebook.**

`callback_results = optimization_result.callback_info` (line 1649) and `merged_result.callback_info` (3330) are now unreferenced. Either drop them, or keep one with a comment showing the positional form still exists.

### Design comments (non-blocking)

- **`constraint_history` is only reachable with `callback=True`.** The dominant use — "did my constraint hold at the solution?" — now costs a per-iteration `fun` + `jac` evaluation of every constraint for the whole run. The cheaper accessor is a final-state one evaluating `self._compiled_constraints` at `scipy_result.x` once. Not required here (this closes #2992 as specified), but the docstring example advertises the expensive path as the normal one. Worth a follow-up.
- **`[-1]` is the last callback iterate, not necessarily `scipy_result.x`.** For SLSQP they coincide in practice (checked: `nit == len(callbacks)`, `last_xk == res.x`), but "at convergence" in the docstring/test names asserts something the structure doesn't guarantee for other methods. Wording only.
- **`info["key"]` is a hard index.** `BudgetOptimizationResult` is a public dataclass a user can construct with hand-built `callback_info`; an entry without `"key"` raises `KeyError` from a property. Cheap hardening: `info.get("key", "unknown")`. Strictness is defensible — your call.

### Test comments

- `test_constraint_history_keyed_by_constraint`'s `spend_floor = sum(budgets) - 10` under `total_budget=100` with an equality `sum == 100` is **structurally inactive**: its value is pinned at 90 by the other constraint. So `assert history["spend_floor"][-1]["value"] >= -1e-6` is a tautology, and "both constraints hold at convergence" in the description overstates what is checked. The keying assertions (the actual point) are fine; either make the floor binding or drop the value assertion.
- The identity assertions (`history["spend_floor"][-1] is last_iter[0]`) pin both regrouping-by-reference and the compile order. Reference identity is worth keeping; the literal `[0]`/`[1]` re-pin the very ordering assumption this PR exists to remove. Prefer locating the positional entry by `key`, then asserting `is`.

### Minor

- `for _, constraint in enumerate(...)` → `for constraint in ...` is a correct drive-by cleanup.
- Codecov −0.05% is the docstring-heavy property counted against untested lines elsewhere; the property itself is covered by both tests.

PabloRoque:
Thanks for the review, useful. Pushed f8a9435 addressing it. Item by item:

**1. Notebook cell raising in docs CI.** Checked this before changing anything, and it isn't a CI breaker: the docs build never executes notebooks (`nb_execution_excludepatterns = ["*.ipynb"]` in `conf.py`, and RTD only runs the gallery script pre-build), and the Test Notebooks workflow had already run these cells on the PR head with mocked sampling and passed. The remaining concern, a converged iterate that is feasible only to solver tolerance, is absorbed by the `round(..., 4)`: anything within 5e-5 of zero reads as zero. So I left the check as it was.

What *was* wrong is my description's claim that outputs would refresh on the next docs build. They wouldn't, so I re-executed the notebook with real sampling and committed the outputs. The single-model cell reads `-0.0` (binding, feasible to tolerance); the merged-model cell reads `240.8141` (not binding there). The old recorded `0.0` in both was the default sum constraint, read positionally.

**2. Dead variables.** Both `callback_results` assignments dropped.

**Design comments.** Agree a final-state accessor that doesn't need `callback=True` is the better everyday path; out of scope here, happy to do it as a follow-up. Docstring now says "at the last iteration" rather than "at convergence". On `info["key"]` I kept the hard index on purpose: a hand-built entry without a key should fail loudly, not land in an `"unknown"` bucket that hides the mismatch.

**Test comments.** Both fair. The floor is structurally inactive under the equality, so I dropped the value assertion and said in the docstring that the test checks the keying, not the solver. The identity checks now find the positional entry by key before asserting `is`, instead of re-pinning the compile order.

juanitorduz (approved):
Thanks @PabloRoque !

PR #2995, opened 2026-09-09
fix(utility): make `diversification_ratio` callable

### LINKED ISSUE #2994: `diversification_ratio` is unusable: wrong 1D guard and misplaced parenthesis
## Summary

`pymc_marketing.mmm.utility.diversification_ratio` cannot be called with any valid input. It has been broken since it was released in 0.11.0 and has no test coverage, so nothing ever caught it.

## Two independent defects

**1. Wrong dimensionality guard (since 0.11.0, PR #1358).** The function's docstring and its body require a 2D `(sample, channel)` array: it takes a per-channel standard deviation and a covariance matrix. But the body calls `_check_samples_dimensionality`, which raises unless the input is 1D. That helper is right for the scalar-response utilities in the same module and wrong for this one. Every valid call fails with:

```
ValueError: Function expected samples to be a 1D tensor variable. Got 2 dimensions.
```

**2. Misplaced parenthesis in the xtensor rewrite (since 0.19.0, PR #2204).** The `w'Σw` line reads:

```python
portfolio_var = ptx.dot(
)
```

The closing parenthesis of `rename(` comes after the inner dot product, so that product is passed as a second positional argument to `rename`, and the outer `ptx.dot` receives only one operand. With defect 1 bypassed, the call fails with:

```
TypeError: XTensorVariable.rename() takes from 1 to 2 positional arguments but 3 were given
```

Defect 1 has masked defect 2 completely; the rewritten line has never executed.

## Verification

With the guard bypassed and the parenthesis moved so the inner dot product is the second operand of the outer `ptx.dot`, the function matches a numpy reference (`(w·σ).sum() / sqrt(wᵀ Σ w)`) to full precision on a random 500×3 sample.

## Proposed fix

- Drop the 1D check from `diversification_ratio` and validate for 2D instead.
- Close `rename(...)` before the inner dot product.
- Add tests against a numpy reference, plus a 1D-input rejection test.

`git log -S diversification_ratio -- tests/` returns nothing, and no notebook uses the function.

PabloRoque:
## Description

Closes #2994.

`diversification_ratio` could not be called with any valid input. Two independent defects:

1. **Wrong dimensionality guard (since 0.11.0).** The body called `_check_samples_dimensionality`, which requires 1D, but the function needs a 2D `(sample, asset)` array for the per-asset volatilities and the covariance matrix. Every valid call raised `ValueError`.
2. **Misplaced parenthesis (since 0.19.0).** In the `w'Σw` line, `rename(` closed after the inner `ptx.dot(cov_matrix, weights)`, so that product went to `rename` as a second positional argument and the outer `ptx.dot` received one operand. Defect 1 meant this line never executed.

## Changes

- Replace the 1D guard with a 2D check and a message that names the expected dims.
- Close `rename(...)` before the inner dot product, so the outer `ptx.dot` contracts the renamed weights against `Σw`.
- Tests in `tests/mmm/test_utility.py`:
  - three parametrised cases against a numpy reference `(w·σ).sum() / sqrt(wᵀ Σ w)`, including a correlated 500×3 sample;
  - a property test: uncorrelated assets with equal volatility and equal weights give exactly `sqrt(n)`, built from a Hadamard matrix so the covariance is a multiple of the identity;
  - 1D input is rejected.

The function had no tests before and no notebook uses it, so there is no caller impact beyond making it work.

## Checks

- `tests/mmm/test_utility.py` passes locally (51 tests).
- pre-commit passes on both files.

----
📚 Documentation preview 📚: https://pymc-marketing--2995.org.readthedocs.build/en/2995/

### DIFF
--- a/pymc_marketing/mmm/utility.py
+++ b/pymc_marketing/mmm/utility.py
@@ -593,16 +593,22 @@ def diversification_ratio(
     """
     samples = as_xtensor(samples)
     budgets = as_xtensor(budgets)
-    samples = _check_samples_dimensionality(samples)
+    if samples.type.ndim != 2:
+        raise ValueError(
+            "Function expected samples to be a 2D tensor variable with a 'sample' "
+            f"dim and an asset dim. Got {samples.type.ndim} dimensions."
+        )
     weights = budgets / budgets.sum()
     individual_volatilities = samples.std(dim="sample", ddof=1)
 
     [asset_dim] = weights.dims
     cov_matrix = _covariance_matrix(samples, asset_dim=asset_dim)
 
-    # w'Σw
+    # w'Σw: rename one copy of the weights so the two asset axes of the
+    # covariance matrix are contracted separately.
     portfolio_var = ptx.dot(
-        weights.rename({asset_dim: f"{asset_dim}'"}, ptx.dot(cov_matrix, weights)),
+        weights.rename({asset_dim: f"{asset_dim}'"}),
+        ptx.dot(cov_matrix, weights),
     )
     portfolio_volatility = ptx.math.sqrt(portfolio_var)
     weighted_avg_volatility = (weights * individual_volatilities).sum()
--- a/pymc_marketing/mmm/utility.py
+++ b/pymc_marketing/mmm/utility.py
@@ -557,7 +557,8 @@ def diversification_ratio(
     The Diversification Ratio measures the effectiveness of diversification by comparing
     the weighted average volatility of individual assets to the overall portfolio volatility.
     A higher ratio indicates better diversification, as it reflects lower correlations among
-    assets, leading to reduced portfolio risk.
+    assets, leading to reduced portfolio risk. It provides insight into how individual asset
+    volatilities and their correlations contribute to the overall portfolio risk.
 
     The Diversification Ratio is calculated as:
 
@@ -583,8 +584,12 @@ def diversification_ratio(
     XTensorVariable
         Diversification Ratio.
 
-    This ratio provides insight into how individual asset volatilities and their correlations
-    contribute to the overall portfolio risk.
+    Raises
+    ------
+    ValueError
+        If ``samples`` is not 2D with a ``sample`` dim, if ``budgets`` is not
+        1D, or if the dim ``budgets`` is labelled with is not a dim of
+        ``samples``.
 
     Notes
     -----
@@ -594,29 +599,42 @@ def diversification_ratio(
     utility function do not have that shape out of the box: the default
     ``response_variable`` is a 1D ``(sample,)`` total, and per-channel
     variables carry a ``date`` dim as well. Reduce over ``date`` before calling
-    this function.
+    this function. The optimizer invokes the utility with keyword arguments,
+    so a wrapper's parameters must be named ``samples`` and ``budgets``.
 
-    Examples
-    --------
-    Optimize for diversification of the per-channel contribution:
-
-    .. code-block:: python
-
-        from pymc_marketing.mmm.utility import diversification_ratio
-
-        optimizer = BudgetOptimizer(
-            model=model,
-            num_periods=num_periods,
-            response_variable="channel_contribution",
-            utility_function=lambda samples, budgets: diversification_ratio(
-                samples.sum(dim="date"), budgets
-            ),
-        )
+    The ratio is undefined when the portfolio volatility is zero: a channel at
+    exactly zero budget contributes zero in every sample, and a channel driven
+    deep into saturation is numerically constant. Both give ``0 / 0`` for the
+    value and an infinite gradient, which the optimizer reports as an opaque
+    solver failure. The default bounds allow zero budgets, so pass
+    ``budget_bounds`` with strictly positive lower bounds when optimizing for
+    diversification.
 
     References
     ----------
     - Choueifaty, Y., & Coignard, Y. (2008). Toward Maximum Diversification. *Journal of Portfolio Management*.
     - Meucci, A. (2009). Managing Diversification. *Risk*, 22(5), 74-79.
+
+    Examples
+    --------
+    Optimize a fitted MMM for diversification of the per-channel contribution:
+
+    .. code-block:: python
+
+        from pymc_marketing.mmm.utility import diversification_ratio
+
+        optimizer = mmm.budget_optimizer(
+            start_date="2025-01-06",
+            end_date="2025-03-31",
+            response_variable="channel_contribution",
+            utility_function=lambda samples, budgets: diversification_ratio(
+                samples.sum(dim="date"), budgets
+            ),
+        )
+        result = optimizer.allocate_budget(
+            total_budget=100.0,
+            budget_bounds={channel: (1.0, 100.0) for channel in mmm.channel_columns},
+        )
     """
     samples = as_xtensor(samples)
     budgets = as_xtensor(budgets)
--- a/pymc_marketing/mmm/utility.py
+++ b/pymc_marketing/mmm/utility.py
@@ -586,6 +586,33 @@ def diversification_ratio(
     This ratio provides insight into how individual asset volatilities and their correlations
     contribute to the overall portfolio risk.
 
+    Notes
+    -----
+    ``samples`` must have exactly two dims: ``sample`` and the asset dim that
+    ``budgets`` is labelled with. The response distributions
+    :class:`~pymc_marketing.mmm.budget_optimizer.BudgetOptimizer` passes to a
+    utility function do not have that shape out of the box: the default
+    ``response_variable`` is a 1D ``(sample,)`` total, and per-channel
+    variables carry a ``date`` dim as well. Reduce over ``date`` before calling
+    this function.
+
+    Examples
+    --------
+    Optimize for diversification of the per-channel contribution:
+
+    .. code-block:: python
+
+        from pymc_marketing.mmm.utility import diversification_ratio
+
+        optimizer = BudgetOptimizer(
+            model=model,
+            num_periods=num_periods,
+            response_variable="channel_contribution",
+            utility_function=lambda samples, budgets: diversification_ratio(
+                samples.sum(dim="date"), budgets
+            ),
+        )
+
     References
     ----------
     - Choueifaty, Y., & Coignard, Y. (2008). Toward Maximum Diversification. *Journal of Portfolio Management*.
@@ -593,15 +620,24 @@ def diversification_ratio(
     """
     samples = as_xtensor(samples)
     budgets = as_xtensor(budgets)
-    if samples.type.ndim != 2:
+    if samples.type.ndim != 2 or "sample" not in samples.dims:
         raise ValueError(
             "Function expected samples to be a 2D tensor variable with a 'sample' "
-            f"dim and an asset dim. Got {samples.type.ndim} dimensions."
+            f"dim and an asset dim. Got dims {samples.dims}."
         )
+    if budgets.type.ndim != 1:
+        raise ValueError(
+            "Function expected budgets to be a 1D tensor variable over the asset "
+            f"dim. Got dims {budgets.dims}."
+        )
+    [asset_dim] = budgets.dims
+    if asset_dim not in samples.dims:
+        raise ValueError(
+            f"budgets dim {asset_dim!r} not found in samples dims {samples.dims}."
+        )
+
     weights = budgets / budgets.sum()
     individual_volatilities = samples.std(dim="sample", ddof=1)
-
-    [asset_dim] = weights.dims
     cov_matrix = _covariance_matrix(samples, asset_dim=asset_dim)
 
     # w'Σw: rename one copy of the weights so the two asset axes of the
@@ -612,5 +648,4 @@ def diversification_ratio(
     )
     portfolio_volatility = ptx.math.sqrt(portfolio_var)
     weighted_avg_volatility = (weights * individual_volatilities).sum()
-    diversification_ratio = weighted_avg_volatility / portfolio_volatility
-    return diversification_ratio
+    return weighted_avg_volatility / portfolio_volatility
--- a/tests/mmm/test_budget_optimizer.py
+++ b/tests/mmm/test_budget_optimizer.py
@@ -740,11 +740,17 @@ def test_constraint_history_keyed_by_constraint(mmm_wrapper):
 
 
 def test_diversification_ratio_through_optimizer(mmm_wrapper):
-    """The docstring recipe runs through BudgetOptimizer to a finite optimum.
+    """The docstring recipe runs through BudgetOptimizer and actually optimizes.
 
     The utility sees the response as ``(sample, date, channel)``; reducing
     over ``date`` gives the ``(sample, channel)`` shape the ratio needs.
+
+    ``total_budget=10`` is deliberate: at 100 the fixture's saturation is so
+    flat that the gradient at the equal split is already within SLSQP's
+    tolerance, and the solver stops at the initial guess, which every
+    assertion here would then satisfy for free.
     """
+    total_budget = 10.0
     optimizer = BudgetOptimizer(
         model=mmm_wrapper,
         num_periods=30,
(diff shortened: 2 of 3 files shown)


juanitorduz:
## Review

**Verdict: approve, with two suggestions worth folding in before merge.** The diagnosis is correct, the fix is minimal and mathematically right, and the tests use an independent oracle rather than re-implementing the graph. Everything below is verified locally on `6c4351d` in a separate worktree.

### Verification I ran

Both defects reproduce on `main` exactly as described:

```
defect 1 -> ValueError Function expected samples to be a 1D tensor variable. Got 2 dimensions.
defect 2 -> TypeError XTensorVariable.rename() takes from 1 to 2 positional arguments but 3 were given
```

(`defect 2` obtained by monkeypatching the 1D guard to identity, confirming the parenthesis bug was genuinely unreachable — `rename(name_dict, extra)` is a `TypeError`, not a silent wrong answer.)

On the branch:

- `tests/mmm/test_utility.py` → **51 passed** ✅ (matches the PR description)
- `pre-commit run --files pymc_marketing/mmm/utility.py tests/mmm/test_utility.py` → all hooks pass (ruff, ruff-format, mypy, API-docs completeness) ✅
- Dim algebra of the corrected contraction checks out: `_covariance_matrix` → `(channel, channel')`; `ptx.dot(cov, weights)` contracts `channel` → `(channel',)`; `ptx.dot(weights.rename(channel→channel'), ·)` contracts `channel'` → scalar. Correct `wᵀΣw`, and correct regardless of `Σ` symmetry.
- Extra oracles beyond the ones in the PR, both of which hold:
  - perfectly correlated assets (`[x, 2x]`) → `0.9999999999999996` (theoretical lower bound `DR = 1`)
  - single asset → `1.0`
- **Optimizer compatibility**: the objective is differentiable through `lower_xtensor`, which the previous code never got to prove. Built the graph the way `BudgetOptimizer._compile_objective_and_grad` does (`budgets_flat` → `as_xtensor` → utility → `rewrite_graph(objective.values, include=("lower_xtensor", "canonicalize", "stabilize"))` → `pt.grad`) and compared against central finite differences:
    ```
    grad: [ 0.18269355  0.18828861 -0.18642359]
    fd  : [ 0.18269355  0.18828861 -0.18642359]
    ```

### 1. The guard's message promises a check it does not perform (important, cheap)

```python
if samples.type.ndim != 2:
    raise ValueError(
        "Function expected samples to be a 2D tensor variable with a 'sample' "
        f"dim and an asset dim. Got {samples.type.ndim} dimensions."
    )
```

Rank is checked; the dim *names* are not. The two ways to get the names wrong both escape into internals:

```
2D, no 'sample' dim   -> ValueError: Reduced dims ('sample',) not found in array dimensions ('chain', 'channel')
samples/budgets dim mismatch -> ValueError: Cannot rename media to media': media not in ('sample', 'channel')
```

The first is tolerable; the second is genuinely confusing (it names a dim the user never asked to rename, and it leaks the internal `'`-suffix convention). Since the error text already commits to the contract, suggest asserting it:

```python
if samples.type.ndim != 2 or "sample" not in samples.dims:
    raise ValueError(
        "Function expected samples to be a 2D tensor variable with a 'sample' dim "
        f"and an asset dim. Got dims {samples.dims}."
    )
[asset_dim] = weights.dims
if asset_dim not in samples.dims:
    raise ValueError(
        f"budgets dim {asset_dim!r} not found in samples dims {samples.dims}."
    )
```

Printing `samples.dims` instead of only the rank makes the common mistake self-diagnosing. `[asset_dim] = weights.dims` similarly surfaces as bare `too many values to unpack (expected 1, got 2)` for 2D budgets — pre-existing, but one line away if you are in there anyway.

### 2. No reachable 2D path through `BudgetOptimizer` (important for the docstring)

`diversification_ratio` is typed as a `UtilityFunctionType`, so the natural next step for a reader of #2994 is to pass it to `optimize_budget`. That still cannot work without a wrapper:

- the default `response_variable="total_media_contribution_original_scale"` is 1D `(sample,)` — cf. `tests/mmm/test_budget_optimizer.py:547`, which feeds it through the *1D* guard — so it now hits the new `Got 1 dimensions` error;
- per-channel variables carry a date dim (`chain, draw, channel, date`, cf. the idata fixtures in `tests/mmm/test_budget_optimizer.py:118`), i.e. 3D after stacking `sample` → `Got 3 dimensions` (confirmed).

So the only usable form is a caller-supplied reduction, e.g. `lambda s, b: diversification_ratio(s.sum(dim="date"), b)`. The function is autosummary-documented (`docs/source/api/index.md` recurses into `pymc_marketing.mmm.utility`), so a two-line `Examples`/`Notes` block naming the required dims and showing that reduction would keep the next user out of the same `ValueError` this PR just replaced. An integration test that actually drives it through `BudgetOptimizer` would be even better, but the docstring note is the high-value/low-cost half.

### Minor

- **`rng` stream shift.** The new `parametrize` block at `tests/mmm/test_utility.py:312` is the first consumer of the module-level `rng`, and it draws at import time *before* the existing decorators at lines 363/392/415+ that pass `random_seed=rng`. Every later randomised case therefore gets different data than on `main`. Harmless here (those tests all compare against a numpy reference or merely assert no-raise), but it means any future insertion silently perturbs unrelated tests. A local `np.random.default_rng(0)` inside the new block would keep the change inert.
- **Two 2D checks, two shapes.** `_check_samples_dimensionality` still exists for the 1D utilities while this one is inline. `_covariance_matrix` also assumes 2D and checks nothing. An optional `expected_ndim: int = 1` parameter on the existing helper would keep one message format and stay backwards compatible with the notebook that imports it (`mmm_budget_allocation_example.ipynb:3310`).
- **`scipy.linalg.hadamard`** is a fine trick and the comment explains why it works — worth keeping since `sqrt(n)` is a genuinely independent oracle rather than a second copy of the formula. Note scipy is only a transitive dependency (via pymc) and is not declared in `pyproject.toml`; it appears only under `[tool.mypy-scipy]`. Not a blocker, but a `np.array([[1, 1], [1, -1], ...])`-style literal or `scipy.stats`-free construction would avoid leaning on it.
- **`test_diversification_ratio_rejects_1d_samples`** reuses the `test_data` fixture, whose `budgets` carry `dims=("sample",)` (the fixture even comments on this oddity). It passes because the rank guard fires first, so the test would pass for the wrong reason if the guard changed. A purpose-built `as_xtensor(np.arange(5.), dims=("sample",))` + `dims=("channel",)` budgets pair is more precise. Adding a 3D case to the same test is one parametrize row and covers the direction users will actually hit.
- **Shadowing.** `diversification_ratio = weighted_avg_volatility / portfolio_volatility` (line 609) rebinds the function name inside its own body. Pre-existing; `return weighted_avg_volatility / portfolio_volatility` is a free cleanup while the lines are already being touched.
- `DR ≥ 1` and `DR = 1` for a single asset both hold (verified above) and are one-liners — `DR = 1` for perfectly correlated assets is arguably the sharpest cheap oracle after `sqrt(n)`.

### Not blocking, out of scope

Zero-volatility samples give `0/0 → nan`, `budgets.sum() == 0` divides by zero, and negative budgets produce `DR < 1` (I got `0.648`), which is meaningless for a ratio whose floor is 1. All pre-existing and unreachable from the optimizer, which constrains budgets to be non-negative.

PabloRoque:
Thanks for the thorough review. Folded everything in at `67e824f2`:

**1. Guard.** Now checks rank, the `sample` dim, and that the budgets dim is present in `samples`, and reports `samples.dims` instead of only the rank. Non-1D budgets get a clear message too, instead of the bare unpack error. Five parametrised rejection cases cover 1D, 3D, missing `sample` dim, asset-dim mismatch, and 2D budgets.

**2. Docstring.** Added a Notes block naming the required dims and stating that the optimizer's response distributions do not have that shape out of the box, plus an Examples block with the `samples.sum(dim="date")` recipe. I also took the "even better" option: `test_diversification_ratio_through_optimizer` drives that recipe through `BudgetOptimizer` on the `channel_contribution` variable of the existing wrapper fixture and asserts convergence, a finite objective, `DR >= 1`, and the budget sum. The fixture's two channels are perfectly correlated, so it lands on the floor of 1, which doubles as one more oracle.

**Minor.** All taken: local `np.random.default_rng(0)` for the new block so the module stream is unchanged; Hadamard columns written as a literal so scipy is not imported; `DR = 1` oracles for `[x, 2x]` and a single asset; the ratio is returned directly rather than rebinding the function name. I left `_check_samples_dimensionality` as is, since the new guard checks dim names that a generic `expected_ndim` parameter would not cover.

Locally: `tests/mmm/test_utility.py` and `tests/mmm/test_budget_optimizer.py` pass (106 tests), pre-commit clean.

juanitorduz:
## Review round 2 — `67e824f2`

All six points from the previous round are addressed, and correctly. Re-verified locally:

- `tests/mmm/test_utility.py` → **56 passed**; `tests/mmm/test_budget_optimizer.py` → **50 passed**; pre-commit clean on all three files.
- `[asset_dim] = budgets.dims` is equivalent to the old `weights.dims` (scalar division preserves dims), and 0D budgets are now caught by the explicit `ndim != 1` check rather than an unpack error. All five rows of `test_diversification_ratio_rejects_bad_dims` hit distinct branches.
- The hand-written 4×3 Hadamard block is correct: columns pairwise orthogonal, each summing to zero with norm² 4, so `Σ = (4/3)·I` and `DR = sqrt(3)`. Dropping the scipy import was the right call.
- `_dr_rng` restores the module `rng` stream, `_diversification_ratio_numpy` keeps the oracle independent, `DR = 1` floor is now pinned.

I then actually drove the new integration test, and it surfaced two things worth fixing before merge.

### 1. `test_diversification_ratio_through_optimizer` converges at the initial guess and asserts nothing about the optimum

```
success True   nit 1   fun -1.0000133394040267   budgets [50. 50.]
```

`[50., 50.]` *is* `x0` (`self._variables.x0(total_budget)`, the equal split, `budget_optimizer.py:2228`). SLSQP terminates on the first iteration. Scanning the compiled `optimizer._objective_and_grad` along the constraint line:

| `channel_1` | DR | ‖grad‖ |
|---|---|---|
| 10 | 1.0000297 | ~1.3e-05 |
| **50 (returned)** | **1.0000133** | ~2.7e-06 |
| 90 | 1.0002141 | — |
| 99 | 1.0177485 | 3.0e-02 |

The returned point is the *flattest and lowest* part of the scan; DR at `[99, 1]` is three orders of magnitude further above the floor. The gradient at `x0` is ~1e-6, i.e. at SLSQP's default `ftol`, so it declares success immediately. Consequently every assertion in the test — `success`, `isfinite`, `-fun >= 1 - 1e-6`, `sum == 100` — is satisfied by the untouched initial allocation, and the docstring's "runs through `BudgetOptimizer` to a finite optimum" is not what happens. `DR >= 1` in particular is a global property of the function, so it cannot discriminate.

The test is still worth keeping: it proves the recipe compiles, the shapes line up, and the objective is differentiable through `lower_xtensor`. But it should either say that, or assert something the initial guess does not already satisfy. `total_budget=10` on the same fixture does move:

```
success True   fun -1.0102441257502   budgets [2.18430972 7.81569028]     (x0 = [5., 5.])
```

That is an interior optimum genuinely distinct from `x0`, so `assert not np.allclose(result.budgets, 5.0)` plus "utility at the solution exceeds utility at `x0`" would actually test the optimization. Why the fixture is so flat at `total_budget=100`: `saturation_beta` and `adstock_alpha` are identical across all four `(chain, draw)` samples (`test_budget_optimizer.py:96-113`), so only `saturation_lam` varies, and at 50 per channel over 30 periods logistic saturation is nearly flat — per-channel volatilities collapse to `[0.136, 0.021]` and the two channels are near-perfectly correlated, hence `DR ≈ 1`. At `[5., 5.]` the volatilities are `[4.61, 6.72]` with a genuinely non-diagonal `Σ`.

### 2. `nan` objective and gradient at zero-budget corners; the recipe crashes at a larger total budget

Same fixture, same recipe, only `total_budget` changed:

```
total_budget=1000 -> MinimizeException: Optimization failed: Inequality constraints incompatible
```

Evaluating `_objective_and_grad` directly explains it:

```
[500., 500.] -> DR 1.0        vol [1.31e-12, 0.0]   cov [1.7e-24, 0, 0, 0]
[999.,   1.] -> DR 1.0        grad [nan, 2.6e-16]
[1000.,  0.] -> DR nan        grad [nan, nan]
[0., 1000.] -> DR nan        grad [nan, nan]
```

A channel at exactly zero budget contributes exactly zero in every sample, so its volatility is zero; fully saturated channels are numerically constant too. Then `weighted_avg_volatility → 0` and `portfolio_var → 0`, giving `0/0` for the value and `d/dx sqrt(x)|_{x=0} = inf` for the gradient. The default bounds are `(0, total_budget)`, so those corners are *feasible*, SLSQP walks into them, and the resulting `nan` surfaces as the thoroughly misleading `Inequality constraints incompatible`.

This is inherent to the diversification ratio, not introduced here — but this PR is what makes DR reachable, and it now ships a docstring recommending it inside `BudgetOptimizer`, so the failure mode belongs in the `Notes` it just added: *DR is undefined when portfolio volatility is zero (a channel at zero budget, or channels driven into saturation); pass `budget_bounds` with strictly positive lower bounds.* Two sentences turn an opaque solver error into a self-diagnosing one. Raising from the utility when `portfolio_var` is not positive would be better still, but that needs a symbolic check and is arguably scope creep.

### Minor

- **The example snippet is not runnable as written.** It imports `diversification_ratio` but not `BudgetOptimizer`, and `model=model` is not the wrapper-plus-`idata` form the real API needs (cf. `mmm_wrapper`). Since the new test now encodes the working recipe, worth aligning the snippet with it.
- **Keyword-call trap.** `BudgetOptimizer` invokes the utility as `self.utility_function(samples=..., budgets=...)` (`budget_optimizer.py:2076`), so the lambda's parameters must literally be named `samples` and `budgets`. Your snippet and test are correct; the abbreviated form fails with `TypeError: <lambda>() got an unexpected keyword argument 'samples'` (I hit it writing my own probe). One clause in `Notes` — "the callable is invoked with keyword arguments" — saves the next reader that detour.
- **numpydoc section order and `Raises`.** Canonical order is `... Notes, References, Examples`; the PR has `Notes, Examples, References`. Ruff's numpy convention does not check ordering so nothing fails, but the function now raises three distinct `ValueError`s and a `Raises` section would document them. Also still floating: the orphan sentence "This ratio provides insight into how individual asset volatilities…" sitting between `Returns` and `Notes` (pre-existing, and now more visibly out of place).

### Summary

`utility.py` is correct and I have no reservations about the fix itself. Item 1 (make the integration test assert something `x0` does not already give you) and item 2 (document the zero-volatility failure mode next to the recipe that triggers it) are the two I would land before merge; the rest are cosmetic.

juanitorduz (approved):
Minor nits

PabloRoque:
Both reproduced exactly as you describe, and both are in a0dbc23 (branch also rebased onto `main` after #2993 merged, which touched the same test file).

**1. Integration test converging at `x0`.** Confirmed: `total_budget=100` gives `nit=1`, `[50, 50]`, gradient at the equal split of order 1e-6. The test now runs at `total_budget=10`, where the fixture has a genuine interior optimum (`[2.18, 7.82]`, `nit=9`), and asserts that the solution is not the equal split and that the objective there is strictly lower than at `x0`, on top of the existing success, finiteness, `DR >= 1` and budget-sum checks. The docstring says why 10 and not 100, so nobody bumps it back.

**2. Zero-volatility corners.** Confirmed: `[1000, 0]` and `[0, 1000]` give `DR = nan` with `nan` gradients, and `total_budget=1000` fails with `Inequality constraints incompatible`. The `Notes` now say the ratio is `0 / 0` with an infinite gradient when portfolio volatility is zero, name the two ways to get there (a channel at zero budget, a channel deep in saturation), say that it surfaces as an opaque solver failure, and tell the reader to pass `budget_bounds` with strictly positive lower bounds. The example does exactly that. I did not add a symbolic check in the utility; agreed it is scope creep here.

**Minor, all taken.** The example is rewritten around `mmm.budget_optimizer(start_date, end_date, response_variable=..., utility_function=...)` followed by `allocate_budget(total_budget, budget_bounds=...)`, so it matches the real entry point and is runnable against a fitted MMM. `Notes` state that the utility is invoked with keyword arguments, so wrapper parameters must be named `samples` and `budgets`. Sections are now `Notes, References, Examples`, a `Raises` section documents the three `ValueError`s, and the orphan sentence is folded into the description paragraph.

Locally: `tests/mmm/test_utility.py` and `tests/mmm/test_budget_optimizer.py` pass (107 tests), pre-commit clean.

PR #2996, opened 2026-09-09
feat: use terms for Bass priors

### LINKED ISSUE #2961: Port the Bass model to Terms
Port the Bass model to Terms, see https://github.com/pymc-labs/pymc-marketing/issues/2960

williambdean:
Closes #2961. Third PR in the terms stack (serialization → Bass components → Bass).

`m`, `p`, and `q` accept term recipes through `BassPriors` /`model_config` and serialize through the terms registry:

```python
from pymc_marketing.terms import Named, Parameter

priors = {
    "m": Parameter("m", prior=Prior("HalfNormal", sigma=500)),
    "p": Parameter("p", prior=Prior("Beta", alpha=1.5, beta=20, dims="product")),
    "q": Named("q", Parameter("q_scale", ...) * 2, dims="product"),
}
model = create_bass_model(t=t, observed=observed, priors=priors, coords=coords)
```

Defaults keep an identical graph: `Parameter(name, prior)` builds exactly like `prior.create_variable(name, xdist=True)`, and the data-driven `m` rescale still applies unless a recipe opts out. `Parameter` gains a `dims` passthrough and `Named` normalizes dims to tuples (mirroring `Prior`) so the existing observed-labeling bookkeeping reads terms transparently — pymc.dims handles all coordination.

Closes the non-CLV consumer story: Bass (dims-native already), then BG/NBD (#2976) and the #2962 slices follow.

----
📚 Documentation preview 📚: https://pymc-marketing--2996.org.readthedocs.build/en/2996/

### DIFF
--- a/pymc_marketing/bass/model.py
+++ b/pymc_marketing/bass/model.py
@@ -168,7 +168,7 @@ from pymc_marketing.bass import plotting
 from pymc_marketing.bass.data import to_bass_dataset
 from pymc_marketing.model_builder import ModelBuilder, SamplingMethod
 from pymc_marketing.model_config import parse_model_config
-from pymc_marketing.terms import ModelTerm, Parameter, build_param
+from pymc_marketing.terms import ModelTerm, Product, Sum, build_param
 from pymc_marketing.version import __version__
 
 #: What :func:`F` and :func:`f` accept for ``t``: a labelled xtensor, or any
@@ -388,28 +388,10 @@ def _observed_dims(
 class BassPriors(TypedDict):
     """Priors for the Bass diffusion model."""
 
-    m: Prior | Censored | VariableFactory | ModelTerm
-    p: Prior | Censored | VariableFactory | ModelTerm
-    q: Prior | Censored | VariableFactory | ModelTerm
+    m: Prior | Censored | VariableFactory | ModelTerm | Sum | Product
+    p: Prior | Censored | VariableFactory | ModelTerm | Sum | Product
+    q: Prior | Censored | VariableFactory | ModelTerm | Sum | Product
     likelihood: Prior | Censored
-
-
-def _build_prior(prior: Any, name: str) -> pmd.XTensorVariable:
-    """Build a prior or term recipe into its tensor.
-
-    ``Parameter(name, prior)`` builds exactly like the previous
-    ``prior.create_variable(name, xdist=True)``, so models configured with
-    plain priors keep an identical graph. Priors must be dims-native:
-    Bass composes them into ``pmd.Deterministic`` outputs.
-    """
-    if isinstance(prior, ModelTerm):
-        return cast("pmd.XTensorVariable", build_param(prior))
-    return cast("pmd.XTensorVariable", build_param(Parameter(name, prior=prior)))
-
-
-def _term_dims(value: Any) -> Any:
-    """Dims declared by a prior or term (``None`` when not annotated)."""
-    return getattr(value, "dims", None)
 
 
 def create_bass_model(
@@ -456,6 +438,11 @@ def create_bass_model(
         - 'p': Innovation coefficient prior or term
         - 'q': Imitation coefficient prior or term
         - 'likelihood': Observation likelihood model
+
+        Terms must build free random variables (``Parameter`` /
+        ``Named``) - data-carrying terms (``Dot``) need
+        ``register_data``/``set_data`` wiring this model does not
+        perform yet.
     coords : dict[str, Any]
         Coordinate values for dimensions in the model, including
         'date' for the time dimension and any other dimensions
@@ -487,26 +474,27 @@ def create_bass_model(
     """
     model = model or pm.Model(coords=coords)
     with model:
+        time = pmd.as_xtensor(t, dims=("T",))
+        m = cast("pmd.XTensorVariable", build_param(priors["m"], "m"))
+        p = cast("pmd.XTensorVariable", build_param(priors["p"], "p"))
+        q = cast("pmd.XTensorVariable", build_param(priors["q"], "q"))
+
         # Declaration order, not set order: `combined_dims` labels the axes of
         # `observed` positionally, so an order that varies between processes
         # would silently mislabel the data. The likelihood comes first because
-        # it is the variable `observed` has to line up with, so an unlabelled
-        # array laid out the way the likelihood declares it is read that way.
+        # it is the variable `observed` has to line up with. `p`, `q` and `m`
+        # report the dims they actually built with, so a term recipe does not
+        # have to re-declare them.
         declared_dims = (
             *(priors["likelihood"].dims or ()),
-            *(_term_dims(priors["p"]) or ()),
-            *(_term_dims(priors["q"]) or ()),
-            *(_term_dims(priors["m"]) or ()),
+            *getattr(p, "dims", ()),
+            *getattr(q, "dims", ()),
+            *getattr(m, "dims", ()),
         )
         combined_dims = (
             "T",
             *(dim for dim in dict.fromkeys(declared_dims) if dim != "T"),
         )
-
-        time = pmd.as_xtensor(t, dims=("T",))
-        m = _build_prior(priors["m"], "m")
-        p = _build_prior(priors["p"], "p")
-        q = _build_prior(priors["q"], "q")
 
         def deterministic(name: str, value: XTensorVariable) -> XTensorVariable:
             """Store ``value`` with the dims it has, in ``combined_dims`` order."""
@@ -549,7 +537,9 @@ class BassModel(ModelBuilder):
     ----------
     model_config : dict, optional
         Dictionary with keys ``"m"``, ``"p"``, ``"q"``, ``"likelihood"``
-        mapping to :class:`~pymc_extras.prior.Prior` (or equivalent dict).
+        mapping to :class:`~pymc_extras.prior.Prior` or a punch-in term
+        recipe (``Parameter`` / ``Named`` composition). The names must
+        match the config keys to keep ``m``/``p``/``q`` in the model.
         See :meth:`default_model_config` for defaults.
     sampler_config : dict, optional
         Dictionary of sampler settings (draws, tune, chains, …).
@@ -690,6 +680,9 @@ class BassModel(ModelBuilder):
         ``model_config``, so every fit is scaled to the data it is looking at. Pass an
         ``m`` prior that differs from the default in ``model_config`` to opt out; a
         prior identical to the default is indistinguishable from not passing one.
+        Wrapping the default prior in a ``Parameter`` term counts as passing one,
+        so the rescale does not apply - the fitted ``m`` scale is the prior's own
+        sigma, not ``2 * observed.sum()``.
         """
         return {
             "m": Prior("HalfNormal", sigma=10),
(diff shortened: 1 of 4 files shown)


juanitorduz:
## Review

Nice, tight PR — the default path is genuinely untouched and the save/load test is the right one to have written. I ran the branch locally: **86/86 `tests/bass/test_model.py` and 84 tests across `TestBassModelTerms` + `tests/test_terms.py` pass as submitted**, and `mypy` on the two changed modules reports no new errors.

Three substantive things below, all in `pymc_marketing/bass/model.py`. Taking #1 and #2 together makes the `terms.py` half of the diff (and the `tests/test_terms.py` churn) unnecessary, which I think is the better shape for this PR.

---

### 1. `_build_prior` re-implements `build_param`, and breaks on `Sum` / `Product`

`build_param(param, name)` already does exactly both branches of `_build_prior` (`ModelTerm` → `create_variable()`, `VariableFactory` → `create_variable(name, xdist=True)`) **and** handles `Sum` / `Product`. `_build_prior` instead wraps everything non-`ModelTerm` in `Parameter(name, prior=...)`, so a bare composition — the most natural thing to write given the operator API this stack is building — dies:

```python
"q": Parameter("q_scale", prior=Prior("Beta", alpha=2, beta=5)) * 2
# AttributeError: 'Product' object has no attribute 'create_variable'
```

(`Sum` and `Product` are not `ModelTerm` subclasses, so `isinstance` misses them and they end up as `Parameter(name="q", prior=Product(...))`.) The PR description's example dodges this only because it wraps in `Named`.

Suggested: drop `_build_prior`, call `build_param(priors["m"], "m")` directly, and widen `BassPriors` to `... | ModelTerm | Sum | Product`.

### 2. `_term_dims` is a shallow `getattr`, so `Named` without explicit `dims` breaks the model

`Named.dims` is optional and `Named.create_variable` is perfectly happy without it (`pmd.Deterministic` infers). But `combined_dims` reads only the annotation, so the dim never reaches the likelihood:

```python
priors = {
    "m": Prior("HalfNormal", sigma=500),
    "p": Prior("Beta", alpha=1.5, beta=20),
    "q": Named("q", Parameter("q_scale", prior=Prior("Beta", alpha=2, beta=5, dims="product")) * 2),
    "likelihood": Prior("Poisson"),
}
create_bass_model(t=t, observed=None, priors=priors, coords=coords)
# ValueError: ('T',) must be a permuted list of ('product', 'T')
```

Same failure with an unlabelled 2-D `observed`. The message names neither `q` nor the missing `dims=`.

There's a fix that needs no annotation-chasing at all: `m`, `p`, `q` are built *before* `combined_dims` is ever consumed, so read the dims off the built variables. That is exact for any term — recipes never have to re-declare what they already built:

```python
        time = pmd.as_xtensor(t, dims=("T",))
        m = build_param(priors["m"], "m")
        p = build_param(priors["p"], "p")
        q = build_param(priors["q"], "q")

        # Declaration order, not set order: `combined_dims` labels the axes of
        # `observed` positionally, so an order that varies between processes
        # would silently mislabel the data. The likelihood comes first because
        # it is the variable `observed` has to line up with. `p`, `q` and `m`
        # report the dims they actually built with, so a term recipe does not
        # have to re-declare them.
        declared_dims = (
            *(priors["likelihood"].dims or ()),
            *getattr(p, "dims", ()),
            *getattr(q, "dims", ()),
            *getattr(m, "dims", ()),
        )
```

Documented declaration order and variable-creation order are both preserved. I applied exactly this locally: it fixes both #1 and #2, and **all 86 `tests/bass/test_model.py` still pass**. It also makes `Parameter.dims`, `Named.__post_init__`, and the `tests/test_terms.py` edit unnecessary for this PR — `bass/model.py` reduces to widening the TypedDict, three `build_param` calls, moving `declared_dims`, and the docstrings.

### 3. `BassPriors` accepts `ModelTerm`, but only free-RV terms actually work

`create_bass_model` never calls `terms.register_data` / `collect_coords`, and `BassModel._data_setter` never calls `terms.set_data`. So any data-carrying term is accepted by the annotation and fails at build time:

```python
"p": Named("p", Dot(var_name="x", prior=Prior("Normal", dims="feature")))
# KeyError: 'x'      # Dot.create_variable does model[self.var_name]
```

Two honest options: wire the lifecycle (`BassModel.build_model` already has `ds`; `_data_setter` already has the new `ds`), or say in the `BassPriors` / `create_bass_model` docstrings that only terms building free random variables are supported. Accepting the type and then raising `KeyError` is the one outcome to avoid.

---

### Minor

**4. The `isinstance(priors["m"], ModelTerm)` rescale guard is dead code.** `Prior.__eq__` returns `False` for any non-`Prior`, so `Parameter(...) == Prior("HalfNormal", sigma=10)` is already `False`. I deleted the clause and the full 86-test bass suite still passes — nothing exercises it. It also misses `Sum`/`Product`, so it would be incomplete if it *were* load-bearing. Drop it, or keep it with a comment saying it documents intent rather than changes behaviour.

**5. `test_m_recipe_opts_out_of_rescale` doesn't test what it says.** `"lognormal" in str(op)` passes whether or not the rescale is skipped, because a `LogNormal` differs from the default regardless. The interesting case is the recipe that wraps the *identical* default prior — the only configuration where the rescale could bite. Using the file's own helper:

```python
model = BassModel(model_config={"m": Parameter("m", prior=Prior("HalfNormal", sigma=10))})
model.build_model(data=y)
assert self._graph_m_sigma(model) == pytest.approx(10.0)  # not 2 * y.sum()
```

Related: that behaviour deserves a line in `default_model_config`'s docstring. Wrapping the default prior in a `Parameter` looks like a no-op refactor but silently changes the fitted `m` scale from `2 * observed.sum()` to `10`.

**6. `Parameter.dims` is a read-only property while `Named.dims` is a mutable field.** `par.dims = ("product",)` raises `AttributeError: property 'dims' of 'Parameter' object has no setter`. `_borrow_dims` mutates `.dims` — only on `likelihood` today, so no live bug, but it's a trap for the BG/NBD and #2962 slices. The two also disagree on normalization: `Named.dims` is coerced to a tuple, `Parameter.dims` passes the prior's value through untouched. If the property stays, at minimum have `get_coords` use `self.dims` instead of repeating `getattr(self.prior, "dims", None)`.

**7. `Named` dims normalization is only asserted through one changed serialization-roundtrip line.** Worth a direct test in `tests/test_terms.py`: `Named("x", 1.0, dims="a").dims == ("a",)` and the list case. (`dims=5` raises `TypeError: 'int' object is not iterable` — fine, just noting it's unguarded.)

**8. Nothing ties a term's `name` to its config key.** `{"m": Parameter("market", ...)}` builds a model with no `m` in it, silently breaking every `var_names=["m", "p", "q"]` example in the `BassModel` docstring. A cheap check or a docstring note would help.

**9. Docs.** `BassModel`'s `model_config` parameter still reads *"mapping to `Prior` (or equivalent dict)"*. That's where users actually look — the term option and one recipe example belong in the class docstring, not only in `create_bass_model`.

williambdean:
The restructure is applied in `4d69128fe`:

- `_build_prior` and `\_term_dims` are gone — `build_param(priors["m"], "m")` handles priors, `Parameter`/`Named`, and bare `Sum`/`Product` compositions; `BassPriors` widened accordingly
- `declared_dims` now reads the dims off the *built* `m`/`p`/`q` variables, so recipes never re-declare; both your #1 and #2 fixed with your exact shape
- The rescale guard relies on `Prior` inequality alone (clause deleted, intent comment added)
- `test_m_recipe_opts_out_of_rescale` now wraps the *identical* default prior and asserts `_graph_m_sigma == 10.0`; `default_model_config` documents the identical-prior caveat
- Docstrings: term option in `BassModel.model_config` with the name-must-match note, and the free-RV-only contract for terms (data-carrying terms documented until lifecycle wiring)

`tests/bass` 86 passed + `tests/test_terms.py` 90+ green after the rebase onto the updated #2991. Also filed operator-completeness as #3001 (pre-existing `Product.__add__` gap you flagged).

juanitorduz:
## Re-review of `4d69128f`

The restructure landed as discussed, and the default path is verifiably untouched rather than just green: I built the default `BassModel` on the base tip and on this branch and diffed `named_vars`, `named_vars_to_dims`, `point_logps()`, and `model.id` — identical. Round-1 #1/#2 are fixed (a bare `Product` prior builds; `Named` without `dims` labels `adopters` as `('T', 'product')`), #4/#5 are fixed (sigma `10.0` for a `Parameter`-wrapped default vs `4036.0 = 2 * y.sum()` for the plain default), `Censored` still routes through `build_param`, and a `Parameter` config survives save/load **in a fresh interpreter**. 174 tests pass across `tests/test_terms.py` + `tests/bass/test_model.py`; mypy reports no new errors on the two modules.

One bookkeeping note first: the effective diff is much smaller than GitHub shows. `442c91a1` duplicates the base tip's `15942f5d` (same subject and author date — cherry-pick leftover), so `git diff feat/terms-bass-components feat/bass-terms` is `bass/model.py` +50/-18, `tests/bass/test_model.py` +71, and `terms.py` **+5** — only the `Parameter.dims` property. A rebase would make the PR diff match its scope.

### 1. `Parameter.dims` is the entire `terms.py` delta, and nothing reads it

The restructure removed its reason to exist: `declared_dims` now reads dims off the built variables, and grep finds no consumer anywhere in `pymc_marketing/` or `tests/`. It is also not a dataclass field (invisible to `to_dict`/`__eq__`), read-only where `Named.dims` is a mutable field, and untested. Deleting it makes this a pure Bass PR — the shape you described in the summary. If it stays as public API, then `get_coords` should read `self.dims` instead of repeating `getattr(self.prior, "dims", None)` (round-1 #6), and it needs a direct test.

### 2. A bare `Sum`/`Product` recipe builds a model with no variable for that key

Widening the annotation makes this the natural thing to write, and it silently drops `q`:

```python
priors = {..., "q": Parameter("q_scale", prior=Prior("Beta", alpha=2, beta=5)) * 2}
sorted(create_bass_model(t=t, observed=None, priors=priors, coords=coords).named_vars)
# ['adopters', 'imitators', 'innovators', 'm', 'p', 'peak', 'q_scale', 'y']    # no 'q'
```

`az.summary(idata, var_names=["m", "p", "q"])` — `model.py` lines 590, 612, 945, 948 — then raises for a config the type accepts. Same silent hole for `{"m": Parameter("market", ...)}` (round-1 #8), which is currently only a docstring "names must match" note. Naming the key at the build site turns the note into an invariant:

```python
def _build(key: str) -> pmd.XTensorVariable:
    spec = priors[key]
    if isinstance(spec, (Sum, Product)):
        spec = Named(key, spec)
    return cast("pmd.XTensorVariable", build_param(spec, key))
```

`Named(key, composition)` with `dims=None` builds and labels correctly — that is exactly the round-1 #2 path, and I confirmed it. Alternatively validate `key in model.named_vars` after each build and raise with the key name. Either is better than a posterior quietly missing `q`.

### 3. `getattr(p, "dims", ())` contradicts the `cast` on the line above

Two runtime paths reach that default: a `ModelTerm` that builds a non-dims-native variable, and a scalar constant. The first fails cryptically a few lines later:

```python
# a term whose create_variable() returns pm.HalfNormal("q", sigma=1.0, dims="product")
TypeError: non-scalar TensorVariable cannot be converted to XTensorVariable without dims
```

— no mention of `q`. The base's `Ref` already rejects non-dims-native variables with the offending name; mirroring that here (`p.dims` directly, plus one `isinstance(..., XTensorVariable)` check per key) makes the `cast` honest *and* puts the key in the message.

### 4. The two bugs fixed in round 1 aren't pinned by tests

`TestBassModelTerms` covers `Parameter`, `Named`-with-dims, the rescale opt-out, and save/load — but neither of the paths `build_param` was adopted to fix: a bare `Product` as a prior, and `Named` without `dims`. Both pass today, so a future regression to a `_build_prior`-style wrapper would still be green. These are the cheapest two tests in the diff.

### Minor

- **"punch-in term recipe"** (`model.py:540`) is the only occurrence of that phrase in the repo — undefined jargon in the docstring users read first. Also, `BassModel.model_config` does not carry the free-RV-only constraint that `create_bass_model` now documents, and neither spot has a runnable recipe example in the class docstring where the other config examples live.
- **Base-branch note for #3001:** `Named`'s `<name>_param` naming only reaches a *direct* bare factory. Two `Named` terms with a nested one collide — `ValueError: Variable name param already exists` — because `build_param` drops `name` in the `Sum`/`Product` recursion. Not this diff, but the same gap family.

Issue #2017, opened 2025-10-19
Consider using pymc.dims


ricardoV94:
PyMC dims submodule should facilitate the handling of dimensions in the models created by PyMC-Marketing without requiring any extra work / knowledge from the users. 

https://www.pymc.io/projects/docs/en/stable/learn/core_notebooks/dims_module.html

This would be a great use-case / driver of the functionality. 

Many of these issue (just searched by keyword), seem relevant: https://github.com/pymc-labs/pymc-marketing/issues?q=is%3Aissue%20state%3Aopen%20dims

The biggest immediate win I see is that most times one doesn't need to ask what are the dims of Deterministics generated by the model, nor worry about aligning dimensions for broadcasting. 

OTOH this wouldn't be a trivial refactor. I wonder if there's a way to gradually adopt the functionality without a giant PR that touches every single line of the codebase. I suspect CLV models would be a good start, as they are small / self-contained. The dims stuff would make it much easier to write the posterior-predictive / summary functions as well, since the batch dimensions of chain/draw wouldn't get in the way of the core logic.

A very valid reply to this issue is: Than you but no thank you :)

williambdean:
Sounds good. This will impact pymc-extras as well since that is where the Prior class and friends are defined.

williambdean:
One remaining blocker for full adoption: the BART-based PIE model (blocked upstream in pymc-devs/pymc-bart#289). Everything else in the core models is on `pymc.dims` already.

ricardoV94:
You should be able to convert dims vars to regular tensor `.values`, and vice-versa `as_xtensor(tensor, dims=...)`, so even if pymc-bart is lagging you can do everything else with dims?

PR #2997, opened 2026-09-09
feat(pie): express the PIE mean function as a terms recipe

williambdean:
Dogsfood the terms framework (#2965) with the model that motivated the swappable-mean-function API in #2017: PIE is now literally a regression with helpers and configuration.

The likelihood mean is a `ModelTerm` recipe. Default stays a BART term, so nothing changes for current users. Swap it out for a plain dot product with:

```python
from pymc_marketing.pie import PIEModel
from pymc_marketing.terms import Dot

model = PIEModel(
    pre_determined_features=[...],
    post_determined_features=[...],
    model_config={
        "sigma": Prior("HalfNormal", sigma=1.0),
        "mu": Dot(var_name="X", name="mu_coef", prior=Prior("Normal", dims="feature")),
    },
)
```

Where else this will impact in the repo:
- New module `pymc_marketing/bart.py`: `Bart` ModelTerm wrapping pymc_bart via the dims bridge (`.values` / `as_xtensor`) from https://github.com/pymc-labs/pymc-marketing/issues/2017#issuecomment-5602576216. Optional pymc-bart dependency, serialization-registered.
- Minimal `to_dict`/`from_dict` for the built-in terms vocabulary (`Parameter`, `Intercept`, `Dot`, `Sum`, `Product`, `Transform`), so recipe artifacts round-trip through `model_config`. Unregistered custom terms fail loudly with `NotImplementedError` at save instead of writing unloadable artifacts.
- `sample_vars` is now a documented `ModelTerm` lifecycle member and `collect_terms` descends into `Transform`.
- `make_split_rules` is a module-level helper; the old private method never shipped so no deprecation cycle was needed.

Release note: `PIEModel` bumped to 0.2.0. Checkpoints saved under 0.1.0 raise `DifferentModelError` on load (`check=False` loads them); rebuild and re-save is recommended. `pymc-bart` now capped at `<0.13`.

Details worth knowing:
- PGBART's logp machinery chokes on xtensor free variables, so `sigma` is a plain (non-xdist) variable lifted dimensionally for the likelihood; verified fit + resized-obs posterior predictive end-to-end for both BART and Dot flavors.
- BART observation dims are derived from the target (with an explicit transpose), so transposed data variables cannot silently swap the feature axis into the trees.
- Saves with an unregistered live recipe fail at `save()` time with a clear message, matching the review's write-side guard.

### DIFF
--- a/docs/source/api/index.md
+++ b/docs/source/api/index.md
@@ -8,6 +8,7 @@
   :toctree: generated/
 
   bass
+  bart
   causal_utils
   clv
   customer_choice
--- a/docs/source/api/index.md
+++ b/docs/source/api/index.md
@@ -7,8 +7,8 @@
   :recursive:
   :toctree: generated/
 
-  bass
   bart
+  bass
   causal_utils
   clv
   customer_choice
--- a/pymc_marketing/__init__.py
+++ b/pymc_marketing/__init__.py
@@ -23,6 +23,8 @@ with _warnings.catch_warnings():
 del _warnings
 
 # Load the data accessor
+# Register the BART term serializer
+import pymc_marketing.bart  # noqa: E402
 import pymc_marketing.data.fivetran  # noqa: E402
 
 # Register R2D2 deserializers
@@ -35,6 +37,7 @@ from pymc_marketing.version import __version__  # noqa: E402
 
 __all__ = [
     "__version__",
+    "bart",
     "bass",
     "clv",
     "customer_choice",
--- a/pymc_marketing/bart.py
+++ b/pymc_marketing/bart.py
@@ -1,0 +1,316 @@
+#   Copyright 2022 - 2026 The PyMC Labs Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""BART term for the ``pymc_marketing.terms`` framework.
+
+``pymc_marketing.bart`` exposes :class:`Bart`, a :class:`~pymc_marketing.terms.ModelTerm`
+that wraps :func:`pymc_bart.pymc_bart.BART` so Bayesian Additive Regression Trees
+can be composed like any other term in the ``pymc_marketing.terms`` vocabulary,
+e.g. interchangeably with :class:`~pymc_marketing.terms.Dot`.
+
+``pymc_bart`` is an optional dependency (``pip install 'pymc-marketing[pie]'``).
+Importing this module never fails when ``pymc_bart`` is missing; constructing
+the tensor raises a ``ImportError`` with an install hint instead. This keeps
+models that swap the BART term for a linear :class:`~pymc_marketing.terms.Dot`
+term runnable without ``pymc_bart`` installed.
+
+Because ``pymc_bart`` does not yet operate on ``pymc.dims`` variables, the term
+bridges the two worlds following the standard recipe: the dimensional shared
+data variable is lowered to a plain tensor with ``.values`` before being handed
+to ``pymc_bart``, and the BART output is lifted back into the dimensional graph
+with ``pytensor.xtensor.type.as_xtensor``. See
+`pymc-labs/pymc-marketing#2017 <https://github.com/pymc-labs/pymc-marketing/issues/2017>`_.
+
+Examples
+--------
+Swap a BART mean function for a linear dot product:
+
+.. code-block:: python
+
+    import pymc.dims as pmd
+    import xarray as xr
+    from pymc_extras.prior import Prior
+
+    from pymc_marketing.bart import Bart
+    from pymc_marketing.terms import (
+        build_param,
+        collect_coords,
+        register_data,
+    )
+
+    ds = xr.Dataset(
+        {
+            "X": (("obs", "feature"), X),
+            "y_obs": (("obs",), y),
+        },
+    )
+    mu = Bart(var_name="X", m=200)
+    # ... or the plain linear alternative:
+    # mu = Dot(var_name="X", prior=Prior("Normal", dims="feature"))
+
+    coords = collect_coords(mu, ds=ds)
+    with pm.Model(coords=coords) as model:
+        register_data(mu, ds=ds)
+        model = pm.modelcontext(None)
+        sigma = Prior("HalfNormal", sigma=1.0).create_variable("sigma", xdist=True)
+        pmd.Normal(
+            "y",
+            mu=build_param(mu),
+            sigma=sigma,
+            observed=model["y_obs"],
+            dims="obs",
+        )
+
+Terms compose, so BART can also be an ingredient of a larger expression:
+
+.. code-block:: python
+
+    from pymc_marketing.terms import Intercept
+
+    mu = Intercept(name="intercept") + Bart(var_name="X", m=200)
+
+The term serializes through ``pymc_marketing.serialization`` (``to_dict`` /
+``from_dict``), so it can be stored in ``model_config`` and survive
+``save()`` / ``load()``.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, cast
+
+import pymc as pm
+import pymc.dims as pmd
+import pytensor.tensor as pt
+import xarray as xr
+from pytensor.xtensor.type import as_xtensor
+
+from pymc_marketing.serialization import serialization
+from pymc_marketing.terms import ModelTerm
+
+try:
+    import pymc_bart as pmb
+except ImportError:  # pragma: no cover
+    pmb = None  # type: ignore[assignment]
+
+__all__ = ["Bart"]
+
+_DEFAULT_M: int = 200
+_DEFAULT_ALPHA: float = 0.95
+_DEFAULT_BETA: float = 2.0
+_DEFAULT_RESPONSE: str = "constant"
+
+
+def _serialize_split_rules(rules: list[Any] | None) -> list[str] | None:
+    """Serialize split rule instances to class names."""
+    if rules is None:
+        return None
+    return [rule.__class__.__name__ for rule in rules]
+
+
+def _load_split_rules(names: list[str] | None) -> list[Any] | None:
+    """Rebuild split rule instances from serialized class names."""
+    if names is None:
+        return None
+    if pmb is None:
+        raise ImportError(
+            "pymc-bart is required to rebuild split rules for a serialized "
+            "Bart term. Install it with: pip install 'pymc-marketing[pie]'"
+        )
+    from pymc_bart import split_rules
+
+    rules = []
+    for name in names:
+        if not hasattr(split_rules, name):
+            raise ValueError(
+                f"Unknown serialized split rule {name!r}. Split rules are "
+                "resolved by class name against pymc_bart.split_rules."
+            )
+        rules.append(getattr(split_rules, name)())
+    return rules
+
+
+@serialization.register
+@dataclass(kw_only=True)
+class Bart(ModelTerm):
+    """Bayesian Additive Regression Trees term.
+
+    Wraps :func:`pymc_bart.pymc_bart.BART` as a composable
+    :class:`~pymc_marketing.terms.ModelTerm`. Interchangeable with
+    :class:`~pymc_marketing.terms.Dot` --- swapping the two changes the model
+    graph but nothing else in the expression or lifecycle.
+
+    The data variable ``var_name`` is registered as ``pymc.dims`` shared data
+    and the target ``y_name`` must also be registered (BART uses ``Y`` to seed
+    the leaf-value prior with residual quantiles). Whoever registers the
+    observed data wins: registration is guarded by ``if name not in model``.
+
+    Parameters
+    ----------
+    var_name : str
+        Name of the feature variable in the dataset, with shape
+        ``(..., feature)``. Registered as ``pmd.Data`` for out-of-sample
+        prediction.
+    y_name : str, optional
+        Name of the target variable in the model. Used only to seed the
+        leaf-value prior at build time. Defaults to ``"y_obs"``.
+    m : int, optional
+        Number of trees. Defaults to 200.
+    alpha : float, optional
+        Tree prior parameter controlling depth. Defaults to 0.95.
+    beta : float, optional
+        Tree prior parameter controlling depth. Defaults to 2.0.
+    response : str, optional
+        Leaf response form, one of ``"constant"``, ``"linear"``, or ``"mix"``.
+        Defaults to ``"constant"``.
+    split_rules : list, optional
+        Split rules, one per feature column. Defaults to ``None``.
+    name : str, optional
+        Name for the PyMC variable. Defaults to ``"bart"``.
+
+    Examples
+    --------
+    .. code-block:: python
+
+        from pymc_extras.prior import Prior
+
+        from pymc_marketing.bart import Bart
+        from pymc_marketing.terms import Dot
+
+        bart_mean = Bart(var_name="X", m=200, name="bart")
+        linear_mean = Dot(
+            var_name="X", name="mu_coef", prior=Prior("Normal", dims="feature")
+        )
+    """
+
+    var_name: str
+    y_name: str = "y_obs"
+    m: int = _DEFAULT_M
+    alpha: float = _DEFAULT_ALPHA
+    beta: float = _DEFAULT_BETA
+    response: str = _DEFAULT_RESPONSE
+    split_rules: list[Any] | None = None
+    name: str = "bart"
+
+    def __post_init__(self) -> None:
+        """Validate configuration."""
+        if self.response not in ("constant", "linear", "mix"):
+            raise ValueError(
+                "Bart response must be 'constant', 'linear', or 'mix', "
+                f"got {self.response!r}."
+            )
+        if self.m < 1:
+            raise ValueError(f"Bart m must be >= 1, got {self.m}.")
+
+    def get_coords(self, ds: xr.Dataset) -> dict[str, Any]:
+        """Extract coordinates from the feature data variable."""
+        return {
+            cast("str", dim): da.coords[dim].values.tolist()
+            for dim, da in ds[self.var_name].coords.items()
+        }
+
+    def register_data(self, ds: xr.Dataset) -> None:
+        """Register the feature and target data variables as ``pmd.Data``."""
+        model = pm.modelcontext(None)
+        if self.var_name not in model:
+            pmd.Data(self.var_name, ds[self.var_name])
+        if self.y_name in ds and self.y_name not in model:
+            pmd.Data(self.y_name, ds[self.y_name])
+
+    def create_variable(self) -> pt.TensorVariable:
+        """Build the BART ensemble tensor.
+
+        Returns
+        -------
+        pt.TensorVariable
+            Dimensional tensor over the observation dimensions.
+        """
+        if pmb is None:
+            raise ImportError(
+                "pymc-bart is required to build a Bart term. "
+                "Install it with: pip install 'pymc-marketing[pie]'"
+            )
+        model = pm.modelcontext(None)
+        X = model[self.var_name]
+        if self.y_name not in model:
+            raise ValueError(
+                f"Bart term requires target data '{self.y_name}' in the model. "
+                "Register it as pmd.Data before building the term."
+            )
+        obs_dims = cast("tuple[str, ...]", X.dims)[:-1]
+        mu_plain = pmb.BART(
+            self.name,
+            X=X.values,
+            Y=model[self.y_name].values,
+            m=self.m,
+            alpha=self.alpha,
+            beta=self.beta,
+            response=self.response,
+            split_rules=self.split_rules,
+            dims=obs_dims,
+        )
+        return as_xtensor(mu_plain, dims=obs_dims)
+
+    def set_data(self, ds: xr.Dataset, model: pm.Model | None = None) -> None:
+        """Update shared data variables for out-of-sample prediction."""
+        if self.var_name in ds:
+            da = ds[self.var_name]
+            coords = {dim: ds[dim].values for dim in da.dims if dim in ds.coords}
+            pm.set_data({self.var_name: da.values}, model=model, coords=coords)
+        if self.y_name in ds:
+            da = ds[self.y_name]
+            coords = {dim: ds[dim].values for dim in da.dims if dim in ds.coords}
+            pm.set_data({self.y_name: da.values}, model=model, coords=coords)
+
+    @property
+    def sample_vars(self) -> list[str]:
+        """Unobserved variables that must be re-sampled in predictive mode.
+
+        The BART ensemble is conditioned on registered data, so when the data
+        are swapped for out-of-sample prediction the likelihood contribution
+        has to be recomputed rather than frozen from the trace.
+        """
+        return [self.name]
+
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize the term configuration.
+
+        Returns
+        -------
+        dict[str, Any]
+            JSON-safe configuration. ``split_rules`` serialize as class names.
+        """
+        return {
+            "var_name": self.var_name,
+            "y_name": self.y_name,
+            "m": self.m,
+            "alpha": self.alpha,
+            "beta": self.beta,
+            "response": self.response,
+            "split_rules": _serialize_split_rules(self.split_rules),
+            "name": self.name,
+        }
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> Bart:
+        """Reconstruct a ``Bart`` term from its serialized configuration."""
+        return cls(
+            var_name=data["var_name"],
+            y_name=data.get("y_name", "y_obs"),
+            m=data.get("m", _DEFAULT_M),
+            alpha=data.get("alpha", _DEFAULT_ALPHA),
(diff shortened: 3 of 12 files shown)


juanitorduz:
## Review

Reviewed at `2e9a56b` against base `f7263b1`. Everything below was reproduced on the PR head (pymc-bart 0.12.0, pymc 6.2.x) unless marked otherwise.

The seam is the right one — `build_model` losing ~40 lines of BART-specific branching in favour of `register_data` / `collect_coords` / `build_param` is a real reduction in coupling, and the error messages are unusually good (the "nested config dicts are replaced wholesale" hint survives verbatim; the missing-`y_name` case raises a clear `ValueError` rather than the obscure failure the description implies). The optional-dependency handling genuinely works: with `pymc_bart` absent, `import pymc_marketing.bart` succeeds, `Bart(...)` constructs, `to_dict()` emits a complete payload, and only `create_variable` raises — that's what makes "swap BART for `Dot` without installing pymc-bart" real rather than aspirational.

But **the headline example in this PR description emits a warning telling the user it was ignored**, and the "nothing changes for current users" claim has a hole. Details below.

---

## Blocking

### 1. `model_config["mu"]` is rejected as an unknown key

`"mu"` is documented as a first-class key (`pie/model.py:194-202`) but appears in neither `default_model_config` nor `_skipped_config_keys`, so `ModelBuilder.__init__` classifies it as a typo. Running the exact snippet from this PR description:

```
UserWarning: The following model_config keys are not used by the model and will be
ignored: ['mu']. Valid keys are: ['bart', 'categorical_split', 'sigma'].
```

The recipe *is* honoured, so the warning is factually wrong and contradicts the docstring it sits next to. All five new `mu` tests emit it; `pyproject.toml`'s `filterwarnings` doesn't promote it, so it only shows up in the pytest warnings summary.

```python
    _model_type = "PIE Model"
    version = "0.2.0"
    _skipped_config_keys: set[str] = {"mu"}
```

Worth pinning with a `warnings.simplefilter("error")` construction so it can't regress.

### 2. A dimensional `sigma` prior no longer builds — on the default path

`pm.Normal` → `pmd.Normal` while `sigma` stays deliberately non-`xdist`. `pymc.dims` distributions reject plain tensors carrying dims, so a `sigma` prior with `dims` — explicitly permitted by the documented `model_config["sigma"]` contract — now fails. **No `mu` recipe required to hit this**; A/B on the same input:

```python
PIEModel(pre_determined_features=["spend"], post_determined_features=["chan"],
         model_config={"sigma": Prior("HalfNormal", sigma=1.0, dims="obs")}).build_model(X, y)
```

| commit | result |
|---|---|
| `f7263b1` (base) | builds; `model["sigma"].type.shape == (None,)` |
| `2e9a56b` (head) | `ValueError: Variable sigma must have dims associated with it.` |

The PGBART constraint quoted in the comment only requires the *free variable* to be a plain tensor. Lifting it for the likelihood preserves that:

```python
sigma = self.model_config["sigma"].create_variable("sigma")
if (sigma_dims := getattr(self.model_config["sigma"], "dims", None)):
    sigma = pmd.as_xtensor(sigma, dims=sigma_dims)
```

### 3. `save()` with a non-`Bart` recipe writes an artifact that cannot be loaded

`"mu"` is now in `_serializable_model_config`, but only `Bart` is `@serialization.register`ed. For a `Dot` mean, `ModelBuilder.create_idata_attrs._json_default` falls through to its `obj.__dict__` branch and **`save()` succeeds**, writing:

```json
"mu": {"var_name": "X", "prior": {"dist": "Normal", "dims": ["feature"]}, "name": "mu_coef"}
```

No `__type__`. Replaying the load path then dies with `ValueError: model_config['mu'] must be a ModelTerm or a serialized term dict with a '__type__' key, got dict.` — an error that blames the user's input for a limitation of the writer. The description acknowledges the #2987 dependency, but shipping a silently-corrupt artifact is a different failure mode from a clear one. Put the check on the write side where it's actionable:

```python
recipe = config.get("mu")
if isinstance(recipe, ModelTerm) and "__type__" not in recipe.to_dict():
    raise NotImplementedError(
        f"{type(recipe).__name__} is not registered for serialization, so a "
        "PIEModel with this 'mu' recipe cannot be saved yet."
    )
```

### 4. `_serialize_split_rules` mangles split rules passed as classes

pymc-bart's own API takes split rules as classes, and this PR description imports them that way (`from pymc_bart.split_rules import OneHotSplitRule`). `rule.__class__.__name__` on a class object yields the metaclass name:

```python
>>> Bart(var_name="X", split_rules=[OneHotSplitRule, ContinuousSplitRule]).to_dict()["split_rules"]
['type', 'type']
>>> Bart.from_dict(_)
ValueError: Unknown serialized split rule 'type'.
```

Instances round-trip fine (`make_split_rules` returns instances, so the PIE default path is safe), but `Bart` is now public and the class spelling is the documented upstream one. Normalise in `__post_init__` (accept both, store one form) and serialise `rule.__name__ if isinstance(rule, type) else type(rule).__name__`.

---

## Important

### 5. `obs_dims = X.dims[:-1]` is positional, which is exactly where `Bart` stops being interchangeable with `Dot`

`bart.py:250`. `pmd.Data` preserves DataArray dim order verbatim, so `ds["X"]` declared as `(("feature", "obs"), X)` yields `obs_dims == ("feature",)`, hands pymc-bart a transposed matrix, and `as_xtensor` mislabels the result. Normally an opaque shape error from inside `BARTRV.signature`; when `n_obs == n_features`, a transposed fit with no error at all. `Dot`'s `data @ beta` is dim-name based and therefore transposition-safe — so `test_bart_dot_swappable` asserts a property the bridge doesn't actually have. Derive from `model[self.y_name].dims` and validate, or expose an explicit `feature_dim` field.

### 6. `sample_vars` is an undeclared sixth lifecycle member, and the walk that collects it is incomplete

`terms.py` documents five lifecycle methods; `Bart.sample_vars` is a sixth, consumed in `pie/model.py:486-490` via `getattr(term, "sample_vars", [])`. Two consequences:

- A custom-term author has no way to discover that this is the hook that makes out-of-sample prediction recompute their contribution. The `getattr` default silently freezes them from the trace instead.
- `collect_terms` recurses through `Sum` and `Product` but not `Transform`:

  ```python
  >>> [type(t).__name__ for t in collect_terms([Transform(dot, func=ptx.math.exp)])]
  ['Transform']
  >>> [type(t).__name__ for t in collect_terms([Intercept(name="i") + dot])]
  ['Intercept', 'Dot']
  ```

  Since the docstring advertises "compositions thereof", `Transform(Bart(...), func=...)` is a supported spelling — and there `_recipe_sample_vars` is empty, so PyMC freezes `bart` at training values after `_data_setter` swapped `X`. Equal-length prediction frames return training fitted values with only an `ImplicitFreezeWarning`.

Declare `sample_vars` on `ModelTerm` (default `[]`, documented in the lifecycle section) and fix `collect_terms` to descend into `Transform.inner`.

### 7. `Bart` isn't registered unless someone imports the module

`pymc_marketing/__init__.py` imports `bass, clv, customer_choice, mmm, pie, terms` — not `bart`. After a bare `import pymc_marketing`, `serialization.deserialize({"__type__": "pymc_marketing.bart.Bart", ...})` raises `SerializationError: Unknown type`. `pie/model.py:417-424` papers over it with a local import right before `deserialize`, but that's incidental. The package already does exactly this for side effects (`import pymc_marketing.r2d2  # Register R2D2 deserializers`), and the import is safe without pymc-bart thanks to the guard at `bart.py:101-104`.

### 8. The new `mu` Deterministic is a duplicate, and it's in the wrong units

For the default recipe `mu` is *exactly* `bart` lifted through `as_xtensor`, so `posterior` now carries two identical `(chain, draw, obs)` arrays; adding `"mu"` to `sample_vars` puts a third copy in `posterior_predictive`. At the documented `m=200` with a few thousand campaigns that's a silent doubling of idata size.

Worse, only `output_var` is rescaled by `_target_scale`. Fitted on `y * 1000` with a `Dot` mean: `list(pp.data_vars) == ['mu', 'y']`, `max|y| = 9573.1`, `max|mu| = 0.97`, `_target_scale = 4549.7`. A variable literally named `mu` sitting next to `y` and off by three orders of magnitude is a trap, and `sample_posterior_predictive`'s "in the original target scale" docstring is now only true of `output_var`. Either rescale it or keep it out of the returned group.

### 9. The version bump breaks every existing artifact

`version = "0.2.0"` is correct given the graph change, but `ModelIO.load_from_idata` **raises** `DifferentModelError` on a version mismatch rather than warning. Every PIEModel saved under 0.1.0 becomes unloadable without `check=False`. Fine — but it needs a release-note line, and if the redundant `mu` Deterministic goes away the graph is otherwise byte-identical to base (verified: `X`, `y_obs`, `bart`, `sigma`, `y`, their dims, and the `constant_data`/`observed_data` groups all match), so 0.1.0 could arguably be kept.

---

## Minor / nits

- **`_make_split_rules` deprecates a method that never existed.** `git log -S"_make_split_rules" --all` returns only `2e9a56b`; base built split rules inline. There's no released API to keep compatible — the shim, the `.. deprecated::` directive with no removal version, and `test_deprecated_make_split_rules_wrapper` all guard a private entry point nobody could have called. Delete all three.
- **`categorical_split` validation regression.** Base validated it unconditionally in `build_model`; now it's only reached via `_default_bart_recipe`, so with a user-supplied `mu` a typo is silently accepted. Validate in `__init__` (already `@validate_call`), or document that `"categorical_split"`/`"bart"` go inert when `"mu"` is set.
- **`make_split_rules` imports before it validates** (`pie/model.py:137` vs `139`), so `make_split_rules("nope", [...])` without pymc-bart raises a bare `ModuleNotFoundError` instead of the documented `ValueError` — and without the install hint every other touch point carries. Also missing the `Examples` section the repo style guide requires for public functions, and it's absent from `pymc_marketing.pie.__all__`.
- **Tests don't distinguish the recipe path from the default.** `test_custom_bart_recipe` and `test_mu_recipe_serialized_dict` both end at `assert "bart" in model.model.named_vars` — but `_default_bart_recipe()` also names its variable `"bart"`, so both pass unchanged if `model_config["mu"]` were ignored entirely, which is the exact regression they exist to catch. They set `m=10` against a default of `200`; assert that. Similarly `assert model.model["mu_coef"] is not None` can never fail (`pm.Model.__getitem__` raises `KeyError` or returns the var) — pin the shape instead.
- **No graph snapshot guards the compatibility claim.** `test_build_model` only does substring `any("bart" in n for n in var_names)` checks and never mentions `mu`. One `assert sorted(model.named_vars) == [...]` would make "nothing changes" checkable.
- **`test_set_data_resizes_obs` re-implements `Bart.create_variable`** line-for-line (`pmb.BART(..., X=model["X"].values, ...)` + `as_xtensor`) instead of calling `build_param(term)`, so a regression in the dims bridge wouldn't be caught there. It also asserts only `model["X"]`'s shape, not that `y_obs` and the `obs` coord resized.
- **`__import__("pymc")`** at `tests/bart/test_term.py:77` — the module already does `import pymc as pm` in four other tests. Same file, `tests/pie/test_model.py:738`/`:775` re-import split rules that are already module-level at line 17.
- **`:func:`pymc_bart.pymc_bart.BART``** (`bart.py:17`, `:148`) — no such module in 0.12.0, and `BART` is a class. Should be `:class:`pymc_bart.BART``. Renders as a dead link in the API docs this PR now generates.
- **Module docstring example isn't runnable** — uses `pm.Model(...)` without importing `pymc`, references `Dot` in the commented alternative without importing it, and `X`/`y` are undefined.
- **`__post_init__` validates `response` and `m` but not `alpha`/`beta`**, which are just as constrained (`alpha ∈ (0,1)`, `beta > 0`) and unvalidated upstream. An `alpha=1.5` typo yields a silently degenerate tree prior.
- **`docs/source/api/index.md`** — the list is otherwise strictly alphabetical; `bart` sorts before `bass`.
- **Forward-looking, not actionable now:** pymc-bart 0.13.x dropped the `pymc_bart.split_rules` module entirely (moved behind the Rust `bartrs` backend), and `pie = ["pymc-bart>=0.12.0"]` has no upper bound. Currently unreachable because 0.13.1 requires `pymc>=6.3.0` while this repo pins `pymc<6.3.0` — but worth capping at `<0.13` before that pin moves.

---

## Verified correct — not worth re-litigating

- **Symbolic `Y`.** `create_variable` now passes `Y=model[y_name].values` (symbolic) where base passed frozen numpy. Safe: `PGBART.__init__` materialises it once via `.eval()` at sampler construction, while `y_obs` still holds training values, and every downstream use (`init_mean`, `leaf_sd`, `sum_trees`) reads that frozen array. `BARTRV.rng_fn` only re-evaluates `cls.Y` in its `not cls.all_trees` fallback, never taken after a fit. Leaf prior is seeded exactly as before.
- **Default graph parity.** Variable names, dims, coords, `m`/`alpha`/`beta`/`response` defaults, split-rule construction, and idata group layout all identical to base. Only `mu` is added.
- **Deprecation mechanics** — right category, `stacklevel=2`, `pytest.warns(DeprecationWarning, match="make_split_rules")`.
- **Resized-obs posterior predictive** is genuinely covered for both flavours; the `Dot` path really does fit and predict in the suite.
- **`Bart` round-trips** through `create_idata_attrs` with `__type__` injected, and the `id` hash is stable across save/load and correctly distinguishes different `split_rules`.
- **`tests/bart/` runs in CI** — `testpaths = "tests"`, and the workflow invokes pytest with no path argument.
- **`tests/bart/__init__.py`** is the licence header only, matching every other test package. Not an import guard, and correct as-is.
- **`docs/source/api/index.md` needs no stub** — the directive is `autosummary :recursive: :toctree: generated/`.
- **No missed callsites.** `_make_split_rules` has no references anywhere in `pymc_marketing/`, `tests/`, `docs/`, or the notebooks; the PIE notebook's `model.model["X"].get_value()` still works because `pmd.Data` returns an `XTensorSharedVariable` subclassing `SharedVariable`, and `pmb.compute_variable_importance(idata, model.model["bart"], ...)` is unaffected.

---

**Assessment:** the design is right and I'd like to see it land. Items 1–4 should be fixed before merge — 1 undermines the feature's own documentation, 2 is a regression on the default path, 3 ships a corrupt artifact, 4 breaks the documented spelling of a public constructor argument. 5–9 are worth resolving or explicitly deferring with an issue.

williambdean:
Addressed the full review in 2e9a56b6e..7fd5e04eb (pytest: 131 passed in tests/pie + tests/bart + tests/test_terms.py; ruff/mypy/pre-commit green).

Blocking 1-4:
1. `_skipped_config_keys = {"mu"}` + warnings-as-error test.
2. Dimensional `sigma` priors restored: plain free variable (PGBART constraint), lifted with `pmd.as_xtensor(sigma, dims=...)` for the likelihood; A/B test added.
3. Serialization: chose in-PR serialization for the built-in terms (`Parameter`/`Intercept`/`Dot`/`Sum`/`Product`/`Transform`) rather than stacking on #2987; the write-side guard remains (`NotImplementedError`) for genuinely unregistered custom terms, and no-loadable-artifact is now unreachable for built-ins. There will be small semantic-only conflicts with #2987 at merge.
4. Split rules normalize class/instance in `__post_init__`; round-trip test uses the upstream class spelling.

Important 5-9: all addressed — dims from target with explicit transpose (5), `sample_vars` lifecycle member + `Transform` descent in `collect_terms` (6), side-effect module import for registration (7), `mu` rescaled by `_target_scale` (8), note in the PR description's Release note section since release notes are labeler-generated here (9).

Minors: shim deleted, `categorical_split` validated at construction, `make_split_rules` validates before importing pymc-bart + got an Examples block, dead `:func:` BART links fixed, docstring example runnable, `alpha`/`beta` validated, graph snapshot in `test_build_model`, `m=10` recipe assertions, resize test now via `build_param`, doc examples/imports tidied, `pymc-bart<0.13` cap + lockfile sync.

juanitorduz:
## Review of `7fd5e04e`

Re-reviewed `2e9a56b6..7fd5e04e` against base `f7263b13` (pymc 6.2.0, pymc-bart 0.12.0, pymc-extras 0.14.0). `tests/pie tests/bart tests/test_terms.py`: **131 passed, 1 skipped**; `ruff` clean. All nine items from the last round are genuinely fixed — I re-verified each, listed at the bottom.

Two new blockers, both introduced by the fixes rather than pre-existing.

---

## Blocking

### 1. The docs job will fail: the `Lifecycle (per term)` underline lost a character

`terms.py:42-43`. Adding the sixth bullet also shortened the underline from 20 to 19 carets under a 20-character title:

```
$ python -c "import docutils.core, pymc_marketing.terms as t; \
  docutils.core.publish_string(t.__doc__, writer_name='null', settings_overrides={'report_level': 2})"
<string>:29: (WARNING/2) Title underline too short.

Lifecycle (per term)
^^^^^^^^^^^^^^^^^^^
```

`.github/workflows/docs.yml:64` runs `sphinx-build … -W --keep-going`, and the paths filter includes `pymc_marketing/**`, so this PR triggers it. That file's own header: "any new warning is a regression and blocks the merge." One caret.

### 2. `Intercept(...) + Bart(...)` — the composition all three docstrings advertise — is rejected

`Sum` and `Product` are not `ModelTerm` subclasses, and `_resolve_mu_recipe` gates on `isinstance(recipe, ModelTerm)` (`pie/model.py:437`):

```python
>>> PIEModel(..., model_config={"mu": Intercept(name="intercept") + Dot(var_name="X", ...)}).build_model(X, y)
ValueError: model_config['mu'] must be a ModelTerm or a serialized term dict with a '__type__' key, got Sum.
```

Same for `Intercept + Bart`. That contradicts:

- `pie/model.py:38-45` — "``model_config['mu']`` takes any term recipe (`Bart`, `Dot`, **compositions thereof**)"
- `pie/model.py:211-219` — the same claim in the `PIEModel` docstring
- `bart.py:80-86` — "Terms compose, so BART can also be an ingredient of a larger expression: `mu = Intercept(name="intercept") + Bart(var_name="X", m=200)`" — not reachable through `PIEModel`

It is only the gate. Monkeypatching it to `isinstance(recipe, (ModelTerm, Sum, Product))` and changing nothing else:

```
Sum(Intercept, Dot)  builds: ['X', 'intercept', 'mu', 'mu_coef', 'sigma', 'y', 'y_obs']
Sum(Intercept, Bart) builds: ['X', 'bart', 'intercept', 'mu', 'sigma', 'y', 'y_obs']
_recipe_sample_vars: ['bart']
serialize(Sum) carries __type__: True
```

`collect_coords` / `register_data` / `build_param` / `collect_terms` already handle `Sum`, and `Sum`/`Product` became `@serialization.register`ed in this PR, so `create_idata_attrs` writes a loadable payload. Fix is the isinstance tuple plus widening `_mu_recipe: ModelTerm | Sum | Product | None`; the `_serializable_model_config` guard (`:364`) needs the same widening, or an unregistered child inside a `Sum` bypasses the actionable `NotImplementedError` (it still fails loudly via `SerializationError` from `_serialize_child`, just with a worse message).

---

## Important

### 3. `mu` and `bart` now sit in `posterior_predictive` in different units

Rescaling `mu` fixed the `mu`-vs-`y` mismatch and created a `mu`-vs-`bart` one. Default recipe, `y` scaled ×1000:

```
_target_scale = 2365.30
posterior            vars: ['bart', 'sigma', 'mu']   bart == mu: True   max|bart| = max|mu| = 0.2169
posterior_predictive vars: ['bart', 'mu', 'y']       max|bart| = 0.217   max|mu| = 513.1   max|y| = 3090.0
```

Two problems. `mu` is still a bit-identical duplicate of `bart` in `posterior` (item 8's size doubling is unchanged), and in `posterior_predictive` the *same quantity* now appears twice, three orders of magnitude apart. Worse, `posterior.mu` and `posterior_predictive.mu` are the same name in different units, so the obvious `idata.posterior.mu` → `idata.posterior_predictive.mu` comparison is silently wrong.

`sample_vars` and `var_names` are separate arguments — `var_names` "controls only which variables appear in the output; does not trigger resampling." So `bart` can be re-sampled without being returned:

```python
post_pred = pm.sample_posterior_predictive(
    self.idata, sample_vars=sample_vars, var_names=["mu", self.output_var], **kwargs
)
```

That leaves one `mu`, in target scale, and drops the duplicate. Nothing pins the pp `mu` scale today either — worth asserting `max|mu| ≈ max|bart| * _target_scale`.

### 4. `collect_terms` *replaces* a `Transform` with its inner term instead of including both

`terms.py:981-984` recurses into `p.inner` but drops `p`, so a `Transform` subclass declaring `sample_vars` is silently frozen — item 6's failure mode, one level up:

```python
class LogBart(Transform):
    @property
    def sample_vars(self): return ["extra"]

>>> [v for t in collect_terms([LogBart(dot, func=ptx.math.exp)]) for v in t.sample_vars]
[]
```

The docstring promises a "flattened list of all individual term instances" and every other branch is additive. `result.append(p)` before recursing into `p.inner` fixes it and costs nothing, since `Transform.sample_vars` is `[]`.

Related: `test_collect_terms_includes_transform` (`tests/test_terms.py:334`) asserts only `len(result) == 1`, so it still passes while collecting the `Intercept` rather than the `Transform` — the test name is no longer true. Assert types.

---

## Minor / nits

- **Duplicated validation block in `make_split_rules`** — `pie/model.py:149-153` and `:156-160` are the identical `if`/`raise`, a leftover from hoisting the check above the import. Delete the second.
- **`terms.py:38` still says "implement five lifecycle methods"** while the list below it has six and `sample_vars`'s own docstring calls it "the sixth lifecycle step".
- **`make_split_rules` still lacks the install hint.** `pie/model.py:154` imports `pymc_bart.split_rules` after validating, so `make_split_rules("onehot", cols)` without pymc-bart raises a bare `ModuleNotFoundError`. PIE's own path is covered by `_default_bart_recipe`'s `pmb is None` check, but the function is public and now in `pymc_marketing.pie.__all__`.
- **`read_bart_m` (`tests/pie/test_model.py:113`) is a heuristic where a direct read exists.** It returns the first scalar RV input that is integer-valued in `(0, 256]`; `beta=2.0` also qualifies, and it only works because `m` precedes `beta` in `['RNG', '[]', X, Y, '13', '0.95', '2.0']`. `m=2` would be indistinguishable from `beta`. `model.model["bart"].owner.op.m` is the value directly (verified: `op.m == 13` for `Bart(m=13)`), and `op.alpha` / `op.beta` / `op.response` / `op.split_rules` are there too if you want to pin the rest of the recipe rather than just the tree count.
- **`test_mu_recipe_key_is_ignored_without_warning`** asserts the opposite of its name — that `mu` is *not* ignored. `test_mu_recipe_key_is_not_warned_about`.
- **The new round-trip tests assert only the type.** `test_serialization_parameter_and_dot_roundtrip` / `test_serialization_sum_product_transform_roundtrip` check `type(rebuilt) is type(term)`; payload fidelity is unasserted (it does hold — `Dot` returns with `var_name`, `name`, and `prior.dims == ('feature',)`). The trailing `assert rebuilt.to_dict() == ...` also reads the leaked loop variable instead of naming `dot`'s rebuild.
- **`pymc_marketing/__init__.py:25-28`** — isort placed `import pymc_marketing.bart` between `# Load the data accessor` and `import pymc_marketing.data.fivetran`, so that comment now labels the BART import.
- **`import pymc_marketing.bart` at package import makes `pymc_bart` unconditional.** 133 ms of the 2.30 s `import pymc_marketing` here (131 ms of it `pymc_bart.split_rules`), paid by every MMM/CLV user who never touches PIE. `serialization.register("pymc_marketing.bart.Bart", …)` from a module that doesn't import `pymc_bart` at top level would avoid it. Flagging the tradeoff, not asking for a change.
- **`tests/bart/test_term.py` imports `pymc as pm` inside six separate test bodies** while `numpy` / `xarray` / `pymc_bart` are module-level. Last round's nit was the reverse; this overshot it.

---

## Verified fixed

Each item from the previous review, re-run on `7fd5e04e`:

1. **`_skipped_config_keys = {"mu"}`** (`pie/model.py:287`) — the hook is real (`model_builder.py:1195`) and `test_mu_recipe_key_is_ignored_without_warning` pins it under `simplefilter("error")`.
2. **Dimensional `sigma` builds again** — plain free variable, lifted with `pmd.as_xtensor(sigma, dims=…)` for the likelihood; `sigma.eval().shape == y.shape` on the default BART path, with no `mu` recipe.
3. **Built-in terms serialize.** `serialize(Dot(...))` → `{'__type__': 'pymc_marketing.terms.Dot', 'var_name': 'x', 'prior': {'dist': 'Normal', 'dims': ('feature',)}, 'name': 'coef'}` and round-trips with dims intact; the write-side `NotImplementedError` fires for an unregistered `ModelTerm`. Registration isn't accidentally inherited either — `_wrapped_to_dict` recomputes `__type__` from `self.__class__`, so an unregistered `Bart` subclass still trips the registry lookup.
4. **Split rules take the upstream class spelling.** `Bart(var_name="X", split_rules=[OneHotSplitRule, ContinuousSplitRule]).to_dict()["split_rules"] == ['OneHotSplitRule', 'ContinuousSplitRule']`, rebuilding as instances.
5. **`obs_dims` derived from the target** with an explicit `X.transpose(*obs_dims, *feature_dims)` and a clear `ValueError` when the feature axis is ambiguous; `test_transposed_data_var_is_ordered_before_bart` catches the `(feature, obs)` declaration.
6. **`sample_vars` is a declared `ModelTerm` member** with a default of `[]`, documented in the lifecycle list (see item 4 for the remaining gap in the walk).
7. **Registration without an explicit import.** `import pymc_marketing` alone: `serialization.deserialize({"__type__": "pymc_marketing.bart.Bart", …})` → `Bart`.
8. **`mu` is rescaled** (see item 3 for what that surfaced).
9. **Release note in the description** — the repo has no changelog file, so that is the right place.

Also spot-checked and fine: `parse_model_config` leaves a serialized `mu` dict untouched on the load path; `_func_name`/`_resolve_func` round-trip symmetrically for `ptx.math.exp`, `ptx.math.sigmoid`, `ptx.linalg.solve`, `ptx.concat`, `ptx.dot`; `xr.DataArray` and numeric-literal children of `Product` round-trip both ways; no `ImplicitFreezeWarning` anywhere in the pie suite; `docs/source/api/index.md` now sorts `bart` before `bass`; the `categorical_split` typo is rejected at construction.

---

**Assessment:** 1 and 2 block — 1 fails the docs job outright, 2 is a one-line gate standing against three docstrings. 3 and 4 should be fixed or explicitly deferred with an issue. The rest are nits. Design still right, and closer than the diff size suggests.

Issue #1386, opened 2025-01-16
We could make BudgetOptimizer num_periods change across calls to `allocate_budget`


ricardoV94:
Write now this is hard-coded as a constant, but it could be a shared variable. Not sure if needed, but I assume the point of the BudgetOptimizer is to allow trying different optimizations without recompiling everything?

Issue #26, opened 2022-05-04
Take the CLV grid to the next level


ricardoV94:
The idea is to follow up to #25, and extend the basic models in interesting ways, e.g., by adding hierarchical effects, time-varying effects... and so on. This should give us a more refined idea not only of what building blocks we need, but how flexible they should be.

This will, potentially,  also be the biggest selling point of the package, as we will be doing things that are not really done out there (or at least not published in neat papers / packages), in large part by taking advantage of working with a fully-fledged PPL (PyMC!)

![image](https://user-images.githubusercontent.com/28983449/166697627-9b4779c4-db9d-402c-b708-1be475d3bab0.png)

Unlike #25, these squares are not yet fixed, and any cool idea you have can be used.

- [ ] Continuous Non-contractual + Hierachical structure (#39) (up for grabs)
- [ ] Continuous Contractual + ??? (up for grabs)
- [ ] Discrete Non-contractual + ??? (up for grabs)
- [ ] Discrete Contractual + cohort / temporal effects (suggested to @drbenvincent), see #35

Possible the grid won't be over the 4 types of models, but perhaps over extensions:
- [ ] Cohort + Temporal effects on lifetime
- [ ] Cohort + Temporal effects on value
- [ ] Complex interactions between Lifetime and Value components
  * E.g., subscription fee affects churn-rate and value, @juanitorduz brought something that resonates with this
 - [ ] THE NEXT BANG IN MARKETING MODELS ???

Issue #24, opened 2022-04-21
Sketch MMM user API


ricardoV94:
In this issue we want to discuss a potential API for the MMM model. For inspiration, take a look into Google's LightWeight MMM https://github.com/google/lightweight_mmm

juanitorduz:
I think https://github.com/pymc-labs/pymc-marketing/pull/123 should close this first iteration for the alpha release.

PR #3002, opened 2026-09-10
feat(optimizer): rebind the compiled objective to a new posterior without recompiling

PabloRoque:
## Description

`extract_response_distribution` substitutes posterior draws into the graph as constants (`pytensor_utils.py`, the `xtensor_constant` replacement). That welds every graph, and every function compiled from it, to one posterior. Any loop that re-solves the optimizer against many posteriors (a resampled, thinned, or updated posterior per iteration) pays a full extract plus compile each time, and compilation dominates the runtime.

This PR lets the draws be swapped under an already compiled function.

## Changes

- `SharedPosterior` (`pymc_marketing/pytensor_utils.py`): holds posterior draws in PyTensor shared variables, one per posterior variable, created on first use and reused by later extractions that pass the same instance. `set_posterior(idata)` rebinds all of them at once under one contract, stated in the class docstring: every bound variable is present with the dtype it was bound with; every non-sample dimension keeps its name and its set of labels, compared by value through `pd.Index` (a reordered `channel` axis is realigned, a datetime axis may change storage unit across a netCDF round trip, new or duplicated labels are refused); a dimension bound without labels must arrive without labels; the number of draws is free. Anything else raises before anything is written: validation, alignment and the dtype cast all happen in a prepare pass, and the write loop only hands over arrays that already fit. A later extraction with the same instance must be conditioned on the draws already bound (NaN-aware comparison), otherwise it raises rather than silently reading them.
- `extract_response_distribution(..., shared_posterior=None)`: keyword-only. When given, the draws are bound through the holder instead of constants. The default constant path is unchanged.
- `BudgetOptimizer.set_posterior(idata)`: the first call creates the holder, compiles the objective and the constraints against the construction posterior, and then moves the new draws in through the same validated rebind path as every later call, so a reordered or resized axis is treated identically on the first call and the tenth; every later call is `set_value` only. A failure at any step of the first call rolls back the holder, `idata`, the compiled objective and the constraints, so the constant-folded graphs stay in place and the next call retries. Optimizers that never call it keep those constant-folded graphs (measured on the repo's dummy MMM: 36 apply nodes on both `main` and this branch, versus 61 when the shared binding was unconditional). Accepts a `DataTree`, a bare posterior `Dataset` (which replaces `self.idata` with a tree holding only `posterior`), or anything the constructor accepts. The rebind is local to the optimizer; an MMM it was built from keeps its own `idata`.
- The budget mask is fixed at construction: it defines the decision vector the graphs were compiled for, so a rebind never re-derives or compares it. A posterior that would have auto-detected a different mask is solved on the existing one, and the docstring says to build a new optimizer to change it. A user-supplied mask is still validated against the new posterior the way construction validates it. Earlier revisions refused a posterior implying a different auto-detected mask; that gate needed a new carve-out every review round (thinned posteriors without `channel_contribution`, dimension order, the all-ones fallback), so it was replaced by pinning the mask and saying so.
- Bug fix, incidental: with arviz 1.2, `az.extract` squeezes a one-variable posterior group to a `DataArray`, so `extract_response_distribution` on `main` fails with `KeyError` for such a posterior. `_posterior_sample_major` normalises it; regression test on the constant path included.

Related: #1386 asks for the same treatment of `num_periods`. Not addressed here: the date axis is frozen into the model before the optimizer builds its graph, so freeing it is a much larger change.

## Benchmark

Ten posteriors of 4 chains × 500 draws, solved on the repo's dummy two-channel MMM, identical allocations both ways (`sandbox`-style script, not committed):

| Approach | Total | Per posterior |
|---|---|---|
| Rebuild the optimizer for each posterior | 15.9 s | 1,594 ms |
| One optimizer, `set_posterior` per posterior | 0.36 s | 36 ms |

The first `set_posterior` recompiles once, about 260 ms here. After that a swap is about 2 ms and a solve about 5 ms, so the per-posterior cost is the solve. The rebuild path pays for model cloning, graph extraction and compilation of the objective and constraints every time. On a larger model the solve grows, so the ratio shrinks, but the rebuild never gets cheaper than the solve it wraps.

## Notes for reviewers

- `az.extract` on a plain posterior `Dataset` is supported (it goes through `convert_to_dataset`), and the optimizer already fed it one on `main` before this PR, exercised by the whole optimizer suite. A `Dataset` that already carries a `sample` dim raises in arviz.
- `arviz.InferenceData` no longer exists on arviz 1.x, so the `_to_datatree` branch in `set_posterior` is the same legacy shim the constructor uses.

## Tests

- `tests/test_pytensor_utils.py`: a graph bound through `SharedPosterior` matches the constant path for each posterior with the same compiled function across a rebind and a changed draw count; two graphs extracted with one instance are compiled, rebound once, and both move; an extraction with different draws is refused, and NaN draws do not make an identical posterior look different; a rebind that fails on the second bound variable, by layout or by an uncastable value, leaves every variable untouched; reordered coordinates realign, new labels are refused with a message that reports counts and a few examples rather than both axes, and labels on one side only are refused in both directions; `[us]` and `[ns]` datetime labels compare equal; missing variables, extra dims and resized non-sample axes are refused; the one-variable posterior regression on the constant path.
- `tests/mmm/test_budget_optimizer.py`: the optimizer starts on constants, the first `set_posterior` recompiles once and the second does not, and after each swap the allocation and objective match a fresh optimizer built on that posterior at an interior optimum; the same reordered posterior gives the same allocation whether it arrives on the first call or a later one, and a resized axis behind a user mask is refused with a full rollback; a failed first call leaves the holder, `idata`, objective and constraints untouched and the next call retries; a custom `Constraint` that calls `optimizer.extract_response_distribution` follows the rebind, and its compiled entry survives the second one; a bare posterior `Dataset` is accepted; a posterior without `channel_contribution` is accepted; a posterior implying a different auto-detected mask is solved on the pinned mask while a fresh optimizer picks the other cell, and an all-ones fallback mask stays all-ones; a posterior missing a bound variable raises.
- `tests/mmm/test_budget_optimizer_mmm.py`: `mmm.budget_optimizer(...)` rebinds, the allocation moves, and the MMM's own `idata` is untouched.
- `tests/test_pytensor_utils.py`, `tests/mmm/test_budget_optimizer.py`, `test_budget_optimizer_mmm.py`, `test_counterfactual.py`, `test_sensitivity_analysis.py`, `test_incrementality.py` pass locally. pre-commit passes.

## Checklist
 - [x] Each commit corresponds to a relevant logical change


PabloRoque:
Thanks. Checked each item before changing anything; a0c5877 takes the minor ones. On the four "before merge" items:

**1. `az.extract` on a Dataset.** Supported: `extract` goes through `convert_to_dataset`, which accepts a Dataset, and this exact call (`_extract_dataset(self.idata, "posterior")` fed to `extract_response_distribution`) is already on `main` at `budget_optimizer.py:2063`, exercised by the whole optimizer suite. A Dataset that already carries a `sample` dim raises a `ValueError` in arviz rather than doing anything implementation-defined. The new test also feeds `set_posterior` a plain `(chain, draw)` Dataset directly.

**2. KeyError message matching.** `pytest.raises(match=...)` uses `re.search` on `str(exc)`. `str(KeyError(msg))` is the repr of the message, so it gains surrounding quotes, but an unanchored search still matches inside them. The test passes for that reason, not by luck, and this doesn't vary by Python version. Kept `KeyError`, since a missing variable name is a missing key.

**3. `set_posterior` before compilation.** The objective is compiled in `model_post_init`, not in `allocate_budget`, so the binding exists the moment the optimizer is constructed and there is no window for a silent no-op. Added an assertion to the rebind test pinning that.

**4. InferenceData path.** `arviz.InferenceData` no longer exists on arviz 1.x (importing it emits a MigrationWarning), so the `_to_datatree` branch is a legacy shim shared with the constructor. Added a test with a duck-typed object exposing `groups()` and a `posterior` attribute, which is what that shim keys on.

**Minor items, taken:** `set_posterior` now raises a clear `ValueError` when a variable's non-sample lengths differ from what the graphs were built against (channels, geos), with a test; `SharedPosterior.__repr__` shows each bound variable and its current shape; a comment records why `borrow=True` is safe.

juanitorduz:
## Review

Reviewed at `a0c5877` against `main`. The mechanism works — I confirmed the shared binding carries no static shape (`Matrix(float64, shape=(?, ?))`), so rebinding 1↔5 draws and length-1 non-sample axes all evaluate correctly; on a real MMM graph a rebound function matched a fresh constant-path extraction to `4.4e-16` (no constant folding, no baked sample length); and the objective, its gradient and a custom `Constraint` whose `constraint_fun` calls `optimizer.extract_response_distribution` all follow `set_posterior` without recompiling (constraint value `-4.866 → -12.837`, jac `[-2.72 -6.59] → [-16.32 -0.66]`, with `c is o._compiled_constraints[1]`). A fitted `MMM` via `mmm.budget_optimizer(...)` rebinds too (`[4.066 5.934] → [7.058 2.942]`).

One blocking issue, then a few that change what the API promises.

### Blocking

**1. The rebind is not atomic.** `SharedPosterior.set_posterior` validates each variable inside the same loop that calls `set_value` (`pytensor_utils.py`, the `for name, var in self._variables.items()` block). A posterior that is compatible on one variable and not on another — a refit that added a channel or a control is the obvious case — leaves the earlier variables already swapped, and `BudgetOptimizer.set_posterior` never reaches `self.idata = idata`. The optimizer then solves against a mixture of two posteriors and reports success:

```
raised: ValueError Posterior variable 'beta' has shape (3,) over dims ('channel',), expected (2,)...
objective under A : -40.0
objective under B : -118.0
objective after the failed set_posterior: -56.0
optimizer.idata reports beta = [1. 1.]
shared beta in use = [1. 1.]   shared gamma in use = [2. 2.]
allocate_budget after the failure -> [5. 5.]  success: True
```

`-56.0` matches neither posterior and nothing warns on the next solve. Two passes — validate every variable, then mutate — fixes it; restoring snapshots on failure also works but costs a copy. No current test can see this: `test_set_posterior_requires_every_bound_variable` exercises the up-front `KeyError`, which is raised *before* the loop, and the dims/shape tests use single-variable posteriors, where partial mutation cannot appear.

**2. `_get_or_create` reuses on the name alone, discarding the second extraction's draws.** The class docstring advertises "one instance can back several graphs", and `BudgetOptimizer` relies on exactly that. But when `name` is already bound, `_get_or_create` returns the existing variable and the *stored* dims and ignores the `posterior_da` it was handed. So a second `extract_response_distribution(model, other_idata, ..., shared_posterior=sp)` silently returns a graph conditioned on the **first** idata — verified with two different prior-predictive seeds: `gB equals A draws? True`, `gB equals B draws? False`. A name collision with a different non-sample length is accepted at extraction time and fails opaquely at call time (`Incompatible vectorized shapes for input 1 and axis 1...`) or, for 2-channel reused into a 3-channel model, silently evaluates the wrong values. dtype is unchecked as well.

`set_posterior` already validates precisely these invariants and raises clearly. Applying the same dims/shape/dtype comparison on the reuse branch turns all three into one explicit `ValueError` at extraction time.

### Important

**3. The auto-detected budget mask keeps the old posterior's answer.** When `budgets_to_optimize` is not supplied, `model_post_init` derives it from the construction posterior's `channel_contribution` (`budget_optimizer.py:1591-1611`). `set_posterior` keeps it — the docstring says so — but the result is that `budgets_to_optimize` and `idata` end up in a combination construction can never produce, and the rebound optimizer can allocate the opposite way from a fresh one:

```
mask at construction (A):    {'channel_1': True, 'channel_2': False}
mask after set_posterior(B): {'channel_1': True, 'channel_2': False}
optimizer.idata mean contribution per channel: [1. 1.]
rebound allocation: {'channel_1': 2.0, 'channel_2': 0.0}
fresh   allocation: {'channel_1': 0.0, 'channel_2': 2.0}
```

This also means `test_set_posterior_rebinds_without_recompile`'s central assertion — agreement with a fresh optimizer built on the new posterior — is not a property of the API. It holds only because `dummy_idata`'s `channel_contribution` is all ones and the update leaves it untouched, so the auto-detected mask cannot move. Re-deriving the mask in `set_posterior` when it was auto-detected and raising (or warning) on a difference would make the equality real; at minimum, pin the divergence in a test so the assertion is not read as a guarantee.

**4. Coordinate labels are not honoured on rebind.** Draws are bound positionally, so a posterior with the same dims and lengths but a permuted `channel` coordinate is accepted and silently mismapped:

```
original   -> [[ 2. 20.]]
reordered coords: ['b', 'a']  values: [10.  1.]
after reorder rebind -> [[20.  2.]]
```

and at the optimizer level: `label-correct {'c1': 10.0, 'c2': 0.0}` vs `permuted {'c1': 0.0, 'c2': 10.0}`, no error. The positional binding is pre-existing — constructing directly from the permuted posterior gives the same reversed allocation — so the PR does not introduce the mechanism. It does introduce the entry point where posteriors arrive from somewhere *other* than the model that built the optimizer, which is where a reordered coordinate is most likely, and it documents a validation contract a reader will assume covers labels. `align_to_model_coords` in this very file already does this for every other coordinate-bearing input; either reindex, or say in the docstring that coordinates match by position.

**5. The shared binding is unconditional, and it costs the majority who never rebind.** `model_post_init` always creates a `SharedPosterior` and `extract_response_distribution` always passes it, so posterior subgraphs lose constant folding and static shapes for every optimizer. Same script on `main` vs this branch, repo dummy MMM fixture:

```
PR   apply nodes in compiled objective+grad: 61   mean objective+grad eval: 5.3 us
main apply nodes in compiled objective+grad: 37   mean objective+grad eval: 4.6 us
```

and at 4 chains × 500 draws: `1.212 ms` vs `1.127 ms` per call. ~+65% graph nodes, +7–15% per solver iteration, paid unconditionally — and a model with a larger posterior-only subgraph (HSGP, time-varying, frozen deterministics) has more to lose, since those subexpressions are now recomputed every iteration instead of folded once. Worth either measuring on a realistic model before making it the default, or making it opt-in (a field, or lazy creation with one recompile on the first `set_posterior`).

### Minor

**6. `_posterior_sample_major` is a real bug fix, not a refactor — and it has no test.** With arviz 1.2.0, `az.extract` squeezes to a `DataArray` when the group holds exactly one data variable, so `main`'s `az.extract(idata).transpose(...)` followed by `posterior[placeholder.name]` crashes. Verified on `main`:

```
arviz 1.2.0
az.extract type for 1-var posterior: DataArray
main extraction FAILED: KeyError 'beta'
```

The PR tree returns the correct values. That deserves a line in the description and a regression test that calls `extract_response_distribution` on a one-variable posterior via the **constant** path (fails before, passes after); right now the branch is only reached incidentally by the single-variable `Dataset` inputs in the validation tests.

**7. `test_set_posterior_accepts_inference_data_like_objects` tests the shim, not the feature.** It only exercises the pre-existing `_to_datatree`, which the constructor already routes every non-`DataTree` idata through, and the object it imitates no longer exists in this dependency set (`az.InferenceData is xr.DataTree` on arviz 1.2.0). Both behaviours the patch actually advertises — custom constraints following the rebind, and the MMM path — are untested at this level, and I verified both work. Swapping this test for either would be a straight upgrade.

**8. Test assertions that pin implementation.** `assert repr(shared_posterior) == "SharedPosterior(beta(3, 2))"` pins separator, parenthesis style and dict insertion order for a test about rebinding; `variables["beta"].get_value(borrow=True).shape == (1, 2)` states the same fact. Same for `assert shared_posterior.dims == {"slope": ("sample",)}`. And `test_shared_posterior_is_reused_across_extractions` asserts only that the variable appears in both graphs' ancestors — neither graph is compiled or evaluated, so the promise it exists to cover ("a single `set_posterior` rebinds all of them at once", and the finding in 2) stays invisible. Compile both, rebind once, assert both outputs moved.

**9. The two `set_posterior` methods disagree on what `idata` may be.** `SharedPosterior.set_posterior` documents and accepts a bare posterior `Dataset` (its own test passes one). The optimizer wrapper calls `_to_datatree` first, so the same object dies with `AttributeError: 'Dataset' object has no attribute 'groups'`. Wrap it, or raise `TypeError` naming the accepted types.

**10. `set_posterior` splits the optimizer from its model.** `optimizer.idata is mmm.idata` before the call; afterwards only the optimizer's reference moves (verified: `mmm.idata untouched: True`). That is the right choice, but MMM-level post-processing of the resulting allocation still runs under the old draws. One sentence in the docstring saying the rebind is local to the optimizer would stop people from combining the two.

---

Nice piece of work overall — the compile-once goal is met and the gradient/constraint sharing is genuinely there. #1 is the one I would not merge without; #2 and #3 are where the API currently promises more than it delivers.

PabloRoque:
Thanks, this was a thorough one. I reproduced every finding before changing anything, and they all held; f326ab4 addresses them (rebased onto main after #2993 merged; the custom-constraint test now finds its compiled entry by key). PR description updated to match.

**1. Atomic rebind.** Two passes now: `set_posterior` validates and lays out every variable first, then writes. A posterior that fails on the second variable leaves the first untouched. Test: a two-variable model where `beta` is valid and `gamma` carries an extra dim; after the `ValueError`, `beta` and the compiled output are unchanged.

**2. Reuse on the name alone.** Reproduced (`gB` evaluated to the first posterior's 2, not 30). `_get_or_create` on the reuse branch now checks dtype, dims, non-sample lengths, coordinates, and the values themselves, and raises telling the caller to extract from the bound draws and use `set_posterior` to move everything. Test: two graphs extracted with one instance are compiled, rebound once, both move; a second extraction with different draws raises; with the same draws it works.

**3. Auto-detected mask.** Reproduced (rebound `[2, 0]` vs fresh `[0, 2]`). The mask fixes the length of the decision vector the graphs were compiled for, so it cannot follow the posterior. `set_posterior` now re-derives it when it was auto-detected and raises, changing nothing, if it would differ; a user-supplied mask is validated against the new posterior the way construction validates it. The rebind test now uses two updates with different draw counts and compares each against a fresh optimizer, and a separate test pins the refusal.

**4. Coordinates.** Reproduced. `SharedPosterior` stores the labels of every non-sample dim at creation and realigns by label on rebind; a posterior with different labels is refused. Positional only when the new posterior carries no coordinates for that dim, and the docstring says so.

**5. Unconditional cost.** Reproduced on the repo's dummy MMM: 61 apply nodes on the branch vs 36 on `main`. Went with lazy creation: the optimizer keeps the constant-folded graphs until the first `set_posterior`, which recompiles the objective and constraints once against shared variables; later calls are `set_value` only. Default optimizers are back to 36 nodes, identical to `main`.

**6.** Yes, it is a fix, not a refactor. Reproduced on `main` (`KeyError: 'beta'` for a one-variable posterior). Regression test on the constant path added, and the description now says so.

**7.** Swapped the shim test for the two you named: a custom `Constraint` calling `optimizer.extract_response_distribution` follows the rebind and its compiled entry survives the second one, and an MMM-level test via `mmm.budget_optimizer(...)` rebinds while `mmm.idata` stays untouched.

**8.** Dropped the repr and `dims` equality pins; the reuse test now compiles both graphs and asserts both outputs moved.

**9.** `BudgetOptimizer.set_posterior` accepts a bare posterior `Dataset` (wrapped into a tree), with a test.

**10.** Docstring now says the rebind is local to the optimizer and that a model it was built from keeps its own draws.

juanitorduz:
## Review, round 2

Reviewed at `5743b310`. Thanks for reproducing everything before changing it — that shows in the diff.

Confirmed fixed, with evidence: **#5** (construction-time apply nodes in the compiled objective+grad are 40 on `main` and 40 on this branch, 64 only after an opt-in rebind — the lazy binding does what it says), **#2** (a second extraction with different draws now raises), **#4** (a reordered `channel` axis realigns, a relabelled one is refused; `np.str_` vs `object` coords compare equal, duplicate labels rejected), **#6** (the one-variable regression test is on the constant path and does fail on `main`: `KeyError: 'beta'` at `016a2cc0`, `[[2. 6.]]` here), **#7** (shim test gone), **#9** (bare `Dataset` accepted), **#10** (docstring + MMM-level test). The mask refactor into `_posterior_channel_mask` / `_auto_detect_mask` / `_validate_mask_against_posterior` is behaviour-identical to `main` across all six construction cases (auto + `channel_contribution`, auto without it, auto with no posterior group, user-valid, user-invalid, user-invalid-without-`channel_contribution`) — same values, dims, dtype, coords, same exception types. The `set_constraints(list(self._constraints.values()))` round trip is identity-preserving (same keys, same order, same `Constraint` object ids, `add_default` correctly `False`), including for a user constraint keyed `default`.

**#1 and #3 are not done, though — the atomicity defect moved rather than disappeared, and it is worse at its new location.**

### Critical

**11. A failed *first* `set_posterior` wedges the optimizer into silently reporting the construction posterior forever.** `SharedPosterior.set_posterior` is atomic now, but `BudgetOptimizer.set_posterior`'s first-call branch is not:

```python
self._shared_posterior = SharedPosterior()
self.idata = idata
self._compile_objective_and_grad()          # can raise
self.set_constraints(constraints=list(self._constraints.values()))
```

If the recompile raises, four pieces of state disagree: `_shared_posterior` exists and is *partially* populated, `self.idata` is already the rejected posterior, and `_objective_and_grad` is still the constant-bound graph from construction. Every later `set_posterior` then takes the fast path — `_shared_posterior is not None` — succeeds, and rebinds variables the compiled objective does not read. Repro, using the repo's own `mmm_wrapper`/`dummy_idata` fixtures:

```
baseline (construction posterior) : [2.22e-16 2.00e+00] -12.90973939184704
first set_posterior raised: KeyError "No variable named 'saturation_beta'..."
_shared_posterior after failure  : SharedPosterior(adstock_alpha(4, 2), saturation_lam(4, 2))
optimizer.idata is the broken one: True | vars: ['adstock_alpha', 'channel_contribution', 'saturation_lam']

after set_posterior(good):          # a valid, different posterior — no error
  rebound alloc : [2.22e-16 2.00e+00] -12.90973939184704
  fresh   alloc : [2.00e+00 9.66e-15] -21.379778664450953
  equals baseline (construction posterior)? True
  equals fresh (claimed contract)?          False
  optimizer.idata is good: True
```

The optimizer is now permanently pinned to the construction draws while `optimizer.idata` advertises the new ones, and it never recompiles again because the holder is non-`None`. Note `saturation_beta` was never bound, so `SharedPosterior.set_posterior` happily rebinds 2 of 3 variables and reports success. This is reachable from exactly the input your own test feeds it: `test_set_posterior_requires_every_bound_variable` asserts the `KeyError` and stops there — it leaves the optimizer in this state and does not notice.

Fix: build into locals and commit only on success — create the holder, compile the objective and the constraints against it, and only then assign `self._shared_posterior`, `self.idata`, `_objective_and_grad` and `_compiled_constraints`. On failure, drop the holder entirely so the next call retries the first-call path. The mask gate already has exactly this discipline (`assert optimizer._shared_posterior is None  # nothing changed`); the compile path needs it too. Worth asserting in the missing-variable test, plus an `allocate_budget` after the failure that still matches the construction posterior.

### Important

**12. The write loop in `SharedPosterior.set_posterior` can still half-write.** Validation fills `prepared`, but the write loop calls `values.astype(var.type.dtype)`, which is fallible *after* validation passed:

```
binding order: ['gamma', 'beta']
RAISED ValueError could not convert string to float: 'x'
after: {'gamma': array([[9., 9.]]), 'beta': array([[1., 1.]])}
fn(): [[9. 9.]]
```

`gamma` moved, `beta` did not, the compiled function returns a cross-posterior product, and no exception is left to observe. Doing the cast inside the `prepared` comprehension makes the write loop infallible — `set_value(..., borrow=True)` on an already-correct-dtype array cannot raise — and removes an allocation from the write phase.

**13. `test_shared_posterior_rebind_is_atomic` does not pin #1.** The comment says "beta is valid and comes first", but `extract_response_distribution` creates the variables in `needed_rvs` order, which here is `['gamma', 'beta']` — the *corrupt* one is validated first, so the round-1 validate-and-write-in-one-loop implementation raises before writing anything and passes this test unchanged. Proven by patching the old implementation back in:

```
with the PR's atomic set_posterior:                        -> PASSES
with the round-1 NON-ATOMIC set_posterior reverted in:     -> STILL PASSES

same non-atomic version, corrupt variable bound SECOND ('beta'):
  raised: Posterior variable 'beta' has dims ('sample', 'extra', 'channel'), ...
  gamma: [[9. 9.]] beta: [[1. 1.]] fn: [[9. 9.]]   <- half-swapped
```

Corrupt the *second* entry of `list(shared_posterior.variables)` instead. That version would also have caught 12.

**14. The auto-mask gate over-corrects: a posterior carrying only the free RVs the graphs read is refused.** `_auto_detect_mask` falls back to all-ones when the new posterior has no `channel_contribution`, and the gate reads that fallback as "the mask is every cell" rather than "the mask cannot be derived from this posterior". So the thinned/simulated posterior this method exists for is rejected whenever construction auto-detected a narrower mask:

```
construction mask: [True, False]  auto: True
thinned vars: ['saturation_lam', 'saturation_beta', 'adstock_alpha']
RAISED ValueError The new posterior implies a different auto-detected budgets_to_optimize
mask than the one this optimizer was built with ([True, False] -> [True, True]).
```

That contradicts the method's own docstring ("The new posterior must hold every variable the graphs read"), and it is the mainstream path: `MMM.optimize_budget` / `mmm.budget_optimizer` always pass `budgets_to_optimize=None`, so `_mask_auto_detected` is `True` for every MMM-built optimizer. When `_posterior_channel_mask(idata)` is `None`, fall through to `_validate_mask_against_posterior` (a no-op in that case) instead of raising.

**15. NaN draws make the reuse check refuse an identical posterior.** `np.array_equal(values, var.get_value(borrow=True))` defaults to `equal_nan=False`, so one NaN anywhere in a bound variable means the array never equals itself:

```
(2a)  NaN draws, second extraction from the SAME posterior
      RAISED ValueError: This SharedPosterior already binds 'beta' to different draws...
(2a') set_posterior with the SAME NaN posterior: OK
```

The reuse path is not exotic — it is the documented custom-constraint pattern (`constraint_fun` calling `optimizer.extract_response_distribution`, which your own new test uses), and the first `set_posterior` runs `_compile_objective_and_grad()` and then `set_constraints(...)` against the holder it just populated. A NaN in a frozen HSGP deterministic or a masked hierarchical cell therefore turns the first rebind into a spurious "different draws" failure, blaming the caller for something they did not do. `equal_nan=True` is invalid for integer arrays, so branch on `np.issubdtype(values.dtype, np.inexact)`.

### Minor

**16. The coordinate-mismatch message dumps both full label lists.** For a variable over `date` — which is exactly what `frozen_deterministics` binds for `time_varying_intercept`/`time_varying_media` — that is unreadable. A 208-week axis:

```
rolling refit rebind RAISED ValueError, message length = 15855 chars
first 220 chars: Posterior variable 'tv_latent' has date coordinates
[datetime.datetime(2020, 2, 2, 0, 0), datetime.datetime(2020, 2, 9, 0, 0), ...
```

Report the symmetric difference and the counts. Related: a rolling refit on a same-length but *shifted* date window is refused by this check. Defensible for a fixed-shape graph, but it makes the message quality matter more, not less.

**17. A dim labelled on one side and unlabelled on the other silently falls back to positional.** `_aligned_values` skips alignment when `dim not in posterior_da.coords`, and `_coords` is captured from the first posterior only, so the reverse asymmetry exists too:

```
coords present at creation, ABSENT on rebind:
  before: [[2. 4.]]   after rebind without coords (values [2,1] meaning b,a): [[4. 2.]]
coords ABSENT at creation, present on rebind:
  after rebind with reversed labels: [[4. 2.]]   <- labels ignored
```

Both directions reintroduce #4 with no warning, and `set_posterior` explicitly accepts a hand-built `Dataset`, which is precisely how an unlabelled posterior arises (bootstrap/resample loops built from raw arrays). Given the class now goes to the trouble of realigning by label, the one case worth raising on is the one where the two sides disagree about whether labels exist at all — it is unverifiable, so guessing is the wrong default.

**18. Dead local import with a false comment.** `set_posterior` does `from pymc_marketing.pytensor_utils import SharedPosterior` with "Local import to avoid circular import at module load time", while this same patch adds that name to the module-level import. The module-level import is the proof that no cycle exists; leaving the local one will make the next reader preserve a workaround for a problem that is not there.

**19. Stray import in the `merge_inference_data` docstring example.** `from pymc_marketing.pytensor_utils import SharedPosterior, merge_models` in copy-pasteable example code that never uses `SharedPosterior` — looks swept up by the edit to the real import.

**20. A bare `Dataset` replaces every other idata group.** `DataTree.from_dict({"/posterior": ds})` assigned wholesale to `self.idata` drops `constant_data`, `observed_data`, `posterior_predictive`:

```
groups before: ['/constant_data', '/observed_data', '/posterior']
groups after bare Dataset: ['/posterior']
```

Nothing downstream breaks today — inside the optimizer `self.idata` is only read for the construction mask and `_extract_dataset(self.idata, "posterior")`, `BudgetOptimizationResult` holds no idata, and `MMM.optimize_budget` reads `self.data.idata` — but `idata` is a public field and the `Parameters` block should say the rest of the tree is discarded.

**21. Move the rebind test off the corner solution rather than loosening `atol`.** At `total_budget=2.0` both channels sit in the linear regime of the logistic saturation, so the optimum is a vertex and the allocation encodes roughly one bit — *which* channel wins. A 10× error in the rebound `saturation_beta` produces the identical allocation:

```
total_budget= 2.0  beta_scale=[3.0, 0.5] : alloc=[2. 0.]          fun=-12.3528
total_budget=60.0  beta_scale=[3.0, 0.5] : alloc=[41.33 18.67]    fun=-58.6563
total_budget=60.0  beta_scale=[30.0, 0.5]: alloc=[50.04  9.96]    fun=-458.8734
```

So `assert_allclose(rebound.values, expected.values)` cannot see a wrong rebind at all, and `rebound_res.fun == approx(...)` is doing all the work in both halves of the test. At `total_budget=60.0` the optimum is interior, the allocation assertion discriminates again, and the `atol=1e-9` escape hatch from `5743b310` becomes unnecessary.

**22. `transpose` missing on the re-derived mask.** Construction normalises with `align_to_model_coords(...).transpose(*self._budget_dims).astype(bool)`; the re-derivation in `set_posterior` calls `align_to_model_coords` only, and `align_to_model_coords` reindexes without reordering dims. For a model with more than one budget dim whose `channel_contribution` declares them in a different order than `channel_data`, `np.array_equal` then compares a transpose — spurious raise for a non-square shape, scrambled comparison for a square one. [INFERENCE] I could not reach this from `MMM` (`channel_contribution` is emitted over the same layout as `channel_data`), so it only bites the hand-built-model path the class documents. One call to close.

---

The lazy binding is the right call and #5 is properly gone. The remaining theme is the same one as round 1, one layer up: the state machine has more commit points than it has rollback points. 11 is the one I would not merge without; 12–14 are the ones that make the documented contract true.

juanitorduz:
## Review, round 3

Reviewed at `d0a44cd1`.

**Round 2 is closed.** Verified on the repo's own fixtures, at `total_budget=60.0` where the optimum is interior:

```
baseline: [34.25358886 25.74641114] -44.55868952462339
first set_posterior raised: KeyError "No variable named 'saturation_beta'..."
all five attributes restored by identity: True
allocation unchanged after failure: True -44.55868952462339
took first-call path again (holder created): True
objective recompiled: True
  rebound: [44.49396916 15.50603084] -66.83933718274788
  fresh  : [44.49396916 15.50603084] -66.83933718274788
  matches fresh: True
```

So **#11** (rollback), **#12** (the cast moved into `_aligned_values`, write loop now infallible — I could not make `set_value(values, borrow=True)` raise), **#14** (`implied_mask is None` skip), **#15** (`_array_equal`, correct on every dtype the class can hold), **#16**, **#17**, **#18**, **#19**, **#20**, **#21** (the interior optimum above is real) and **#22** are done. **#13** is fixed properly: the reworked atomicity test corrupts the second-bound variable and fails when the round-2 implementation is patched back in, and each of the other reworked tests also fails when its round-3 fix is reverted. Mask-gate behaviour was re-walked across all nine (construction × new posterior) combinations and matches the intended rule everywhere except the one noted in 25 below.

One new Critical, though — and it is the same contract, unguarded on the one path nobody had tested.

### Critical

**23. The *first* `set_posterior` validates nothing.** The first-call path creates an empty `SharedPosterior()` and then compiles against the **new** `idata` (`self.idata = idata` before `_compile_objective_and_grad()`). With an empty holder, `_get_or_create` takes the *create* branch for every variable and simply adopts whatever dims, lengths and coordinate order the incoming posterior happens to have — `_aligned_values` never runs. So every guarantee the method's own docstring makes is delivered only from the second call onwards:

> "Coordinates are matched by label when present, so a reordered `channel` axis is realigned"
> "ValueError: If a variable's non-sample dimensions, lengths or coordinate labels differ from the construction posterior"

Reproduced on `mmm_wrapper`/`dummy_idata` with a posterior that is *identical* except that its `channel` axis is stored in the other order:

```
ordered   channels: ['channel_1', 'channel_2']
reordered channels: ['channel_2', 'channel_1']
reference (ordered)      : [44.49396916 15.50603084] -66.83933718274788
reordered on FIRST  call : [15.50603084 44.49396916] -66.83933718274788
reordered on SECOND call : [44.49396916 15.50603084] -66.83933718274788
first  == reference: False
second == reference: True
```

The budget is silently attributed to the wrong channels, and the same posterior gives a different answer depending on whether it arrives on the first call or a later one. This is exactly the hazard round-1 #4 was about, still live on the entry path.

The shape half is worse, because nothing raises inside the `try`, so the round-3 rollback never fires. With a user-supplied `budgets_to_optimize` (which bypasses the mask gate that would otherwise catch a changed channel axis):

```
--- user mask, first call with a 3-channel posterior ---
bound shapes: {'adstock_alpha': (4, 3), 'saturation_lam': (4, 3), 'saturation_beta': (4, 3)}
allocate RAISED: ValueError | Vectorized input 1 has an incompatible shape in axis 0.
--- user mask, first call with a 1-channel posterior ---
bound shapes: {'adstock_alpha': (4, 1), ...}
allocate RAISED: ValueError | Incompatible vectorized shapes for input 1 and axis 0. [...]
--- SECOND call rejects the same thing, for contrast ---
second call RAISED: ValueError | Posterior variable 'adstock_alpha' has different channel
coordinates: 1 labels given, 2 bound; 0 not bound, 1 missing ('channel_2').
```

`set_posterior` returns normally, keeps the new `idata`, the new holder and the newly compiled objective, and every later `allocate_budget` dies inside pytensor — bricked, with no rollback, because the rollback can only trigger on an exception and there isn't one.

The fix is small and makes the first call structurally identical to the rest: compile against the **construction** posterior, then reach the new one through the already-validated rebind path.

```python
self._shared_posterior = SharedPosterior()
try:
    self._compile_objective_and_grad()
    self.set_constraints(constraints=list(self._constraints.values()))
    self._shared_posterior.set_posterior(_extract_dataset(idata, "posterior"))
    self.idata = idata
except BaseException:
    ...
```

That ordering was checked end-to-end: the reordered first call then matches the reference allocation, the 1-channel posterior raises the labelled `ValueError` and rolls back cleanly, and the existing behaviours still hold (a 1-draw rebind matches a fresh optimizer; a thinned posterior without `channel_contribution` is still accepted with the mask kept). Note it also removes the need for the comment explaining why the new state must be visible while compiling — with this ordering it no longer has to be. Worth a test asserting the *same* posterior gives the same allocation whether it arrives first or second.

### Important

**24. Coordinate labels are compared in a datetime-unit-dependent way, so a saved-and-reloaded posterior is refused.** `_aligned_values` uses `set(new_labels.tolist()) != set(bound_labels.tolist())`, and `.tolist()` on a `datetime64` coord is unit-dependent: `datetime64[us]` yields `datetime.datetime` objects, `datetime64[ns]` yields raw integer nanoseconds. An in-memory `pd.date_range` coord is `[us]`; after a netCDF round trip it is `[ns]` — which is precisely `ModelBuilder.save` (`DataTree.to_netcdf`) followed by `ModelBuilder.load` (`xr.open_datatree`). Rebinding onto a reloaded posterior with a datetime-typed model dim then raises although every label is identical:

```
date dtype: datetime64[us] -> datetime64[ns]
values identical: True True
ValueError: Posterior variable 'tvp' has different date coordinates: 3 labels given, 3 bound;
3 not bound (1578182400000000000, 1578787200000000000, 1579392000000000000),
3 missing (datetime.datetime(2020, 1, 5, 0, 0), ...)
```

String coords survive the round trip, so the stock MMM (`channel`/`country`/`control`) is unaffected; it needs a model dim whose coord is datetime-like — `frozen_deterministics` over `date` for the time-varying components is the realistic case. `.sel` already accepts cross-unit labels (verified both directions), so the whole fix is in how the two label sets are built: `set(pd.Index(new_labels)) != set(pd.Index(bound_labels))`, plus the same normalisation in `_describe_label_mismatch` — which also turns those nanosecond integers back into readable timestamps.

### Minor

**25. The mask gate still treats the all-ones fallback as a claim in one corner, and the comment says it doesn't.** Full matrix (construction × new posterior), measured:

```
A auto+cc [True, False] -> new cc same             PASS
A auto+cc [True, False] -> new cc different        RAISE
A auto+cc [True, False] -> new cc wider            RAISE
A auto+cc [True, False] -> new NO cc               PASS   (round-3 fix)
B auto+fallback [True, True] -> new cc NARROWER    RAISE   <-- odd one out
B auto+fallback [True, True] -> new cc all         PASS
B auto+fallback [True, True] -> new NO cc          PASS
C user [True, False] -> new cc narrower            RAISE  (subset rule)
C user [True, False] -> new cc wider               PASS
C user [True, False] -> new NO cc                  PASS
```

`B1` raises on the same all-ones fallback that the new comment declares to be "a construction default, not a claim about this posterior", and the message tells the user the mask was "auto-detected" from a posterior that never carried `channel_contribution`. Either skip the comparison when the mask came from the fallback (track that alongside `_mask_auto_detected`), or keep the raise and reword the comment and the docstring to say the mask is pinned once chosen, wherever it came from. Related: `test_set_posterior_accepts_a_posterior_without_channel_contribution`'s docstring says "the all-ones fallback is a construction default" while the mask it actually builds and asserts is `[True, False]` — a posterior-derived mask, not the fallback.

**26. `test_shared_posterior_coordinate_mismatch_message_stays_short` does not reach the truncation it exists to pin.** It shifts the 208 weekly dates by exactly one week, so 207 labels still overlap and the message lists one label per side — 189 characters, and the `limit=5` branch in `_describe_label_mismatch` is never taken. The `< 300` bound therefore holds for a reason unrelated to the feature; the worst case the test claims to guard, a wholly disjoint 208-label axis, produces 503 characters and would fail that very assertion. Shift by a day instead of a week (all 208 labels differ), assert `"..."` appears, and set the bound to the real ceiling.

---

23 is the merge blocker and it is a five-line reordering. 24 is the one I would fix in the same pass, since it turns a legitimate rebind into an error on the save/load path. Everything else from rounds 1 and 2 is genuinely closed, and the reworked tests now have teeth — the atomicity test in particular.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -2098,7 +2106,131 @@ def extract_response_distribution(self, response_variable: str) -> XTensorVariab
...
+            self.idata,
+            self._objective_and_grad,
+            self._constraints,
+            self._compiled_constraints,
         )
+        self._shared_posterior = SharedPosterior()
+        self.idata = idata
+        try:
+            self._compile_objective_and_grad()
+            self.set_constraints(constraints=list(self._constraints.values()))
```
juanitorduz, line 2224: **Critical (finding 23): the first call validates nothing.** The holder is empty here, so `SharedPosterior._get_or_create` takes the *create* branch for every variable and adopts whatever dims, lengths and coordinate order the incoming posterior has — `_aligned_values` never runs. None of the guarantees in this method's own docstring (`Raises: ValueError - If a variable's non-sample dimensions, lengths or coordinate labels differ from the construction posterior`) hold on the first call.

A posterior that is identical except that its `channel` axis is stored in the other order:

```
reference (ordered)      : [44.49396916 15.50603084] -66.83933718274788
reordered on FIRST  call : [15.50603084 44.49396916] -66.83933718274788
reordered on SECOND call : [44.49396916 15.50603084] -66.83933718274788
```

Same posterior, transposed allocation, no error — and the answer depends on whether it arrives first or second. The shape half is worse: with a user-supplied `budgets_to_optimize` (which bypasses the mask gate above), a 1- or 3-channel posterior is bound happily, `set_posterior` returns normally, and every later `allocate_budget` dies inside pytensor — the rollback cannot fire because nothing raises.

Fix: compile against the **construction** posterior, then let the new draws arrive through the same validated rebind path as every later call. The comment about the new state having to be visible while compiling then no longer applies.

Verified with this exact body: the reordered first call now matches the reference allocation (`first == reference: True`), a 1-channel posterior raises `Posterior variable 'adstock_alpha' has different channel coordinates: 1 labels given, 2 bound...` and rolls back (`rolled back: True`), a 1-draw rebind still matches a fresh optimizer, and a thinned posterior without `channel_contribution` is still accepted with the mask kept.

```suggestion
        # First rebind: bind the draws through shared variables and compile once
        # against them; later calls are set_value only.  The compile runs on the
        # posterior the optimizer was built with, and the new draws then arrive
        # through the same validated rebind path as every later call -- so a
        # reordered axis is realigned and a resized one refused here exactly as
        # it would be on the second call.  Anything that raises rolls every
        # piece back, so a failed first call leaves the constant-folded graphs
        # and the previous idata in place, and the next call retries this path
        # rather than rebinding variables the compiled objective never reads.
        previous = (
            self._shared_posterior,
            self.idata,
            self._objective_and_grad,
            self._constraints,
            self._compiled_constraints,
        )
        self._shared_posterior = SharedPosterior()
        try:
            self._compile_objective_and_grad()
            self.set_constraints(constraints=list(self._constraints.values()))
            self._shared_posterior.set_posterior(_extract_dataset(idata, "posterior"))
            self.idata = idata
```

Worth a test asserting the same posterior gives the same allocation whether it arrives on the first call or a later one.

pymc_marketing/pytensor_utils.py
```diff
@@ -243,12 +246,236 @@ def validate_unique_value_vars(model: Model) -> None:
...
+                        if bound_labels is not None
+                        else f"has {dim} coordinates, but was bound without them"
+                    )
+                    + ". Label both sides or neither; positions cannot be "
+                    "checked against labels."
+                )
+            if bound_labels is None:
+                continue
+            new_labels = np.asarray(posterior_da.coords[dim].values)
+            if set(new_labels.tolist()) != set(bound_labels.tolist()):
```
juanitorduz, line 383: **Important (finding 24): this comparison is datetime-unit-dependent, so a saved-and-reloaded posterior is refused.** `.tolist()` on a `datetime64[us]` coord yields `datetime.datetime` objects; on `datetime64[ns]` it yields raw integer nanoseconds. An in-memory `pd.date_range` coord is `[us]`, but after a netCDF round trip it is `[ns]` — i.e. `ModelBuilder.save` (`DataTree.to_netcdf`) followed by `ModelBuilder.load` (`xr.open_datatree`). Rebinding onto a reloaded posterior with a datetime-typed model dim then raises although every label is identical:

```
date dtype: datetime64[us] -> datetime64[ns]
values identical: True True
ValueError: Posterior variable 'tvp' has different date coordinates: 3 labels given, 3 bound;
3 not bound (1578182400000000000, 1578787200000000000, 1579392000000000000),
3 missing (datetime.datetime(2020, 1, 5, 0, 0), ...)
```

String coords survive the round trip, so `channel`/`country`/`control` are unaffected; this needs a model dim whose coord is datetime-like — `frozen_deterministics` over `date` for the time-varying components is the realistic case.

`pd.Index` normalises the units, and `.sel` already accepts cross-unit labels (verified: `set(pd.Index(us)) == set(pd.Index(ns)) -> True`, `da.sel(date=pd.Index(us))` on an `ns` index returns `[0. 1. 2.]`), so the whole fix is in how the two label sets are built:

```suggestion
            if set(pd.Index(new_labels)) != set(pd.Index(bound_labels)):
```

pymc_marketing/pytensor_utils.py
```diff
@@ -243,12 +246,236 @@ def validate_unique_value_vars(model: Model) -> None:
...
+        b.dtype, np.inexact
+    )
+    return bool(np.array_equal(a, b, equal_nan=equal_nan))
+
+
+def _describe_label_mismatch(
+    new_labels: np.ndarray, bound_labels: np.ndarray, limit: int = 5
+) -> str:
+    """Summarise how two coordinate label sets differ without printing both in full."""
+    new_set, bound_set = set(new_labels.tolist()), set(bound_labels.tolist())
```
juanitorduz, line 270: Same normalisation is needed here, or the message that reports the mismatch prints raw nanosecond integers on one side and `datetime.datetime(...)` on the other (see the output in the comment on line 383). `pd.Index` also makes the labels render as readable timestamps.

```suggestion
    new_set, bound_set = set(pd.Index(new_labels)), set(pd.Index(bound_labels))
```

tests/test_pytensor_utils.py
```diff
@@ -712,3 +713,343 @@ def conditioned_slope(graph):
...
+    )
+    shared_posterior = SharedPosterior()
+    extract_response_distribution(
+        model, posterior, "out", shared_posterior=shared_posterior
+    )
+
+    shifted = posterior.assign_coords(date=list(dates + pd.Timedelta(weeks=1)))
+    with pytest.raises(ValueError, match=r"208 labels given, 208 bound") as info:
+        shared_posterior.set_posterior(shifted)
+    assert len(str(info.value)) < 300
```
juanitorduz, line 917: **Minor (finding 26): this test never reaches the truncation it exists to pin.** Shifting 208 weekly dates by exactly one week leaves 207 labels overlapping, so the message lists one label per side — 189 characters — and the `limit=5` branch in `_describe_label_mismatch` is never taken. The `< 300` bound therefore holds for a reason unrelated to the feature. The worst case the test claims to guard, a wholly disjoint axis, produces 503 characters and would fail this very assertion.

Shift by a day so all 208 labels differ, then assert the truncation marker and a bound that reflects the real ceiling:

```suggestion
    shifted = posterior.assign_coords(date=list(dates + pd.Timedelta(days=1)))
    with pytest.raises(ValueError, match=r"208 labels given, 208 bound") as info:
        shared_posterior.set_posterior(shifted)
    assert "..." in str(info.value)
    assert len(str(info.value)) < 600
```

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -2098,7 +2106,131 @@ def extract_response_distribution(self, response_variable: str) -> XTensorVariab
...
+        # The mask fixes the decision vector the graphs were compiled for, so
+        # it cannot follow the posterior; it can only be checked against it.
+        current_mask = cast(DataArray, self.budgets_to_optimize)
+        if self._mask_auto_detected:
+            implied_mask = self._posterior_channel_mask(idata)
+            # A posterior without channel_contribution (thinned to the free RVs
+            # the graphs read) implies nothing about the mask, so there is
+            # nothing to compare; the all-ones fallback is a construction
+            # default, not a claim about this posterior.
+            if implied_mask is not None:
```
juanitorduz, line 2180: **Minor (finding 25): the code does not hold to this comment in one corner.** When the *construction* posterior had no `channel_contribution`, the mask in use **is** the all-ones fallback — and if the new posterior then carries `channel_contribution` implying something narrower, that same fallback is compared as if it were a claim and the call is refused, with a message telling the user the mask was "auto-detected" from a posterior that never carried the variable. Full matrix, measured:

```
A auto+cc [True, False] -> new cc same             PASS
A auto+cc [True, False] -> new cc different        RAISE
A auto+cc [True, False] -> new cc wider            RAISE
A auto+cc [True, False] -> new NO cc               PASS   (this fix)
B auto+fallback [True, True] -> new cc NARROWER    RAISE   <-- odd one out
B auto+fallback [True, True] -> new cc all         PASS
B auto+fallback [True, True] -> new NO cc          PASS
C user [True, False] -> new cc narrower            RAISE  (subset rule)
C user [True, False] -> new cc wider               PASS
C user [True, False] -> new NO cc                  PASS
```

Either track "the mask came from the fallback" alongside `_mask_auto_detected` and skip the comparison in that case too, or keep the raise and reword the comment plus the `Raises` docstring to say the mask is pinned once chosen, wherever it came from. Related: `test_set_posterior_accepts_a_posterior_without_channel_contribution`'s docstring says "the all-ones fallback is a construction default", but the mask it builds and asserts is `[True, False]` — posterior-derived, not the fallback.

juanitorduz (changes requested):
Line-anchored follow-up to my round-3 comment, with committable suggestions.

Round 2 is closed and verified — the rollback restores all five attributes by identity, the allocation is unchanged after a failed call, and the next call retries the first-call path and matches a fresh optimizer exactly. The reworked tests have teeth: each fails when its round-3 fix is patched back to the round-2 behaviour.

What is left, in priority order:

1. **`budget_optimizer.py`, the first-call block** — the first `set_posterior` binds straight from the incoming posterior with an empty holder, so none of the dim/length/label checks the docstring promises actually run on that call. A reordered `channel` axis is silently transposed into the allocation; with a user-supplied mask, a resized axis bricks the optimizer with no rollback because nothing raises. The suggested four-line reordering (compile against the construction posterior, then rebind through the validated path) fixes both; I ran it end-to-end and the existing behaviours still hold.
2. **`pytensor_utils.py:383` and `:270`** — coordinate labels are compared through `.tolist()`, which is datetime-unit-dependent, so an identical posterior that went through `ModelBuilder.save`/`load` (`[us]` → `[ns]`) is refused. `pd.Index` on both sides fixes it and makes the error message readable.
3. Two minor ones: the mask-gate comment does not match behaviour for the all-ones-fallback case, and the "message stays short" test never reaches the truncation it exists to pin.

1 is the merge blocker; 2 I would fix in the same pass since it turns a legitimate rebind into an error.

juanitorduz:
Hola @PabloRoque ! I left a last review with the agent. Seems we are very close :) . Thanks!

PabloRoque:
Thanks. I reproduced all four before changing anything, and #23 came out exactly as you showed: the reordered posterior gave a transposed allocation on the first call and the right one on the second, and a 1-channel posterior with a user mask bound cleanly and then died in pytensor.

Rather than patch again, I stepped back. Three rounds in, most of the diff was corner-case machinery around a small feature, and each fix was adding a branch the next round then probed. So this commit fixes #23, #24 and #26 as you suggested, and replaces the coordinate handling and the mask gate with a contract that fits in the docstring. That is a deliberate change to one behaviour you asked for in round 1, so I want to flag it clearly.

**The contract.** `set_posterior` accepts a posterior when every bound variable is present with its bound dtype; every non-sample dim has the same name and the same *set* of labels, compared by value; a dim bound without labels arrives without labels; the draw count is free. Anything else raises before anything is written. The budget mask is fixed at construction and is never compared or re-derived by a rebind.

**#23, first call validated.** Your four-line reordering, as suggested: compile against the construction posterior, then rebind through the validated path, then commit `idata`. Test: the same reordered posterior gives the same allocation whether it arrives first or second, and a 1-channel posterior with a user mask raises and rolls back all five attributes with the construction allocation unchanged. I confirmed the test fails against the round-2 ordering.

**#24, datetime units.** `_coords` now stores `pd.Index` objects and the comparison is `bound.sort_values().equals(given.sort_values())`, so `[us]` and `[ns]` compare equal and the message renders timestamps. Test: a `[us]`-bound variable rebinds from `[ns]`, and a one-day-off `[ns]` axis is refused with both dates readable in the message.

**#26.** Day shift, as suggested.

**#16, #17, #24 together, simplified.** `_describe_label_mismatch` is gone. The mismatch message is `N not bound, M missing (e.g. [...] / [...])` with three examples per side from `Index.difference`, 297 characters for a wholly disjoint 208-week axis. The labelled-on-one-side check is one condition with one message for both directions. Duplicate labels still fail, since they change the sorted index.

**#3, #14, #22, #25, the mask gate, removed.** This is the change that needs your agreement. The round-1 finding was real: an auto-detected mask does not follow the posterior. But the gate that refuses a posterior implying a different mask has needed a carve-out in every round since (thinned posteriors, the transpose, the fallback case), and each carve-out was a new place for the code and its comment to disagree. The mask is now simply pinned: the docstring says a posterior that would have auto-detected a different mask is solved on the existing decision vector, and to build a new optimizer to change it. That is what a caller who wants the other decision problem has to do anyway, and `budgets_to_optimize` is a public field they can read. A user-supplied mask is still validated against the new posterior the way construction validates it. Test: an optimizer built with mask `[True, False]` accepts a posterior whose contributions are swapped, keeps its mask, allocates the whole budget to the first cell, and a fresh optimizer on that posterior auto-detects `[False, True]`. The fallback case is a second test: built without `channel_contribution`, the all-ones mask stays all-ones when a narrower posterior arrives. `_mask_from_fallback` is gone.

**Net.** Source is 90 lines smaller than the round-3 patch would have been; `SharedPosterior` is 182 lines including docstrings; two private attributes and one helper removed; three mask tests replaced by one. Everything you verified in rounds 2 and 3 still holds: I reran both repro scripts under the rewrite and every finding stays closed, with the mask ones now closed by construction rather than by a branch.

If you would rather keep the refusal, the fallback is one flag and one comparison on top of this, and I will add it.

PabloRoque:
Two small follow-ups on top of `1d8d2b09`, both coverage and contract rather than behaviour.

**Multi-dim alignment is now tested.** The deleted dim-order test only ever bound `beta` in the order it was declared; the permutation was in `channel_contribution`, which is a deterministic and never bound. So `transpose(*dims)` on a variable with more than one non-sample dim was untested before as well. `test_shared_posterior_realigns_permuted_multi_dim_variables` binds `beta` as `(country, channel)`, rebinds from a posterior stored as `(channel, country)` with the countries reversed, and checks the compiled output. It fails with a shape error when the transpose is removed.

**Mixed-type labels raise ValueError, not TypeError.** The label check is now a length comparison plus `Index.difference`, so it never sorts. Duplicates still fail on length, and a `["a", 1]` axis compares by set and refuses a changed label with the documented ValueError. Tested.

pymc_marketing/mmm/budget_optimizer.py
```diff
@@ -2098,8 +2106,118 @@ def extract_response_distribution(self, response_variable: str) -> XTensorVariab
...
+
+        # The mask is fixed at construction: it defines the decision vector the
+        # graphs were compiled for, so a rebind never re-derives it.  A
+        # user-supplied mask is still checked against the new posterior the way
+        # construction checks it, so it cannot optimize a cell the new draws
+        # carry no information about.
+        if not self._mask_auto_detected:
+            self._validate_mask_against_posterior(
+                cast(DataArray, self.budgets_to_optimize), idata
+            )
```
juanitorduz, line 2182: **On the mask reversal: agreed on dropping the *raise*, not on dropping all *signal*.** Two things, both small.

**(a) The `if not self._mask_auto_detected:` guard makes the optimizer accept, on rebind, exactly the configuration construction refuses when a caller spells it out.** `_validate_mask_against_posterior` is a superset check — "do not optimize a cell the posterior has no information about" — so running it unconditionally does not re-derive anything and does not undo the pinning: a pinned mask that is *narrower* than the new posterior implies still passes. All it catches is the case where the pinned mask optimizes cells the new draws say are dead. Today that case is silent and degenerate:

```
construction posterior informs both channels  -> auto mask [True, True]
rebound posterior: channel_2 provably dead (saturation_beta=[1.0, 0.0])
pinned  at total_budget=5000: [2500.0, 2500.0]  success=True  fun=-30.0   (grad on channel_2 == 0.)
fresh   at total_budget=5000: [5000.0,    0.0]  success=True  fun=-30.0
```

Half the budget parked on a dead channel, reported as success. Pass that same mask explicitly to the constructor and you get `ValidationError: budgets_to_optimize mask contains True values at coordinates where the model has no information.` Removing the branch is strictly less code than what is here now:

```suggestion
        # The mask is fixed at construction: it defines the decision vector the
        # graphs were compiled for, so a rebind never re-derives it.  It is
        # still checked against the new posterior the way construction checks
        # it, so a pinned mask cannot optimize a cell the new draws carry no
        # information about -- the configuration construction refuses outright.
        self._validate_mask_against_posterior(
            cast(DataArray, self.budgets_to_optimize), idata
        )
```

**(b) The other direction — pinned mask *narrower* than the new posterior implies — needs a warning.** This is the one that costs money in the loop this PR exists for, and nothing in the code, the result object or the docs surfaces it:

```
construction mask: [True, False]      (channel_2 uninformed at build time)
rebound onto a posterior where both channels are informed and channel_2 is the better buy:
  set_posterior raised nothing; mask still [True, False]
  rebound : [60. 0.]                    fun   -3.74  success True
  fresh   : [22.70856378 37.29143622]   fun -123.40  success True
```

33x worse objective, `success=True`, and `BudgetOptimizationResult` carries `budgets / scipy_result / optimized_vars / spend_var_names / callback_info` — no mask — so a caller iterating over posteriors has nothing to look at. `MMM.optimize_budget` discards its allocator entirely, so `optimizer.budgets_to_optimize` is not reachable there either. A documented default is fine for a value a user chose; this one was inferred from draws the optimizer no longer holds.

A `UserWarning` is enough and needs none of the carve-outs the old gate accumulated: `_posterior_channel_mask(idata)` returning `None` means there is nothing to compare (thinned posteriors stay silent for free), and a warning states a fact rather than a policy, so the fallback-vs-claim wording problem from finding 25 disappears too. Something like, after (a):

```python
implied = self._posterior_channel_mask(idata)
if self._mask_auto_detected and implied is not None:
    implied = (
        align_to_model_coords(implied, self._budget_coords, label="budgets_to_optimize")
        .transpose(*self._budget_dims)
        .astype(bool)
    )
    if not np.array_equal(implied.values, current_mask.values):
        warnings.warn(
            "The new posterior has information about cells the pinned "
            f"budgets_to_optimize excludes ({current_mask.values.tolist()} in use, "
            f"{implied.values.tolist()} implied). Those cells stay out of the "
            "decision vector; build a new optimizer to include them.",
            UserWarning,
            stacklevel=2,
        )
```

With (a) + (b) I am happy with the pinned-mask contract: the raise is gone, no legitimate rebind is blocked, and neither direction is silent.

pymc_marketing/pytensor_utils.py
```diff
@@ -243,12 +246,214 @@ def validate_unique_value_vars(model: Model) -> None:
...
+            # length check catches duplicates, the set check is order-free and
+            # does not need the labels to be sortable (a mixed-type object axis
+            # would raise TypeError from sort_values).
+            if len(bound) != len(given) or not bound.difference(given).empty:
+                extra, missing = given.difference(bound), bound.difference(given)
+                raise ValueError(
+                    f"Posterior variable {name!r} has different {dim} labels: "
+                    f"{len(extra)} not bound, {len(missing)} missing "
+                    f"(e.g. {list(extra[:3])} / {list(missing[:3])})."
+                )
```
juanitorduz, line 367: **Minor: the comment is half true, and two label shapes escape as pandas errors instead of the documented `ValueError`.**

"duplicates change the length and fail it" holds only when *one* side is duplicated. When both sides carry the same duplicates, `sort_values().equals(...)` returns `True`, the guard is bypassed, and `.sel({dim: bound})` hits a non-unique index:

```
[dup on both sides]  InvalidIndexError: Reindexing only valid with uniquely valued Index objects
[dup on given only]  ValueError: Posterior variable 'beta' has different channel labels: 0 not bound, 1 missing (e.g. [] / ['b']).
```

Note the second message is also slightly wrong — it says `0 not bound` although the given axis carries an extra `'a'` — because `Index.difference` de-duplicates.

Separately, sorting is new in round 4 and `Index.sort_values()` raises on an unsortable object-dtype coord, so a case round 3 accepted now surfaces a pandas error:

```
[mixed-object] TypeError: '<' not supported between instances of 'int' and 'str'
```

Both are contained — the raise happens in the validate phase, so the binding stays untouched (confirmed) — and both need a fairly pathological coord. But `set_posterior`'s `Raises` section promises `ValueError` for label problems. An up-front `if not bound.is_unique or not given.is_unique:` raising the same `ValueError`, and either a `try` around the sort or comparing through `Index.symmetric_difference`/`get_indexer` instead of sorting, makes the comment and the docstring true.

tests/mmm/test_budget_optimizer.py
```diff
@@ -1601,3 +1601,346 @@ def test_spend_var_allocations_excludes_levers():
...
+
+
+def test_set_posterior_accepts_a_posterior_without_channel_contribution(
+    mmm_wrapper, dummy_idata
+):
+    """A posterior thinned to the free RVs the graphs read is accepted.
+
+    Without ``channel_contribution`` the new posterior implies nothing about
+    the auto-detected mask, so there is nothing to compare, and the
+    posterior-derived mask in use is kept.
```
juanitorduz, line 1863: Nit: this rationale is vestigial — round 4 removed the comparison, so an auto-detected mask is never compared against the new posterior, `channel_contribution` present or not. The assertion itself is worth keeping; only the stated reason will send the next reader looking for a comparison in `set_posterior` that is not there.

```suggestion
    The graphs read only the free RVs, so ``SharedPosterior`` never needs
    ``channel_contribution``, and the mask is pinned as on any other rebind.
    """
```

juanitorduz (changes requested):
## Final review — `1d8d2b09`

The simplification was the right instinct. `SharedPosterior` reads as one contract now instead of four rounds of sediment, and I verified the three fixes rather than taking them on trust.

**#23 closed.** The reordered posterior that produced a transposed allocation on the first call now matches the reference exactly, and a resized axis raises the labelled error and rolls back all five attributes:

```
[23] reference          : [44.49396916 15.50603084]
[23] reordered FIRST    : [44.49396916 15.50603084]
[23] first == reference : True
[23] 1-channel raised   : ValueError Posterior variable 'adstock_alpha' has different channel labels: 0 not bound, 1 ...
[23] rolled back        : True
```

**#24 closed properly** — not just "does not raise": a real `to_netcdf`/`open_dataset` `[us]`→`[ns]` round trip rebinds, the realignment across units is value-correct, a one-day-off axis is refused with `Timestamp`s rendered rather than nanosecond integers, and the test fails against the round-3 implementation with exactly the message from the original report. **#26 closed** — the week shift produced `1 not bound, 1 missing` and never reached the truncation branch; the day shift yields 208/208 and a 297-character message. Atomicity still holds with a label failure on the second-bound variable, and the write loop is still infallible. `.coords` → `.indexes` is symmetric (both binding and rebinding go through `_posterior_sample_major`), so the stricter refusal only reaches artificial posteriors. `tests/test_pytensor_utils.py` + `tests/mmm/test_budget_optimizer.py`: **84 passed**. No dead code left behind by the gate removal; the user-mask branch is behaviourally identical to construction and mutates nothing when it raises.

### On the mask reversal, since you asked

Agreed on removing the **raise** — it had accumulated a carve-out per round and each one was a new place for code and comment to drift. Not agreed on removing all **signal**. Two small things, both in the inline comment on the mask block:

**(a)** The `if not self._mask_auto_detected:` guard means a rebind now accepts exactly the configuration construction refuses when a caller spells the mask out. Running `_validate_mask_against_posterior` unconditionally is *less* code than what is there, does not re-derive anything, and does not undo the pinning — it is a superset check, so a pinned mask narrower than implied still passes. Without it: `[2500.0, 2500.0]`, `success=True`, zero gradient on a channel the rebound draws say is dead.

**(b)** The narrower direction deserves a `UserWarning`. It is the direction that costs money in the loop this PR exists for:

```
construction mask [True, False]; rebound onto a posterior informing both channels
  rebound : [60. 0.]                    fun   -3.74  success True
  fresh   : [22.70856378 37.29143622]   fun -123.40  success True
```

33x worse objective, reported as success, and nothing surfaces it: `BudgetOptimizationResult` carries no mask, and `MMM.optimize_budget` discards its allocator, so `optimizer.budgets_to_optimize` is not even reachable on that path. A documented default is fine for a value the user chose; this one was inferred from draws the optimizer no longer holds. A warning needs none of the old carve-outs — `_posterior_channel_mask` returning `None` means there is nothing to compare, so thinned posteriors stay silent for free, and a warning states a fact rather than a policy, which also dissolves the finding-25 wording problem.

### Nits

- Duplicate labels present on both sides bypass the equality guard and escape as `InvalidIndexError`; an unsortable object-dtype coord escapes as `TypeError` — both contradict the `Raises` section, both contained to the validate phase. Comment on `_aligned_values`.
- `test_set_posterior_accepts_a_posterior_without_channel_contribution`'s docstring still explains the outcome through the comparison round 4 deleted.
- The `budgets_to_optimize` field description, its class-docstring entry and the module `Notes` all present auto-detection as the rule without qualification; after this patch it holds at construction only. One sentence on the field (it is resolved once at construction and `set_posterior` keeps it) closes the gap for the reader who is looking at the mask rather than at `set_posterior`. Couldn't anchor this one — the lines are outside the diff.

(a) and (b), and I am happy to merge. Everything else here is optional.

juanitorduz (approved):
Minor comments from the agent :) 

Feel free to merge after

PabloRoque:
Good catch, reproduced: a variable bound with a `["a", "a"]` axis accepted a rebind from `["a", "b"]` and evaluated to `[[10., 10.]]`, column `a` twice.

Fixed at bind time as suggested: `_get_or_create` refuses a non-unique index with `ValueError: Posterior variable 'beta' has duplicated k labels: ['a'].`, which makes the "set of labels" contract literal. With the bound side unique, equal lengths plus an empty `Index.difference` is a genuine set comparison, and a duplicate on the rebind side shows up as a missing label and is refused by the existing check. The constant path is untouched: it never realigns by label. All three cases are in `test_shared_posterior_refuses_duplicated_labels_at_bind_time`.

PR #3003, opened 2026-09-10
docs: point budget docstrings at sample_response_distribution

anevolbap:
The `samples` argument of `BudgetPlots.contribution_over_time` and `BudgetSummaryFactory.contribution_over_time` is documented as the output of `allocate_budget_to_maximize_response(...)`. That method is not in the package. The current name is `sample_response_distribution`, which samples `channel_contribution_original_scale`, the variable both docstrings require.

Both files already use the current name in the `allocation_roas` docstring above, so this aligns the two remaining lines. The plotting one also loses its `mmm.` receiver, since the method is defined on `BudgetOptimizerWrapper`, not on `MMM`.

Docstrings only, no behaviour change.

### DIFF
--- a/pymc_marketing/mmm/plotting/budget.py
+++ b/pymc_marketing/mmm/plotting/budget.py
@@ -145,7 +145,7 @@ class BudgetPlots:
         Parameters
         ----------
         samples : xr.Dataset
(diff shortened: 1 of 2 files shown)


juanitorduz (approved):
Thanks!

PR #3010, opened 2026-09-12
fix(terms): complete arithmetic operators on Sum and Product

### LINKED ISSUE #3001: terms: Product/Sum operator completes (Product has no __add__)
- `-Intercept("a") + 3 * Parameter("b")` raises `TypeError: unsupported operand type(s) for +: 'Product' and 'Product'` — `ModelTerm.__neg__` returns a `Product`, and `Product`/`Sum` lack the additive `__add__`/`__radd__` pair.

The terms module advertises these compositions as serializable recipes, so operator completeness matters once recipes are round-tripping through `model_config`. Found during the serialization review pass on #2987 (not a regression — pre-existing on `main`).

anevolbap:
Fixes #3001

`Product` defined only `__mul__` and `__rmul__`. So `product + x` worked only when `x` was a `ModelTerm` or a `Sum`, where the reflected `__radd__` on those classes handled it, and raised `TypeError` for another `Product`, an `int` or a `float`. `x + product` raised for every non-term `x`. `ModelTerm.__neg__` returns a `Product`, so `-Intercept("a") + 3 * Parameter("b")` hit the first case with `unsupported operand type(s) for +: 'Product' and 'Product'`. `Sum` and `Product` also had no `__neg__`, `__sub__` or `__rsub__`, so `term - product`, `product - product` and `term - sum` raised too.

The operators now live in one `_TermOps` mixin shared by `ModelTerm`, `Sum` and `Product`. A `Sum` operand of `+` is flattened on either side, and `0 + x` short-circuits so builtin `sum()` works over products and sums.

Shape changes to expressions that already worked: `term + sum` and `sum - term` now give a flat `Sum` instead of a nested one. `build_param` recurses into nested sums, so both shapes build the same sum.

Not a regression: the file has had this shape since `pymc_marketing.terms` landed in #2965.

### DIFF
--- a/pymc_marketing/terms.py
+++ b/pymc_marketing/terms.py
@@ -431,6 +431,16 @@ class Product:
             **get_coords(self.right, ds),
         }
 
+    def __add__(self, other: Any) -> Sum:
+        """Compose additively with another term or product."""
+        return Sum([self, other])
+
+    def __radd__(self, other: Any) -> Sum | Product:
+        """Compose additively from the left; ``0 + product`` returns the product."""
+        if isinstance(other, int) and other == 0:
+            return self
+        return Sum([other, self])
+
     def __mul__(self, other: Any) -> Product:
         """Compose multiplicatively with another term."""
         return Product(self, other)
--- a/pymc_marketing/terms.py
+++ b/pymc_marketing/terms.py
@@ -432,7 +432,9 @@ class Product:
         }
 
     def __add__(self, other: Any) -> Sum:
-        """Compose additively with another term or product."""
+        """Compose additively with another term, product or sum."""
+        if isinstance(other, Sum):
+            return Sum([self, *other.terms])
         return Sum([self, other])
 
     def __radd__(self, other: Any) -> Sum | Product:
--- a/tests/test_terms.py
+++ b/tests/test_terms.py
@@ -461,6 +461,14 @@ def test_product_add():
     assert result.terms[1] == Intercept(name="b")
 
 
+def test_product_add_sum_flattens():
+    p = Product(Intercept(name="a"), 2.0)
+    result = p + Sum([Intercept(name="b"), Intercept(name="c")])
+    assert isinstance(result, Sum)
+    assert len(result.terms) == 3
+    assert result.terms[0] == p
+
+
 def test_product_radd_int():
     p = Product(Intercept(name="a"), 2.0)
     result = 3 + p
--- a/tests/test_terms.py
+++ b/tests/test_terms.py
@@ -453,6 +453,35 @@ def test_product_rmul():
     assert isinstance(result, Product)
 
 
+def test_product_add():
+    p = Product(Intercept(name="a"), 2.0)
+    result = p + Intercept(name="b")
+    assert isinstance(result, Sum)
+    assert result.terms[0] == p
+    assert result.terms[1] == Intercept(name="b")
+
+
+def test_product_radd_int():
+    p = Product(Intercept(name="a"), 2.0)
+    result = 3 + p
+    assert isinstance(result, Sum)
+    assert result.terms[0] == 3
+    assert result.terms[1] == p
+
+
+def test_product_radd_zero():
+    p = Product(Intercept(name="a"), 2.0)
+    result = 0 + p
+    assert result == p
+
+
+def test_product_add_product():
+    result = -Intercept(name="a") + 3 * Parameter(name="b", prior=Prior("Normal"))
+    assert isinstance(result, Sum)
+    assert len(result.terms) == 2
+    assert all(isinstance(term, Product) for term in result.terms)
+
+
 def test_sum_rmul():
     s = Sum([Intercept(name="a")])
     result = 3 * s


juanitorduz (commented):
## Review

**Verdict: correct, verified, approve-with-suggestions.** The diagnosis is right, the fix is minimal, and the second commit is a real catch. One substantive push: the `__neg__` question raised at the bottom of the description should be folded in, not deferred — and doing so cleanly argues for collapsing the three duplicated operator sets into one mixin.

### Verification (worktree at `561105b5`, probed against `main`)

Operator matrix, with `p`/`q` = `Product`, `s` = `Sum`, `a`/`b` = `ModelTerm`:

| expr | main | this PR |
|---|---|---|
| `p + q` | `TypeError` | `Sum[Prod, Prod]` |
| `3 + p` / `p + 3` | `TypeError` | `Sum[int, Prod]` / `Sum[Prod, int]` |
| `0 + p`, `sum([p, q])` | `TypeError` | short-circuits / `Sum[Prod, Prod]` |
| `-a + 3 * b` (#3001) | `TypeError` | `Sum[Prod, Prod]` |
| `p + s` | `Sum[Prod, a, b]` | `Sum[Prod, a, b]` ← **unchanged** |

End-to-end `collect_coords` → `register_data` → `build_param` on `-Intercept("a") + 3 * Parameter("b")` raises on `main` and returns an `XTensorVariable` on this branch. The user-facing bug is genuinely closed.

The second commit is validated: without the `isinstance(other, Sum)` branch, `p + s` would have nested. Flat before, flat after. Good self-review.

### Main point: fold in `__neg__` / `__sub__` / `__rsub__`

Deferring ships this asymmetry:

```
p - b   -> Sum[Prod, Prod]     # works (ModelTerm.__rsub__ catches it)
b - p   -> TypeError: bad operand type for unary -: 'Product'
p - q   -> TypeError
b - s   -> TypeError: bad operand type for unary -: 'Sum'
```

Same root cause as #3001 — the operator protocol lives only on `ModelTerm`, so the containers fall through — and the same confusing error class. Shipping `+` but not `-` means users hit `unsupported operand` on the very next expression they write. Not scope creep; it is the rest of the bug.

### Secondary: this is the third copy of the operator set

`__mul__`/`__rmul__` now appear three times verbatim; `__add__`/`__radd__` twice. They have already drifted — after this PR, `Product.__add__` flattens a `Sum` operand but `ModelTerm.__add__` does not:

```
p + s  -> Sum[Prod, a, b]        # flat
a + s  -> Sum[a, Sum[a, b]]      # nested
```

Functionally equivalent (`build_param` recurses), but expression shape then depends on which side is a container, which makes structural assertions and reprs unpredictable. Adding `__neg__`/`__sub__`/`__rsub__` to two more classes would add six more copies and two more drift opportunities.

I prototyped the alternative and ran it: a `TermOps` mixin on `ModelTerm`, `Sum`, `Product`, with one `_addends(value)` helper that flattens `Sum`. **All 71 tests in `tests/test_terms.py` (including the five new ones here) pass unmodified**, every `TypeError` above disappears, and `a + s` becomes flat like `p + s`:

```python
def _addends(value: Any) -> list[Any]:
    """Return the addends of ``value``, flattening a ``Sum``."""
    return list(value.terms) if isinstance(value, Sum) else [value]


class TermOps:
    def __add__(self, other: Any) -> Sum:
        return Sum(_addends(self) + _addends(other))

    def __radd__(self, other: Any) -> Sum | TermOps:
        if isinstance(other, int) and other == 0:
            return self
        return Sum(_addends(other) + _addends(self))

    def __sub__(self, other: Any) -> Sum:
        return Sum(_addends(self) + [-other])

    def __rsub__(self, other: Any) -> Sum:
        return Sum(_addends(other) + [-self])

    def __mul__(self, other: Any) -> Product:
        return Product(self, other)

    def __rmul__(self, other: Any) -> Product:
        return Product(other, self)

    def __neg__(self) -> Product:
        return Product(-1, self)
```

Net effect on `terms.py`: −29 lines of duplication, three classes with one arithmetic contract. Keeping this PR surgical is defensible — but then the follow-up should be the mixin refactor rather than six more hand-copied methods.

### Nits

- `test_product_add_product` asserts only `all(isinstance(term, Product))`. It would pass if the operands were swapped or wrapped. Assert the operands instead: `result.terms == [Product(-1, Intercept(name="a")), Product(3, Parameter(...))]`.
- `test_product_radd_zero` uses `result == p`; `result is p` is what the short-circuit actually promises and is stricter (dataclass `__eq__` would also accept a rebuilt copy).
- The new tests are purely structural. #3001 surfaced at build time — one test running `build_param(-Intercept("a") + 3 * Parameter("b"))` inside a `pm.Model` would pin the behavior users actually hit. Precedent: `test_build_param_with_subtraction`.
- `isinstance(other, int) and other == 0` misses `np.int64(0)`. Pre-existing in `ModelTerm`/`Sum`, consistent, not this PR's job — just noting it.

### Answer to the open question

Fold `__neg__`/`__sub__`/`__rsub__` in. Either as three more method pairs on `Sum` and `Product`, or — preferably — via the mixin above, which is confirmed green against the suite on this branch.

anevolbap:
Folded in through the `TermOps` mixin. `__neg__`/`__sub__`/`__rsub__` now work on `Sum` and `Product`, and `a + s` and `s - a` are flat like `p + s`. Added the subtraction tests and a `build_param` test for `-Intercept("a") + 3 * Parameter("b")`, and tightened the two assertions. The `np.int64(0)` case is left as is.

juanitorduz (commented):
## Re-review at `a3ee5b6b`

All four review points are addressed and the implementation is clean. One small API-hygiene item left, plus a correction to something I got wrong last round.

### Verified on this head

- `uv run pytest tests/test_terms.py` — **82 passed**.
- `pre-commit run --files pymc_marketing/terms.py tests/test_terms.py` — all hooks pass, including `mypy` and `validate-api-docs`.
- Every expression that raised before now builds end-to-end (`collect_coords` → `register_data` → `build_param`, each returning an `XTensorVariable`): `-term + scalar*term` (#3001), `term - product`, `product - product`, `term - sum`, `-(sum)`, `sum([...])` over products, `3 - sum`, and a flattened `term + (dot + term)` whose `feature` coord survives the flattening.
- Shape matrix confirms the intended contract, with `+`/`-` now producing a flat `Sum` from either side:

```
a + s  -> Sum[a, b]                  s + a -> Sum[a, b, a]
p + s  -> Sum[p, a, b]               s + p -> Sum[a, b, p]
s + s  -> Sum[a, b, a, b]            sum([p, q, s]) -> Sum[p, q, a, b]
s - a  -> Sum[a, b, Prod(-1, a)]     b - s -> Sum[b, Prod(-1, s)]
0 + p  -> p (identity)               0 + s -> s
```

- Blast radius checked: nothing in `pymc_marketing/` imports `terms` outside its own docstrings, and `tests/test_terms.py` is the only consumer. The `term + sum` / `sum - term` shape change cannot reach anything else.
- `__sub__` correctly does *not* flatten its right operand (it is negated first), and `p - 3` yields `Sum[p, -3]` rather than a pointless `Product(-1, 3)`.

### Correction: my `np.int64(0)` nit was wrong

I claimed the `isinstance(other, int)` guard misses `np.int64(0)`. It does not, in practice. `isinstance(np.int64(0), int)` is indeed `False`, but NumPy's object-dtype loop converts the scalar to a Python `int` before delegating, so `__radd__` receives a real `int`:

```
np.int64(0) + p -> Product   (identity short-circuit fires)
np.int64(3) + p -> Sum
False + p       -> Product   (bool subclasses int, False == 0)
True + p        -> Sum[True, p]
```

Leaving it as is was the right call. Sorry for the noise.

### One item before merge: `TermOps` visibility

`TermOps` is a public name that is not in `__all__`, so `autosummary` will not document it, while `_addends` directly above it is underscore-private. The module is internally inconsistent about which of the two new symbols is API.

Pick one:

- **Rename to `_TermOps`** — matches `_addends`, and the documented extension path is already "subclass `ModelTerm`", which inherits the operators anyway. My preference.
- **Add `"TermOps"` to `__all__`** and a line in the module docstring's *Extending* section — justified only if a custom *container* (a third peer of `Sum`/`Product`) is a path you want to support.

### Nit

`TermOps.__doc__` is "Arithmetic operators shared by terms, sums and products." The non-obvious half of the contract — that `+` and `-` always flatten to a single-level `Sum`, which is the entire reason `_addends` exists — is not stated anywhere in the source. One extra line there would keep the next person from re-introducing a nested shape.

Good work on the parametrized `test_sub_negates_right`; the six rows genuinely exercise three distinct dispatch paths (`ModelTerm.__sub__`, `Product.__sub__`, `__rsub__`) rather than padding. `test_build_param_product_add_product` pins the failure mode from #3001 at the layer where users actually hit it.

Approving once `TermOps` vs `_TermOps` is settled.

pymc_marketing/terms.py
```diff
@@ -263,8 +263,47 @@ def add_coords(self, ds):
 ]
 
 
+def _addends(value: Any) -> list[Any]:
+    """Return the addends of ``value``, flattening a ``Sum``."""
+    return list(value.terms) if isinstance(value, Sum) else [value]
+
+
+class TermOps:
+    """Arithmetic operators shared by terms, sums and products."""
```
juanitorduz, line 272: Two things to settle here, both about this class being the new home of the module's arithmetic contract.

**1. The flattening invariant is undocumented.** That `+` and `-` always produce a single-level `Sum` is the entire reason `_addends` exists, and it is the property that silently regressed in the first round of this PR. Nothing in the source states it, so the next person to touch an operator has no way to know it is load-bearing. Directly applicable:

```suggestion
class TermOps:
    """Arithmetic operators shared by terms, sums and products.

    ``+`` and ``-`` always produce a flat :class:`Sum`: a ``Sum`` operand
    contributes its own terms rather than nesting.  ``*`` and unary ``-``
    produce a :class:`Product`.  Defining the operators once here keeps
    ``ModelTerm``, ``Sum`` and ``Product`` from drifting apart.
    """
```

**2. `TermOps` is public-named but not public API.** It is absent from `__all__`, so `autosummary` will not document it, while `_addends` two lines above is underscore-private. The module is inconsistent about which of the two new symbols is API.

My preference is to rename it `_TermOps`, matching `_addends`: the documented extension path in the module docstring is already "subclass `ModelTerm`", and that inherits the operators for free. A custom *container* — a third peer of `Sum`/`Product` — is the only case that would want `TermOps` directly, and that is not a path this module advertises today.

The rename touches five lines, so it is not atomically suggestible:

- L271 `class TermOps:` → `class _TermOps:`
- L278 `-> Sum | TermOps:` → `-> Sum | _TermOps:`
- L306 `class ModelTerm(TermOps):` → `(_TermOps)`
- L365 `class Sum(TermOps):` → `(_TermOps)`
- L400 `class Product(TermOps):` → `(_TermOps)`

If you would rather keep it public, the equivalent fix is adding `"TermOps"` to `__all__` plus a sentence in the module docstring's *Extending* section. Either is fine — what should not ship is a public name that the docs will not render.

juanitorduz (commented):
One inline comment on `TermOps` with an applyable suggestion for the docstring, and the exact line list for the `_TermOps` rename if you go that way. Everything else from the last pass is verified green at `a3ee5b6b`.

juanitorduz (changes requested):
minor changes :)

pymc_marketing/terms.py
anevolbap: Both done. Renamed to `_TermOps` (665f9bfc), so it matches `_addends`. The docstring (b80b9cee) follows the suggestion with two corrections: the right operand of `-` is negated as a whole, so `a - s` gives `Sum[a, Product(-1, Sum[b, c])]` and not a flat sum of `b` and `c`, and `0 + term` returns the term. Should `a - s` also flatten into `Sum[a, -b, -c]`, or is the current shape fine?

juanitorduz:
## Re-review at `665f9bfc`

Both items landed. Verified on this head:

- `pytest tests/test_terms.py` — **82 passed**.
- `pre-commit run --files pymc_marketing/terms.py tests/test_terms.py` — all hooks pass, including `ruff`, `mypy`, and `validate-api-docs`.
- No stale `TermOps` references anywhere in the tree (`.py`, `.md`, `.ipynb`) — the rename is complete.
- Every claim in the new `_TermOps` docstring is exactly true as written:

```
a + s  -> Sum[a, b, c]                          # flattened
s - a  -> Sum[d, e, Prod(-1, a)]                # left operand flattened
a - s  -> Sum[a, Prod(-1, Sum[b, c])]           # right operand negated whole
-s     -> Prod(-1, Sum[b, c])
0 + a  -> a   (identity, `is`)                  0 + s -> s   (identity, `is`)
```

`_TermOps` now matches `_addends`, and since it is out of `__all__` it correctly stays out of the generated API docs.

### On `a - s`: keep the current shape

Good question, and I think the answer is no — do not distribute. Four reasons, in order of weight:

**1. It is not the same kind of transformation.** Flattening `+` exploits associativity of the *container*: `Sum[a, Sum[b, c]]` and `Sum[a, b, c]` compile to the identical graph, so normalizing the shape is free. Distributing `-` over a `Sum` is an algebraic rewrite that changes the graph.

**2. It costs one multiply per term instead of one, total.** Measured on a five-term `Sum` through `build_param`:

| shape | apply nodes | multiplies |
|---|---|---|
| `Prod(-1, Sum[...])` (current) | 14 | 1 |
| distributed `Sum[a, -c0, ..., -c4]` | 17 | 5 |

That is O(n) multiplies where O(1) suffices, and there is no guarantee PyTensor's rewrites re-collapse them.

**3. It would force `Sum.__neg__` to change contract.** If `a - s` flattened but `-s` still returned `Product(-1, s)`, you would have reintroduced exactly the kind of shape-depends-on-which-operator inconsistency this PR set out to remove. Staying consistent would mean `Sum.__neg__` returning a `Sum` of negated terms — which breaks the `__neg__ -> Product` annotation shared with `ModelTerm`, and `test_sum_neg` with it. Not worth it.

**4. Nesting is invisible to every consumer.** `build_param`, `get_coords`, `register_data`, `set_data` and `collect_terms` all recurse, so the nested `Product(-1, Sum)` is a representational detail with no behavioral consequence. The current docstring already states the rule plainly, which is the part that actually mattered.

### One cosmetic nit

`pymc_marketing/terms.py` L276 is 120 characters — right at `max-line-length`, so `ruff` passes, but it is a lone long line inside a block otherwise wrapped at ~75. Reflowing keeps the docstring readable in the source:

```
    ``+`` and ``-`` produce a flat :class:`Sum` (except ``0 + term``, which
    returns the term): a ``Sum`` operand of ``+``, or the left operand of
    ``-``, contributes its own terms rather than nesting. The right operand
    of ``-`` is negated as a whole, so a ``Sum`` there enters as one
    ``Product``. ``*`` and unary ``-`` produce a :class:`Product`. Defining
    the operators once here keeps ``ModelTerm``, ``Sum`` and ``Product``
    from drifting apart.
```

Non-blocking. Everything substantive is resolved — LGTM.

juanitorduz (approved):
Thanks @anevolbap

PR #3011, opened 2026-09-14
docs: add MMM prior sensitivity case study

TeemuSailynoja:
## Summary
- add an executed MMM case study that introduces ArviZ power-scaling prior sensitivity diagnostics
- compare observational and lift-calibrated models, distinguishing internal parameter sensitivity from all-time ROAS sensitivity
- add the notebook to the MMM gallery with an extracted thumbnail and final watermark

## Validation
- `uv run ruff check docs/source/notebooks/mmm/mmm_prior_sensitivity.ipynb`
- `uv run python scripts/run_notebooks/runner.py --notebooks mmm/mmm_prior_sensitivity.ipynb`
- real execution of `mmm_prior_sensitivity.ipynb` with 6 chains × 2,000 draws; zero divergences and max R-hat 1.00 for both fits using Nutpie `draw_diag` adaptation
- `uv run python scripts/generate_gallery.py --check --no-thumbnails`

No full documentation build was run.

----
📚 Documentation preview 📚: https://pymc-marketing--3011.org.readthedocs.build/en/3011/


daimon-pymclabs:
Reviewed the notebook end to end. The narrative arc is the right one. Four things I'd fix before merge, two of them real bugs visible in the committed outputs.

**`roas_summary` silently drops the HDI.** `interval_columns = [c for c in summary.columns if c.startswith("hdi_")]` matches nothing in arviz 1.2, so the comparison table in cell 38 prints only `mean` and `sd` despite `ci_prob=0.94, ci_kind="hdi"`. The intervals are the whole point of that table. The same cell also shows `sd` as `9.6, 11, 5.84, 13.8`, so the column is object dtype and `.round(1)` isn't applying.

**The `seasonality_prior` column is exactly 0.000 for all three targets.** Perturbing the `gamma_fourier` prior moving the `gamma_fourier` posterior by literally zero is not plausible, especially when the baseline prior moves it 0.133. That looks like a silent no-op (name not resolving inside `psense_summary`) rather than a finding. Worth confirming before the text leans on "ROAS remains insensitive to both prior blocks".

**The lift model's flags get worse, and the text doesn't say so.** `media_parameters` likelihood sensitivity goes 0.152 to 0.447, and `adstock_alpha[x1]` and `saturation_lam[x2]` newly flag as prior-data conflict. Cell 42's "some internal parameters remain sensitive" undersells it. The honest version is that sharp lift evidence creates new tension with the weak media priors while the reporting target stabilizes, which is a better lesson anyway.

**Lift-test provenance.** `delta_y = 23.34703279570842` with `sigma=3` is presented as "suppose the team runs two lift tests" but is clearly derived from the DGP. Say where it came from, otherwise the clean resolution looks rigged.

Smaller points:

- Typo "observtions" in cell 42.
- `row["likelihood"]` is recomputed and overwritten on every block iteration in `psense_by_target_and_block`. It doesn't depend on `prior_var_names`, so that's 4x wasted work.
- `business_prior` stays in `log_likelihood` after being copied to `log_prior`, so a reader who omits `likelihood_var_names` double-counts it. `del` it.
- Cells 39 and 45 build the identical `xr.concat`.
- The business prior numbers live in both `biz_prior_x1`/`biz_prior_x2` and `business_prior_df`.
- Two `zip(..., strict=False)` should be `strict=True`.

juanitorduz (changes requested):
AGENT REVIEW:

Reviewed against the committed outputs, arviz 1.2.0 source, the `mmm_roas` notebook, and the RTD preview.

## Blocking (visible in committed outputs)

**1. Every figure renders twice in the docs.** Cells 7, 16, 18, 20, 37, 39, 46 end with a bare `fig`, so each cell emits both `display_data` (inline backend) and `execute_result` (repr). The notebook JSON has two `image/png` outputs per cell, and the RTD preview page has 18 `` tags for 9 figure cells. Fix: drop the trailing `fig` / end with `;` (the `mmm_roas` convention). Cells 29/30 similarly print the `PlotCollection` repr above the plot.

**2. `roas_summary` silently drops the HDI.** arviz 1.2.0 `az.summary(ci_prob=0.94, ci_kind="hdi")` returns columns `['mean','sd','hdi94_lb','hdi94_ub',...]`, so `column.startswith("hdi_")` matches nothing and cell 38 shows only `mean`/`sd`. Also `sd` prints as `9.6, 11, 5.84, 13.8`: summary returns sig-fig-rounded object dtype, so `.round(1)` is a no-op. Use `startswith("hdi")` (or explicit names) and `pd.to_numeric` before rounding, as `convergence_summary` already does.

**3. Lift-test provenance.** `23.34703279570842 / 0.25 = 93.388 = true_roas_x1`; the values are `true_roas_x1 * delta_x`, copied from `mmm_roas` cell 37 where this is stated explicitly in an admonition. Here they are magic constants introduced as "suppose the team runs two lift tests", which contradicts the "truth stays hidden until the end" framing. Say where they come from (same admonition as `mmm_roas`), or compute them alongside cell 44 and acknowledge the reveal.

## Should fix

**4. `seasonality_prior` column is exactly 0.000 everywhere.** The `_get_power_scale_weights` -> `get_log_prior(var_names=["gamma_fourier"])` path is wired correctly and `gamma_fourier` is a free RV, so I don't think it's a no-op. More likely `Normal(0, 2)` is essentially flat over a posterior with coefficients much smaller than 2, so the log-prior barely varies across draws and the PSIS weights are uniform. That is a useful teaching point (a vacuous prior has ~zero sensitivity by construction) but the text should say it. Confirm with `idata.log_prior["gamma_fourier"].std("draw")` before leaning on "ROAS remains insensitive to both prior blocks".

**5. Cell 42 undersells the lift model.** Cell 41 shows `media_parameters` likelihood sensitivity 0.152 -> 0.447, and `adstock_alpha[x1]` and `saturation_lam[x2]` newly flagged as prior-data conflict. "Some internal parameters remain sensitive" is wrong in direction: sharp lift evidence creates *new* tension with the weak media priors while ROAS stabilizes. Worth stating; it's the better lesson.

**6. `business_prior` double-counting hazard.** It is copied to `log_prior` but left in `log_likelihood`. `psense_by_target_and_block` dodges this only because it whitelists `["y", "lift_measurements"]`; anyone calling `az.psense_summary` without `likelihood_var_names` (as cell 22's pattern invites) counts it on both sides. `del idata.log_likelihood["business_prior"]` after the move.

## Minor

- `finish_model(..., business_prior_df=None)` default is dead: `add_cost_per_target_calibration(calibration_data=None)` would fail. Make it required.
- `row["likelihood"]` is recomputed/overwritten per prior block (4x redundant psense calls per target).
- Cells 39 and 45 build the identical `xr.concat`; drop cell 45.
- Business prior duplicated in `biz_prior_x1/x2` and `business_prior_df`; derive one from the other.
- `zip(..., strict=False)` x2 -> `strict=True`.
- Typo "observtions" (cell 42).
- Gallery `.md`/`.yaml` entries and thumbnail look fine and consistent with neighbors.

Not run locally: docs build or real notebook execution. Convergence outputs look sound (0 divergences, min ESS > 2900).

juanitorduz:
This is great @TeemuSailynoja ! Some minor comments above. 

I would expand the introduction to explain why this is important and what we will be learning. Also, for each figure and table, we should have a clear learning, observation or takeaway :)

TeemuSailynoja:
<p>added</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/pymc-marketing/pull/3011/discussion/#comment-5664664966'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz (commented):
Round 2 on `d9bcbeb` + `5418c14`. Checked the notebook JSON outputs directly.

## Resolved

- Duplicate figures: every plot cell now emits a single `display_data`. `PlotCollection` repr suppressed. Retina config added.
- `roas_summary`: HDI columns present (`hdi94_lb`/`hdi94_ub`), numeric `sd`.
- `business_prior` removed from `log_likelihood` after the move.
- `finish_model` signature, `zip(strict=True)`, typo, docstring, seasonality-prior `0.000` explanation, lift-test provenance sentence.

## Should fix before merge

**1. The retrospective figure (cell 47) has no takeaway, and the honest one matters.** True ROAS from the data is `x1 = 93.4`, `x2 = 171.4`. Lift-calibrated posteriors are `x1 = 109.3 ± 5.8` (94% HDI ~98-120) and `x2 = 142.3 ± 13.8` (~116-168). Neither interval covers the truth. `x1` moved most of the way from 170.7 toward 93.4 but the `Normal(100, 20)` business prior still holds it above the truth; `x2` barely moved (143.3 -> 142.3) because it received no lift tests and is still pulled toward the business prior's 150. Cell 43's "help correct the resulting ROAS bias" and the checklist's "no longer materially sensitive to the original prior choices" read as a clean resolution; the figure shows a partial one. State this explicitly: it is the strongest argument in the notebook for why the diagnostic is about *sensitivity*, not correctness, and why the `x2` result should not be reported as validated.

**2. HDI bounds lose precision.** `az.summary` rounds to 2 significant figures by default, so the table shows `150 190 / 120 160 / 100 120 / 120 170` while `mean` has one decimal. Pass `round_to=1` (or `"none"`) to `az.summary` and drop the post-hoc `.round(1)`.

**3. Filler takeaways.** Cell 8 ("This quick visualization provides context for the model inputs and target.") says nothing; either state an observation (e.g. `x2` spend trends up with the target, which is the confounding to worry about) or delete. Cells 31, 38 and 40 still have no interpretation; @juanitorduz asked for one per figure.

**4. Cell 43 still soft-pedals the parameter-level change.** "Still contains warnings" hides that `media_parameters` likelihood sensitivity rose 0.152 -> 0.447 and `adstock_alpha[x1]` / `saturation_lam[x2]` newly flag as prior-data conflict. One sentence: sharp lift evidence creates new tension with the weak media priors even as the reporting target stabilizes.

## Minor (unchanged from round 1)

- `row["likelihood"]` still recomputed per prior block in `psense_by_target_and_block`; hoist it or compute once per target.
- Cell 46 `posterior_roas` duplicates cell 40 `roas_model_comparison`; reuse.
- `biz_prior_x1/x2` and `business_prior_df` still encode the same numbers twice.
- Lift-test `delta_y` values are still bare magic constants; `true_roas * delta_x` as in `mmm_roas` cell 37 (or a comment saying so) would make the provenance sentence concrete.

Not run locally: docs build, real execution.

juanitorduz:
## Review checklist (round 2, at `5418c14`)

### Resolved
- [x] Duplicate figures (bare `fig` at cell end); `PlotCollection` repr; retina config
- [x] `roas_summary` dropped HDI columns (`hdi_` vs `hdi94_lb`); `sd` object dtype
- [x] `business_prior` left in `log_likelihood` after move to `log_prior` (double counting)
- [x] `finish_model(business_prior_df=None)` dead default; `zip(strict=True)`; typo; helper docstring
- [x] `seasonality_prior` column `0.000` explained (broad `Normal(0, 2)` prior)
- [x] Lift-test provenance sentence

### Should fix before merge
- [ ] **Retrospective figure (cell 47) needs an honest takeaway.** True ROAS: `x1 = 93.4`, `x2 = 171.4`. Lift-calibrated: `x1 = 109.3 ± 5.8` (94% HDI ~98–120), `x2 = 142.3 ± 13.8` (~116–168). Neither interval covers the truth. `x1` moved most of the way from 170.7 but the `Normal(100, 20)` business prior still holds it above the truth; `x2` barely moved (143.3 → 142.3) because it received no lift tests. Cell 43 ("help correct the resulting ROAS bias") and the checklist bullet ("no longer materially sensitive to the original prior choices") read as a clean resolution; the figure shows a partial one. State that the diagnostic is about sensitivity, not correctness, and that `x2` should not be reported as validated.
- [ ] **HDI bounds lose precision.** `az.summary` rounds to 2 significant figures (`150 190 / 120 160 / 100 120 / 120 170`) while `mean` has one decimal. Pass `round_to=1` (or `"none"`) and drop the post-hoc `.round(1)`.
- [ ] **Filler / missing takeaways.** Cell 8 says nothing; state an observation (e.g. `x2` spend trends up with the target, which is the confounding to worry about) or delete. Cells 31, 38, 40 have no interpretation.
- [ ] **Cell 43 soft-pedals the parameter-level change.** `media_parameters` likelihood sensitivity rose 0.152 → 0.447 and `adstock_alpha[x1]` / `saturation_lam[x2]` newly flag as prior-data conflict. One sentence: sharp lift evidence creates new tension with the weak media priors even as the reporting target stabilizes.

### Minor
- [ ] `row["likelihood"]` recomputed per prior block in `psense_by_target_and_block`
- [ ] Cell 46 `posterior_roas` duplicates cell 40 `roas_model_comparison`
- [ ] `biz_prior_x1/x2` and `business_prior_df` encode the same numbers twice
- [ ] Lift-test `delta_y` are bare magic constants; use `true_roas * delta_x` as in `mmm_roas` cell 37, or comment that this is what they are

juanitorduz (approved):

PR #3015, opened 2026-09-14
Run self-contained docstring examples as doctests in CI

### LINKED ISSUE #578: Run example code in doc strings in CI
We have quite some examples in doc strings which we currently don't run. When we change things they easily break and we don't notice. There is some tooling but not sure how to do that.

felipebridge:
## Summary

Closes #578.

- Adds `tests/test_docstring_examples.py`, which runs `doctest` against a curated list of modules and fails CI if an example stops matching its documented output.
- Starts the list with `pymc_marketing.customer_choice.nested_logit` and `pymc_marketing.mmm.tvp` — these are the two modules I found where every `Examples` block is either self-contained (no fitted model / external data needed) or has been marked `# doctest: +SKIP` because it genuinely requires an active `pymc.Model` context (`tvp.create_hsgp_from_config`).
- Fixes `NestedLogit._parse_nesting`'s own docstring example along the way: as written it raises `NameError` (missing the `NestedLogit.` prefix, since it's a staticmethod) and its expected output used double quotes where Python's actual dict repr uses single quotes. Running it caught both.

## Why this scope

Most other docstrings with `>>>` examples in the codebase (in `mmm.py`, `plot.py`, `plot_interactive.py`, `summary/*.py`, `time_slice_cross_validation.py`, `r2d2.py`, etc.) illustrate usage against a fitted `MMM`/`NestedLogit` model, an `idata`, or other objects that don't exist in a bare doctest namespace — they're intentionally not literal, runnable transcripts. Forcing `--doctest-modules` over the whole package would either require rewriting a large number of docstrings with model-fitting fixtures (slow, and a much bigger PR) or blanket-skipping them (which defeats the point of the issue). I kept this first PR to the modules whose examples were already runnable as documentation, verified locally with `python -m doctest`, and left a comment in the test file explaining the scope so it's easy to extend later.

## Test plan

- [x] `pytest tests/test_docstring_examples.py -v` — 2 passed
- [x] `pytest tests/mmm/test_tvp.py tests/test_docstring_examples.py -q` — 46 passed (no regressions in existing `tvp` tests)
- [x] `ruff check` / `ruff format --check` on all changed files — clean

----
📚 Documentation preview 📚: https://pymc-marketing--3015.org.readthedocs.build/en/3015/


daimon-pymclabs:
Thanks for picking this up, this is a good first cut and the scoping call in your description is the right one. Forcing `--doctest-modules` over the whole package would have been a much worse PR. Nice catch on `_parse_nesting` too: that example was broken twice over and nothing would have surfaced it without actually running it.

Three things I'd like to see before merge, then I'm happy to approve.

**1. The examples aren't actually verified as self-contained.** `doctest.testmod(module)` uses the module's `__dict__` as the globals namespace, so every example runs with whatever that module already imported at the top. `np.arange(52)` in `tvp.py` and the bare `is_hsgp_kwargs_format(...)` call both pass in CI, but a reader who copies them into a fresh session gets a `NameError`. Since "self-contained" is the property the test is meant to protect, passing `globs={}` and adding the missing import lines to the docstrings is what makes it real. It will likely fail at first, which is the useful part.

**2. The assertion can pass vacuously.** `assert results.failed == 0` is satisfied when `results.attempted == 0`, so a renamed module or a deleted `Examples` block turns the test green rather than red, which is the exact rot the test exists to catch. Adding `assert results.attempted > 0` closes that.

**3. `tvp`'s `Examples` section is now entirely `+SKIP`.** That means including `tvp` in the list only exercises `is_hsgp_kwargs_format`. `create_hsgp_from_config` runs fine inside `with pm.Model(coords={"date": range(52)}):`, so wrapping the example is both cheaper than the skip and better documentation: the reader gets something they can paste, instead of prose telling them what to wrap it in.

Two smaller, take-them-or-leave-them notes:

- pytest ships `--doctest-modules`, so this could live as a scoped entry in `[tool.pytest.ini_options]` rather than a hand-maintained module list in a test file. The list is fine for two modules, but worth considering if this grows.
- `doctest.testmod` writes failure detail to stdout, so it only reaches the log via pytest's capture. Using a `DocTestRunner` with a `StringIO` would let you put the failing example directly in the assertion message.

None of this is structural, the shape of the PR is right. Happy to re-review as soon as you've pushed.

PR #2204, opened 2026-01-20
Implement MultiDimensionalMMM with pymc.dims

ricardoV94:
TODO: 
- [ ] Dev migration guide (separate PR https://github.com/pymc-labs/pymc-marketing/pull/2392)
- [x] Adapt new incrementality code
- [x] Censored Prior
- [x] Deprecate dims to Transformation.apply

Closes #1981 
Closes #1514 
Closes #1630 (just a regression test, it worked fine before)
related to #2017 

----
📚 Documentation preview 📚: https://pymc-marketing--2204.org.readthedocs.build/en/2204/

PR #1358, opened 2025-01-08
Custom optimizer constraints

cetagostini:
## Description

Allowing optimizer to use custom constraints around any intermediate variable (e.g: Response, Specific channel contributions or budgets).

It also allows optimizing a subset of budgets by passing a boolean `opt_mask`

Build on top of #1357 by @ricardoV94 

## Related Issue

- [x] Closes #1331 
- [x] Related to #1357 
- [x] Closes #1387  

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [x] Included tests that prove the fix is effective or that the new feature works
- [x] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Modules affected

- [x] MMM
- [ ] CLV
- [ ] Customer Choice

## Type of change

- [x] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc-marketing--1358.org.readthedocs.build/en/1358/

PR #3016, opened 2026-09-15
fix: diversification_ratio has been unusable since release

### LINKED ISSUE #2994: `diversification_ratio` is unusable: wrong 1D guard and misplaced parenthesis

felipebridge:
## Summary

Closes #2994.

`pymc_marketing.mmm.utility.diversification_ratio` has been unusable since it shipped in 0.11.0, due to two independent, compounding bugs:

1. **Wrong dimensionality guard** (since 0.11.0, #1358). The function requires a 2D `(sample, channel)` array per its own docstring and body (`samples.std(dim="sample", ...)`, `_covariance_matrix(samples, ...)`), but calls `_check_samples_dimensionality`, which raises unless the input is 1D. That check is correct for the scalar-response utilities elsewhere in this module and wrong for this one. Added `_check_samples_dimensionality_2d` and use it here instead.
2. **Misplaced parenthesis** in the `w'Σw` computation (since 0.19.0, #2204): the closing paren of `weights.rename(...)` landed after the inner `ptx.dot(cov_matrix, weights)` call, so that product was passed as a second positional argument to `rename()` instead of as the second operand of the outer `ptx.dot`. Bug 1 masked this completely — the rewritten line never actually executed until now.

## Test plan

- Verified the fixed function matches a numpy reference implementation (`(w·σ).sum() / sqrt(wᵀ Σ w)`) to full precision, parametrized across a few sample/asset shapes (500×3, 50×1, 25×5).
- Verified 1D input is now rejected with a clear, correct error (`... expected samples to be a 2D tensor variable ...`) instead of failing further down the call stack with an unrelated `rename()` error.
- Confirmed both new tests fail against the pre-fix code, each reproducing one of the two original bugs, and pass with the fix.
- `pytest tests/mmm/test_utility.py` — 50 passed (no regressions in the other utility functions in this module).
- `ruff check` / `ruff format --check` — clean.

There was no prior test coverage for this function at all (`git log -S diversification_ratio -- tests/` returns nothing), which is exactly why it shipped broken and stayed broken across two releases without anyone noticing.

----
📚 Documentation preview 📚: https://pymc-marketing--3016.org.readthedocs.build/en/3016/


felipebridge:
Closing — I missed that #2995 already fixes this exact bug (opened 2026-09-09, before this PR). Apologies for the duplicate; deferring to the existing PR.

PR #3017, opened 2026-09-15
fix: delayed_adstock returns NaN for alpha=0 with non-integer theta

### LINKED ISSUE #2889: delayed_adstock returns NaN for alpha=0 with non-integer theta

felipebridge:
## Summary

Closes #2889.

The weights in `delayed_adstock` are `w = alpha ** ((lag - theta) ** 2)`. With `alpha` exactly `0` and a non-integer `theta`, no lag's exponent lands on exactly `0`, so every weight is `0`, `w.sum()` is `0`, and `normalize=True` divides `0/0` into `NaN`. `alpha=0` is also the function's own default argument, and values that merely underflow to `0` (e.g. via `DelayedAdstock`'s half-life parametrization, per the issue) hit the same failure mode.

`_check_alpha`'s own error message (`"0 < alpha <= 1"`) already implies `alpha=0` should be invalid, but the actual check tests `alpha >= 0`, so `0` passes a guard whose own message excludes it.

## Approach

Per the two options laid out in the issue, I went with the non-breaking one: floor `alpha` at the smallest positive value for its dtype right after the existing check, rather than making `_check_alpha` reject `0` outright (that's a behavior change for anyone currently relying on `alpha=0` with an *integer* `theta`, which is finite today, plus it would need a new default since `alpha=0.0` is the function's own default argument).

One deviation from the issue's suggested snippet: I tied the floor's dtype to **`x.dtype`** (matching how `lags` already does `pt.arange(l_max, dtype=x.dtype)` two lines below it), not to `pytensor.config.floatX` as suggested. Testing the float32 case directly showed that using `pytensor.config.floatX` mixes dtypes when the graph's actual dtype differs from the global config default, and crashes under the numba backend — `x.dtype` is the dtype the rest of the function already anchors to, so it doesn't have that problem.

## Test plan

- Verified `alpha=0.0, theta=0.5` (the issue's repro) now returns finite values instead of `NaN`.
- Verified `alpha=0.0, theta=5` (integer, already finite before this change) is numerically unchanged — `allclose` against the pre-fix output, with only a ~1e-309 perturbation exactly at the zero-crossing lag.
- Verified a normal case (`alpha=0.5`) is unaffected.
- Verified both float32 and float64 graphs stay finite and preserve their own dtype (this is what caught the `pytensor.config.floatX` dtype-mismatch issue above).
- Added 4 regression tests covering all of the above in `tests/mmm/test_transformers.py`. Confirmed they fail against the pre-fix code — the two finiteness tests reproduce the original `NaN`, and the dtype test reproduces the numba crash a naive `config.floatX`-based fix would have introduced — and pass with the fix as implemented.
- `pytest tests/mmm/test_transformers.py` — 148 passed (no regressions elsewhere in the adstock/saturation transformer suite).
- `ruff check` / `ruff format --check` — clean.

----
📚 Documentation preview 📚: https://pymc-marketing--3017.org.readthedocs.build/en/3017/


felipebridge:
Closing — I missed that #2899 already fixes this exact bug (opened 2026-08-18, well before this PR). Same root cause, same non-breaking floor approach. Apologies for the duplicate; deferring to the existing PR.

PR #3018, opened 2026-09-15
fix: log_model_graph leaves the DOT source behind when rendering fails

### LINKED ISSUE #2968: log_model_graph leaves the DOT source in the working directory when graphviz is not installed

felipebridge:
## Summary

Closes #2968.

`graph.render(path)` (`pymc_marketing/mlflow.py`) writes the DOT source to disk *before* invoking the `dot` executable. When `dot` isn't on `PATH` (or rendering otherwise fails), `render()` raises, the `except` branch logs and returns, and the `os.remove(path)` meant to clean up only ran in the success (`else`) branch -- so the DOT source was left behind. The call site passes a relative path (`log_model_graph(model, "model_graph")`), so the leftover lands wherever the process happens to be running from, untracked and uncovered by `.gitignore`.

## Fix

Renders into a `tempfile.TemporaryDirectory()` instead, the same pattern already used by `log_error()` a few hundred lines above in this file. Nothing is ever written outside the temp directory, so there's nothing to clean up manually and nothing to leak regardless of where rendering fails.

## Test plan

- Added `test_log_model_graph_no_leftover_when_dot_executable_missing`, which hides any real `dot` executable from `PATH` rather than mocking `graph.render()` itself -- mocking `render()` directly would skip past the real DOT-source write entirely and pass either way, so it wouldn't actually catch this bug (I checked: the existing `render_error`-mocked case in `test_log_model_graph_no_graphviz` passes identically with or without this fix, since it never reaches a real filesystem write).
- Confirmed the new test fails against the pre-fix code, reproducing the exact leftover file, and passes with the fix.
- Ran the full `tests/test_mlflow.py` suite. The fix introduces no new failures -- the 13 that fail in this sandbox are pre-existing and unrelated (missing system graphviz/`dot` binaries, missing NetCDF4 backend, missing optional `jax`/`nutpie`/`blackjax` samplers), confirmed identical on unmodified `main`. One of those (`test_autolog_pymc_model[pymc]`) actually demonstrated the bug live during this check: running it against the pre-fix code left an untracked `model_graph` file sitting in my own working directory afterward.
- `ruff check` / `ruff format --check` -- clean.

----
📚 Documentation preview 📚: https://pymc-marketing--3018.org.readthedocs.build/en/3018/


felipebridge:
Closing — I missed that #2969 already fixes this exact bug (opened 2026-09-08, before this PR). Same fix: render into a temporary directory. Apologies for the duplicate; deferring to the existing PR.

PR #3019, opened 2026-09-15
docs(mmm): address Lance's review follow-ups on Case Study II

daimon-pymclabs:
Follow-ups to Lance's review comments on #2936 (merged).

### 1. Notation (`s_sem` -> `s_spd`)
The spend normalizer is renamed to `s_spd` so it no longer collides with "SEM". Both constants are now written out explicitly in the prose: `s_spd = max_t(sem_spend_t)` and `s_imp = max_t(sem_impressions_t)`.

### 2. Calibration callout
The lift-test callout now states that an upper-funnel channel calibrated against a total-effect experiment needs a target spanning both the direct and the mediated path, and that this is the subject of a separate notebook. Next step 2 is reworded to match (it previously pointed at `mmm_lift_test` only).

### 3. CRPS fold axis (library fix, not a notebook fix)
The mislabelled fold axis came from `MMMCVPlotSuite.crps` in `pymc_marketing/mmm/plotting/cv.py`, which plotted folds as positional floats and discarded the fold labels it already had.

- Discrete, labelled ticks on the fold axis.
- Markers on by default so single folds are visible.
- New `combine_splits=True` option drawing train and test in one panel on a shared y-scale (Lance's merge suggestion).

Seven new tests in `tests/mmm/plotting/test_cv.py` (the seventh covers the guard for backends without matplotlib's tick API, which Codecov flagged as the one uncovered line); all 39 tests in that file pass.

### 4. Budget range plot
New figure in the budget section showing where the plan sits inside its guardrails. Now drawn with `azp.plot_forest` (per Juan's 501c469) so it matches the notebook's other forest plots: each channel's +/-50% band is the full-range ETI of a uniform grid over the band, whose mean marks the historical spend, and the funnel-aware optimum is a second dataset in the same `model` dimension. The x axis stays log-scaled so every +/-50% band has the same visual width, and the cell is followed by a printed count of media cells sitting at a bound.

### Notebook outputs
The notebook was re-executed end to end on this branch (nutpie, zero errors, ~2 min). Every quoted number reproduces against the previous execution; the only outputs that moved are the CRPS values, whose posterior predictive is unseeded upstream, and they stay inside the ranges the prose quotes (train 17 to 18M; test about 15M for Folds 0 and 2, about 29M for Fold 1, level with train for Fold 3).


juanitorduz (approved):

juanitorduz:
@lancechua I will merge this one as I wanna do additional improvements, so we have faster iterations.

lancechua:
> @lancechua I will merge this one as I wanna do additional improvements, so we have faster iterations.

Sure sure, didn't have any more feedback anyway. Looks good. Thanks!

PR #3020, opened 2026-09-15
fix(mmm): load constant adstock and saturation parameters from dict

### LINKED ISSUE #1613: Cannot set a constant value as the parameter of an Adstock or Saturation transformation
[This line](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/components/adstock.py#L339) for the adstock and [this line](https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/components/saturation.py#L469) for the saturation transformations assume that the transformation `priors` must always be a `Prior` instance or something that `deserialize` knows how to handle. It would be great if we could actually set the prior parameter to a constant value, e.g. set `alpha` for a `GeometricAdstock` to 0.5, and simply try to infer the other MMM parameters. With the current deserialization, we cannot do that while loading the MMM hyperpriors from a configuration file.

anevolbap:
Fixes #1613

## Description
`AdstockTransformation.from_dict`, `SaturationTransformation.from_dict` and `Basis.from_dict` passed every prior value to `deserialize`. `to_dict` writes a constant parameter as a plain float or list (`_serialize_value`), so loading it back raised `DeserializableError: Couldn't deserialize 0.5`.

A new `_deserialize_value` in `components/base.py` sends only dicts to `deserialize` and passes constants through unchanged. `HSGP.from_dict` uses the same dict check. `from_dict` now lives on `Transformation`, which replaces the three identical overrides.

```python
GeometricAdstock.from_dict({"l_max": 4, "priors": {"alpha": 0.5}})  # raised before, works now
```

## Checklist


pymc_marketing/mmm/components/base.py
```diff
@@ -775,3 +776,12 @@ def _serialize_value(value: Any) -> Any:
         return value.tolist()
 
     return value
+
+
+def _deserialize_value(value: Any) -> Any:
```
anevolbap, line 794: Only dicts go to `deserialize`. `_serialize_value` writes a `Prior` as a dict and a constant as a float or list, and `SupportedPrior` (L57-64) already accepts `float`, `list` and `np.ndarray`, so constants go back to the constructor unchanged. The list case in `test_roundtrip_constant_prior` samples the restored adstock without any list-to-array conversion.

`SerializationRegistry._deserialize_config_value` looks close, but it only builds a `Prior` from the `{"dist": ...}` shape. A hand-written `{"distribution": "HalfNormal", "sigma": 1}` stays a plain dict there, while `deserialize` accepts both.

pymc_marketing/mmm/events.py
```diff
@@ -133,9 +133,9 @@ def from_dict(cls, data: dict) -> "Basis":
         data.pop("lookup_name", None)
 
         if "priors" in data:
-            from pymc_extras.deserialize import deserialize
-
-            data["priors"] = {k: deserialize(v) for k, v in data["priors"].items()}
+            data["priors"] = {
+                k: _deserialize_value(v) for k, v in data["priors"].items()
```
anevolbap: Same change as in `adstock.py` and `saturation.py`. The local import is gone because `deserialize` is already imported at module level (L113).

pymc_marketing/mmm/components/adstock.py
```diff
@@ -147,7 +147,9 @@ def from_dict(cls, data: dict) -> AdstockTransformation:
         data.pop("__type__", None)
 
         if "priors" in data:
-            data["priors"] = {k: deserialize(v) for k, v in data["priors"].items()}
+            data["priors"] = {
+                k: _deserialize_value(v) for k, v in data["priors"].items()
```
anevolbap: The `from_dict` overrides in adstock, saturation and events are now identical except for the `lookup_name` pop in `Basis`. Move `from_dict` onto `Transformation` in a follow-up, or keep the overrides?

juanitorduz:
Reviewed the diff against `origin/main` and ran the new tests locally (`tests/mmm/components/test_adstock.py`, `test_saturation.py`, `tests/mmm/test_events.py -k "constant_prior or Roundtrip"`: 33 passed).

**Fix is correct and minimal.** `_deserialize_value` is the exact inverse of `_serialize_value`: `Prior`/`VariableFactory` serialize to a dict and go back through `deserialize`; float/list/ndarray/`Variable` serialize to a float or list and are accepted as-is by `SupportedPrior`. Since `Transformation.__eq__` compares `to_dict()`, an `np.ndarray` original and its restored `list` also compare equal, so the roundtrip assertions are meaningful. Using `deserialize` (rather than `Prior.from_dict` or `SerializationRegistry._deserialize_config_value`) is the right call: it keeps hand-written `{"distribution": ...}` configs working. Dropping the local import in `events.py` is a good cleanup.

**One request before merge — consolidate the three identical `from_dict` overrides now rather than in a follow-up.** `Transformation` currently has `to_dict` but no `from_dict`, so the natural home exists and the change is ~10 lines:

```python
# components/base.py, on Transformation
@classmethod
def from_dict(cls, data: dict) -> Self:
    data = data.copy()
    data.pop("__type__", None)
    if "priors" in data:
        data["priors"] = {k: _deserialize_value(v) for k, v in data["priors"].items()}
    return cls(**data)
```

Then `AdstockTransformation.from_dict` / `SaturationTransformation.from_dict` are deleted, and `Basis.from_dict` becomes `data = data.copy(); data.pop("lookup_name", None); return super().from_dict(data)`. That also removes the cross-module imports of the private `_deserialize_value` (and lets it stay module-private in `base.py`), which is the only slightly awkward part of the current shape.

**Minor / non-blocking**
- `test_roundtrip_constant_prior` in `test_adstock.py` also samples the restored adstock with `coords`, which is the strongest of the three tests because it proves the restored constant is usable, not just equal. The saturation and basis variants only check equality; fine, since the mechanism is shared.
- No changelog file in the repo, so nothing to add there; the docstring-less private helper is consistent with `_serialize_value`.

LGTM once `from_dict` lives on `Transformation`.

anevolbap:
Moved `from_dict` onto `Transformation` in 08cbeda9. The adstock and saturation overrides are gone, `Basis.from_dict` only drops `lookup_name` and calls `super()`, and `_deserialize_value` is now used only inside `base.py`.

Running `from_dict` for every concrete adstock, saturation and basis class (Prior, constant, list, hand-written `distribution` config, extra keys) gives the same output before and after. `test_from_dict` in `test_base.py` covers a custom subclass, which had no `from_dict` before.

daimon-pymclabs (approved):
Approving. `_deserialize_value` is the exact inverse of `_serialize_value` (a dict means the value had a `to_dict`, anything else was a constant), and collapsing the three identical overrides onto `Transformation` is the right cleanup. Non-blocking follow-ups:

- `serialization._deserialize_config_value` already inverts the same format for `model_config`, and it differs: it falls back to a plain dict when a dict is not a prior spec, and it rebuilds numeric lists as `np.ndarray`. Round-trips agree today because `Transformation.__eq__` compares `to_dict()` output, but two inverses of one format will drift. A comment cross-referencing the other would help.
- `Basis.from_dict` now exists only to pop `lookup_name`. Filtering `data` to `cls.__init__`'s signature in the base would remove the override entirely and make legacy configs with stale keys load instead of raising `TypeError`.
- Two untested paths: the `lookup_name` pop (no test emits that key), and a constant prior under the half-life parametrization, e.g. `GeometricAdstock(l_max=4, priors={"halflife": 2.0})`, which exercises the mixin's `to_dict` branch together with the new one.
- `test_from_dict_constant_prior` asserts `function_priors["alpha"] == 0.5`, which passes on the constructor alone. The parametrized round-trip above it already covers the real behaviour.

juanitorduz (approved):
Thanks @anevolbap !

PR #3022, opened 2026-09-15
Use pymc.pytensorf.resolve_shapes in deterministics_to_flat

### LINKED ISSUE #2387: get_symbolic_rv_shape should be replaced by pymc.pytensorf implementation
There are a couple of PRs right now implemeting something like `get_symbolic_rv_shape` between pymc / pymc-extras, and after #2204 there is one here.

Given the obvious demand, it makes sense for it to leave upstream of pymc-marketing, and we should switch it when it is in a stable place

juanitorduz:
Closes https://github.com/pymc-labs/pymc-marketing/issues/2387

## Description

Removes the local `pymc_marketing.model_graph.get_symbolic_rv_shape` helper (a stopgap copied in with #2204, tagged `# TODO: Remove this when utility is available in pymc.pytensorf`) and has `deterministics_to_flat` call the upstream implementation.

Upstream status:
- pymc-devs/pymc#8302 added `pymc.pytensorf.resolve_shapes` / `get_symbolic_rv_shapes` (same `ShapeFeature` + `infer_shape_db` pass as our copy). First released in pymc **v6.0.1**; pymc itself now uses it in `logprob/tensor.py` and `shape_utils.maybe_resize`.
- pymc-extras already switched (pymc-devs/pymc-extras#707).
- We pin `pymc>=6.2.0,<6.3.0`, so no dependency change is needed.

Implementation notes:
- `resolve_shapes(list(underlying_var.shape))` is used instead of `get_symbolic_rv_shapes([underlying_var])` because `XTensorVariable.shape` (the `pymc.dims` path used by `MMM.frozen_deterministics` / `SoftPlusHSGP`) is a tuple of scalar dim lengths, not a shape vector. This matches what the deleted helper did for both branches.
- The guard that the resolved shape no longer depends on the replaced variable is kept inline. Upstream has no equivalent, and without it the `Flat` would silently keep the replaced subgraph alive through its shape.
- Upstream builds its `FunctionGraph` with `copy_inputs=False`, so the resolved shape now references the model's own dim-length shared variables instead of clones sharing the same container. Functionally equivalent; the rebuilt model just no longer carries duplicate `SharedVariable` objects.
- Adds one regression test for the `pymc.dims` branch of `deterministics_to_flat` (the existing tests only covered the tensor branch).

Verification: `tests/test_model_graph.py` (4 passed), `tests/mmm/test_hsgp.py` + `tests/mmm/test_incrementality.py` (229 passed, 6 skipped), `tests/mmm/test_mmm.py -k "flat or deterministic or hsgp or new_data or out_of_sample"` (23 passed), and a smoke run of the `deterministics_to_flat` docstring example.

## Related Issue
- [x] Closes #2387
- [ ] Related to #

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--3022.org.readthedocs.build/en/3022/


pymc_marketing/model_graph.py
```diff
@@ -129,7 +95,14 @@ def deterministics_to_flat(model: pm.Model, names: list[str]) -> pm.Model:
         model_var = memo[variable]
         dims = model_var.owner.op.dims
         underlying_var = model_var.owner.inputs[0]
-        underlying_shape = get_symbolic_rv_shape(underlying_var)
+        # Shape expressed via the op's inputs so the Flat does not keep the
+        # replaced computation alive through its shape graph.
+        underlying_shape = resolve_shapes(list(underlying_var.shape))
```
daimon-pymclabs: Please put the reason for calling `resolve_shapes(list(var.shape))` rather than the more obvious `get_symbolic_rv_shapes([underlying_var])` into the comment here. The PR description explains it well (`XTensorVariable.shape` is a tuple of scalar dim lengths, not a shape vector, so the upstream RV wrapper would not fit the `pymc.dims` branch), but that reasoning is invisible at the call site and this is the one line someone will want to "tidy up" later.

Something like:

```suggestion
        # `resolve_shapes` rather than `get_symbolic_rv_shapes([underlying_var])`:
        # `XTensorVariable.shape` is a tuple of scalar dim lengths, not a shape
        # vector, so the upstream RV wrapper does not fit the `pymc.dims` branch.
        # Resolving expresses the shape via the op's inputs, so the Flat does not
        # keep the replaced computation alive through its shape graph.
        underlying_shape = resolve_shapes(list(underlying_var.shape))
```

pymc_marketing/model_graph.py
```diff
@@ -129,7 +95,14 @@ def deterministics_to_flat(model: pm.Model, names: list[str]) -> pm.Model:
...
+        if underlying_var in set(ancestors(underlying_shape)):
```
daimon-pymclabs: Two small things.

`set(ancestors(...))` walks the whole ancestor graph before testing membership, while `underlying_var in ancestors(underlying_shape)` short-circuits on the generator. Irrelevant for the toy models, less so for a wide HSGP graph, and it reads the same. The deleted helper used the generator form.

More importantly, this `raise` is the line codecov flags as uncovered, and the equivalent guard in the old helper was never covered either. You are keeping it deliberately (upstream has no equivalent, and without it the `Flat` silently retains the replaced subgraph), which is the right call, but that makes it worth pinning down. If you can construct a deterministic whose shape genuinely will not resolve, a `pytest.raises` test would lock in the behavior. If you cannot construct one cheaply, a one-line comment saying so is more useful than leaving a future reader to guess whether the branch is reachable.

tests/test_model_graph.py
```diff
@@ -67,3 +67,24 @@ def test_no_inputs(toy_model_given_x_mean):
         "x_mean": set(),
         "centered": {"x", "x_mean"},
     }
+
+
+def test_dims_deterministic_to_flat_keeps_dims_and_drops_dependency():
+    import pymc.dims as pmd
+    from pytensor.graph import ancestors
```
daimon-pymclabs: Move these to module level with the other imports. Every other import in this file is top-level, and there is no cycle or optional-dependency reason to defer them: `pymc.dims` is imported unconditionally by `pymc_marketing.model_graph` already.

tests/test_model_graph.py
```diff
@@ -67,3 +67,24 @@ def test_no_inputs(toy_model_given_x_mean):
...
+    with pm.Model(coords=coords) as model:
+        x = pmd.Normal("x", dims=("time", "covariate"))
+        x_mean = pmd.Deterministic("x_mean", x.mean(dim="time"))
+        pmd.Deterministic("centered", x - x_mean)
+
+    new_model = deterministics_to_flat(model, ["x_mean"])
+
+    flat = new_model["x_mean"]
+    # pmd.Flat builds an XRV op wrapping the tensor FlatRV
+    assert flat.owner.op.core_op.__class__.__name__ == "FlatRV"
```
daimon-pymclabs: Comparing `__class__.__name__` to a string is brittle in both directions: a rename upstream gives a failure message that points at the assertion rather than the cause, and a typo in the literal gives a test that can never pass. Prefer `isinstance(flat.owner.op.core_op, pm.FlatRV)` (or wherever `FlatRV` is importable from in the pinned pymc).

If you want to keep the name-based style for symmetry with the tensor tests, consider teaching the existing `get_rv_class_name` helper above to unwrap `core_op`, so both branches assert through one path instead of two spellings.

tests/test_model_graph.py
```diff
@@ -67,3 +67,24 @@ def test_no_inputs(toy_model_given_x_mean):
         "x_mean": set(),
         "centered": {"x", "x_mean"},
     }
+
+
+def test_dims_deterministic_to_flat_keeps_dims_and_drops_dependency():
```
daimon-pymclabs: Consider parametrizing the existing `toy_model` fixture over the tensor and dims backends instead of adding a separate single test. The model here is the `pmd` transliteration of `toy_model`, and the assertions overlap with `test_is_flat_distribution` and `test_no_inputs`.

A `@pytest.fixture(params=["tensor", "dims"])` on `toy_model` would run all four existing tests against both branches, which is broader coverage than this test gives and less code than the current duplication. The genuinely new assertions are the dims preservation and the `initial_point` shape, which would stay as their own tests.

Also, the four asserts here cover four distinct behaviors, so a failure message will not say which one broke. The rest of the file is one behavior per test.

daimon-pymclabs (commented):
## Review

Nice de-duplication. I verified the substantive claims in the description and they hold up:

- `resolve_shapes` and `get_symbolic_rv_shapes` are present at `pymc/pytensorf.py` in both `v6.0.1` and `v6.2.0`, and the current pin is `pymc>=6.2.0,<6.3.0`, so no dependency bump is needed.
- The upstream `resolve_shapes` body is the same `FunctionGraph(features=[ShapeFeature()], clone=True)` + `infer_shape_db.default_query.rewrite` pass under `optdb__max_use_ratio=10, cxx=""` as the deleted copy, differing only in `copy_inputs=False`. Dropping `copy_inputs=True` is an improvement here: the resolved shape now references the fgraph's own dim-length variables instead of clones that `replace_all(import_missing=True)` had to import as duplicates.
- The kept guard is semantically identical to the deleted one (both test whether the variable passed in is an ancestor of the resolved shape), just with a better message.
- `get_symbolic_rv_shape` had no other in-repo callers, and `docs/source/api/index.md` autodocs the module rather than naming the function, so nothing dangles.

### CI

The `build` / readthedocs failure is **not** from this PR. The only entry in the `docs-warnings` artifact is an ambiguous `'model'` cross-reference from `budget_optimizer.BuildMergedModel`, and the Docs workflow also failed on the latest `main` push. Worth confirming against a rebase, but nothing to fix here.

The codecov patch miss is real though: the uncovered line is the new `raise ValueError`, and neither the old helper's guard nor the new one has ever had a test. See the inline note.

### Main asks

1. Move the `resolve_shapes` vs `get_symbolic_rv_shapes` rationale from the PR description into a code comment. It is the one non-obvious choice in the diff and the next reader will be tempted to "simplify" it.
2. Consider a regression test for resizing, since `copy_inputs=False` changes which shared variables the shape graph points at. The dim-length aliasing should make it strictly better, but `SoftPlusHSGP` + `set_data` / new-data flows are exactly where a divergence would show, and asserting it directly is cheaper than re-deriving the argument later.
3. A few test-style points inline: module-level imports, the stringly-typed op assertion, and reusing the existing fixture pattern.

None of these are blockers. The core change looks correct to me.

pymc_marketing/model_graph.py
ricardoV94: I believe resolve_shape calls list itself but doesn't matter much

pymc_marketing/model_graph.py
ricardoV94: I wouldn't worry about this but if you do I'd as the bot says.

tests/test_model_graph.py
ricardoV94: why the new test?

ricardoV94 (commented):
I'd just keep it simple

tests/test_model_graph.py
juanitorduz: sorry, some slop I did not catch

juanitorduz:
Thanks @ricardoV94 , I simplified it :)

ricardoV94 (approved):

Issue #706, opened 2024-05-30
Model config wrapper around `pm.find_contrained_prior`


wd60622:
It might be convenient to support wrapper around [find_contrained_prior](https://www.pymc.io/projects/docs/en/latest/api/generated/pymc.find_constrained_prior.html) from the dictionary model config somehow. 

For instance, 

```python
model_config = {
    "beta_channel": {
        "dist": "Normal", 
        "kwargs": {"mu": 0, "sigma": 1}, 
        # Just add this 
        "bounds": {"lower": -3, "upper": 3}, 
    }, 
    ...
}
mmm = DelayedSaturatedMMM(..., model_config=model_config, ...)

# Which would be converted to via some internal logic
beta_channel_parameters = pm.find_contrained_prior(pm.Normal, lower=-3, upper=3, init_guess={"mu": 0, "sigma": 1})
beta_channel = pm.Normal("beta_channel", **beta_channel_parameters, dims="channel")
```
Though a user can do this outside of the model, I'd imagine it would have to be done many times so this would be convenience.

Would love to hear any thoughts on this

juanitorduz:
Yes! I think this would be a great feature!

wd60622:
> Yes! I think this would be a great feature!

## 1. How should the dictionary look?

Do you think that other parameters of the find_contrained_prior function should be exposed as well? 
Should the "kwargs" be for that functions then? 
i.e. 
```python
beta_channel_config = {
    "dist": "Normal", 
    "init_guess": {"mu": 0, "sigma": 1}, 
    # kwargs for the find_contrained_prior
    "kwargs": dict(lower=-3, upper=3, mass=0.90), 
}
```
Think that there would need to be a new key / different set of keys from in order to distinguish the branch of logic to take. So this could look many different ways.

## 2. Should it be internal or external of class initialization? 

Also might be a good util function instead so it is done once and outside of the model initialization. 
i.e. 
```python
model_config = ... # Whatever dictionary format is agreed upon
model_config = pymc_marketing.utils.find_contrained_prior(model_config)

# model_config now has "kwargs" that will fit the format as before
# Models would only take the already done format of the model_config

juanitorduz:
I like the external approach better 👍

PR #3024, opened 2026-09-15
docs: register class attributes in the toctree so numpydoc links resolve in class scope

### LINKED ISSUE #3023: NumPyDoc release is causing problems
It seems this release https://github.com/numpy/numpydoc/releases/tag/v1.11.0 is causing problems with our docs, see https://github.com/pymc-labs/pymc-marketing/actions/runs/34995463449/job/104470639726?pr=3022

daimon-pymclabs:
Closes #3023.

## Problem

numpydoc 1.11.0 ([#706](https://github.com/numpy/numpydoc/pull/706)) changed `link_prefix` in `_process_param` from `""` to `"."`, so every Attributes entry now renders as `` :obj:`model <.model>` ``. The leading dot sets `refspecific`, which puts Sphinx's Python domain into fuzzy search: class scope first, then any object whose name ends in `.model`.

`docs/source/_templates/autosummary/class.rst` built the Attributes autosummary without a `:toctree:`, so no Python domain object entry was ever registered for class attributes. The class-scope lookup misses and the fuzzy fallback matches two modules, `pymc_marketing.bass.model` and `pymc_marketing.pie.model`:

```
WARNING: more than one target found for cross-reference 'model' [ref.python]
```

`sphinx-build -W` in `.github/workflows/docs.yml` makes that single warning fatal.

## Fix

Add `:toctree: classattributes` to the Attributes block, mirroring what the Methods block already does with `:toctree: classmethods`. Attributes become real Python domain objects, so the new `.model` reference resolves in class scope and never falls through to the module search.

This removes the whole collision class rather than the one `BuildMergedModel.model` instance, and it needs no ongoing maintenance: no version pin to bump, no warning baseline to refresh.

The alternative of pinning `numpydoc<1.11` only defers the same failure to whoever bumps it next.

## Notes

Generated stubs land under `docs/source/api/generated/`, which is already gitignored, so this adds no tracked files. Docs build time grows by the stub-generation pass over class attributes.

----
📚 Documentation preview 📚: https://pymc-marketing--3024.org.readthedocs.build/en/3024/


daimon-pymclabs:
Good catch, fixed in aaef1ce. The `classattributes` toctree now has the same three companion exclusions the `classmethods` stubs already had: `remove_from_toctrees`, `sitemap_excludes`, and the four locale `Disallow` lines in `robots.txt`.

juanitorduz (approved):

juanitorduz:
Reviewed. Mechanism checks out: numpy/numpydoc#706 does flip `link_prefix` to `"."`, and `autoclass` without `:members:` never registered attribute objects, so `:toctree:` on the Attributes block is the right fix. The original `more than one target found for cross-reference 'model'` warning is gone from the build log on aaef1ce.

But the Docs workflow on aaef1ce still fails (`-W --keep-going`, exactly 1 warning): https://github.com/pymc-labs/pymc-marketing/actions/runs/35011194725/job/104523624110

```
pymc_marketing/data/idata/mmm_wrapper.py:docstring of pymc_marketing.data.idata.mmm_wrapper.MMMIDataWrapper.cost_per_unit:14: WARNING: Inline emphasis start-string without end-string. [docutils]
```

Cause: `pymc_marketing/data/idata/mmm_wrapper.py:364`

```
Cost per unit values with dims ("date", *custom_dims, "channel").
```

The bare `*custom_dims` is an unclosed RST emphasis. It was latent because the Attributes table only rendered the docstring summary line; the new `classattributes` stub pages render the full docstring. Wrapping it in double backticks (``` ``("date", *custom_dims, "channel")`` ```) or escaping `\*custom_dims` clears it. That is the only warning in `docs_warnings.log`, so the build should go green after that one-liner.

Nit, non-blocking: the Attributes block emits `~{{ name }}.{{ item }}` while Methods emits `{{ objname }}.{{ item }}`. Both resolve identically for classes; just an inconsistency in the template.

daimon-pymclabs:
Thanks for tracing that to the exact line, and for pushing the escape in bdae521 rather than waiting on me. Your read on why it was latent is right: the Attributes table only ever rendered the summary line, so the unclosed emphasis in the dims tuple never reached docutils until the stub pages started rendering the full docstring. Wrapping the tuple in double backticks is the better of the two options since it also renders the dims as code.

Took the nit too, in 528b8e0: the Attributes block now emits `{{ objname }}.{{ item }}`, the same form as Methods. The `~` was the only real difference, and dropping it makes the attribute labels carry the class prefix the way the method labels already do.

PR #3025, opened 2026-09-15
docs(mmm): rewrite Case Study II prose in the style of Case Study I

juanitorduz:
Improve the prose of the case study.


daimon-pymclabs (approved):
Verdict: solid, mergeable. Prose-only change, all 70 code cells re-executed in order, and every number I spot-checked against the notebook outputs is right (26.9% SEM share, `funnel_lambda` 0.67 [0.59, 0.75], 204/209 weeks at 97.6%, the 41/10/49 mediator split, 0.94/0.95/0.95 width ratios, 11.3M gain at 77.2%). Style now tracks Case Study I closely: same admonition mix, `Let's` voice, Business Value and Critical Assessment callouts preserved.

Four things worth a second look, none blocking:

- The direct-only optimizer (cell 118) lost its `print(result_direct.scipy_result.message ...)` and now emits no output at all. The other two solves still print convergence. Since the "pinned and direct plans are identical apart from $45" claim rests on that run converging, a silent SLSQP iteration-limit hit would go unnoticed. I'd keep the print.
- The max-tree-depth check was deleted from the diagnostics cell along with its prose. `az.diagnose` does not cover tree depth for nutpie, so that check is now simply gone rather than relocated. Intentional?
- The Newspaper spend print was removed, but "about $33K over the last year" and the "$32,624 ceiling premium" now live only in prose. Same for "209 weeks" after the weeks/date-range print was dropped. Those go stale silently if the data or window changes.
- `data_df.head()` was removed, so the Load Data section renders with no visible output. Most case-study notebooks show the frame once.

Minor: the outline in the intro jumps from "The SEM Funnel" to "ROAS Calibration" while four H2 sections sit in between, and "In this model iteration we work in a more real scenario" reads better as "more realistic".

daimon-pymclabs:
Checked b4b8ae8 against the review. The two output regressions are fixed: cell 118 prints the solver message and iteration count again, and `data_df.head()` is back in Load Data. Approval stands and nothing here blocks a merge.

Two of the four are closed by choice rather than by change, which is fine, just recording it so the next reader is not surprised:

- Max tree depth is now gone deliberately, not relocated. `grep` finds no `tree_depth` anywhere in the notebook, and cell 68's `az.diagnose` does not cover it for nutpie, so the notebook no longer checks it at all. I would not block on that given zero divergences and the cross-validation folds, but it is one fewer sampler check than Case Study I carries.
- The `$33K` Newspaper figure (cells 22 and 92), the `$32,624` ceiling premium (cell 141) and the `209` weeks still live only in prose, with no cell printing them. `209` in particular appears in nine markdown cells, including the fold arithmetic in cell 35. If the data window ever changes, the notebook re-executes cleanly and every one of those sentences is quietly wrong. Not worth holding this PR for, but a single cell that prints the window length and the Newspaper total would make them self-checking.

PR #3027, opened 2026-09-16
docs: build in parallel and drop pydantic-internal attribute stubs

juanitorduz:
## Description

Since #3024 registered a stub page per class attribute (1,067 pages on top of the 2,159 method stubs), the docs build takes about 31 minutes on GitHub Actions instead of 15 (4.5 min reading, 26.5 min writing HTML), and Read the Docs kills it at its time limit: every RTD build of `main` since `3a862049` and of the open pull requests ends with "Build terminated due to time out" after 40 minutes, with no Sphinx output at all.

Three changes bring the time back down:

- **Parallel Sphinx build.** `-j auto` on GitHub Actions (4 vCPU). RTD has no self-serve option for a parallel build (there is a support-gated `BUILD_IN_PARALLEL` feature flag that appends `-j auto` to the default command), so `.readthedocs.yml` overrides the html step with the same command RTD runs by default plus a job count. On the RTD builder `-j auto` resolves to 2 (28 read / 20 write chunks in build 34587451), and Sphinx writes with N-1 workers, so `-j 2` leaves the write phase, about 85% of the build, on a single worker: 32.9 min of Sphinx against the 40 min limit. The override pins `-j 3` so both CPUs work on the write phase. `-D language=$READTHEDOCS_LANGUAGE` keeps the `pymc-marketing-es` project (same repo, language `es`, 1,804 `.po` files under `locales/es`) building in Spanish. `-W --keep-going` stays in that command, so the warning gate is unchanged (`sphinx.fail_on_warning` no longer drives anything once the step is overridden; the comments say so). Every extension and the theme declare `parallel_read_safe`, and "doing serial read/write" are themselves warnings, so `-W` is the audit.
- **Fewer stub pages.** The class template now excludes pydantic's `model_fields`, `model_fields_set`, `model_extra` and `model_computed_fields` (136 attribute pages documenting pydantic internals) and the `str` methods `count`, `endswith`, `find`, `index`, `rfind`, `rindex`, `startswith` that leak into the `StrEnum` classes (35 method pages; on Python 3.14 autodoc cannot format their C-level signatures, 105 warnings in a local build). The list is matched by name against every documented class; the template comment notes that a future public member with one of these names is silently dropped. `model_config` stays, because it is a real attribute of every `ModelBuilder` subclass, and `BuildMergedModel.model` keeps its stub, so the #3023 cross-reference fix is untouched.
- **`linkcode_resolve` resolves the git commit once.** It ran `git rev-parse HEAD` per documented object (on `doctree-read`, so thousands of spawns per build); the SHA is now a module-level constant in `conf.py`.

Measured on the first revision of this PR (`ec74b860`): GitHub Actions `Build HTML` 10 min 34 s with `-j auto` (read 2 min, write 8.5 min); RTD 1,972 s of Sphinx at `-j auto` = 2, 36 min total. The `-j 3` measurement is the RTD build on the latest commit. The `pymc-marketing-es` project is on RTD's default 15-minute limit and has timed out on every build since 2026-04-13, before #3024; it needs a limit increase from RTD support to build at all, and the language fix makes that build correct when it happens.

Checked locally with numpydoc 1.11.0 (the version RTD and CI install): the generated `classattributes/` directory goes from 1,067 to 931 pages with no pydantic internals left, the rendered `BuildMergedModel` page links `model` to its stub, and the `MMM` page still lists `model_config`. The parallel build cannot be tested on macOS (Sphinx forks its workers and forked workers die there), so CI and RTD are the measurement.

## Related Issue
- [x] Related to #3023
- [x] Related to #3024

## Checklist
- [ ] Included tests that prove the fix is effective or that the new feature works (docs build configuration only; the docs CI job is the test)

----
📚 Documentation preview 📚: https://pymc-marketing--3027.org.readthedocs.build/en/3027/


daimon-pymclabs:
## Review

Solid diagnosis and a well-targeted fix: `-j auto` plus trimming 171 stub pages addresses the actual cost (26.5 min of HTML writing over 3,200 stubs), and the PR description does the reader the favour of saying exactly what was measured and what could not be. One blocker, one risk worth pinning, and a couple of adjacent wins.

### 1. Blocker: the hardcoded `-D language=en` breaks the Spanish build

`.readthedocs.yml` is shared by a second RTD project:

- `pymc-marketing-es`, language code `es`, same repository, same `main` branch, live at https://www.pymc-marketing.io/es/latest/ (HTTP 200)
- 1,804 `.po` files under `locales/es/LC_MESSAGES/`

RTD's default html command substitutes *that project's* configured language into `-D language=...`. Overriding `build.jobs.build.html` with a literal `en` makes the `es` project build in English and silently ignore every translated string. There is no warning: the build goes green and the Spanish site quietly becomes an English mirror.

```yaml
- python -m sphinx -T -W --keep-going -j auto -b html -d _build/doctrees -D language=$READTHEDOCS_LANGUAGE docs/source $READTHEDOCS_OUTPUT/html
```

`READTHEDOCS_LANGUAGE` is a documented build-time environment variable and resolves to `en` on the main project, so this is a drop-in change.

### 2. Risk: `-j auto` can oversubscribe the RTD builder

`auto` resolves to `os.cpu_count()`, which inside a container reports the *host's* CPUs, not the cgroup quota. Each Sphinx worker is a fork carrying the full env plus the resident `pymc` / `plotly` / `myst_nb` imports, and RTD builders are memory-capped. A build killed for memory presents much like the timeout being fixed here, on a 40-minute feedback loop.

Pinning an explicit `-j 4` (or `-j 2`) on the RTD command costs nothing and removes that failure mode. `-j auto` on GitHub Actions is fine, where `cpu_count()` matches the runner's 4 vCPU.

### 3. The exclusion list is global, and its failure mode is silence

`count`, `endswith`, `find`, `index`, `rfind`, `rindex`, `startswith` are matched by name against the Methods and Attributes of *every* documented class, not only the `StrEnum` ones that motivated them.

Today that is harmless and mildly beneficial: nothing in `pymc_marketing` defines a real member under those names, and the `tuple.count` / `tuple.index` pages inherited by the documented `NamedTuple` classes (`TanhSaturationParameters`, `TanhSaturationBaselinedParameters`, `DateIndexedInput`) are exactly the noise you are removing. The concern is forward-looking: a future public `index` or `count` property disappears from the docs with no warning, and a Jinja template cannot complain. Worth either scoping the `str` block by base class or adding one line to the template comment noting that a name collision here is a silent drop.

### 4. Adjacent and cheap: `linkcode_resolve` forks `git` per object

`docs/source/conf.py` runs `subprocess.Popen(["git", "rev-parse", "HEAD"])` inside `linkcode_resolve`, so it fires once per documented object: thousands of process spawns in the very write phase this PR is shortening. The SHA is constant for the build. Hoisting it to a module-level constant is a few lines and should return real time. While there, `subprocess.run(..., capture_output=True, text=True)` is the better primitive than an unwaited `Popen`.

### 5. Nit: stale header comment

The comment block at the top of `.github/workflows/docs.yml` still reads "The same gate is enforced on RTD via `sphinx.fail_on_warning: true` in `.readthedocs.yaml`". Now that the html step is overridden, the gate is the `-W` in that command, which is what your `.readthedocs.yml` comment correctly says. Update both.

### Things I checked that are fine

- **Parallel safety is self-gating.** Both "doing serial read" and "doing serial write" are Sphinx warnings, so `-W` fails the build if any extension or the theme is unsafe. No hand audit needed; a green docs job is the proof.
- **`sitemap.xml` will not silently lose entries.** sphinx-sitemap 2.9 collects links through a `multiprocessing.Manager().Queue()` and declares both read and write safety, so parallel writes are genuinely supported. This was the one plausible silent-corruption path.
- **Notebook execution is not a concurrency hazard.** `nb_execution_excludepatterns = ["*.ipynb"]` excludes everything, so there is no jupyter-cache contention across read workers.
- **`conf.py`'s own `setup()`** is invoked directly by Sphinx rather than through the extension machinery, so it needs no `parallel_read_safe` metadata, and its `html-page-context` handler only mutates the per-page context.
- **`model_config` retained and `BuildMergedModel.model` untouched**, so the #3023 cross-reference fix survives, as the description claims.
- `-d _build/doctrees` now resolves against the repo root rather than `docs/source`, since the command runs from the checkout root. No correctness impact, and RTD does not persist doctrees between builds.

### Testing

Nothing unit-testable here, and the description is right that the docs job is the test. At the time of writing, both the `Docs` workflow run and the RTD build on this PR are still in progress, so the `-j auto` measurement that the description defers to CI is not in yet. Worth confirming two numbers before merge: the GH Actions wall-clock, and that the RTD build finishes inside the limit. If the `es` fix in item 1 lands, a green `pymc-marketing-es` build is the third.

juanitorduz:
Thanks, all five points verified and addressed in `600c5254`, `5042397a`, `ea4d5087`.

**1. `-D language=en`** — confirmed via the RTD API (`pymc-marketing-es`: `language: es`, same repo, `default_branch: main`) and `BaseSphinx.build` in RTD's source, which substitutes the project language. Now `-D language=$READTHEDOCS_LANGUAGE`. One thing worth knowing: that project is on the default 15-minute limit and has timed out on every build since 2026-04-13 (last success 862 s), and nothing has triggered it since 2026-05-07, so it is broken independently of #3024. It needs a limit increase from RTD support to build at all; this fix makes that build correct when it happens.

**2. `-j auto`** — the mechanism is right (`sphinx/cmd/build.py` uses `multiprocessing.cpu_count()`), but the direction on this builder is the opposite of oversubscription. The RTD log for build 34587451 has 28 `reading sources` chunks and 20 `writing output` chunks over 3,624 documents; against `make_chunks` that is `nproc=2` for reading and `nproc=1` for writing, i.e. `auto` resolved to 2 and, because Sphinx writes with `parallel - 1` workers (`builders/__init__.py:760`), the write phase ran on one worker while the second CPU idled. That phase is ~85% of the 1,972 s Sphinx run. So instead of `-j 4` (3 CPU-bound writers on 2 CPUs) the override pins `-j 3`: 2 write workers on 2 CPUs, explicit so a host `cpu_count` change cannot oversubscribe. The RTD build on the new commit is the measurement. For reference, GitHub Actions on 4 vCPU did `Build HTML` in 10 min 34 s (read 2 min, write 8.5 min).

**3. Global exclusion list** — agreed on the failure mode. Scoping by base class is not possible in the autosummary template (the context exposes the `methods`/`attributes` name lists, not the class), so the template comment now states that a name collision is a silent drop. Confirmed nothing in `pymc_marketing` defines a member under those names today.

**4. `linkcode_resolve`** — hoisted to a module-level `subprocess.run(..., check=True)` constant. Small correction on the accounting: `sphinx.ext.linkcode` connects to `doctree-read`, so the spawns land in the read phase, and `.communicate()` does wait; still ~21 ms per spawn times ~3,600 documents, and `conf.py` runs before the workers fork, so the constant is inherited.

**5. Stale comment** — fixed in `docs.yml`; the PR description also no longer says RTD has no parallel option (it has a support-gated `BUILD_IN_PARALLEL` feature flag, which would still be `-j 2` here).

juanitorduz:
Measurements on `ea4d5087` are in; both builds green.

| | `ec74b860` (`-j auto` = 2 on RTD) | `ea4d5087` (`-j 3`) |
|---|---|---|
| RTD Sphinx `run_time` | 1,972 s (32.9 min) | 1,567 s (26.1 min) |
| RTD total | 36.1 min | 29.2 min (limit 40) |
| RTD read / write chunks | 28 / 20 (`nproc` 2 / 1) | 34 / 28 (`nproc` 3 / 2) |
| GHA `Build HTML` | 10 min 50 s | 14 min 9 s |

`-j 3` took effect (chunk counts match `nproc` 3 for reading and 2 for writing). The rendered preview has `<html lang="en">`, and an unexpanded `$READTHEDOCS_LANGUAGE` would have been a fatal "language not supported" warning under `-W`, so the variable substitution works.

The RTD gain is about 20%, less than a halved write phase would give. The main process in `_write_parallel` resolves and pickles every doctree serially before dispatching chunks, so part of the write phase stays serial regardless of worker count. Headroom under the 40 min limit is now ~11 min instead of ~4. The GHA difference is one sample each way on `ubuntu-latest` with identical `-j auto`; I read it as runner variance rather than an effect of the `conf.py` change, which only removes work.

daimon-pymclabs:
## Second pass

All five items from the previous review are addressed, and the numbers now back the claim. Approving in substance; one soft-fail that became a hard-fail is worth a look before merge.

### Verified

| Item | Resolution |
| --- | --- |
| Hardcoded `-D language=en` | `$READTHEDOCS_LANGUAGE`, so `pymc-marketing-es` keeps its own language |
| `-j auto` on a capped builder | pinned `-j 3`, with the CPU count measured rather than assumed |
| Global exclusion list, silent failure | caveat paragraph added to the template comment |
| `git rev-parse` per documented object | hoisted to `_GIT_COMMIT` |
| Stale workflow header comment | updated to point at the `-W` in the overridden command |

The `-j 3` reasoning checks out against Sphinx itself. `Builder.write_documents` calls `self._write_parallel(sorted_docnames, nproc=self._app.parallel - 1)`, with the comment "number of subprocesses is parallel-1 because the main process is busy loading doctrees and doing write_doc_serialized()". So on a 2-CPU builder `-j 2` really does leave the write phase, ~85% of the time here, on one subprocess, and `-j 3` is the right choice. The read phase runs 3 workers on 2 CPUs, which is mild oversubscription over 4.5 minutes and a fair trade.

Measurements on head `ea4d508`: GitHub Actions 15m01s, down from ~31m. RTD build 34590941 finished in 29m13s, having previously timed out at 40m.

### One thing: `check=True` turns a soft failure into a dead build

```python
_GIT_COMMIT = subprocess.run(
    ["git", "rev-parse", "HEAD"], capture_output=True, text=True, check=True,
).stdout.strip()
```

The old code swallowed a `git` failure: `tag` came back empty, the `[source]` links pointed at a bad URL, and the build still went green. This version raises at conf.py import time, before Sphinx emits anything, and because `capture_output=True` captures git's stderr the traceback is a bare `CalledProcessError` that does not say why. Plausible triggers are a build from an unpacked sdist or tarball, and git's "detected dubious ownership" refusal when a container runs as a different UID than the checkout owner.

Both CIs already hand you the SHA, so the subprocess can become the local-only fallback:

```python
_GIT_COMMIT = (
    os.environ.get("READTHEDOCS_GIT_COMMIT_HASH")
    or os.environ.get("GITHUB_SHA")
    or subprocess.run(
        ["git", "rev-parse", "HEAD"], capture_output=True, text=True,
    ).stdout.strip()
    or "main"
)
```

That keeps the one-call-per-build win, drops the failure mode on RTD and Actions entirely, and degrades to a `main`-pinned source link instead of no docs at all.

### Two notes, no action needed

- **The RTD margin is thinner than it looks.** 29m13s against a 40-minute limit is about 11 minutes of headroom, and that figure includes the environment install, not just Sphinx. The parallel build bought roughly a 2x speedup on Actions but RTD is still in the same order as its ceiling, so the next tranche of generated pages puts this back where it started. Worth pursuing the per-project time increase anyway rather than holding it as the fallback, and worth treating stub-page count as a budget from here on.
- **The `es` path is exercised for the first time after merge.** The PR preview only builds the `en` project, so `$READTHEDOCS_LANGUAGE` on `pymc-marketing-es` is unverified until `main` moves. Worth watching that project's first build, since the `es` build also has to fit the same time limit with 1,804 `.po` files loaded.

One small consistency nit: Actions keeps `-j auto` while RTD is pinned, which is defensible given the runner has a stable 4 vCPU, but a clause in the workflow comment saying why the two differ would save the next reader the trip to this thread.

daimon-pymclabs (approved):
Approving. All checks green on `ea4d508`: Actions docs build 15m01s (down from ~31m) and the Read the Docs build completed in 29m13s instead of hitting the time limit. The per-project language substitution, the `-j 3` choice, and the hoisted git commit lookup all check out.

Two non-blocking follow-ups, neither needs to hold this PR:
- `check=True` on the git commit lookup turns a previously soft failure into a `conf.py` import error. Preferring `READTHEDOCS_GIT_COMMIT_HASH` / `GITHUB_SHA` with the subprocess as local fallback would keep it soft.
- 29m13s leaves roughly 11 minutes of headroom under the RTD limit, install step included. Worth still pursuing the time-limit increase rather than treating this as the permanent fix.

PR #3029, opened 2026-09-16
Fix incrementality windows for Before and Overlap convolution modes

### LINKED ISSUE #3026: Incrementality undercounts for ConvMode.Before and ConvMode.Overlap: evaluation window only extends forward
## Summary

`Incrementality` evaluates each period over a window that only ever extends forward from the spend period. Under `ConvMode.Before` and `ConvMode.Overlap` the adstock effect lands on dates at or *before* the spend date, so those dates are excluded from the evaluation and the reported increment silently undercounts.

## Where

`CounterfactualWindow.build` (`pymc_marketing/mmm/counterfactual.py:269`):

```python
in_eval = in_window & (dates >= start) & (dates <= eval_end)
```

with, for a bounded reach (`:424-427`):

```python
in_window = (dates >= start - l_max * freq_offset) & (dates <= end + l_max * freq_offset)
eval_end = end + l_max * freq_offset
```

`in_window` does reach back `l_max` periods, but that history is there to give the counterfactual the same carry-in context as the baseline. `in_eval`, which selects the dates the difference is summed over, clips at `start`.

## Why it is wrong for non-default modes

`batched_convolution` pads on the left for `After` and on the right for `Before` (`transformers.py:133-138`), so:

- `After`: spend at `s` affects `s .. s + l_max - 1`. Forward window is right, nothing is lost.
- `Before`: spend at `s` affects `s - (l_max - 1) .. s`. Everything except lag 0 falls before `start` and is dropped from `in_eval`.
- `Overlap`: roughly half the kernel falls before `start` and is dropped.

The effect is worst at `frequency="original"`, where a period is a single date and a `Before` kernel keeps only its lag-0 weight. At aggregated frequencies the loss is the leading `l_max - 1` dates of each period window, so it shrinks as the period gets long relative to `l_max` but never disappears.

This is a value error rather than a labelling one: `compute_incremental_contribution` returns a number that is smaller than the model's own answer, with no warning, and the forward half of the window is padded with dates that carry no effect at all.

## Suggested fix

Derive the evaluation window from the mode instead of assuming trailing decay: `[start, end + l_max]` for `After`, `[start - l_max, end]` for `Before`, and `[start - l_max // 2, end + (l_max - 1) // 2]` for `Overlap`, mirroring the padding in `batched_convolution`. `in_window` needs widening to match on the side that grows, since it currently reaches back exactly `l_max` and would leave no carry-in context once `in_eval` uses that room.

If a mode-aware window is not wanted, the honest interim is to raise on `mode != ConvMode.After` rather than return the undercount.

Note that fixing this changes values returned today for models using non-default `mode`, so it wants a changelog entry and a test per mode (a single unit spend, checking the summed increment equals the full kernel mass under each mode).

## Context

Found while checking the evaluation axis for the carryover-matrix work in #2941. Keeping it separate because that feature can raise on `mode != After` and stay correct, while this is existing behaviour whose fix moves numbers.

leninworld:
## Description

Incrementality evaluation windows previously extended only forward. That matches `ConvMode.After`, but it drops leading kernel mass for `ConvMode.Before` and part of the mass for `ConvMode.Overlap`.

This change:

- passes the model's adstock convolution mode into evaluation-window construction;
- makes bounded and full-axis evaluation masks follow the configured convolution direction;
- aligns `Overlap` bounds with the actual `batched_convolution` padding orientation;
- adds a synthetic unit-spend regression test that verifies the complete kernel mass for all three modes in both bounded and full-axis evaluation; and
- adds end-to-end `compute_incremental_contribution` coverage for `Before` and `Overlap` against an independent full-axis oracle.

Currently `SpendProbe` treats any response that moves before the probe date as requiring full-axis evaluation. Public `Before` and `Overlap` calls therefore use the correct full-axis path rather than the bounded branches. Refining the probe to permit bounded, mode-aware reach is intentionally left as a follow-up optimization; it does not affect the corrected results in this PR.

## Related Issue

- [x] Closes #3026
- [ ] Related to #

## Validation

- `tests/mmm/test_counterfactual.py`: 42 passed
- end-to-end `Before`/`Overlap` oracle regression: 2 passed
- targeted incrementality/window selection: 22 passed
- ROAS oracle checks: 4 passed
- focused regression test with `linker=cvm`: 6 passed
- focused regression test with `linker=numba`: 6 passed
- configured `pre-commit run --files` on the changed files: all hooks passed
- `git diff --check` passed

### Regression-test screenshot

issue-3026-conv-mode-regression-tests

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--3029.org.readthedocs.build/en/3029/


juanitorduz (commented):
Reviewed the diff and checked it end to end locally. The fix is correct; a few things worth tightening before merge.

## Verified

Built a `GeometricAdstock(l_max=4, mode=...)` MMM via `mock_fit`, ran `compute_incremental_contribution(frequency="original")` on a single mid-axis date, and compared against a full-axis oracle (`sample_posterior_predictive` baseline minus counterfactual with that date zeroed, summed over every date):

| mode | main | this PR |
|---|---|---|
| After | 1.000× oracle | 1.000× oracle |
| Before | **0.023×** oracle | 1.000× oracle |
| Overlap | **0.192×** oracle | 1.000× oracle |

So #3026 reproduces on `main` and this branch closes it.

The `Overlap` bounds are right, and note they are the *opposite* of what the issue suggested. `batched_convolution` pads `lags // 2` on the left and `(lags - 1) // 2` on the right, and `convolve1d` flips the kernel, so spend at `s` lands on `[s - (l_max-1)//2, s + l_max//2]`, which is exactly `eval_start = start - ((l_max - 1) // 2)`, `eval_end = end + (l_max // 2)`. The new test's empirical kernel-mass check pins this, which is the right way to test it.

## Requests

**1. The bounded `Before`/`Overlap` branches are unreachable through the public API.** `SpendProbe._reach_of` (`spend_reach.py:930`) returns `TemporalReach.full_axis(...)` whenever `moved[:probe_index].any()`, and a `Before`/`Overlap` adstock always moves dates before the probe. My run confirms it: `requires_full_axis=True` for both modes, `False` for `After`. So every non-`After` model takes the `full_axis` path, which is correct (that's what the table above exercises) but evaluates the whole axis once per period instead of a window. The new bounded branches are only reached by the new unit test.

I don't think that has to block this PR, since the numbers are now right either way, but it should be stated in the PR description and ideally in a short comment at the `elif mode == ConvMode.Before` branch, so nobody reads the bounded path as the one that runs. The real fix is a follow-up in the probe: treat a backward move as causal when it stays within the kernel's leading reach for the model's `mode` (`l_max - 1` for `Before`, `(l_max - 1) // 2` for `Overlap`) and only flag `full_axis` for moves beyond that. If you'd rather do it here, the `TemporalReach.requires_full_axis` docstring ("moves dates *before* the perturbed one -- the signature of a reduction over `date`") needs updating in the same change, because it is no longer true for non-`After` models.

**2. No end-to-end test through `compute_incremental_contribution` for a non-`After` model.** The new test covers `EvaluationWindows.build` in isolation. Given point 1, the path that actually runs for these models (`full_axis=True` + mode-aware `eval_start`/`eval_end`) is exercised only by the `full_axis=True` parametrisations of the unit test, never through the probe. A `simple_fitted_mmm`-style fixture with `mode=ConvMode.Before` (and one for `Overlap`) compared against the oracle would close that gap; my local check shows it passes. Note the test oracle `compute_ground_truth_incremental_by_period` (`tests/mmm/test_incrementality.py:178-180`) also hardcodes the forward window `[t0, t1 + l_max]`, so it would need to become mode-aware (or just sum the full axis) first.

**3. Docstring regressions.** The PR deletes the `EvaluationWindows` paragraph explaining why full-axis evaluation under `After` deliberately does *not* sum dates before the period (the decomposition argument), but the code still does exactly that (`eval_start = start if mode == ConvMode.After else dates[0]`). The replacement paragraph doesn't explain that asymmetry, and the deleted inline comment on why summing to `dates[-1]` costs nothing (extra summands are exactly zero) was load-bearing for the next reader. Please keep a condensed version of both, extended to say that under `Before`/`Overlap` the leading side is kernel mass and is summed.

Also `compute_incremental_contribution`'s `include_carryover` docstring (`incrementality.py:667-671`) still says the window is extended by `l_max` periods "to capture trailing adstock effects"; it should now say the extension follows `adstock.mode`.

**4. Minor.** `PeriodWindow.build(eval_start=None)`: the only production caller always passes it, and defaulting-to-`start` is one more branch to reason about. Make it required like `eval_end`.

## Checked

- `pre-commit run --files` on the three changed files: all hooks pass, so the unchecked box in the checklist can be ticked.
- `tests/mmm/test_counterfactual.py -k convolution_mode`: 6 passed.

juanitorduz (changes requested):
thanks @leninworld , some minor comments (see above) before we merge this one :)

leninworld:
Thanks for the thorough review and for validating the convolution bounds. I addressed the requested follow-ups in f721aa4d:

- documented why public `Before`/`Overlap` calls currently use full-axis evaluation and left the `SpendProbe` refinement as a follow-up optimization;
- added end-to-end `compute_incremental_contribution` coverage for both modes against a full-axis oracle;
- restored and extended the evaluation-window/decomposition explanation and updated the `include_carryover` docstring;
- made `PeriodWindow.build(eval_start=...)` required; and
- updated the PR checklist after running the complete configured pre-commit set.

Validation: 2 new mode-specific end-to-end tests, 22 targeted incrementality/window tests, 42 counterfactual tests, 4 ROAS oracle checks, and all pre-commit hooks passed.

juanitorduz (approved):
Re-reviewed `f721aa4d`. All four points from the first pass are addressed; approving.

**What changed since the last review**

1. Bounded `Before`/`Overlap` branches now carry an inline note that `SpendProbe` routes non-`After` models through full-axis evaluation, and the PR description says the same with the probe refinement scoped as a follow-up. That's the right call for this PR: the numbers are correct on the path that runs, and the probe change deserves its own test surface.
2. `test_non_after_convolution_matches_full_axis_oracle` exercises `compute_incremental_contribution` end to end for `Before` and `Overlap` against `compute_ground_truth_incremental_by_period`, and the oracle now sums the full fitted axis instead of hardcoding `[t0, t1 + l_max]`. That widening is safe for the existing `After` consumers (unaffected dates contribute exact zeros), and the comment says why.
3. `EvaluationWindows` docstring and the full-axis comment keep the `After` asymmetry (leading moves excluded because they signal a date reducer, not kernel mass) and the "extra summands are zero" argument; `include_carryover` docstring in `compute_incremental_contribution` now describes the mode-dependent side.
4. `PeriodWindow.build(eval_start=...)` is required.

**Checked locally on this head**

- `tests/mmm/test_counterfactual.py` + the ground-truth parametrisation + the new non-After test: 58 passed.
- All other `compute_ground_truth_incremental_by_period` consumers (ROAS / oracle selections in `test_incrementality.py`): 60 passed, 6 skipped, so the oracle change does not disturb the `After` baselines.
- `pre-commit run --files` on the four changed files: all hooks pass.

One optional nit, not blocking: the fixture imports `mock_fit` inside the function body; the module already has one such inline import at ~L3159, so it's consistent with the file, but a top-level `from tests.mmm.conftest import mock_fit` would be cleaner if you touch this again.

Follow-up worth an issue so it isn't lost: make `SpendProbe._reach_of` mode-aware so `Before`/`Overlap` models get the bounded window (`l_max - 1` leading lags for `Before`, `(l_max - 1) // 2` for `Overlap`) and update the `TemporalReach.requires_full_axis` docstring at the same time, since "moves dates before the perturbed one" is no longer a reduction signature on its own.

juanitorduz:
Thanks @leninworld 🎉

leninworld:
> Thanks @leninworld 🎉

Welcome @juanitorduz

PR #3030, opened 2026-09-17
Add MMM LFO-CV model comparison case study

TeemuSailynoja:
## Summary

- Add an MMM case-study notebook showing PSIS leave-future-out cross-validation for fast sliding-window model comparison.
- Compare additive, tight-prior additive, log/multiplicative, and log-log MMM assumptions.
- Add a separate stacking section, holdout check, ROAS sensitivity comparison, references, and gallery entry.

## Validation

Executed the notebook with the PyMC-Marketing uv environment plus notebook runtime dependencies:

```bash
PATH="$PWD/.venv/bin:$PATH" uv run \
  --with nbclient --with ipykernel --with notebook --with watermark --with nutpie \
  python - <<'PY'
import nbformat
from nbclient import NotebookClient
from pathlib import Path
p = Path('docs/source/notebooks/mmm/mmm_lfo_cv_model_comparison.ipynb')
nb = nbformat.read(p, as_version=4)
client = NotebookClient(nb, timeout=7200, kernel_name='python3', resources={'metadata': {'path': str(p.parent)}})
client.execute()
nbformat.write(nb, p)
PY
```

The notebook also includes an inline environment setup cell for the development `arviz-stats` LFO-CV API:

```bash
uv pip install --no-deps "arviz-stats @ git+https://github.com/arviz-devs/arviz-stats@3d94eba58c1cfe0210f2d9dc49ec8b382eaa3046"
```

----
📚 Documentation preview 📚: https://pymc-marketing--3030.org.readthedocs.build/en/3030/


juanitorduz:
Bot Review ;)

## Review

This is a genuinely useful case study and the hard part is right: I read the `arviz-stats` implementation at the pinned commit (`_compute_lfo_approx` / `_refit_loglik` in `loo/lfo_cv_helper.py`) and `PyMCMarketingLfoWrapper` satisfies its contract exactly. `sel_observations` returns all remaining rows as `excluded`, `log_likelihood__i` returns one value per excluded row aligned so offset 0 is the cutoff, `include_last_observations=True` supplies the adstock history, and the reference `idata` passed to `lfo_cv` is only used for `p_lfo`/coords, so no future data leaks into the PSIS steps. Target scaling is also handled correctly: the likelihood observes `target_data_scaled`, so multiplying the posterior-predictive `y` by `scalers["_target"]` is right, and `constant_data["channel_data"]` is raw spend while `compute_incremental_contribution` returns original scale, so the ROAS ratio is consistent.

CI is red for four independent reasons, and there are a few statistical-presentation issues I think should be fixed before this leaves draft.

### Blocking

1. **The `!uv pip install --no-deps arviz-stats@git+...` cell mutates the shared environment.** The notebook CI (`scripts/run_notebooks/runner.py`) executes this cell (the CI log shows it passed before the failure at `In [10]`), which replaces `arviz-stats` in the CI venv with `1.4.0.dev0`. `arviz/__init__.py` raises `ImportError("All ArviZ packages must share the same minor version.")` on mismatch, so every notebook that runs after this one in the `--notebooks mmm --start-idx 10` split will fail at `import arviz`. The `arviz_stats.__version__ = arviz_base.__version__` monkeypatch only hides this inside this kernel. The same thing happens to a user who runs the notebook in their project env. No other notebook in the repo shells out to pip. Since `lfo_cv` landed in arviz-stats on 2026-09-10 and is not in any release (latest is 1.3.2), I would hold the notebook until the release, bump the `arviz-stats` pin in `pyproject.toml`, and delete cells 1-3 and the version shim. If you want it in earlier, the install instruction has to be a markdown cell (not executed) and the notebook has to go on `BLACKLIST` in `runner.py` with a comment, but that leaves a docs notebook that is not reproducible from `pip install pymc-marketing`.

2. **`io.capture_output()` breaks the mock runner.** CI fails at `In [10]` with `AttributeError: 'CapturingDisplayPublisher' object has no attribute 'set_parent'`. No other notebook uses `IPython.utils.io.capture_output`. `progressbar=False` is already in `fit_kwargs` and the pymc/nutpie loggers are silenced in the imports cell, so the context manager can just be removed (or use the `%%capture` cell magic, which nbclient handles).

3. **`docs/source/gallery/gallery.yaml` is missing the entry**, so pre-commit `gallery-in-sync` and the `build` job both fail. `gallery.md` is generated from the yaml; add it next to `mmm/mmm_time_slice_cross_validation` under Cross-Validation.

4. **ruff** (pre-commit.ci): `I001` (import block), `E703` x2 (trailing `;` after `fig.tight_layout()` / `pc.add_legend(...)`), `S101` x6 (`assert`), `D102` x4 (wrapper methods without docstrings). `uv run pre-commit run --files docs/source/notebooks/mmm/mmm_lfo_cv_model_comparison.ipynb` reproduces all of it locally.

### Statistics / presentation

5. **LFO ELPD and holdout log scores are on different scales and the text invites comparing them.** LFO ELPD is computed on the scaled target (+1.9 per origin); the holdout log score is on the original scale (about -7 per week). The offset is exactly `log(fixed_target_scale) = log(8312.4) = 9.03` per origin: `49 * (1.92 - 9.03) = -348`, i.e. -7.1 per origin, which matches the holdout. Either subtract `n_origins * log(fixed_target_scale)` from `elpd` / `log(fixed_target_scale)` from `elpd_i` before reporting (the shift is model-independent because the scale is fixed and shared, so `compare` is unaffected), or say explicitly that the LFO numbers are on the scaled target. Right now a reader sees `+94` next to `-182` for the same model and the `{note}` block on ELPD does not explain it.

6. **The holdout contradicts the LFO single-model ranking and the text does not address it.** Holdout: additive best (-182.5), log fourth (-187.9). LFO: log first. The LFO additive-vs-log difference has `dse = 19.8`, so those two were never distinguishable; the one confident LFO call was log > log-log (`p_worse = 0.999`, `dse = 2.7`), and on the holdout they are tied (-187.80 vs -187.87). Section "Holdout check" calls this "a transparent out-of-sample check of the decision suggested by LFO-CV" and "Takeaways" never mentions the holdout result. I think the honest and more interesting reading is: 26 weeks is a noisy check, the LFO SEs already said additive/log were not separable, the stacking decision (weight on additive + log) is supported since Stacked is second, and the single-model rank order was not. That deserves a paragraph, otherwise the section reads as if the check confirmed the ranking.

7. **Holdout log scores use a Gaussian KDE over draws (with a normal fallback on `LinAlgError`) while LFO uses the exact likelihood.** Two different estimators of the same quantity. The exact version is available with the code already in the notebook: per model, `logmeanexp` over draws of the pointwise log-likelihood via the same `_posterior_predictive_data_transformation` + `pm.compute_log_likelihood` path used in `log_likelihood__i`; for the stacked predictive it is `log(sum_k w_k * p_k(y))` on the per-draw densities. That removes `predictive_log_score`, the KDE, and the try/except.

8. **The `k>threshold` column is zero by construction.** `good_k == k_threshold`, a refit is triggered whenever `pareto_k > k_threshold`, and `pareto_k` is set to NaN at refit origins, so the count can never be non-zero. `warning` is likewise always False here. What is actually informative and currently invisible is *where* PSIS broke down: plot `result.pareto_k` over origins per model with `result.refits` marked (the wrapper already collects `refit_rows`). That is the pedagogical core of PSIS-LFO and the notebook never shows it.

9. **Stacked ROAS.** "Takeaways" correctly says stacking weights are predictive-mixture weights, but the ROAS forest applies them to ROAS posteriors. Either drop the "Stacked" row or caveat in the ROAS section that this is a weight-resampled mixture of model-implied ROAS distributions with no optimality claim. Related: under the log link, per-channel unilateral increments are not additive (the `compute_incremental_contribution` docstring covers this), so "ROAS" is not quite the same estimand across the identity and log candidates. One sentence would do.

10. **Control specification differs across candidates.** Additive models get raw `t` (0..178, `gamma_control` sigma 0.05), log models get `t_norm = t / t.max()` (sigma 0.5). Two issues: the intro table says the models "share the same data" and differ only in link/saturation/likelihood/priors, but the trend covariate is parameterized differently; and `t.max()` uses the full dataset length including the holdout, a tiny but real leak in a case study whose selling point is being leakage-aware. Use one normalized trend for all models (normalize by `MIN_OBSERVATIONS` or `lfo_cutoff`), and put the prior difference where it belongs, in `model_config`.

### Nits

- Dead code: `lfo_result_to_dataframe` / `lfo_origin_scores`, `reference_idata_by_model`, `num_roas_samples`, `refit_payloads` (stores a full `idata` per refit and is never read; memory), `plot_num_samples = min(3999, ...)` (magic number).
- `mmm_i.build_model(...)` before `mmm_i.fit(...)` in both the wrapper and `fit_reference_model` is redundant; `fit` builds when no model exists.
- `samples.isel(date=slice(-len(holdout_data), None))` in `holdout_predictions` is redundant; `sample_posterior_predictive(include_last_observations=True)` already drops the first `l_max` rows.
- `\(k\)` inline math (intro and Takeaways) will not render: `conf.py` enables `dollarmath` only. Use `$k$` like the other MMM notebooks.
- The link to the multiplicative notebook is a hard-coded `/en/stable/` URL; use a `{ref}` to `mmm_multiplicative` so it stays in-version.
- Section "LFO wrapper" says the wrapper "fits only when PSIS diagnostics require an exact refit"; the wrapper does not decide that, `lfo_cv` does.
- Section "Stacking predictive distributions" hard-codes the observed outcome ("additive is only third ... receives the largest stacking weight"). Fine if you re-check after every re-execution, but phrasing it conditionally would age better.
- `random_seed=int(np.random.default_rng(8675309).integers(1, 2**30))` in `holdout_predictions` can just be `SEED`.
- Cells 1 and 2 (markdown) say the same thing twice.
- The `additive` Normal likelihood uses `sigma ~ HalfNormal(6)` on a target scaled to roughly [0, 1]; not wrong, just very wide relative to the other priors in the same config.

Happy to re-review once the environment question is settled; the wrapper itself is in good shape.

TeemuSailynoja:
> ### Blocking
> 1. **The `!uv pip install --no-deps arviz-stats@git+...` cell mutates the shared environment.** The notebook CI (`scripts/run_notebooks/runner.py`) executes this cell, which replaces `arviz-stats` in the CI venv with `1.4.0.dev0`. `arviz/__init__.py` raises `ImportError("All ArviZ packages must share the same minor version.")` on mismatch, so every notebook that runs after this one in the `--notebooks mmm --start-idx 10` split will fail at `import arviz`. The `arviz_stats.__version__ = arviz_base.__version__` monkeypatch only hides this inside this kernel.

@juanitorduz, do you think we need to hold this notebook? @OriolAbril how long is the typical release cycle, do you think we are talking days, weeks, or a month here? Roughly.

> If you want it in earlier, the install instruction has to be a markdown cell (not executed) and the notebook has to go on `BLACKLIST` in `runner.py` with a comment, but that leaves a docs notebook that is not reproducible from `pip install pymc-marketing`.

Or do you think we should do this?

TeemuSailynoja:
@juanitorduz, thanks for the detailed review. I’ve pushed follow-up commits addressing most of the notebook-level issues and the CI environment issue.

## Addressed

1. **Executed `uv pip install` cell / environment mutation**  
   Addressed by removing the executable `!uv pip install ...` cell from the notebook. The install command is now markdown-only. I also added the notebook to the notebook runner `BLACKLIST` with a comment explaining that it requires the development `arviz-stats` LFO-CV API until that API is released. This avoids mutating the shared CI environment and prevents later notebooks from failing due to ArviZ package version mismatch.

2. **`io.capture_output()` breaks the mock runner**  
   Removed `io.capture_output()`. The LFO fitting cell now uses `%%capture`, which avoids the empty `Output()` widget artifacts and should be compatible with notebook execution.

3. **Missing `gallery.yaml` entry**  
   Added the notebook to `docs/source/gallery/gallery.yaml` under Model Evaluation, next to the cross-validation notebook. `gallery.md` is in sync.

4. **ruff / pre-commit issues**  
   Fixed the notebook lint issues:
   - sorted imports,
   - removed semicolons,
   - replaced or removed `assert` checks,
   - added docstrings to wrapper methods,
   - fixed formatting / EOF issues.  
   `uv run --with pre-commit pre-commit run --files docs/source/notebooks/mmm/mmm_lfo_cv_model_comparison.ipynb scripts/run_notebooks/runner.py` passes locally.

5. **LFO ELPD and holdout log scores on different scales**  
   Changed the holdout evaluation to use the same scaled target scale as the model likelihood. The holdout section now explicitly states that holdout log scores are computed on the same scaled target used internally by the likelihood, so they are comparable to the LFO log scores.

6. **Holdout contradicts the single-model LFO ranking**  
   Updated the holdout narrative and takeaways. The notebook now frames the 26-week holdout as a noisy check to be read together with the paired LFO standard errors, not as a definitive re-ranking. The takeaways also state that single-model rankings should be interpreted with paired uncertainty.

7. **Holdout log scores used KDE while LFO used exact likelihood**  
   Replaced the KDE-based holdout log score with exact likelihood evaluation using the same model/data transformation path as the LFO wrapper and `pm.compute_log_likelihood`. The stacked holdout log score is now computed as `log(sum_k w_k * p_k(y))`.

8. **`k>threshold` column is uninformative**  
   Removed the `k>threshold` column from the summary. Added a PSIS diagnostics section with a Pareto-$k$ plot over forecast origins and vertical markers for exact refits.

9. **Stacked ROAS interpretation**  
   Added a caveat in the ROAS section: the stacked row is a weight-resampled mixture of model-implied ROAS distributions using prediction-optimal stacking weights; it is not a separate causal model and carries no ROAS-optimality claim. Also noted that log-link unilateral increments are not additive in the same way as identity-link increments, so ROAS should be read as sensitivity analysis across assumptions.

10. **Control specification / trend leakage**  
    Changed the normalized trend to use `TREND_SCALE = MIN_OBSERVATIONS - 1` instead of `t.max()`, avoiding holdout leakage. Also updated the candidate-model narrative to say models share target/media/adstock/seasonality/scaling, but differ in link, saturation, likelihood, priors, and trend parameterization.

11. **Dead code**  
    Removed several unused pieces:
    - `lfo_result_to_dataframe`,
    - `lfo_origin_scores`,
    - `reference_idata_by_model`,
    - `refit_payloads`,
    - `num_roas_samples`.
    `plot_num_samples = min(3999, ...)` remains because it is used to standardize ROAS plotting sample sizes.

12. **Redundant `build_model` calls**  
    Removed redundant `build_model(...)` calls before `fit(...)`.

13. **Potentially redundant holdout date slicing**  
    Replaced the silent slice with an explicit length check. If posterior predictive length is not the holdout length, the notebook raises an error instead of slicing.

14. **Inline math syntax**  
    Replaced `\(k\)` with `$k$`.

15. **Hard-coded stable docs link**  
    Replaced the hard-coded multiplicative notebook URL with `{ref}` to `mmm_multiplicative`.

16. **Wrapper wording**  
    Updated the wording: `lfo_cv` decides when refits are needed; the wrapper exposes the method that `lfo_cv` calls for exact refits.

17. **Stacking text hard-coded observed result**  
    Rephrased the stacking narrative more generally so it does not hard-code a particular future run’s ranking/weights.

18. **Holdout prediction random seed**  
    Replaced the ad hoc RNG seed with `SEED`.

19. **Duplicate setup markdown**  
    Removed the duplication around the setup/install instructions. The development dependency instruction is now markdown-only.

20. **Additive likelihood prior too wide**  
    Changed additive Normal likelihood sigma prior from `HalfNormal(6)` to `HalfNormal(0.5)` on the scaled target.

## Still open / intentionally deferred

- The notebook is currently blacklisted from CI execution until the `arviz-stats` LFO-CV API is available in a normal release and can be expressed as a regular project dependency. Once that release is available, we should remove the markdown-only development install instruction and remove the notebook from the blacklist.

juanitorduz:
> > ### Blocking
> > 
> > 1. **The `!uv pip install --no-deps arviz-stats@git+...` cell mutates the shared environment.** The notebook CI (`scripts/run_notebooks/runner.py`) executes this cell, which replaces `arviz-stats` in the CI venv with `1.4.0.dev0`. `arviz/__init__.py` raises `ImportError("All ArviZ packages must share the same minor version.")` on mismatch, so every notebook that runs after this one in the `--notebooks mmm --start-idx 10` split will fail at `import arviz`. The `arviz_stats.__version__ = arviz_base.__version__` monkeypatch only hides this inside this kernel.
> 
> @juanitorduz, do you think we need to hold this notebook? @OriolAbril how long is the typical release cycle, do you think we are talking days, weeks, or a month here? Roughly.
> 
> > If you want it in earlier, the install instruction has to be a markdown cell (not executed) and the notebook has to go on `BLACKLIST` in `runner.py` with a comment, but that leaves a docs notebook that is not reproducible from `pip install pymc-marketing`.
> 
> Or do you think we should do this?

I think for now is fine, but please create an issue so that we can remove this once we have a new release

juanitorduz:
## Re-review (7a60fc4)

Went through the new notebook cell by cell against the first review. Everything listed as addressed is actually addressed in the code, not just in the comment: the install cell is markdown-only and the notebook is on `BLACKLIST`; `%%capture` replaces `io.capture_output()`; `gallery.yaml` has the entry and pre-commit.ci is green; the holdout is now scored with `pm.compute_log_likelihood` through the same data-transformation path as the wrapper, on the scaled target, with the stacked score as `logsumexp(log w_k + lpd_k)`; the reconciled numbers confirm the scale fix (holdout `+52` over 26 weeks = 2.0/week vs LFO 1.9/origin, where before it was `-182` vs `+94`); the Pareto-$k$ section is there; `TREND_SCALE = MIN_OBSERVATIONS - 1`; dead code, redundant `build_model`, `\(k\)`, hard-coded URL, ad hoc seed, and the wide `HalfNormal(6)` are all gone. `{ref}\`mmm_multiplicative\`` resolves (the label exists in `mmm_multiplicative.ipynb`).

The blacklist approach is fine with me given @juanitorduz's OK. Remaining items are small; none change the conclusions.

### Should fix

1. **The Pareto-$k$ plot marks the initial fit as a refit.** `refit_rows` gets a row for every `sample()` call, including the first fit at `cutoff = MIN_OBSERVATIONS`, so the first red line in every panel (2020-04) is not a PSIS-triggered refit, and the red line has no legend entry. `result.refits` already holds exactly the PSIS-triggered cutoffs (the docstring says the initial fit is excluded), so `lfo_data[DATE_COLUMN].iloc[result.refits]` gives the marker dates directly. With that, `refit_rows`, `fit_count`, `n_time`, and the `forecast_date` bookkeeping in the wrapper can be deleted; the scale checks in `sample` are the only bookkeeping worth keeping.

2. **The follow-up issue @juanitorduz asked for does not exist yet** (searched open issues for lfo / arviz-stats). It should reference `BLACKLIST` in `runner.py`, the markdown install cell, the `arviz_stats.__version__` shim in the imports cell, and the `arviz-stats` pin in `pyproject.toml`, so whoever picks it up after the release removes all four.

3. **"Prepare notebook" (cell 1) is stale.** It says "The compatibility shim below mirrors the exploratory notebook and should be removed once synchronized releases are available." Readers have no exploratory notebook; the shim is now in the imports cell, and cell 2 already says everything cell 1 says. Fold cell 1 into cell 2.

### Nits

- `holdout_crps` is in scaled-target units (0.018) and the column is not labeled as such. CRPS is scale-equivariant, so multiplying by `fixed_target_scale` gives sales units (about 153) for free; the log score should stay on the scaled target to match LFO, which the text already explains.
- Cell 6 now prints `[Text(0.5, 0, 'Date'), ...]` because the trailing `;` was removed; end the cell with `fig.tight_layout()` or assign the `set(...)` result to `_`.
- `np.log(weights)` with the two zero weights raises a `RuntimeWarning` that only the global `warnings.filterwarnings("ignore")` hides; `np.log(weights, where=weights > 0, out=np.full_like(weights, -np.inf))` or `sp.xlogy`-style masking makes it explicit.
- `warning` in `lfo_summary` is always `False` here (it only flips when refits exceed half the origins); it can go, or be worth one sentence in the text if you keep it.
- On the additive panel, the second refit comes one origin after the first (k = 0.03 at the single PSIS step in between). That is the influential-observation signature behind the additive model's large SE, and it is a good teaching moment for the PSIS-diagnostics section if you want one sentence on it.

Once (1) and (2) are in, this is good to merge from my side.

juanitorduz:
Thanks @TeemuSailynoja ! Looks like a great addition! I left some comments about adding more explanations and more "takeaways" for each plot and diagnostic.

juanitorduz (changes requested):
minor comments

PR #3032, opened 2026-09-18
Update pymc version constraint in pyproject.toml

juanitorduz:
Update pymc version constraint in pyproject.toml

----
📚 Documentation preview 📚: https://pymc-marketing--3032.org.readthedocs.build/en/3032/


daimon-pymclabs (approved):
Reviewed the dependency bump and the split-rule migration. Verified against the pymc-bart 0.13.0 sdist that the string `split_rules` API and `response="mix"` both exist at the declared floor, and that the `"OneHotSplit"`/`"ContinuousSplit"` literals match what bartrs exports. `pymc-extras` is capped at `<0.16.0` to match the `pymc` pin, and the response test is parametrized over constant/linear/mix.

Approving on the code. Two non-blocking notes: the title and `docs` label understate the change (it also bumps pymc-bart and migrates the split-rule API), and the test and notebook jobs were still running at approval time.

PR #3033, opened 2026-09-18
fix(mmm): reject prior names the transformation does not use

### LINKED ISSUE #1294: Media transformation allow for arbitrary prior names
```python
import pymc as pm

from pymc_marketing.mmm import LogisticSaturation


saturation = LogisticSaturation(
    priors={
        # alpha isn't used in the model....
        "alpha": Prior("HalfNormal"),
        "beta": Prior("HalfNormal"),
    }
)
```

Ideally this should raise an something. However, it might not do this because of how it is used in the MMMs

anevolbap:
Fixes #1294

Priors for a parameter the function does not take were merged into `function_priors` and then ignored. A misspelled name (`priors={"alpah": ...}`) or a model variable name (`priors={"saturation_beta": ...}`) left the default prior in place with no error, and `to_dict` serialized the unused key.

The `function_priors` setter now raises `ValueError` listing the unknown names and the actual parameters. When a name ends in `_<parameter>`, the message suggests the parameter name. The setter is the single entry point, so `__init__` and `from_dict` are both covered. `_AlphaHalfLifeParametrization` already rejects the inactive parameter this way (#2839), and its more specific error still comes first.

The `prior_predictive` notebook had this bug: its saturation cell passed `saturation_lam` and `saturation_beta`, so the curve used the default priors. The cell now uses `lam` and `beta`, with `dims="channel"` on the per-channel `beta` prior, and the figure shows one panel per channel instead of one.

The MMM path is unaffected: `MMM.__init__` passes the whole `model_config` through `with_updated_priors`, which filters keys by prefix in `update_priors` and never reaches the setter.

One behavior change: a saved transformation whose `priors` dict holds an unused key no longer loads. The key never reached the model. Is a `DeprecationWarning` preferred over an error for the 1.x line?

`tests/mmm` passes.


juanitorduz (approved):
Thanks!

Issue #3034, opened 2026-09-19
Make MuEffect variables available to causal adjustment-set identification


leninworld:
## Summary

Variables supplied through `MuEffect` cannot currently count as available variables when `MMM` computes a causal adjustment set.

`MMM.__init__` calls `CausalGraphModel.compute_adjustment_sets` before `self.mu_effects` is initialized and before callers can invoke `add_mu_effect`. In addition, the `MuEffect` abstract base class does not expose the causal variable names represented by an effect.

This means a DAG can contain a valid backdoor adjustment set represented by a `MuEffect`, while the causal-identification check still reports the effect as unidentified because it only sees channels, controls, and built-in model features.

## Current ordering

1. `MMM.__init__` builds `CausalGraphModel` and calls `compute_adjustment_sets`.
2. `MMM.__init__` initializes `self.mu_effects` later.
3. The user calls `MMM.add_mu_effect(...)` after construction.

At step 1, neither the effect instances nor a general effect-variable contract is available.

## Expected direction

Provide a way for causal adjustment-set computation to include variables represented by attached `MuEffect` objects, without guessing from PyMC variable names.

Possible requirements:

- `MuEffect` exposes the causal/data variable names it represents, or a causal graph fragment from which those names can be derived.
- Causal identification is recomputed after effects are attached, or deferred until the full model specification is available.
- Existing channels, controls, yearly seasonality, and custom effect variables share one explicit availability contract.
- Save/load and chained `add_mu_effect` calls preserve the same identification result.

The exact API should align with the composable causal-graph work proposed in #2886 rather than introducing a second naming convention.

## Relationship to other issues

- Follow-up to #2778, which corrects minimal adjustment-set computation and makes unavailable identification explicit.
- Related to #2886, especially the proposed `MuEffect` causal-fragment protocol and the open question of variable identity across effects.

## Acceptance criteria

- A DAG whose required adjustment variable is supplied by a `MuEffect` can be identified after the effect is attached.
- A `MuEffect` variable is not silently inferred from its PyMC deterministic or prefix name.
- Adding or loading effects cannot leave a stale `is_backdoor_identified` result.
- Tests cover an identified effect variable, a missing effect variable, chained effect addition, and serialization round-trip behavior.

PR #3035, opened 2026-09-19
Fix causal adjustment set identification

### LINKED ISSUE #2778: get_unique_adjustment_nodes documents and names a "minimal adjustment set" but returns a non-minimal union
### Summary

`CausalGraphModel.get_unique_adjustment_nodes` is documented as *"Compute the minimal adjustment set required for backdoor adjustment across all treatments"*, and `compute_adjustment_sets` stores its result on `self.minimal_adjustment_set`. What it actually returns is the **flat union of every node appearing on any backdoor path**, minus the treatments and the outcome:

https://github.com/pymc-labs/pymc-marketing/blob/main/pymc_marketing/mmm/causal.py#L1344-L1360

There is no d-separation test, no minimality check, and no distinction between confounders, descendants of the treatment and colliders.

### Why it matters

The union of the nodes on the backdoor paths is a heuristic superset, and the two things the backdoor criterion is most often got wrong on are exactly the two it does not check:

- **descendants of the treatment** — condition (i) forbids conditioning on them; doing so blocks part of the causal path and changes the estimand;
- **colliders** — conditioning on one *opens* the paths that meet there.

Concrete example, the DAG from `mmm_funnel_mueffect_advanced` (12 nodes, treatment `U1`, outcome `Y`):

```python
adjustment = model.get_unique_adjustment_nodes()
# ['C', 'D', 'MS', 'N', 'U2', 'season', 't']
```

against `networkx`:

```
descendants(U1)                       = ['D', 'M', 'MS', 'S', 'Y']
every minimal d-separator of (U1, Y)  = ['C', 'U2', 'season']
                                        ['D', 'U2', 'season']
                                        ['MS', 'U2', 'season']
                                        ['N', 'season', 't']
```

The returned set has seven nodes where the minimal ones have three; two of the seven (`D`, `MS`) are descendants of the treatment, and one (`N`) is a latent no dataset can carry. A user who reads the docstring, or the `minimal_adjustment_set` attribute name, and reports that set as the minimal adjustment set would be wrong on all three counts.

### Suggested fix

Either make the docstring and the attribute name describe what the method does — "every node on a backdoor path", explicitly not minimal and not checked for admissibility — or compute a real one (`networkx.algorithms.d_separation.find_minimal_d_separator`, restricted to observed non-descendants, is a few lines and is what the notebook does). Renaming is the cheaper half and fixes the misleading part; the docstring is what will mislead someone first.

Found while writing #2769, which carries a note describing the current behaviour accurately for readers who reach for `CausalGraphModel` after that notebook's hand-rolled d-separation section.

leninworld:
This PR fixes [issue #2778](https://github.com/pymc-labs/pymc-marketing/issues/2778).

## Description

`get_unique_adjustment_nodes` currently returns a union of nodes on backdoor paths, which can include unnecessary variables and treatment descendants. This PR uses NetworkX minimal d-separation after removing outgoing treatment edges and excluding treatments, outcomes, and treatment descendants from adjustment candidates.

The implementation distinguishes graph-level, model-available, indispensable, and best-effort sets and exposes `is_backdoor_identified`. When the model cannot represent a valid adjustment set, it emits `UnidentifiedCausalEffectWarning` and preserves supplied controls unchanged. Structurally impossible adjustment also warns; invalid treatment/outcome configuration still raises `ValueError`. Strict pipelines can promote the dedicated warning category with `warnings.filterwarnings`.

Yearly seasonality is treated as an available model feature and is removed only when an identified adjustment set excludes it. The causal-identification notebook has been updated and re-executed to explain the unmeasured `competitor_offers` path and why a time-varying proxy/modeling assumption does not establish identification. The dependency requirement is now `networkx>=3.3` in the DAG, docs, and test extras.

### Warning policy

Warn-not-raise follows [anevolbap's proposed direction](https://github.com/pymc-labs/pymc-marketing/issues/2778#issuecomment-5647509724). That comment was a proposal rather than a maintainer decision; confirmation of this policy is welcome during review. The public `get_backdoor_paths` method is retained and covered by a regression test.

### Behavior changes / release notes

- `minimal_adjustment_set` now contains the selected adjustment variables, excluding treatment nodes, and can be `None` when the represented model variables do not identify the effect. `[]` represents a valid empty adjustment set.
- Supplied controls are preserved rather than filtered when the effect is unidentified.
- A configured `outcome_node` absent from the DAG now raises `ValueError` instead of silently producing an empty set. Omitting `outcome_node` retains the existing constructor behavior.

### Validation

- `86 passed` in `tests/mmm/test_causal.py` and `265 passed` in `tests/mmm/test_mmm.py`.
- Combined run after the coverage follow-up: **351 passed**. Local coverage confirms the public indispensable-set error and seasonality-removal paths are exercised.
- Full `pre-commit run --all-files` passed.
- Causal-identification notebook executed in place with no error outputs.

Rendered summary of the earlier 349-test local run, before adding the two coverage regression tests (not a direct terminal screenshot):

Rendered summary of local pytest results: 349 passed

## Related Issue

- [x] Closes #2778
- [x] Related to #3034 — separately tracks MuEffect variable availability for causal identification.

## Checklist

- [x] Included tests that prove the fix is effective or that the new feature works.
- [x] Added necessary documentation (docstrings and example notebook) using [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [x] Each commit corresponds to a relevant logical change.

----
📚 Documentation preview 📚: https://pymc-marketing--3035.org.readthedocs.build/en/3035/


juanitorduz:
## Review

**Verdict: the algorithm is correct and a real fix for #2778. Two blocking issues before merge — the notebook was reverted out of the PR, and `minimal_adjustment_set` is still not minimal.**

### Evidence gathered

Ran against the PR head (`e29a04f2`) in a clean worktree with the repo venv:

| Check | Result |
|---|---|
| `tests/mmm/test_causal.py` | 88 passed |
| `tests/mmm/test_mmm.py` | 265 passed |
| `ruff` + `mypy` on changed files | clean (no `causal.py` findings) |
| Single-treatment property test vs textbook backdoor oracle, 86 random DAGs | 0 mismatches (set validity, inclusion-minimality, indispensability) |
| Two-treatment property test vs generalized adjustment criterion (path-enumeration oracle), 600 random DAGs | 0 mismatches (candidate pool, minimality, indispensability) |
| Three-treatment test of `restricted`/`preferred`/`is_valid_adjustment_set`, 400 random DAGs | 0 mismatches |
| #2778's own example (`mmm_funnel_mueffect_advanced` DAG, treatment `U1`) | old: `['C','D','MS','N','U2','season','t']` → new: `['N','season','t']`; `D`/`MS` correctly rejected as `De(U1)`; `{C,U2,season}` confirmed valid |
| Selection stability across `PYTHONHASHSEED` in {0,1,2,3,7,12345} | stable |

The proper-backdoor-graph construction in `_backdoor_parts` (per-treatment node removal for *proper* paths, `forb = De(cn) ∪ X`, first-edge removal) matches Perković's generalized adjustment criterion. That part is right.

### Blocking

**1. The notebook is not in the PR — the final commit reverted it, and the shipped outputs now contradict the code.**

The description says the causal-identification notebook "has been updated and re-executed". It has not: `9b6bbc66` changed it, `e29a04f2` ("docs: remove stale causal notebook outputs") restored it to the merge-base blob. Verified identical: `b26e47d4:docs/source/notebooks/mmm/mmm_causal_identification.ipynb` and `e29a04f2:…` are both blob `e405c5ff`. `git diff main...pr-3035` touches 6 files, none under `docs/`.

Because `docs/source/conf.py:126` sets `nb_execution_excludepatterns = ["*.ipynb"]`, the stored outputs are what gets published:

| Cell | Committed output | Actual under this PR |
|---|---|---|
| 44 | `([], ['x1', 'x2'])` | `([], [])` |
| 46 | `(['holiday_signal'], ['holiday_signal', 'x1', 'x2'])` | `(['holiday_signal'], ['holiday_signal'])` |
| 57 | `(['holiday_signal', 'competitor_offers'], ['holiday_signal', 'x1', 'x2'])` | `UnidentifiedCausalEffectWarning` + `minimal_adjustment_set is None` |

Markdown cell 58 ("our adjustment set clearly indicates the need to control for both **competitors** and **holidays** … we set `time_varying_intercept` to `True` so that our model can accommodate the unobserved confounder") narrates an output the code no longer produces, and asserts exactly the proxy-equals-identification claim the PR description says the notebook was rewritten to refute. Restore the `9b6bbc66` notebook work (re-executed), or drop the claim from the description and open a follow-up.

**2. `minimal_adjustment_set` is not minimal — `preferred` reintroduces #2778's naming defect.**

`compute_adjustment_sets` passes `preferred=model_features`, which becomes `included=` in `find_minimal_d_separator`, forcing `yearly_seasonality` into the result whenever a valid set containing it exists. End-to-end, DAG `yearly_seasonality -> y; holiday -> x1; holiday -> y; x1 -> y`, `control_columns=["holiday"]`, `yearly_seasonality=2`:

```
main:  minimal_adjustment_set = ['holiday', 'x1']                  yearly_seasonality dropped
PR:    minimal_adjustment_set = ['holiday', 'yearly_seasonality']  yearly_seasonality kept
truly minimal: ['holiday']
```

Keeping seasonality is the right *outcome*; inflating the reported set to get it is not. #2778 is titled "documents and names a 'minimal adjustment set' but returns a non-minimal union" — this ships a smaller version of the same defect, and it is user-visible (the notebook prints it).

Suggested split: compute `minimal_adjustment_set` with no preference, then decide retention separately, e.g. keep the feature iff `is_backdoor_identified and is_valid_adjustment_set(set(minimal) | {"yearly_seasonality"})`. Same behavior, honest attribute. Side effect worth checking: as written, the "Yearly seasonality excluded as it's not required for adjustment" branch is unreachable whenever `yearly_seasonality` is an admissible DAG node — it now fires only when the node is absent from the DAG or is forbidden (a treatment descendant).

### Should fix

**3. Identified models now drop more user controls than main, and it is not in the release notes.** When several minimal sets exist, the arbitrary winner prunes the rest:

```python
dag = "digraph { A -> X; A -> S; S -> Y; X -> Y; }"          # {A}, {S}, {A,S} all valid
compute_adjustment_sets(channel_columns=["X"], control_columns=["A", "S"])
# selected ['A'], returns ['A'], warns "Columns {'S'} … Controls are being modified."
```

`S` is admissible, valid alongside `A`, and possibly predictive — main's union kept it. The `preferred` machinery already exists; passing the supplied controls (or keeping any control whose addition preserves `is_valid_adjustment_set`) would make selection respect user intent instead of networkx's tie-break.

**4. The unidentified warning lists variables nobody needs.** `unavailable = sorted(candidates - available)` is the whole admissible pool, so the notebook DAG yields *"Eligible graph variables not represented by the model: ['competitor_offers', 'market_growth']"* — `market_growth` is a pure `y`-parent that appears in no adjustment set. Narrow it to variables occurring in at least one minimal separator; `missing_indispensable` is already correct.

**5. Docstrings for the new public surface.** `CausalGraphModel` gained five public attributes (`adjustment_set` semantics *changed*, plus `minimal_adjustment_set`, `indispensable_adjustment_nodes`, `best_effort_adjustment_nodes`, `is_backdoor_identified`) with no `Attributes` section; `compute_adjustment_sets` documents neither the `Warns` behavior nor the attributes it mutates. Repo convention requires numpydoc for public API. Also worth defining `best_effort_adjustment_nodes` in docs — it is `arbitrary_graph_level_set ∩ available`, so with alternative graph-level sets it is itself arbitrary, which the name oversells.

**6. Missing test for the headline new behavior.** No test covers "`yearly_seasonality` is an admissible DAG node but unnecessary → retained"; `test_mmm_removes_yearly_seasonality_when_identified_and_unneeded` uses a DAG without that node, so the `model_features`/`preferred` path is only exercised indirectly.

**7. Undocumented behavior change: `control_columns=None` now triggers computation.** Main early-returned before touching the graph; the PR always computes. Verified main vs PR on `holiday -> x1; holiday -> y; x1 -> y` with `control_columns=None`: main silent, PR emits `UnidentifiedCausalEffectWarning`. Reasonable, but it belongs in the behavior-changes list, and it means seasonality removal can now fire for control-free models.

### Nits

- `_backdoor_parts()` runs 4x per `compute_adjustment_sets` (measured ~10 ms on a 48-node/8-treatment DAG, so cheap — but thread the tuple through). The `candidates` computed before the `try` is only used in the unidentified branch; move it there.
- `restricted=["Z"]` / `preferred=["Z"]` raise `TypeError: unsupported operand type(s) for &=: 'set' and 'list'`, while `is_valid_adjustment_set` accepts both. Coerce with `set(...)`.
- `treatment="x1"` (str instead of list) → `ValueError Nodes ['1', 'x'] are not in the causal graph`. Normalize in `__init__`.
- `mmm.py:630-631`: `is_backdoor_identified and minimal_adjustment_set is not None` is redundant — both come from the same expression.
- `UnidentifiedCausalEffectWarning` docstring says "Raised as a warning"; it is issued, not raised.
- `get_unique_adjustment_nodes` no longer describes what it does ("unique" referred to the union dedup). Rename with a deprecation shim, or note it in the docstring.
- `stacklevel=2` points at `mmm.py:622` rather than user code when called from `MMM.__init__` (pre-existing).

### Questions for maintainers

- **Warn vs raise**: worth confirming explicitly rather than inheriting it from a comment. `UnidentifiedCausalEffectWarning(UserWarning)` is the right base — pipelines already running `-W error::UserWarning` keep failing, and the dedicated category gives everyone else a targeted filter.
- **Fourier seasonality as "represented"**: `model_features={"yearly_seasonality"}` treats a k-order Fourier basis as the DAG node itself and flips `is_backdoor_identified` to `True` on that basis, while `time_varying_intercept` is (correctly) not counted. A truncated basis spans only part of a seasonal confounder — is that the identification claim you want to make, given the PR argues the opposite for time-varying proxies? Related to #3034.

leninworld:
@juanitorduz Thank you for the detailed review and the independent property testing. I addressed the two blockers and the feedback-scoped items in [`66bdcb57`](https://github.com/pymc-labs/pymc-marketing/commit/66bdcb57d702de9282de887bd631d3f368ae1e53):

- `minimal_adjustment_set` is now computed independently of retention policy. Valid user controls and safe model features are retained separately, so seasonality can remain without inflating the reported minimal set.
- The PR description now states that the notebook is unchanged and was not re-executed or regenerated; the stale notebook-output claim was removed.
- The unidentified warning now reports one explicitly non-exhaustive example minimal set, rather than presenting the full admissible candidate pool as required variables.
- The new public attributes, warning behavior, and `control_columns=None` behavior are documented.
- String/set inputs are normalized, repeated backdoor decomposition is reduced, and regression coverage was added for minimality, control retention, seasonality, warning content, and input shapes.

For that feedback revision, `tests/mmm/test_causal.py` passed 92 tests and `tests/mmm/test_mmm.py` passed 266 tests; the changed-file lint/format checks also passed. The current PR head additionally merges the latest `main`, while the effective PR diff remains scoped to the same six files.

The pre-existing `stacklevel` item is unchanged. The warn-vs-raise and Fourier-representation questions remain open for maintainer direction.

Issue #2937, opened 2026-08-31
Budget optimizer: public API for objective/gradient evaluation at a labelled plan


juanitorduz:
The funnel-aware case study notebook (#2936) runs a didactic gradient-decomposition check (funnel-aware gradient minus direct-only gradient equals the marginal mediated response) that requires `BudgetOptimizer._objective_and_grad` and `optimizer.optimization_variables.pack`. The PR review (https://github.com/pymc-labs/pymc-marketing/pull/2936#pullrequestreview-5068287770, item 10) suggested making this pattern first-class instead of pinning a flagship notebook to internals.

Proposal: a public method to evaluate the compiled objective and gradient at a labeled plan, e.g. `optimizer.evaluate_plan(plan: dict[str, xr.DataArray]) -> (float, xr.Dataset)`, built on the existing packing machinery in `optimization_variables`. This would let docs and users compare objectives, decompose gradients, and score externally proposed plans without touching private attributes.

daimon-pymclabs:
Reviewed the proposal. Worth doing, but it is scoped narrower than the title suggests, and there are four design details worth pinning down first.

`optimization_variables` is already a public property (`budget_optimizer.py:1793`), and `pack`/`unpack` are public methods on it (`optimization_variables.py:855, 893`). The only private thing the notebook reaches for is `_objective_and_grad`. So "decision-vector packing" is not actually a gap, and the title could drop it.

On the method itself:

- **Sign convention.** `_compile_objective_and_grad` builds `objective = -utility_function(...)`, so what comes back is negated utility and the gradient points the minimizer's way. A public `evaluate_plan` returning that raw will trip up people doing the didactic decomposition. Either name the outputs `objective`/`objective_grad` and document the negation loudly, or return utility and flip the sign.
- **Input safety.** The compiled function sets `trust_input = True` (`budget_optimizer.py:2244`), so it performs no shape or dtype checking. A public entry point should route strictly through `optimization_variables.pack`, which already yields a float64 vector of length `size`, rather than accepting a raw array from the caller. Otherwise a wrong-dtype input is undefined behavior instead of a clean error.
- **Return type.** `unpack` returns `dict[str, DataArray]`, not `xr.Dataset`. Returning the dict keeps the method symmetric with the packing API it is built on; a `Dataset` needs an extra conversion and only works cleanly when the variables share dims.
[...]

PR #3045, opened 2026-09-22
Budget optimizer: spend-dependent cost per unit via PowerPriceResponse

### LINKED ISSUE #3036: Budget optimizer: allow cost per unit to depend on the spend level, not only on date
## Feature request

Allow the budget optimizer's cost-per-unit conversion to depend on the spend level being evaluated, not only on date and dims.

## Motivation

`cost_per_unit` currently answers "what does one unit of this channel cost in week *t*". That is enough for price drift over time, and it is what the [Cost-per-Unit](https://www.pymc-marketing.io/en/latest/notebooks/mmm/mmm_cost_per_unit.html) notebook demonstrates. It cannot express a second, very common fact about media buying: **the price of a unit depends on how much you buy.**

Two mechanisms produce this, and both bite exactly where budget optimization is used:

- **Auction bid-up.** In biddable inventory, raising the budget raises the clearing price. The marginal impression costs more than the average impression already bought.
- **Inventory tiers and late buying.** In reserved inventory, the cheap tier is bought first (upfront, committed, negotiated), and the increment clears at scatter prices. The rate card is a step function of volume, not a constant.

The consequence is one-sided, not noise. The optimizer computes marginal return as the response slope times units-per-dollar, with units-per-dollar fixed at the average price. If the increment actually clears above that price, every marginal return the optimizer sees is too high, and it is too high precisely on the channels it is choosing to grow. Concretely, a 20% higher clearing price on the increment makes the true marginal return about 17% below the reported one, which is enough to flip a channel across a hurdle-rate threshold with nothing in the output indicating it.

This also means the model cannot be queried honestly for the question planners most often ask: "how much more can I spend here before it stops being worth it". The answer is a function of a price that the optimizer treats as constant.

## Where this would plug in

The conversion is a single PyTensor op inside the differentiable graph, in `pymc_marketing/mmm/optimization_variables.py`:

```python
# Convert from monetary units to original units using date-specific
# rates. Applied AFTER time distribution so each period uses its own
# cost rate.
if self.cost_per_unit_tensor is not None:
    repeated_budgets = repeated_budgets / self.cost_per_unit_tensor
```

At that point `repeated_budgets` is per-period spend in money, which is exactly the quantity the price should be allowed to depend on. Generalizing the division to a monotone spend-to-delivery map keeps everything downstream untouched:

$$u_t = \frac{s_t}{p_t(s_t)}$$

A convenient first parameterization is iso-elastic:

$$p_t(s) = p_{0,t}\left(\frac{s}{s_\text{ref}}\right)^{\gamma}, \qquad \gamma \ge 0 \quad\Longrightarrow\quad u_t \propto s_t^{\,1-\gamma}$$

with $\gamma = 0$ reproducing today's behavior exactly. It is smooth, monotone, concave for $0 < \gamma < 1$, has a closed-form inverse, and PyTensor differentiates it for free.

Two properties make this a smaller change than it first appears:

1. **Constraints are unaffected.** Decision variables stay in money and the budget sum constraint stays in money. Only the map into model units changes, so no constraint machinery moves.
2. **The model graph is unaffected.** Adstock, saturation, `MuEffect`s, and the objective all receive model units exactly as they do now.

## Suggested API

```python
optimal, res = wrapper.optimize_budget(
    budget=...,
    cost_per_unit=cpu_df,            # base price, date x channel, unchanged
    price_response=PowerPriceResponse(
        elasticity={"tv": 0.25, "display": 0.10},   # gamma per channel, default 0
        reference_spend=weekly_hist,                 # where the base price applies
    ),
)
```

Ship a small closed set of parametric forms rather than an arbitrary callable:

- `PowerPriceResponse(elasticity, reference_spend)` as above.
- `TieredPriceResponse(breakpoints, prices, smoothing)` for rate cards, with the kinks smoothed so SLSQP does not stall on a non-differentiable point.

An open callable is tempting but creates the inverse-map problem below, so it should come later if at all.

## Three things that need care

**1. The inverse map.** `OptimizationVariable` needs money-from-units for reporting and for packing externally supplied plans. Today that is a multiplication. With a power law it stays closed form; with an arbitrary callable it becomes a numerical solve. This is the main argument for restricting the first version to invertible, monotone, parametric families.

**2. Double counting against saturation.** If the model was fitted on nominal spend, part of the price curvature is already inside the fitted saturation curve, because price and volume moved together historically. Adding a concave price map on top of that bends the same curve twice and understates marginal return. The feature is only sound when the model was fitted on delivery units, or on spend restated into constant prices. That precondition belongs in the docstring and in the docs, not only in a notebook, because the users who most need this feature are exactly the ones most likely to have fitted on nominal spend.

**3. The elasticity is an input, not something the MMM estimates.** The model never observes price, so $\gamma$ has to come from buying data: realized cost against volume, per channel. That regression has its own endogeneity, since prices and demand both move with seasonality and competitive pressure. The honest posture in the docs is a sensitivity sweep over $\gamma$, in the same spirit as the existing cost-per-impression sweep in the Cost-per-Unit notebook, rather than a single point value presented as known.

## Test plan

- $\gamma = 0$ reproduces the current allocation bit for bit (regression guard).
- Gradient of the objective matches finite differences under a non-zero $\gamma$.
- Round-trip: money to units to money is the identity for every shipped parameterization.
- Behavior: with two channels of equal average ROAS and unequal elasticity, the optimizer allocates less to the more elastic channel, and the gap widens with total budget.
- Tiered form: the optimum respects a rate card whose marginal price steps up, and the solver does not stall at a breakpoint.

## Alternatives available today

Worth documenting even if this lands, because they need no new machinery:

- **Fixed-point iteration.** Optimize at an assumed price, read the implied volume, reprice from a rate card or an elasticity estimate, pass the updated `cost_per_unit`, re-solve. A few passes converge for modest elasticity, and `set_constraints` reuses the compiled optimizer so the repeats are cheap.
- **Tiered bounds.** Cap each channel at the volume its cheap tier supports with `budget_bounds`, forcing the expensive increment into a separate, explicit decision.

Both are workable, and both leave the marginal-return numbers the optimizer reports internally still computed at a constant price, which is the part this request is really about.

## Related

- #2799 (date-level budget decisions) interacts: if the optimizer also chooses *when* to buy, a spend-dependent and time-dependent price becomes more valuable, since buying late is a large part of why the increment is expensive.
- #2937 (public objective and gradient evaluation) would make the behavior tests above expressible without touching private attributes.

juanitorduz:
## Description

Closes #3036. The budget optimizer's money-to-units conversion can now depend on the money spent in the period, not only on the date.

```python
from pymc_marketing.mmm import PowerPriceResponse

mmm.set_cost_per_unit(historical_cpu_df)  # records that the fit is in delivery units
    start, end,
    cost_per_unit=window_cpu,                                  # base price p0, date x channel
    price_response=PowerPriceResponse(elasticity={"tv": 0.25, "display": 0.10}),
)
result = optimizer.allocate_budget(total_budget=weekly_budget)
result.budgets, result.implied_delivery, result.implied_price, result.implied_marginal_price
```

Scope is the thread's: `PowerPriceResponse` only, shipped correctly. `TieredPriceResponse` stays its own issue.

### What lands

- **`pymc_marketing/mmm/price_response.py`** (new). `PriceResponse` (pydantic ABC) / `PowerPriceResponse` declare the map; `ResolvedPowerPriceResponse` binds it to one variable's cell layout. The map is iso-elastic above a floor and a **quadratic pinned at `u(0) = 0`** below it (C1, concave, increasing); the floor is derived from `max_slope_ratio`, the spread of marginal returns the solver can meet on one cell. Coefficients use the closed forms under a double `where`, so a mixed array with `gamma = 0` cells resolves finite and warning-free (a `nan` in a dead branch is harmless on C and poisons the gradient under JAX). Both `where` branches receive a clipped input: the power branch would have an infinite derivative at 0, and the quadratic price branches have a pole in the region where they are not selected.
- **`MediaVariable`** applies `to_delivery` to **unscaled per-period money before `channel_scales`**, with `cost_per_unit` as base price: `(s/c)**(1-γ) ≠ s**(1-γ)/c`. When no response is passed, or it is the identity, the graph is the pre-feature graph operation for operation (pinned bitwise). `delivery_report` evaluates the same symbolic maps at the solution, compiled once per variable.
- **`BudgetOptimizer.price_response`** — a bare response, or a dict keyed by decision variable when `spend_vars` are declared (a bare object then raises: media and a spend variable draw from one pot through one constraint, so pricing one on a curve and the other at a constant must be explicit).
- **The gate.** A non-identity response is only sound when the model was fitted on delivery units — a saturation curve fitted on nominal spend has already absorbed part of the price curvature, and a concave map on top bends it twice. The optimizer checks each optimized channel against the historical `cost_per_unit` table on the fitted artifact (`idata.attrs["cost_per_unit"]`, per **column**, because a partial `set_cost_per_unit` writes `channel_spend` for every channel) and refuses otherwise, naming the channels and both remedies. Opt-out: `assume_delivery_units=True` with an explicit `reference_spend`.
- **`reference_spend`** is per-period money, the units of `result.budgets` and `total_budget`. Derived by default as the mean of `channel_spend` over the periods each cell was **on air** (a flighted channel is anchored where it bought, not 4x below); a supplied value is guarded against the derived one (`reference_spend_tolerance`, default 10x) so a window total mistaken for a per-period rate is refused naming both numbers.
- **Three result fields**: `implied_delivery` (channel units, what `sample_response_distribution` wants), `implied_price` and `implied_marginal_price` (`nan` at zero money). Exact on spent cells: `(implied_delivery * implied_price).sum(date) == budgets * num_periods`; on the power branch `implied_marginal_price / implied_price == 1 / (1 - γ)`.
- `BudgetOptimizerWrapper.optimize_budget(price_response=...)` forwards, since the issue's suggested API and the Cost-per-Unit notebook use the wrapper.

### Design decisions (the thread's three passes, plus the plan's own review)

- Marginal price is **piecewise** (`p0 / (a + 2bs)` below the floor), continuous at `s_f`, `== p/(1-γ)` on the power branch only, `m(0) == p(0) == p0/a` — per the 2026-09-22 correction.
- **Fixed-point iteration is not equivalent** under heterogeneous elasticity: it converges to `R'(u)/p = const`, missing the `(1-γ)` factor. Tested via the first-order condition through the public `evaluate_plan` (#3038): gradient equal across cells at the interior optimum, unequal at the fixed point with heterogeneous γ, equal with homogeneous γ. The comparator needs a fresh optimizer per pass — the issue body's "`set_constraints` makes repeats cheap" was wrong.
- **`total_budget` is per period**, so window length at fixed `total_budget` does not move the price; the thread's "shorter window raises the price" bullet was dropped as false. The concentrated-`budget_distribution_over_period` change is real (Jensen) and tested at the `MediaVariable` level.
- **Anchoring.** Below the reference the power law says units are *cheaper* than `p0`; "the more elastic channel receives less" is a statement at or above the reference. Measured on the fixture: at a budget ~5x below history the marginal-return factor on the elastic channel is 1.3–1.4, i.e. more, not less. The end-to-end test anchors `reference_spend` at the constant-price optimum; a companion test pins the other direction exactly through the report; the docstring has an **Extrapolation** paragraph.
- `_to_datatree` root-attrs fix from the thread **not** applied: `arviz.InferenceData` *is* `xarray.DataTree` on this stack, so that branch is unreachable. `merge_inference_data` does drop root attrs; merged models take the documented opt-out.
- Coordinate-less `elasticity`/`reference_spend` DataArrays are refused (the #3038 hazard one level down); elasticity is zeroed outside the mask; the tolerance guard is NaN-aware and reports the cells it could not check; a priced fit with no window `cost_per_unit` warns that money would reach the model as units.

### Tests

```
uv run pytest tests/mmm/test_price_response.py tests/mmm/test_optimization_variables.py \
  tests/mmm/test_cost_per_unit.py tests/mmm/test_budget_optimizer.py tests/mmm/test_budget_optimizer_mmm.py -q
306 passed
uv run pytest tests/mmm -q -k "budget or optim or allocat or pack or cost_per_unit or price"
422 passed, 1 skipped
```

New: `tests/mmm/test_price_response.py` (34: the map's shape, floor, continuity, gradient at exactly 0, slope cap, bitwise γ=0, monotone/concave; declaration validation, mapping/DataArray resolution, coordinate-less refusal, tolerance guard); `test_optimization_variables.py` (+6: ordering guard on both spreading paths, identity graph bitwise, Jensen, labelled report with `nan` at zero money); `test_cost_per_unit.py` (+17: gate matrix incl. partial table, on-air reference, opt-out, window-granularity guard, missing-window-price warning, γ=0 regression, money identity, FD gradient, FOC vs fixed point, zero start, both anchoring directions); `test_budget_optimizer_mmm.py` (+4: `spend_vars` keying, wrapper forwarding).

Smoke run on the fixture (γ sweep on `channel_2`, `total_budget=600`, history ≈ 750 per period on that channel):

```
gamma=0.00  budgets=[220.7  53.3 326.1]  price=[2.  3.  1.5]  marginal=[2.  3.  1.5]
gamma=0.15  budgets=[184.2 114.3 301.5]  price=[2.  2.26 1.5]  marginal=[2.  2.66 1.5]
gamma=0.30  budgets=[178.9 123.1 298. ]  price=[2.  1.74 1.5]  marginal=[2.  2.49 1.5]
```

The baseline reports `implied_price == p0` so it tables against the sweep; the elastic channel below its reference gets cheaper and receives more — the anchoring caveat, visible.

### Follow-ups

- Notebook section in `mmm_cost_per_unit.ipynb` (γ sweep, allocation shift, implied clearing prices, the precondition) — stacked PR.
- `TieredPriceResponse` taking **marginal** (bracket) rates — own issue.
- Labelled per-variable delivery reports for `spend_vars` on the result (reachable today via `optimization_variables.variables[i].delivery_report`).

## Related Issue

- [x] Closes #3036
- [x] Related to #2937

## Checklist

----
📚 Documentation preview 📚: https://pymc-marketing--3045.org.readthedocs.build/en/3045/


daimon-pymclabs:
## Review

Solid piece of work. The math is right, the gate is the correct design call, and the test matrix is unusually thorough for a feature this size. I verified the closed forms by hand and found no correctness bug in the map itself. Three things I'd want changed before merge, all small.

### What I checked and confirms out

- **Power branch.** `u(s) = s_ref^γ s^(1-γ)/p0` gives `p = s/u = p0 (s/s_ref)^γ` and `m = 1/u'(s) = p/(1-γ)`. ✓
- **Quadratic floor.** With `a = (1+γ) scale s_f^-γ` and `b = -γ scale s_f^(-γ-1)`, value and slope both match at `s_f` (`scale s_f^(1-γ)` and `(1-γ) scale s_f^-γ`), `u(0)=0`, `b<0` so concave, `u'(s_f)>0` so increasing. ✓ C1 as claimed.
- **Floor derivation.** `u'(0)/u'(s_ref) = (1+γ)/(1-γ) · (s_f/s_ref)^-γ = M` ⇒ `s_f/s_ref = (M(1-γ)/(1+γ))^(-1/γ)`. ✓ Matches the code and the `max_slope_ratio > (1+γ)/(1-γ)` validator.
- **Pole safety.** `a + bs = 0` at `s = ((1+γ)/γ)·s_f` and `a + 2bs = 0` at `s = ((1+γ)/2γ)·s_f`, both strictly above `s_f` for `γ<1`, so clipping `s_quad = min(s, s_f)` keeps the price branches off their poles. The clip on `s_power` likewise keeps the infinite derivative at 0 out of the dead branch. ✓ The reasoning in the comment is correct and worth keeping.
- **The double `where` on `γ=0`.** `safe_gamma`/`safe_inner`/`safe_floor` do keep `-1/γ` and `s_f^-γ` finite. ✓

### 1. The `sample_response_distribution` claim in the docstrings is wrong (blocking)

Two places say `implied_delivery` is "what `sample_response_distribution` wants as channel data":

- `budget_optimizer.py`, `BudgetOptimizationResult.implied_delivery`
- `price_response.py`, `PowerPriceResponse` Notes → "Reading the result"

`sample_response_distribution` passes `allocation_strategy` straight into `create_zero_dataset(channel_xr=...)`, whose contract is explicit: *"Dimensions must be a subset of `model.dims` and must **not** include the date dimension. Values are broadcast across every date."* `implied_delivery` has dims `(date_dim, *budget_dims)`, so it cannot be handed over as-is. A user following the docstring hits a dim error at best; at worst they reduce over date themselves and pick `sum` where `mean` is meant, or pass `budget_distribution_over_period` alongside and apply the concentration twice.

No test covers the round trip, which is why this survived. Either state the reduction explicitly (`implied_delivery.mean(date_dim)`, with a note that the per-period spread is already baked in) or add a test that actually feeds a result into `sample_response_distribution`. Given that method is marked deprecated in `create_zero_dataset`'s own docstring, I'd lean toward the first plus a pointer at the non-deprecated path.

### 2. Misleading gate message when the artifact exists but `channel_spend` is unusable

In `_media_price_reference`:

```python
if unpriced is None or unpriced or spend is None:
    who = (
        f"channels {unpriced}"
        if unpriced
        else "every optimized channel (no historical cost_per_unit table on the fitted model)"
    )
```

When the table exists and prices every optimized channel (`unpriced == []`) but `spend` was set to `None` by the dims check just above, the branch is taken and the message tells the user there is no `cost_per_unit` table. There is one. They will go set it again and get the same error.

Split the third condition out, or make `who` distinguish "no table" from "table is fine, `constant_data['channel_spend']` has dims I cannot use (got X, expected `(date, *budget_dims)`)". The second wording also tells them what to fix.

### 3. `delivery_report` ignores `compile_kwargs`

`MediaVariable._compile_delivery_report` calls `pytensor.function([z], outputs)` with no mode, while everything else in the optimizer compiles through `BudgetOptimizer.compile_kwargs`. For a user running `mode="JAX"` or `"NUMBA"` the report silently compiles on the default backend. Values agree, so this is not a correctness bug, but it is an inconsistency that will bite the first time an op in the price graph is backend-sensitive, and the whole point of the double-`where` comment is JAX gradient behaviour. `MediaVariable` has no handle on `compile_kwargs` today; threading it through the constructor is a two-line change.

### Smaller notes (non-blocking)

- **`is_identity` is evaluated on the declaration, not the resolution.** `_resolve_price_responses` gates on `response.is_identity`, but `resolve()` zeroes γ outside the mask. A response naming only masked-out channels resolves to the identity yet still trips the fitted-artifact gate. Conservative, so arguably fine, but the error message will read oddly ("your channels were fitted on nominal spend") for a response that would have been a no-op.
- **"exactly" is an overstatement.** `(implied_delivery * implied_price).sum(date) == budgets * num_periods` holds algebraically (`scale·s^(1-γ) · (s/s_ref)^γ = s`), not in floating point. The docstrings say "exactly" in three places. `implied_delivery * implied_price` at `s=0` is also `0 * nan = nan`, not 0, which the "on the spent cells" qualifier covers but only just.
- **JSON duplication.** `_unpriced_optimized_channels` inlines `json.loads(idata.attrs["cost_per_unit"])` while `mmm.py` already has `_deserialize_cost_per_unit`. Two readers of one serialization format will drift.
- **`spend_vars` get no gate.** Only `channel_data_var` is checked against the artifact; a spend variable needs a `reference_spend` and nothing else. The code comments acknowledge this and it is defensible (there is no artifact to check), but the asymmetry deserves a line in the `price_response` field description so it is not read as an oversight.
- Docstring quality across the new module is genuinely high, especially the Extrapolation and "Why not fixed-point iteration" notes. The FOC-vs-fixed-point test is the right way to pin that argument.

### Test coverage

306 passing on the targeted files is good, and the shape/continuity/gradient-at-zero/γ=0-bitwise set on `test_price_response.py` covers the map properly. Gaps I'd add: the `sample_response_distribution` round trip (item 1), and the `unpriced == [] and spend is None` gate path (item 2), which is currently unreachable from the test matrix as far as I can tell.

juanitorduz:
Addressed in `6c5d6a33`. Each item was checked against the code before changing anything; the three change requests all held, one of the smaller notes turned into a code change, and one is pushed back on with reasoning.

### 1. `sample_response_distribution` guidance — confirmed wrong, fixed

`create_zero_dataset` raises `"channel_xr must NOT include the date dimension."` (`utils.py:494-495`) and broadcasts date-less values across the window (`np.broadcast_to(vals, (n_dates, *vals.shape))`), so `implied_delivery` cannot be handed over as-is. Both docstrings (`BudgetOptimizationResult.implied_delivery` and the `PowerPriceResponse` "Reading the result" note) now say: `implied_delivery` is per period; score the plan's posterior response with `evaluate_response_distribution`, which runs the same graph the solver used, price map included; the deprecated path needs `implied_delivery.mean(date_dim)` and is exact only under a uniform `budget_distribution_over_period`, and with a non-uniform one the spread is already inside `implied_delivery` and must not be applied twice.

Test added rather than a round trip through the deprecated method: `test_response_distribution_sees_the_price_map` pins that `evaluate_response_distribution` at the same plan differs between the constant-price and the aware optimizer (measured mean ratio 1.045 on the fixture), and that `implied_delivery.mean("date")` is date-less and equals every period under the uniform spread.

### 2. Gate message when the table exists but `channel_spend` is unusable — confirmed, fixed

The `unpriced == [] and spend is None` path did fall into the "no historical cost_per_unit table" message. The two checks are now sequential: the nominal-spend refusal (with the two-part opt-out) fires only on a missing table or unpriced channels; when the table prices every optimized channel but `constant_data['channel_spend']` is missing or has the wrong dims, the error names that array and the dims it found, and asks for an explicit `reference_spend` only — not `assume_delivery_units`, because the artifact does vouch for the units in that case. `test_priced_table_with_unusable_channel_spend_names_the_real_problem` deletes `channel_spend` after `set_cost_per_unit` and asserts the message mentions `channel_spend`, does not mention "no historical cost_per_unit table" or "fitted on nominal spend", and that a supplied `reference_spend` alone is accepted.

### 3. `delivery_report` ignored `compile_kwargs` — confirmed, fixed

Worse than cosmetic on this stack: pytensor's default linker here is `NumbaLinker`, so a `cvm`-compiled objective was paired with a numba-compiled report. `MediaVariable` takes `compile_kwargs` and `_compile_delivery_report` passes them to `function`; `BudgetOptimizer` forwards its own for both the media variable and every `spend_vars` entry. `test_delivery_report_compiles_with_the_optimizer_compile_kwargs` builds with `Mode(linker="py")` and checks the compiled function's linker.

### Smaller notes

- **`is_identity` on the declaration, not the resolution** — changed. `PriceResponse.is_identity_on(dims, coords, mask, date_dim)` (default: the declaration; `PowerPriceResponse` overrides by zeroing outside the mask) is what `_resolve_price_responses` asks for the media variable, and `resolve()` branches on the masked gamma too. `test_a_response_that_only_prices_masked_channels_is_the_identity`: `elasticity={"channel_3": 0.4}` with `channel_3` masked out on an unpriced model now builds and resolves to the identity instead of being refused over channels it never touches.
- **"exactly"** — qualified in all three places as the algebraic identity `u(s) · p(s) = s`, holding to floating point, with the zero-money cells masked first.
- **`spend_vars` not gated** — stated in the `price_response` field description with the reason (no fitted price artifact exists for a node that is not channel data).
- **JSON duplication** — not changed, for two reasons. `_deserialize_cost_per_unit` lives in `mmm.py`, which imports `budget_optimizer.py` at module level, so reusing it here means either a lazy import or moving the helper; and what the gate reads is the top-level `columns` key of pandas' `orient="split"` envelope, i.e. pandas' documented format rather than an in-house encoding, while the `mmm.py` helper's added value is parsing the frame and normalising tz-aware dates, neither of which the gate needs. A comment at the read site now says both things so the two readers are at least aware of each other. If you would rather see the helper moved to a shared module, say so and I will do it in this PR.

310 passing on the five touched test modules (306 + 4).

daimon-pymclabs:
All three change requests confirmed fixed in `6c5d6a33`, and the pushback on the JSON note is the right call. Checked each against the code rather than the description.

**1. `sample_response_distribution` guidance.** Both docstrings now point at `evaluate_response_distribution` (`budget_optimizer.py:2685`, public and not deprecated) and state the `.mean(date_dim)` reduction plus the uniform-spread precondition for the deprecated path. The "must not be applied twice" sentence is the part that was actually load-bearing and it is there. `test_response_distribution_sees_the_price_map` pins the thing worth pinning: that the price map reaches the posterior score, not just the solver.

**2. Gate message.** The two checks are genuinely sequential now, and the second error names `channel_spend` and the dims it found. The reasoning for asking only for `reference_spend` and not `assume_delivery_units` on that path is correct: the table does vouch for the units, only the reference is missing.

**3. `compile_kwargs`.** Threaded through `MediaVariable` and every `spend_vars` entry. Good catch that it was worse than cosmetic on this stack; the linker assertion in both directions is the right shape of test.

`is_identity_on` is a cleaner resolution of the masked-channel note than I expected: putting the override on `PowerPriceResponse` rather than teaching the optimizer about elasticity layouts keeps the knowledge where it belongs, and switching `resolve()`'s branch to `not np.any(gamma)` makes the two agree by construction instead of by convention.

**On the JSON duplication: agreed, leave it.** The import direction argument settles it, and your second point is the stronger one: the gate reads pandas' documented `orient="split"` envelope, not an in-house encoding, so there is no shared format to drift from. The comment at the read site is enough. Not worth a lazy import or a new module for six lines.

---

One thing the fix to item 2 introduced, and it is the same shape as the bug it fixed:

```python
if spend is None:
    if response.reference_spend is None:
        raise ValueError(...)
    return None

if self.cost_per_unit is None:
    warnings.warn("... no cost_per_unit for the window, so the base price is 1 ...")
```

The new early return sits above the window-price warning, so a user with a priced fit, an explicit `reference_spend`, a missing or malformed `channel_spend`, and no window `cost_per_unit` gets no warning at all and their money silently reaches the model as units. Reachable, and precisely the case where the user is already off the happy path and least able to notice. The warning is about the window price and is independent of where the reference came from, so hoisting it above the `if spend is None:` block fixes it. One line, and the existing missing-window-price test would extend to cover it with a `channel_spend` deletion.

Non-blocking, but I would take it in this PR rather than as a follow-up. Otherwise this is ready from my side.

daimon-pymclabs:
## Second review round

The fix commit (`6c5d6a3`) does what it says: the gate now separates "no table / unpriced channels" from "table is fine but `constant_data['channel_spend']` is unusable", `delivery_report` compiles under `compile_kwargs` for both the media variable and `spend_vars`, and `is_identity_on` judges the declaration after masking. All three are covered by tests.

I also ran the branch this time rather than reading it. `tests/mmm/test_price_response.py`, `test_optimization_variables.py` and `test_cost_per_unit.py` pass under `-W error::RuntimeWarning`, which rules out the numpy warning hygiene I wanted to check on the `spend.where(spend > 0).mean(date)` path for an all-off channel.

Two probes of my own, outside what the suite covers:

- **Money identity under a non-uniform `budget_distribution_over_period` at the optimizer level** (the suite pins Jensen and the identity at the `MediaVariable` level, not through `allocate_budget`). Holds, including a period whose factor is exactly `0`: `(implied_delivery * implied_price).sum(date)` matches `budgets * num_periods` to float, the zero-factor period reports `nan` price, and the `gamma = 0` channel reports a flat `p0` across periods while the priced ones sweep `1.66 -> 0.93` as the concentrated buy clears down.
- **`implied_delivery` against the value the model node actually receives.** They differ by `channel_scales` (measured exactly `2.0` and `5.0` on a two-channel variable with those scales). See the second item below.

Three things left, none of them the math.

### 1. The dict form does not enforce the contract its own error message states

A bare response with `spend_vars` declared is refused, with:

> Pass a dict keyed by variable name and name every variable, including the ones that stay at a constant price [...] pricing one on a curve and leaving the other constant must be explicit.

But `_resolve_price_responses` only checks for *unknown* keys. `price_response={"some_spend_var": PowerPriceResponse(elasticity=0.3)}`, with `channel_data_var` omitted, is accepted: `price_responses.get(self.channel_data_var, (None, None))` hands the media variable `None`, media is priced at a constant, the spend variable is on a curve, and the two compete for the same pot through the same constraint. That is precisely the configuration the bare-object branch exists to refuse, reachable by omission instead of by a bare object.

The result is also silently thin there: `implied_delivery` / `implied_price` / `implied_marginal_price` come back `None` because they are only populated from the media variable's report, so the run gives no signal that anything was priced at all.

Either enforce what the message says (require every optimized monetary variable to appear as a key once any of them does, an explicit identity response being the way to say "constant"), or soften the message to describe what is actually checked. I would enforce: the asymmetry is the hazard, and the dict path is the one people will use.

No test covers a dict that omits `channel_data_var`; `test_budget_optimizer_mmm.py`'s `spend_vars` keying tests all name both.

### 2. `implied_delivery` is pre-`channel_scales`, and the docstring says otherwise

`delivery_report` evaluates `price_response.to_delivery(money, base_price=p0)`, while `to_model` divides that same quantity by `channel_scales` before it reaches the node. Measured on a `MediaVariable` with `channel_scales = [2.0, 5.0]`:

```
model_input[0]:      [42.77,  24.00]
implied_delivery[0]: [85.54, 120.00]
ratio:               [ 2.0,    5.0 ]
```

Pre-scale is the right choice: it is the number a planner wants (impressions, not scaled impressions), and the money identity `u(s) * p(s) = s` depends on it. But `BudgetOptimizationResult.implied_delivery` is documented as "In the model's channel units, per period", and the model's channel units are the post-scale ones. `MMM` pins `_channel_scales = 1.0` so nothing bites there, but `CustomModelWrapper` exposes `channel_scales`, and a `spend_vars` entry with `spend_var_scales` has the same gap through `variables[i].delivery_report(...)`.

Docstring fix rather than a code fix: say "in the delivery units the money buys, before `channel_scales`", and in `MediaVariable.delivery_report` note that the node receives `implied_delivery / channel_scales`.

### 3. The remaining branch where the gate message can still lie

`_unpriced_optimized_channels` returns `None` for two different reasons: no table on `idata.attrs`, and no `"channel"` budget dim to attribute the table's columns to. `_media_price_reference` collapses both into

> every optimized channel (no historical cost_per_unit table on the fitted model)

which is false in the second case, and is the same class of misleading message the fix commit just removed for `channel_spend`. I could not construct a path that reaches it through `MMM` (channel data always carries a `channel` dim, and a wrapper model would not carry the attr), so this is consistency rather than a live bug. Splitting it costs two lines: return a sentinel or raise from the "cannot attribute" case and name it ("the fitted model prices channels but this optimization's budget dims are `[...]`, with no `channel` dim to match the table's columns against").

### Smaller notes

- `PowerPriceResponse.is_identity_on` calls `_resolve_elasticity(template, date_dim, "price_response")` with a hardcoded label, while `resolve` passes `f"{name}: price_response"`. Since the gate now runs first, a malformed `elasticity` (date dim present, unknown dim, partial coordinate coverage) on the media variable reports the generic label instead of the variable-qualified one. Thread the label through.
- `reference_spend_tolerance` defaults to `10`, and the docstring is honest that it cannot see a mis-scale smaller than the tolerance, with `test_reference_spend_at_window_granularity_is_refused` pinning `4x` as passing. The failure mode the guard exists for is a window total mistaken for a per-period rate, which is off by exactly `num_periods` -- and the typical window is 4 to 13 periods, i.e. under the default. The optimizer knows `num_periods`, so it can test that specific hypothesis directly and independently of the tolerance: when `supplied / derived` is within a few percent of `num_periods`, say so by name rather than waiting for the generic factor check. That converts the common case from silent to caught.

juanitorduz:
Addressed in `fa87e08c`. All three held on inspection; both smaller notes turned into code.

### 1. Dict form did not enforce its own message — confirmed, enforced

Only unknown keys were checked, so `{"lf_budget": ...}` with `channel_data_var` omitted (or `{"channel_data": ...}` with a spend var omitted) was accepted and reproduced the one-pot asymmetry by omission, returning a result with no report at all. `_resolve_price_responses` now requires every monetary variable to appear once any does, and the error names the missing ones and how to say "constant" (`PowerPriceResponse(elasticity=0.0)`). `test_a_dict_must_name_every_monetary_variable` covers both omissions; the existing spend-var test now passes an explicit identity for media and asserts the media report is present and flat at `p0 = 1`, which is the side benefit of saying it explicitly.

### 2. `implied_delivery` is pre-`channel_scales` — confirmed, docstrings fixed

Pre-scale stays, for the reasons you give. `BudgetOptimizationResult.implied_delivery`, `MediaVariable.delivery_report` and the `PowerPriceResponse` "Reading the result" note now say "the delivery units the money buys, before `channel_scales`; the node receives `implied_delivery / channel_scales`, which coincides for an `MMM` (scales are 1)". `test_the_node_receives_implied_delivery_over_channel_scales` pins the ratio at the `MediaVariable` level with `channel_scales = [1, 10, 1000]`.

### 3. "No table" reported for "no `channel` dim" — confirmed, split

`_unpriced_optimized_channels` is now two steps: `_priced_channels()` (`None` only when there is no table) and the per-channel diff, and `_media_price_reference` words three facts separately: no table; a table but budget dims `[...]` with no `'channel'` dim to match its columns against; channels absent from the table. You said the second was unreachable through `MMM` — it is reachable through any custom model with the attr set, so `test_price_gate_names_a_missing_channel_dim_rather_than_a_missing_table` builds the six-line `pm.Model` with a `media` dim from `TestDateAxisMustMatchTheBlocks`, sets `idata.attrs["cost_per_unit"]`, and asserts the message names the dims and does not say "no historical cost_per_unit table".

### Smaller notes

- **Hardcoded label in `is_identity_on`** — threaded through (`label` keyword, default `"price_response"`); the optimizer passes `f"{name}: price_response"`. Asserted from `is_identity_on` directly with a date-dim elasticity. One thing that surfaced writing it: an *all-zero* declaration short-circuits as the identity before its dims are read, so `is_identity_on` does not raise for it; `resolve` still does, with the same label, so nothing malformed gets through.
- **`num_periods` hypothesis** — done, as an error rather than a widening of the tolerance. `resolve()` takes `num_periods` (`MediaVariable` passes its own); when a supplied reference is within 5% of `num_periods ×` the derived one on **every** comparable optimized cell, it is refused naming the window total and `num_periods`, independently of `reference_spend_tolerance`. The escape hatch is the knob that already exists: setting `reference_spend_tolerance` explicitly, to any value, is read (via `model_fields_set`) as the assertion that the scale is intended, and skips the hypothesis; the generic factor check then applies alone. Requiring every cell keeps a legitimate scale-up plan that happens to be `num_periods`-fold on one channel out of the trap; the coincidence on all channels is the mistake's signature. `test_a_reference_that_is_num_periods_times_the_fitted_spend_is_named_as_a_window_total` covers exact, 2.5% off, a mixed reference (passes), explicit tolerance (passes), and `num_periods=None` (passes); `test_reference_spend_at_window_granularity_is_refused` now asserts the 4-week case is caught at the default and passes with an explicit `reference_spend_tolerance=10.0`.

On the JSON read, one more fact than last round: `test_budget_optimizer_has_no_marketing_imports` bans this module from importing `pymc_marketing.mmm.mmm` at all, so reusing `_deserialize_cost_per_unit` is ruled out by policy as well as by the cycle; the comment at the read site now says so.

314 passing on the five touched test modules (310 + 4).

daimon-pymclabs:
## Review

Reviewed the full diff. The math is right and the test suite is unusually strong. I re-derived the `PowerPriceResponse` closed forms independently and they all hold:

- `u(s) = s_ref^γ s^(1-γ) / p0` gives `p(s) = p0 (s/s_ref)^γ` and `m(s) = 1/u'(s) = p(s)/(1-γ)`. ✓
- Quadratic floor: `a = (1+γ) scale s_f^-γ`, `b = -γ scale s_f^(-γ-1)` pins `u(0)=0`, matches value *and* slope at `s_f` (`a s_f + b s_f² = scale s_f^(1-γ)`, `a + 2b s_f = (1-γ) scale s_f^-γ`), stays concave (`b<0`) and increasing on `[0, s_f]`. ✓
- `u'(0)/u'(s_ref) = (1+γ)/(1-γ) · (s_f/s_ref)^-γ = M` inverts to the floor formula in the docstring. ✓
- Price continuity at the floor: `p0/(a + b s_f) = p0 (s_f/s_ref)^γ`. ✓
- `γ=0` collapses to `a=1, b=0, s_f=0` and the power branch is `1.0 * s ** 1.0`, which is exact in floating point, so the "bitwise identity" claim survives even in a mixed array.

The double-`where` on the coefficients (to keep `-1/γ` and `s_f**-1` off the dead branch) is the right call for JAX, and clipping *both* `where` inputs is the right call for the gradient. The `MediaVariable` ordering fix, `(s/c)^(1-γ) ≠ s^(1-γ)/c`, is the substantive correctness point of the PR and it is tested on both spreading paths.

The gate design (fitted `cost_per_unit` table as the proof of delivery units, per column, with three distinct failure messages and a named opt-out) is the part I'd have expected to get wrong, and it's careful. No cycle introduced: `price_response` imports `optimization_variables`, which imports back only under `TYPE_CHECKING`.

### Issues

**1. `sample_response_distribution` has no guard and is now easier to misuse.** It takes a raw allocation and feeds it to `create_zero_dataset` with no knowledge of the optimizer. `mmm.sample_response_distribution(result.budgets)` after a priced run silently scores money as units. This footgun predates the PR (it is already wrong under a plain `cost_per_unit`), but a non-identity `price_response` makes the correct input, `implied_delivery.mean(date_dim)`, non-obvious enough that prose in three docstrings is thin protection. Consider stamping something on `result.budgets.attrs` (you already stamp `pymc_marketing_version`) and warning from `sample_response_distribution` when it sees it.

**2. `"reference_spend_tolerance" not in self.model_fields_set` as the escape hatch for the window-total check.** Passing `reference_spend_tolerance=10.0`, the default value, changes behaviour. That is hard to discover and hard to reason about from a repr, and it means the only way to plan at exactly `num_periods ×` history is to set an unrelated-looking knob. Two alternatives: make the window-total check a `UserWarning` rather than a `ValueError` (the generic tolerance check still catches the real unit error at longer windows), or give it its own boolean.

**3. `json.loads(raw)` in `_priced_channels` is unguarded.** A malformed or non-pandas `idata.attrs["cost_per_unit"]` raises a bare `JSONDecodeError` out of the middle of the gate, bypassing all three of the curated messages the method exists to produce. A `try/except (ValueError, KeyError, TypeError)` falling back to `None` (treat as no table) keeps the error surface to the three you wrote.

**4. `set(table["columns"]) - {"date"} - set(self._budget_dims)`.** What is the budget-dims subtraction defending against? If the table is always wide with channel names as columns, it is dead; if it can ever be long, the `"date"` subtraction is also doing something. As written, a channel literally named like a budget dim (`"channel"`, `"country"`) would be dropped from `priced` and then refused by the gate as unpriced, which is the one failure mode with no clear message. Narrow, but a comment or a tighter read of the frame's orientation would settle it.

**5. `zero = money == 0.0` for the NaN mask.** SLSQP can land a cell at `1e-16` rather than exactly `0`, in which case `implied_price` reports the finite low-spend price `p0/a` for a cell that bought nothing, and the money identity picks up a cell the docs say is masked out. A tolerance relative to `reference_spend` (say `money <= 1e-12 * reference`) would make the reported mask match the intent. Low severity since the identity still holds algebraically.

**6. Doc/type mismatch on `MediaVariable.price_response`.** The class docstring documents the parameter as `PriceResponse or None`, but after `__init__` the attribute holds a `ResolvedPriceResponse`. Anyone reading `self.price_response.is_identity` in `to_model` has to work that out. Either name the attribute `_resolved_price_response` or document the attribute separately from the parameter.

**7. `delivery_report` compiles unconditionally on the identity path.** `allocate_budget` always calls it when any `price_response` was passed, so an `elasticity=0.0` baseline run pays a full extra `pytensor.function` compile to report `implied_price == p0`. Given the docstring explicitly recommends running the `γ=0` baseline beside the believed values as a sensitivity sweep, that is the common case. Worth making the three fields lazy (a property on the result holding the closure) or short-circuiting the identity case with a closed-form report.

### Style

**Docstring volume.** `price_response.py` is 747 lines and a large share is narrative rationale, some of it referencing the design process rather than the code ("the thread's three passes", "the issue body's ... was wrong", "per the 2026-09-22 correction"). The math, the precondition, the units contract and the extrapolation warning all earn their place in the module; the process history belongs in the PR description, which already has it. A reader six months out will be grateful for the `Units` and `Precondition` sections and puzzled by the rest. Same for a few of the inline comments in `budget_optimizer.py`, which argue with alternatives that were never in the code.

**Minor.** `_decision_date_coords` falls back to `list(range(num_periods))`, so on a model with no date coord `implied_*` carry integer date labels while `budgets` carries none. Harmless, slightly inconsistent.

### Tests

Coverage is the strongest part of the PR: C1 continuity across the floor, gradient finite at exactly zero, slope spread equal to `max_slope_ratio`, the money identity in both regions, the bitwise `γ=0` regression, FD gradient check, and the FOC-versus-fixed-point test that justifies not shipping the iteration. The gate matrix covers the partial table, the missing-`channel`-dim case and the unusable `channel_spend` case separately, which matches the three messages.

Two gaps I'd add:
- A non-uniform `budget_distribution_over_period` end to end through `allocate_budget`, not only at the `MediaVariable` level. The Jensen behaviour change is the one that bites existing users the moment they set a non-zero elasticity, and the interaction with `implied_delivery`'s already-spread values is exactly where the `sample_response_distribution` caveat lives.
- A `channel_scales != 1` case through the optimizer (a `CustomModelWrapper`), not just the node-level assertion. The ordering fix is the heart of the PR and scales are the other half of it.

### Verdict

Approve after 1 and 2 get a response; 3 through 7 are cheap and can ride along. Nothing here is a blocker on correctness.

juanitorduz:
Addressed in `729db355`. Items 1–6 and both test gaps are in; 7 and one style point get a measured response instead.

### 1. `sample_response_distribution` guard — done

`allocate_budget` stamps `result.budgets.attrs["price_response"]` (the resolved class name) whenever a response was applied, identity included, since the identity still reports and the money-vs-units question is the same. `BudgetOptimizerWrapper.sample_response_distribution` warns when it sees the stamp, naming both correct inputs: `result.implied_delivery.mean('date')` (exact under a uniform spread) and `evaluate_response_distribution`. Scoped to the price-response case rather than the pre-existing plain-`cost_per_unit` footgun you note: that one would warn on every `cost_per_unit` user of the deprecated method today, and there is no stamp to key it on without deciding what "converted units" means for a model fitted on spend; it belongs in its own issue. Test: `test_wrapper_optimize_budget_forwards_price_response` asserts the stamp and the warning through the wrapper. The stamp is lost under xarray arithmetic (`keep_attrs` is off by default), which is the limit of what an attr can do.

### 2. `model_fields_set` escape hatch — agreed, replaced

Passing the default value changing behaviour is the kind of thing nobody finds from a repr. The window-total check is now a `UserWarning` and the `model_fields_set` condition is gone: the generic factor check still refuses the unit error on any window longer than the tolerance, and a genuine `num_periods`-fold plan is no longer blocked. Unit test converted to `pytest.warns` with the no-warning rows under `simplefilter("error")`; the optimizer-level test asserts the warning through `mmm.budget_optimizer` directly.

### 3. Unguarded `json.loads` — done

`try/except (ValueError, KeyError, TypeError)` around both the parse and the `["columns"]` read; either failure reads as no usable table and the no-table wording now says "no usable". Test with a non-JSON string and with valid JSON that is not a split frame.

### 4. The budget-dims subtraction — narrowed, and the orientation stated

The frame is wide: `_parse_cost_per_unit_df` takes `["date", *custom_dims]` as dim columns and every other column as a channel, so the subtraction was defending against `geo` / `country` columns in a multidimensional model and was live there. `"channel"` itself is the axis the columns are labels of, so it is now excluded from the subtraction and a channel literally named `"channel"` stays priced. Comment at the read site says all of this.

### 5. `money == 0.0` — done

`money <= 1e-12 * reference_spend` (relative, so it has the units of money). Test extended: a cell at `1e-15` reports `nan`. The money-identity test's own `spent` mask was `budgets > 0` and disagreed with the report's mask the moment SLSQP left a cell at `1e-16` — which it did on this fixture — so it now reads "spent" off the report, which is the consumer-visible contract.

### 6. Parameter-vs-attribute type — documented

An `Attributes` section on `MediaVariable` states that the parameter is the declaration and the attribute is its `resolve()` result. Kept the name: `optimizer.optimization_variables.variables[i].price_response` is the documented route to a spend variable's report, and a leading underscore would hide it.

### 7. Report compile on the identity path — measured, leaving as is

On `simple_fitted_mmm`: optimizer construction and compile 239 ms, the identity report's compile **27 ms**, once per optimizer and then cached (a second `allocate_budget` costs 8 ms end to end). That is about 10% of the cost of building the optimizer, paid once, and a sweep over budgets on one optimizer pays it once too. Making the three fields lazy means turning dataclass fields into properties holding a closure, which changes how `BudgetOptimizationResult` is constructed and pickled and makes the identity a special case in two places. For 27 ms I would rather not; happy to revisit if the number is different on a real model.

### Style

The three phrases quoted — "the thread's three passes", "the issue body's ... was wrong", "per the 2026-09-22 correction" — are in the PR description, not in `price_response.py` or any comment; `grep` for `thread`, `issue body`, `2026`, `pass` over the three modules finds none of them. The Notes that are there (Precondition, Extrapolation, Units, Behaviour change, "The elasticity is an input", "Why not fixed-point iteration", "Reading the result") are the ones the issue explicitly asked to live in the docstring rather than only in a notebook, and each states a property of the code. I did reword one `to_model` comment that referred to "before price responses existed". If the maintainers want the two longer Notes moved to a docs page, that is a docs PR, and I would rather do it deliberately than trim here.

**Minor.** `_decision_date_coords` returns `None` for a length-only date dim, so the report carries no date labels the budgets do not have.

### Tests

- `test_non_uniform_distribution_end_to_end`: through `allocate_budget` with factors `[0.4, 0.3, 0.2, 0.1]` / uniform / `[0.7, 0.3, 0, 0]` across the three channels — money identity holds, the two dark periods report `nan` and zero delivery, the `gamma = 0` channel is flat at `p0`, and the front-loaded channel's price falls across the window. Your probe, as a test.
- `test_price_response_and_channel_scales_compose_through_the_optimizer`: on the `CustomModelWrapper` fixture, `channel_scales = [2, 5]` vs `1.0` with the same response — `implied_delivery` is identical across the two (the delivery the money buys does not depend on the scales) and the node input equals `implied_delivery / channel_scales`.

317 passing on the five touched test modules (314 + 3).

daimon-pymclabs:
## Re-review of `729db35`

All seven items from the previous round are addressed, and I re-checked each fix against the code rather than the commit message:

- `sample_response_distribution` now warns when it is handed a stamped allocation (`mmm.py:4649`), naming `implied_delivery.mean('date')` and `evaluate_response_distribution`. A warning rather than a refusal is the right call for a deprecated path that previously accepted this array silently.
- The `model_fields_set` escape hatch is gone; the window-total check is a `UserWarning` and the generic factor check still refuses the real unit error on any window longer than the tolerance. Passing the default tolerance no longer changes behaviour, which was the actual defect.
- Malformed `idata.attrs["cost_per_unit"]` is caught for `ValueError`/`KeyError`/`TypeError` and reads as "no usable table"; a JSON scalar or a list both land on the curated refusal rather than escaping the gate.
- `columns - {"date"} - (set(self._budget_dims) - {"channel"})` keeps a channel literally named `channel` priced.
- The nan mask is now relative (`money <= 1e-12 * reference_spend`) and dimensionally sound: resolved `reference_spend` is an `np.ndarray` of the dims shape, so it broadcasts against `(date, *dims)` money correctly.
- `MediaVariable.price_response` the attribute is documented apart from the parameter.
- The report graph compiles lazily now (`_delivery_report_fn is None`), so the compile is paid only when a report is actually produced.

New tests look right: the non-uniform `budget_distribution_over_period` case exercises a zero-factor period, a `gamma = 0` channel, and a monotone price under front-loaded money, and `channel_scales` now goes through `BudgetOptimizer`'s own wiring rather than a hand-built variable. `spent = result.implied_price.notnull().any("date")` is a better spent-predicate than `budgets > 0`.

CI is green on this head; only `test_slow (cvm, 2)` is still running.

### One thing worth fixing before merge

**The stamp fires on the identity response, so the `elasticity=0.0` baseline is no longer a no-op.** `delivery_report` returns `{}` only when `price_response is None` (`optimization_variables.py:667`), so an identity response still produces a report, and `budget_optimizer.py:3219` stamps `attrs["price_response"]` on any non-empty report. Concretely:

```python
identity = optimizer_with(PowerPriceResponse(elasticity=0.0)).allocate_budget(total_budget=300.0)
wrapper.sample_response_distribution(identity.budgets)
# UserWarning: allocation_strategy was optimized under a spend-dependent price ...
```

The price is constant there, and the run is meant to be operation-for-operation identical to a no-response run (that is what the `is_identity` short-circuit in `to_model` buys). Now it differs in two observable ways: the budgets carry an extra attr (so `xr.testing.assert_identical` against a baseline fails, and any attr-preserving round-trip carries it), and the recommended baseline triggers a warning telling the user their money is not money. `test_zero_elasticity_reproduces_the_constant_price_allocation` (`tests/mmm/test_cost_per_unit.py:1195`) passes only because `assert_allclose` ignores attrs.

Gating the stamp on `not self._media_variable.price_response.is_identity` fixes it, and matches the condition `to_model` already uses.

### Two nits

- The stamp value is `type(self._media_variable.price_response).__name__`, which resolves to `ResolvedPowerPriceResponse`. The warning then reads "optimized under a spend-dependent price (ResolvedPowerPriceResponse)", naming a class the user never wrote. The declaration's class name would be more useful.
- `optimization_variables.py:649` still documents `date_coords` as "positional `0..num_periods-1` when omitted", but `_decision_date_coords` now returns `None` for a length-only dim and the `DataArray` is built without a `date` coord at all. The return-section doc is correct; the parameter doc is stale.

juanitorduz:
Fixed in `60fe5acf`. Both the blocker and the two nits.

### The identity stamp

Reproduced before changing anything — `assert "price_response" not in identity.budgets.attrs` failed with `{'pymc_marketing_version': '1.1.0', 'price_response': 'ResolvedPowerPriceResponse'}`, and the warning fired on the recommended baseline. Your reading of the cause is exact: `delivery_report` returns `{}` only for `price_response is None`, and the stamp was gated on the report being non-empty rather than on the price actually varying.

The stamp is now gated on `resolved.is_identity`, the same condition `to_model` branches on, so the two agree by construction. `test_zero_elasticity_reproduces_the_constant_price_allocation` gained the assertion you named — `xr.testing.assert_identical(identity.budgets, baseline.budgets)` alongside the existing `assert_allclose`, which as you say could never have caught it — plus the explicit `"price_response" not in attrs`.

### Nits

- Stamp value is now the declaration's class (`PowerPriceResponse`), carried on a private attr set beside `_media_variable`, so the warning names what the user wrote. The wrapper test asserts the exact string, and asserts the identity run is *not* stamped, which is the same fact from the other side.
- `date_coords`'s parameter doc now says "the dim carries no coordinates when omitted", matching `_decision_date_coords` returning `None` and the Returns section.

One thing that surfaced writing the priced half of the wrapper test: `fitted_mmm` is geo × channel, so a channel-only `reference_spend` is refused with `must have exactly the budget dims ['geo', 'channel']`. Working as intended — that is the per-cell contract the multidimensional case needs — and the test now passes a `(geo, channel)` reference.

317 passing on the five touched test modules.

daimon-pymclabs:
## Re-review of `60fe5ac` — approving

The stamp is now gated on `resolved.is_identity`, the same predicate `to_model` branches on, so the two cannot drift apart. I checked the two things that would have made the fix incomplete:

- `_media_price_declaration` is assigned on the same path that assigns `_media_variable` (`budget_optimizer.py:1884-1887`), from the same `price_responses.get(self.channel_data_var, ...)` tuple that produced the resolved object. There is no path where the resolved response is non-identity and the declaration is `None`, so `type(...).__name__` cannot report `NoneType`.
- `assert_identical` on the identity-vs-baseline budgets is the right assertion: it compares attrs, which `assert_allclose` does not, so the regression that shipped in `729db35` now has a test that fails on it.

Stamping the declaration's class rather than the resolved one is the better choice for a message the user reads. The wrapper test asserting both sides (identity not stamped, `elasticity=0.3` stamped and warning) is stronger than the single-direction assertion it replaced.

CI is green on this head; only `test_slow (cvm, 2)` is still running, and it was green on the previous head.

One observation, not a change request: the stamp keys off the media channel's response only, so a run that prices a `spend_vars` entry and leaves the media channel at constant price produces unstamped budgets. That is correct for `sample_response_distribution`, which consumes the channel allocation, but if labelled per-variable reports land on the result later (your third follow-up) the stamp will want to widen with them.

Approving. Nothing outstanding from my side.

daimon-pymclabs:
## Review round 3

I re-ran the branch rather than reading the diff: the full three new test files pass (93 tests, 7m41s, under `-W error::RuntimeWarning`), and I re-probed the three items from round 2 plus six new hypotheses against a fitted `MMM`.

**The round-2 items are all closed.**

* The dict form of `price_response` now enforces the contract its error message states.
* `implied_delivery` is documented as pre-`channel_scales` ("the units the money buys -- before `channel_scales`; the model node receives `implied_delivery / channel_scales`"), which matches what I measured.
* The gate now separates the three facts it used to collapse: no usable table, a table that cannot be matched because the budget dims have no `channel`, and a table that leaves some optimized channels unpriced.

**What held under probing** (elasticity `{"tv": 0.3, "radio": 0.1}`, 8-period window):

* The money identity `sum_date(implied_delivery * implied_price) == budgets * n_periods` is exact to 1.1e-16 uniform and 4.4e-16 under a `budget_distribution_over_period` that is zero in half the periods.
* `implied_marginal_price / implied_price` is `1.4285714285714286` on `tv` against `1 / (1 - 0.3)`, and exactly `1.0` on the unpriced channel.
* `elasticity=0.0` is bitwise-identical to no price response at all, leaves no `price_response` attr on the budgets, and still attaches the report at price 1. The identity stamp fix does what it says.
* `implied_price` carries the window's real dates, and `evaluate_response_distribution` runs on the returned budgets.

Two things left, both small and neither in the math.

### 1. `budget_optimizer.py:2786` cross-references a method that does not exist on `MMM`

The `evaluate_response_distribution` docstring says it is "not `:meth:`~pymc_marketing.mmm.mmm.MMM.sample_response_distribution``". That target does not resolve: `hasattr(MMM, "sample_response_distribution")` is `False`, and a user who follows the pointer gets `AttributeError: 'MMM' object has no attribute 'sample_response_distribution'` (I hit this writing the probe). The method lives on `BudgetOptimizerWrapper`, which is exactly how `price_response.py:442` spells the same reference. Worth matching the two, since a nitpick cross-ref here is also a broken sphinx link.

### 2. The `implied_*` report is silently NaN wherever spend is zero, and nothing says so

Two ways to get there, both verified:

* a `budget_distribution_over_period` that is zero in some periods: 12 of 24 `implied_price` entries come back NaN, delivery 0 in those cells;
* a channel bounded to `(0.0, 0.0)`: its whole `implied_price` row is NaN, delivery 0.

The value is defensible, price at zero spend is genuinely undefined, but the docs describe the report as if it were dense, and the obvious summarization (`result.implied_price.mean()`, an `az.hdi`, a plot) gives NaN for the whole channel with no hint why. A sentence in the "Reading the result" block and one test pinning the zero-spend cell to NaN would make it a contract rather than a surprise. A non-uniform `budget_distribution_over_period` with a zero factor is not exotic: it is the flight-planning case the parameter exists for.

